PR #17, opened 2018-11-29
docs(Forma 36 Components): Changed the title from UI Component Librar…

wichniowski:
This PR will change the storybook title from UI Component Library to Forma 36 Components. Forma 36 React Components is to long and will break the title

image


wichniowski:
@m10l ok i'll close it 👍

PR #20, opened 2018-11-30
Storybook V4 Upgrade

m10l:
Hey, apologies pretty huge PR. This PR deals with updating Storybook and related dependencies and dealing with the ramifications.

Some changes that might be questionable - how builds are created, and the updated snapshots due to the new version of Jest.


packages/forma-36-react-components/src/components/Card/CardLoading/__snapshots__/CardLoading.test.js.snap
```diff
@@ -412,9 +410,7 @@ exports[`renders the component with an additional class name 1`] = `
       timeout={100}
       unmountOnExit={true}
     >
-      <React.Fragment>
-        CardLoading
-      </React.Fragment>
+      CardLoading
```
wichniowski: Did we throw out the Fragment or is this just the way the snapshot is rendered now?

m10l: It seems to be just the way fragments are rendered now in Jest

PR #39, opened 2019-01-09
fix(Card): Added fixed height to Card Loading, Reference Card and Ass…

wichniowski:
This PR will introduce a property height to Card Loading, Reference Card and Asset Card to prevent scrolling of the rich text editor while loading embeds.

### DIFF
--- a/packages/forma-36-react-components/src/components/Card/CardLoading/CardLoading.js
+++ b/packages/forma-36-react-components/src/components/Card/CardLoading/CardLoading.js
@@ -2,7 +2,6 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import cn from 'classnames';
 import { CSSTransition } from 'react-transition-group';
-import AnimateHeight from 'react-animate-height';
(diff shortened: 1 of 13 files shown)


packages/forma-36-react-components/src/components/Card/CardLoading/CardLoading.js
```diff
@@ -42,24 +63,31 @@ class CardLoading extends React.Component {
...
+      height,
       ...otherProps
     } = this.props;
 
     const classNames = cn(styles.CardLoading, extraClassNames, {
-      [styles['CardLoading--is-loading']]: isLoading,
+      [styles['CardLoading--is-loading']]: isLoading && !height,
     });
 
+    const Container = height ? React.Fragment : AnimateHeightContainer;
```
sbezludny: this condition appears in multiple places (see L71, 74, 82). I think we should start to think about better abstraction.

e.g.
use inversion of control and passing container as prop

please, tell me what do you think

PR #45, opened 2019-01-11
feat(Skeleton): Added Skeleton components

wichniowski:
Provides Skeleton components for display text and body text.

## Todo

- [x] Add API for body and display text
- [x] Add Tests
- [x] Create Image Skeleton component


packages/forma-36-react-components/src/components/Card/AssetCard/AssetCard.css
```diff
@@ -10,7 +10,7 @@
 }
 
 .AssetCard__asset {
-  height: calc(1rem * (253 / var(--rem-base)));
+  height: calc(1rem * (254 / var(--rem-base)));
```
m10l: Where does this number come from? Adding a comment with an explanation would be good

packages/forma-36-react-components/src/components/Skeleton/SkeletonContainer/SkeletonContainer.js
```diff
@@ -0,0 +1,146 @@
...
+    extraClassNames: PropTypes.string,
+    children: PropTypes.node.isRequired,
+    testId: PropTypes.string,
+    ariaLabel: PropTypes.string,
+    width: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+    height: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+    preserveAspectRatio: PropTypes.string,
+    clipId: PropTypes.string,
+    gradientId: PropTypes.string,
+    primaryColor: PropTypes.string,
```
m10l: I think instead of `primaryColor`/`secondaryColor` we should use `backgroundColor`/`highlightColor`, or `backgroundColor`/`foregroundColor`. Makes the application of the colour clearer to me

PR #56, opened 2019-01-24
Add typescript type def

Khaledgarbaya:
This PR adds typescript type definition for all of the forma-36 components


sbezludny:
tested with typescript@latest -- works just fine

PR #57, opened 2019-01-24
Add warning note type

m10l:
This PR adds a warning note type. It also adds a warning icon type, and "peach" colour tokens fitting with existing colours for other notes


domarku:
Alles gut! 👍

PR #58, opened 2019-01-28
Added line-height to Textarea

domarku:
## Change

- Added line-height to the Textarea component

## Before
before

## After
after


domarku:
I can't merge so feel free to do it whenever you're ready :)

PR #60, opened 2019-01-28
Typescript Support. Converted Icon, Spinner and Button components

suevalov:
## Added Typescript support

Most of the changes are in yarn.lock file, so it shouldn't be hard to review.

- [x] Configure TSLint for Typescript
- [x] Generate `.d.ts` files automatically on build
- [x] Check types as prepush hook
- [x] Update plop templates to generate TS files for new components

## Other updates

* Updated Node.js to 8.12.0 to be compatible with new ESLint version
* Updated ESLint and fixed a bunch of new found violations
    * Missing `type="button"`
    * Reading `this.state` while using `this.setState` was incorrect. Updated to use callback.
* Replaced `sinon.fake` with `jest.fn()`


wichniowski:
wunderbar! 👍  I think we'd also need to adjust the plop templates once this is done

PR #73, opened 2019-02-06
chore(React): update React to 16.8 which introduces hooks

wichniowski:
# Purpose of PR

Just updating react to at least use 16.8 so we can use hooks if we want to

## PR Checklist

- [x] Tests are passing


suevalov:
Using hooks is a breaking change and it requires a bump of peer dependencies. We can use it for stories and/or marketing website.

wichniowski:
We'll not update to prevent using 16.8 minimum requirement

PR #93, opened 2019-02-14
feat(Tabs): Created Tabs component

suevalov:
# Purpose of PR

This PR introduces new set of components: `Tabs`, `Tab` and `TabPanel`.

`Tabs` and `TabPanel` have no styles and used only for logical composition and hiding all required ARIA-attributes.

tabs

## PR Checklist

- [x] Tests are passing


packages/forma-36-react-components/src/components/Tabs/Tab.tsx
```diff
@@ -0,0 +1,86 @@
...
+      elementProps['aria-disabled'] = true;
+    }
+    if (href) {
+      elementProps['href'] = href;
+      if (selected) {
+        elementProps['aria-current'] = 'page';
+      }
+      return <a {...elementProps}>{children}</a>;
+    } else {
+      elementProps['aria-selected'] = selected;
```
wichniowski: this is clever :)

packages/forma-36-react-components/src/components/Tabs/Tabs.css
```diff
@@ -0,0 +1,54 @@
...
+  position: absolute;
+  background: transparent;
+  bottom: 0;
+  left: 0;
+  right: 0;
+  height: 3px;
+}
+
+.Tab:hover:before {
+  background: #a7e2ff;
```
wichniowski: don't we have a token for that? Isn't that primary blue?

wichniowski: --color-primary

wichniowski: or --color-blue-dark i guess instead of this value

suevalov: it's quite different actually, maybe I could use token and play with opacity

packages/forma-36-react-components/src/components/Tabs/TabPanel.tsx
```diff
@@ -0,0 +1,35 @@
+import React, { Component } from 'react';
+
+interface TabPanelProps
+  extends React.DetailedHTMLProps<
+    React.HTMLAttributes<HTMLDivElement>,
+    HTMLDivElement
+  > {
```
guilebarbosa: Why do you need this? Just curious.

suevalov: Because API allows using (with `...rest`) all HTML properties that you would normally use with `<div />`

PR #95, opened 2019-02-18
Typography container component

m10l:
# Purpose of PR

A wrapper to automatically add bottom margins to Typography components

The `<Typography>` component should be used to wrap a group of typography components (`<Heading>`, `<Paragraph>`, etc) for cases where you wish to display these components with a set of default margins.

We follow the principle that a component should only be responsible for its own internal spacing - never its external spacing. This means that we're flexible in where our components can be used without having to override margins.

Using bottom margins only makes it easier to control layout and avoid adding any unintentional spacing

## PR Checklist

- [x] Tests are passing


packages/forma-36-react-components/tools/.storybook/webpack.config.js
```diff
@@ -16,8 +16,31 @@ module.exports = {
           require.resolve('react-docgen-typescript-loader'),
         ],
       },
+      {
```
wichniowski: those fcss classes will be available globally right?

m10l: Yup! Also good catch - I needed to make this available to production builds. I've updated the PR

PR #97, opened 2019-02-19
TextInput refactor

m10l:
# Purpose of PR

Migrates the TextInput component to TypeScript.

Also:

- Creates TypeScript defs for CSS (using [typings-for-css-modules-loader](https://github.com/Jimdo/typings-for-css-modules-loader)) so that we can import CSS just using import (also has component development benefits!)
- Creates leaner builds (we were shipping stories and tests)
- Ignores `dist-storybook` directory in TypeScript config
- Adds an `inputRef` prop for setting a ref on the component

## PR Checklist

- [x] Tests are passing


packages/forma-36-react-components/src/components/TextInput/__snapshots__/TextInput.test.tsx.snap
```diff
@@ -2,11 +2,10 @@
 
 exports[`renders the component as required 1`] = `
 <div
-  className="TextInput TextInput--full"
+  className=""
```
suevalov: Looks like this is unexpected update and something is broken with styles import

packages/forma-36-react-components/src/components/TextInput/TextInput.tsx
```diff
@@ -1,36 +1,30 @@
...
+  testId?: string;
+  maxLength?: number;
+  onBlur?: (...args: any[]) => any;
+  onCopy?: (...args: any[]) => any;
+  value?: string;
+  error?: boolean;
+  required?: boolean;
+}
 
+class TextInput extends React.Component<TextInputPropTypes> {
```
suevalov: ```export class TextInput extends Component<TextInputPropTypes> {}```

it's the only way to build properties table properly in storybook

packages/forma-36-react-components/package.json
```diff
@@ -120,7 +121,7 @@
     "start": "start-storybook -p 9001 -c ./tools/.storybook",
     "test": "jest",
     "build:clean": "rm -rf dist",
-    "build:scripts": "BABEL_ENV=build yarn babel src --out-dir dist --ignore *.stories.js,*.test.js --extensions '.ts','.tsx','.js'",
+    "build:scripts": "BABEL_ENV=build yarn babel src --out-dir dist --ignore 'src/**/*.stories.js','src/**/*.stories.tsx','src/**/*.test.tsx','src/**/*.test.js','src/**/*.css.d.ts' --extensions '.ts','.tsx','.js'",
```
suevalov: 👌

packages/forma-36-react-components/src/components/Asset/Asset.css.d.ts
```diff
@@ -0,0 +1,14 @@
+// This file is automatically generated by typings-for-css-modules.
+// Please do not change this file!
+export interface IAssetCss {
```
suevalov: Cool, I really like it

packages/forma-36-react-components/src/components/TextInput/__snapshots__/TextInput.test.tsx.snap
m10l: Good catch, didn't need to import styles as I initially thought, fixed

packages/forma-36-react-components/src/components/TextInput/TextInput.tsx
```diff
@@ -20,6 +20,7 @@ export interface TextInputPropTypes {
   onBlur?: (...args: any[]) => any;
   onCopy?: (...args: any[]) => any;
   value?: string;
+  inputRef?: any;
```
m10l: @suevalov I'm not sure what this type should be. Setting it to `any` isn't ideal, but also makes TypeScript upset in the Storybook example:

```
Property 'focus' does not exist on type '{}'.
```

Also, maybe this isn't the right way to handle refs, just seemed the simplest for most use cases.

suevalov: I believe it should be `RefObject<HTMLInputElement>`

suevalov: `import React, { RefObject } from 'react';`

packages/forma-36-react-components/src/components/TextInput/TextInput.tsx
```diff
@@ -42,6 +43,7 @@ class TextInput extends Component<TextInputPropTypes> {
     disabled: false,
     required: false,
     width: 'full',
+    inputRef: undefined,
```
suevalov: listing undefined properties is optional

packages/forma-36-react-components/src/components/TextInput/TextInput.tsx
m10l: Ahh, I also needed to update how I created the ref in the story.

```
const textInputRef = React.createRef<HTMLInputElement>();
```

PR #103, opened 2019-02-26
fix(Notification): Fixed exception when showing notifications

suevalov:
# Purpose of PR

Fixes error happened when notification was showing before container is mounted in React Portal.

## PR Checklist

- [x] Tests are passing


packages/forma-36-react-components/src/components/Notification/index.tsx
```diff
@@ -7,42 +9,43 @@ import NotificationManager, {
...
   internalAPI[fnName] = fn;
 }
 
-function createRoot() {
+function createRoot(callback) {
   const container = document.createElement('div');
   document.body.appendChild(container);
 
-  render(<NotificationManager register={registerAPI} />, container);
+  render(<NotificationManager register={registerAPI} />, container, callback);
```
MarkusLaut: @suevalov 
Could this also be wrapped in a Promise, like `afterInit`?

suevalov: This function is not exposed as external API and uses callback internally

PR #115, opened 2019-03-01
fix(Storybook): Fix broken global styles

m10l:
# Purpose of PR

Our styles from F36 FCSS aren't currently working as expected. This PR fixes that

## PR Checklist

- [ ] Tests are passing


packages/forma-36-react-components/tools/.storybook/storybook.global.css.d.ts
```diff
@@ -0,0 +1,230 @@
+// This file is automatically generated by typings-for-css-modules.
+// Please do not change this file!
+export interface IStorybookGlobalCss {
```
suevalov: Why it's produced? I should not happen according to the configuration I see

m10l: Good catch, was an artefact from trying something else. I've also removed the `font.css.d.ts` file

PR #128, opened 2019-03-12
fix: 🐛 Fallback font collection of the sans

azer:
# Purpose of PR

Typography components look not nice as the primary sans font does not load in extensions.

This PR replaces "Helvetica" with a collection of sans fonts that look close to "'Avenir Next W01'".

Here is how SectionHeading looks like in my extension: 

It looks bigger, thicker, and uglier. Note that I use Linux which doesn't have Helvetica, so I get default system `sans-serif` font which leaves a bitter taste in your mouth.  

This is what we expect to see (when the font loads):

After adding the change in this PR, it looks as below: 

It's not identical, but close enough.

## PR Checklist

- [x] Tests are passing


wichniowski:
cool so many fallbacks :) but maybe we should decide on one of the open source fonts and distribute it with the component library so we can be sure it will look a certain way if Avenir is not available. For that i'd also throw Lato in the mix :D

azer:
@wichniowski I think this list was specifically crafted for selecting the best possible sans font in the default configurations of common operating systems, so they're good as fallback fonts. I know that changing the primary font is huge work, so just wanted to extend the fallback collection in this PR.

m10l:
@azer Nice - seems like a good interim solution. Thanks for the PR 🎉

domarku:
@wichniowski Choosing an open source font to distribute with the library is a separate effort we were considering. This is more of a reasonable interim solution, as @m10l said :)

PR #130, opened 2019-03-13
feat(Modal): Accept non-string titles

MarkusLaut:
More flexibility, in our case to handle long titles with ellipses, but also e.g. svg in titles.

# Purpose of PR

Allow jsx elements as title of modal dialog.
We need it to shorten long titles with ellipses, but other use cases are imaginable.

## PR Checklist

- [x] Tests are passing
- [x] Storybook stories are added/updated/not required (a string is still a valid title)


MarkusLaut:
As discussed with @m10l , we will handle ellipses specifically instead

PR #158, opened 2019-04-01
Add switch button

azer:
# Purpose of PR

Adds a Switch button. To see my motivation to add Switch to Forma36, please see following document:
[Checkboxes vs Switches](https://docs.google.com/document/d/1Iu1E0RzvsSa9UzaOgEwoWw0f7NRoHjFDQqE5pKIw05g/edit?usp=sharing)

# How does it look?

Enabled:

Disabled:

### DIFF
--- a/packages/forma-36-react-components/src/components/Switch/Switch.tsx
+++ b/packages/forma-36-react-components/src/components/Switch/Switch.tsx
@@ -1,0 +1,50 @@
+import React, { FunctionComponent, KeyboardEvent } from 'react';
+import FormLabel from '../Typography/Paragraph';
+import classNames from 'classnames';
+import styles from './Switch.css';
+
+export interface SwitchProps {
+  onToggle?: (isChecked: boolean) => void;
+  isChecked?: boolean;
+  isDisabled?: boolean;
+  labelText: string;
+  className?: string;
+}
+
+export const Switch: FunctionComponent<SwitchProps> = (props: SwitchProps) => {
+  return (
+    <div className={classNames(styles['Switch__wrapper'], props.className)}>
+      <input
+        type="checkbox"
+        onClick={onToggle}
+        onKeyUp={onKeyUp}
+        checked={props.isChecked}
+        className={classNames(styles['Switch'], {
+          [styles['Switch--checked']]: props.isChecked,
+          [styles['Switch--disabled']]: props.isDisabled,
+        })}
+      />
+      <FormLabel
+        className={classNames(styles['Switch__label'], {
+          [styles['Switch__label--disabled']]: props.isDisabled,
+        })}
+      >
+        {props.labelText}
+      </FormLabel>
+    </div>
+  );
+
+  function onToggle() {
+    if (props.onToggle) {
+      props.onToggle(!props.isChecked);
+    }
+  }
+
+  function onKeyUp(e: KeyboardEvent<HTMLElement>) {
+    if (props.onToggle && e.key === 'Enter') {
+      props.onToggle(!props.isChecked);
+    }
+  }
+};
+
+export default Switch;
--- a/packages/forma-36-react-components/src/components/Switch/Switch.css
+++ b/packages/forma-36-react-components/src/components/Switch/Switch.css
@@ -1,0 +1,64 @@
+@import 'settings/colors';
+@import 'settings/dimensions';
+@import 'settings/transitions';
+
+.Switch__wrapper {
+  display: flex;
+}
+
+.Switch {
+  box-sizing: content-box;
+  width: calc(1rem * (34 / var(--font-base-default)));
+  height: calc(1rem * (17 / var(--font-base-default)));
+  background-color: var(--color-element-mid);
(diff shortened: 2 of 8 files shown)


packages/forma-36-react-components/src/components/Switch/Switch.tsx
```diff
@@ -0,0 +1,51 @@
...
+  disabled?: boolean;
+  extraClassNames?: string;
+}
+
+export const Switch: FunctionComponent<SwitchProps> = (props: SwitchProps) => {
+  return (
+    <main
+      className={classNames(styles['Switch__wrapper'], props.extraClassNames)}
+    >
+      <aside
```
wichniowski: Why are you using those semantic elements? Why does it make sense that the switch is an aside?

azer: To make it easier to understand the general structure; one container and an inner box on the side next to the paragraph

packages/forma-36-react-components/src/components/Switch/style.css
```diff
@@ -0,0 +1,55 @@
...
+
+.Switch:before {
+  position: absolute;
+  content: '';
+  width: 17px;
+  height: 17px;
+  left: 2px;
+  top: 2px;
+  background: var(--color-white);
+  -webkit-transition: 0.1s;
```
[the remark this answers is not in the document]
m10l: @azer +1 for tokens. Also you can use unprefixed css properties here, so just transition. I'd also recommend specifying which properties you want to target for the transition.

```css
transition: left var(--transition-easing-default) var(--transition-duration-default);
```

In general I think there's a better approach for styling this component. Here's an example: https://codepen.io/anon/pen/mgOeyL. This uses border for the spacing between the inner and outer components, which means you can easy to understand values for the transition - which should use `transform` rather than `left`/`right` for performance purposes

azer:
I've updated the PR, hopefully addressed the feedback.

packages/forma-36-react-components/src/components/Switch/style.css
```diff
@@ -0,0 +1,56 @@
+@import 'settings/colors';
+
+.Switch__wrapper {
+  display: flex;
+}
+
+.Switch {
+  width: 35px;
```
m10l: Please use rems for dimensions. We use `calc` and a formula for translating pixel values to rem

**Example to render 35px to equivalent rem value based on our font base size:**

```css
calc(1rem * (35/ var(--font-base-default)));
```

packages/forma-36-react-components/src/components/Switch/style.css
```diff
@@ -0,0 +1,56 @@
...
+  border: 0;
+  appearance: none;
+}
+
+.Switch:before {
+  position: absolute;
+  content: '';
+  width: 17px;
+  height: 17px;
+  left: 2px;
```
m10l: Would recommend adding a border to the main `.Switch` element so we don't have to play around with positioning like this. https://codepen.io/anon/pen/NmgEMX

packages/forma-36-react-components/src/components/Switch/Switch.tsx
```diff
@@ -0,0 +1,51 @@
...
+import FormLabel from '../Typography/Paragraph';
+import classNames from 'classnames';
+import styles from './style.css';
+
+export interface SwitchProps {
+  onToggle?: (isChecked: boolean) => void;
+  isChecked?: boolean;
+  isDisabled?: boolean;
+  labelText: string;
+  extraClassNames?: string;
```
m10l: Please use `className` - we made a change in f36 v3

packages/forma-36-react-components/src/components/Switch/Switch.tsx
```diff
@@ -0,0 +1,51 @@
...
+}
+
+export const Switch: FunctionComponent<SwitchProps> = (props: SwitchProps) => {
+  return (
+    <div
+      className={classNames(styles['Switch__wrapper'], props.extraClassNames)}
```
m10l: `props.className`

packages/forma-36-react-components/src/components/Switch/Switch.tsx
```diff
@@ -0,0 +1,49 @@
...
+    </div>
+  );
+
+  function onToggle() {
+    if (props.onToggle) {
+      props.onToggle(!props.isChecked);
+    }
+  }
+
+  function onKeyUp(e: KeyboardEvent<HTMLElement>) {
```
wichniowski: I'm a bit worried with how we use the checkbox here. In this case enter would not set the checkbox to checked since natively it would be space which checks the input. But as we implement it now it does not matter since we're just returning the !props.isChecked which means the checkbox elements state and what it returns will get out of sync.. can this result in unexpected behaviour? So basically we're just using the checkbox to render some interactive element..

m10l: Good catch, just updated as well as adding more snapshot tests

PR #175, opened 2019-04-12
WIP: docs(TextField): Add TextField component documenation

wichniowski:
TextField component documentation.

## PR Checklist

- [x] Tests are passing

## Doc Checklist

- [x] Provide all configurations
- [ ] Provide copy guidelines

### DIFF
--- a/packages/forma-36-website/src/pages/components/text-field/index.mdx
+++ b/packages/forma-36-website/src/pages/components/text-field/index.mdx
@@ -6,11 +6,11 @@ storybook: 'https://f36-storybook.contentful.com/?path=/story/components-textfie
 ---
 
 The `TextField` component is a composition of the `FormLabel`, `TextInput`, `HelpText` and `ValidationMessage` components.
-It allows the user to input text or numbers and should be used in a Form context.
+It allows users to input text or numbers and should be used in a form context.
 
 ## Best practices
 
-- Provide a labelText, name and id to ensure accessibility
+- To ensure the TextField component meets accessibility standards, provide labelText, name and id.
 
 ## Copy considerations
 
@@ -18,11 +18,12 @@ Todo: Provide copy considerations
 
 ## TextField variations
 
-The TextField component can be configured in a number of different ways. Here is a guide to when to use certain variations.
+The TextField component can be configured in a number of different ways. Use this guidance to determine the correct variation based on the context. 
 
 ### Compositional properties
 
-Since the TextField is a composition of multiple components it is possible to pass child component specific properties. For this purpose the TextField exposes the follwing properties:
+Since the TextField is a composition of multiple components, it is possible to pass child component specific properties. 
+For this purpose, the TextField exposes the following properties:
 
 ```jsx
 <React.Fragment>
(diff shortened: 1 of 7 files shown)


wichniowski:
@m10l I'd do the the docs of all components which are part of TextField and then document the TextField as a composition of those components wdyt?

m10l:
> @m10l I'd do the the docs of all components which are part of TextField and then document the TextField as a composition of those components wdyt?

Sounds good!

wichniowski:
@m10l yeah I'm meeting later today with Sara

wichniowski:
@m10l @joaoramos I added docs for HelpText, ValidationMessage, FormLabel, TextInput and TextField as the composition out of those components. I'll add TextArea as well. Other component docs like CopyButton which is also used in the TextField component will happen in a sperate PR. @sarakalinoski is reviewing copy and opened up her own branch. So lets get the review started :). Some open questions:

- Should we add a section for a11y consideration per component, right now I've added it to the best practices.
- How to call "compositional properties" like textInputProps.. maybe compositional properties is ok.
- For all interactive elements, we need to provide some kind of a note that event handlers are provided, so we don't have to mention obvious props like onClick every time for button or links. I think this should stay part of the storybook wdyt.

sarakalinoski:
> * Should we add a section for a11y consideration per component, right now I've added it to the best practices.

Can we see how many A11y considerations we have for each component and decide from there? If it's just one it makes sense to wrap it under another section, but if we have more, why not make it it's own section.

wichniowski:
@sarakalinoski yeah I think this is the best way to deal with it.

packages/forma-36-website/src/pages/components/text-field/index.mdx
```diff
@@ -0,0 +1,172 @@
...
+github: 'https://github.com/contentful/forma-36/tree/master/packages/forma-36-react-components/src/components/TextField'
+storybook: 'https://f36-storybook.contentful.com/?path=/story/components-textfield--default'
+---
+
+The `TextField` component is a composition of the `FormLabel`, `TextInput`, `HelpText` and `ValidationMessage` components.
+It allows the user to input text or numbers and should be used in a Form context.
+
+## Best practices
+
+- Provide a labelText, name and id to ensure accessibility
```
sarakalinoski: Don't need "a" and start with objective:

To ensure TextField meets accessibility standards, provide labelText, name and id.

packages/forma-36-website/src/pages/components/text-field/index.mdx
```diff
@@ -0,0 +1,172 @@
...
+
+## Copy considerations
+
+Todo: Provide copy considerations
+
+## TextField variations
+
+The TextField component can be configured in a number of different ways. Here is a guide to when to use certain variations.
```
sarakalinoski: The TextField component can be configured in a number of different ways. Use this guidance to determine which variation to use based on the context.

wichniowski:
@sarakalinoski wow thanks a lot for your review and all those great commits :)

PR #186, opened 2019-04-24
feat(EmptyState): EmptyState component

thebesson:
# Purpose of PR

This is the component for displaying empty states on various pages. 

## PR Checklist

- [x] Tests are passing


packages/forma-36-react-components/src/components/EmptyState/EmptyState.css
```diff
@@ -0,0 +1,29 @@
+@import 'settings/colors';
+@import 'settings/typography';
+@import 'settings/transitions';
```
m10l: We can remove the color and transition imports as we won't be using them.

Please also add `@import 'settings/dimensions';` so that we can use our f36 spacing variables

packages/forma-36-react-components/src/components/EmptyState/EmptyState.css
```diff
@@ -0,0 +1,29 @@
...
+  align-items: center;
+}
+.EmptyState_illustration {
+  background-repeat: no-repeat;
+  background-size: cover;
+  background-position: center;
+}
+
+.EmptyState_element {
+  max-width: 500px;
```
m10l: We don't have any nice helpers, and this is a good reminder to add documentation - we try to use REMs for all spacing so that if a user has their browser font size set to a different size, the UI will resize to accommodate.

We've been using `calc` to help us write these in a way that you can see the original desired pixel value.

tl;dr: please change to `max-width: calc(1rem * (500 / var(--font-base-default)));`

packages/forma-36-react-components/src/components/EmptyState/EmptyState.css
```diff
@@ -0,0 +1,29 @@
...
+  margin-bottom: 16px;
```
m10l: Please change to `margin-bottom: var(--spacing-m);`

packages/forma-36-react-components/src/components/EmptyState/EmptyState.tsx
```diff
@@ -0,0 +1,115 @@
...
+   */
+  children?: React.ReactNode;
+  /**
+   * An ID used for testing purposes applied as a data attribute (data-test-id)
+   */
+  testId?: string;
+  /**
+   * Props for image block
+   */
+  image?: ImageProps;
```
m10l: Maybe rename to `imageProps` - we use this convention so that it's clearer what kind of prop this is just from the name

packages/forma-36-react-components/src/components/EmptyState/EmptyState.tsx
```diff
@@ -0,0 +1,115 @@
...
+  description: string;
+} & typeof defaultProps;
+
+interface ImageProps {
+  /**
+   * Url of the illustration for empty state
+   */
+  url: string;
+  /**
+   * Width of the illustration for empty state
```
m10l: Maybe update comment with an example value to make it clear that `width`/`height` expects the unit too:

```
Width of the illustration for empty state including unit (e.g. '250px')
```

packages/forma-36-react-components/src/components/EmptyState/EmptyState.tsx
```diff
@@ -0,0 +1,115 @@
...
+              className={cn(
+                image.className,
+                styles['EmptyState_element'],
+                styles['EmptyState_illustration'],
+              )}
+              aria-label={image.description}
+              role="img"
+            />
+          )}
+          <Heading element={'h1'} className={styles['EmptyState_element']}>
```
m10l: You can remove the `element` prop here as h1 is the default. Then I'd recommend adding a `headingProps` object to allow developers to override the default element if needed.

packages/forma-36-react-components/src/components/EmptyState/EmptyState.tsx
```diff
@@ -0,0 +1,115 @@
...
+            {heading}
+          </Heading>
+          <Paragraph
+            element={'p'}
```
m10l: You can remove the `element` prop here as p is the default. Then I'd recommend adding a `paragraphProps` object to allow developers to override the default element if needed.

packages/forma-36-react-components/src/components/EmptyState/EmptyState.tsx
thebesson: I don't think it's necessary, the design I got from Sara has consistent heading and paragraph sizings, the only customization I could trace is different image and child components - buttons or other controls, for example. I'll remove the element prop though.

packages/forma-36-react-components/src/components/EmptyState/EmptyState.tsx
thebesson: I have the same point as the one above. I'll discuss with Sara possibility of customisation of this component from the design perspective.

packages/forma-36-react-components/src/components/EmptyState/EmptyState.tsx
m10l: @thebesson Ahh, this prop doesn't have any impact on visual appearance - it just changes the semantic element used to render the heading. It's useful if the title of the page is rendered in a H1 and using this component as a H1 is incorrect semantically.

packages/forma-36-react-components/src/components/EmptyState/EmptyState.tsx
m10l: @thebesson Same point as above (https://github.com/contentful/forma-36/pull/186#discussion_r278215530), but, this one probably needed actually. Would be nice from a consistency point of view, but, in reality I can't imagine it being used.

packages/forma-36-react-components/src/components/EmptyState/EmptyState.tsx
thebesson: I see. Makes sense, but I don't want to overcomplicate the api for this component. Semantically speaking, if it's a state for when this component is the only thing displayed, does it makes sense to have it as H1?

m10l:
If the containing element is an `article` then yup! Maybe this would be a
good change. The prop would be optional anyway though so wouldn’t over
complicate the api in my opinion

packages/forma-36-react-components/src/components/EmptyState/EmptyState.tsx
m10l: Oop - tried replying via email but it didn't make this thread. Reposting here for better clarity.

If the containing element is an `article` then H1 would be fine. Maybe this would be a
good change anyway. The prop would be optional anyway though so wouldn’t over
complicate the api in my opinion. I quite strongly believe this change should be done, as we should have the convention that it's possible to pass props to all child f36 components

thebesson:
@m10l I updated the component.

PR #187, opened 2019-04-24
Refactor card drag handles

m10l:
# Purpose of PR

This PR introduces a breaking change: the `withDragHandle` prop is now deprecated. To render a drag handle on a component you must pass a CardDragHandle via the `cardDragHandleComponent` prop. This change is necessary for integrating drag and drop libraries such as `react-sortable-hoc` (https://github.com/clauderic/react-sortable-hoc/blob/master/examples/drag-handle.js#L10) - it'll allow the user to pass a HOC drag and drop handle.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Card/AssetCard/AssetCard.tsx
+++ b/packages/forma-36-react-components/src/components/Card/AssetCard/AssetCard.tsx
@@ -47,17 +47,21 @@ export type AssetCardProps = {
    */
   type?: AssetType;
   /**
-   * Renders a drag handle for the component for use in drag and drop contexts
+   * Prop to pass a custom CardDragHandle component to for use in drag and drop contexts
+   */
+  cardDragHandleComponent?: React.ReactNode;
+  /**
+   * Renders a default drag handle for the component for use in drag and drop contexts
    */
   withDragHandle?: boolean;
+  /**
+   * Props to pass down to the default CardDragHandle component (does not work with cardDragHandleComponent prop)
+   */
+  cardDragHandleProps?: Partial<CardDragHandlePropTypes>;
   /**
    * Applies styling for when the component is actively being dragged by the user
    */
   isDragActive?: boolean;
-  /**
-   * Props to pass down to the CardDragHandle component
-   */
-  cardDragHandleProps?: Partial<CardDragHandlePropTypes>;
 } & typeof defaultProps;
 
 const defaultProps = {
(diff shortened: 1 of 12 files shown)


packages/forma-36-react-components/src/components/Card/AssetCard/AssetCard.tsx
```diff
@@ -47,17 +44,13 @@ export type AssetCardProps = {
    */
   type?: AssetType;
   /**
-   * Renders a drag handle for the component for use in drag and drop contexts
+   * Prop to pass a CardDragHandle component to for use in drag and drop contexts
    */
-  withDragHandle?: boolean;
+  cardDragHandleComponent?: React.ReactNode;
```
suevalov: Isn't it a breaking change?

m10l: Yeah sadly, it's flagged in the commit message so should be picked up in release notes / version bump.

suevalov: I'd rather leave `withDragHandle` method and add ability to use custom component when needed. By default, `CardDragHandle` should be used.

packages/forma-36-react-components/src/components/Card/AssetCard/AssetCard.tsx
```diff
@@ -134,14 +126,7 @@ export class AssetCard extends Component<AssetCardProps> {
...
           <React.Fragment>
-            {withDragHandle && (
-              <CardDragHandle
-                isDragActive={isDragActive}
-                {...cardDragHandleProps}
-              >
-                Reorder entry
-              </CardDragHandle>
-            )}
+            {cardDragHandleComponent && cardDragHandleComponent}
```
suevalov: Just `{cardDragHandleComponent}` should be sufficient.

packages/forma-36-react-components/src/components/Card/AssetCard/AssetCard.tsx
m10l: @suevalov I like this approach as it'll save us from making a breaking change, however, the drag handles are currently pretty useless without this change and the `cardDragHandleComponent` would be the prop used in pretty much all cases.

packages/forma-36-react-components/src/components/Card/AssetCard/AssetCard.tsx
m10l: @suevalov nice, updated

packages/forma-36-react-components/src/components/Card/AssetCard/AssetCard.tsx
m10l: Just to clarify to anyone reading - went with Alex's solution, we use both props, so no breaking changes

PR #227, opened 2019-06-12
docs(AlphaNote): Alpha note on home splash added

wichniowski:
This PR will add the an alpha note to the documentation website:

image

# Purpose of PR

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/packages/forma-36-website/src/components/HomeSplash.js
+++ b/packages/forma-36-website/src/components/HomeSplash.js
@@ -1,9 +1,17 @@
 import React from 'react';
-import { DisplayText, Paragraph } from '@contentful/forma-36-react-components';
+import {
+  DisplayText,
+  Paragraph,
+  Note,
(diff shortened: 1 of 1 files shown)


joaoramos:
@m10l has some good arguments. If the intention is to manage expectations while gathering contributions, maybe there's something we can do with less prominence and less cheesy copy.

Screenshot 2019-06-13 at 11 33 49

I'd still argue if we should show anything at all though. As @m10l said, we should make people trust us. If we cool about being _in progress_, that will end up influencing the perception of others too.

wichniowski:
@joaoramos @m10l 
> It should be expected that design systems change and evolve. I think we should focus more on communicating these updates rather than having a disclaimer that things are in flux.

good point, thinking about it again I guess the message is kind of redundant. I'll close this PR thanks for your feedback!

PR #234, opened 2019-07-08
feat(DropDown): Add dropdownContainerClassName

wichniowski:
This PR will add `dropdownContainerClassName` prop to the dropdown for more flexibility overwriting portal container styles. (e.g override zIndex when used in modals)

### DIFF
--- a/packages/forma-36-react-components/src/components/Dropdown/Dropdown/Dropdown.tsx
+++ b/packages/forma-36-react-components/src/components/Dropdown/Dropdown/Dropdown.tsx
@@ -22,7 +22,6 @@ export type DropdownProps = {
   onClose?: Function;
   testId?: string;
   getContainerRef?: (ref: HTMLElement | null) => void;
-  dropdownContainerClassName?: string;
   className?: string;
   children: React.ReactNode;
 } & typeof defaultProps;
@@ -160,7 +159,6 @@ export class Dropdown extends Component<DropdownProps, DropdownState> {
       getContainerRef,
       children,
       isOpen,
-      dropdownContainerClassName,
       ...otherProps
     } = this.props;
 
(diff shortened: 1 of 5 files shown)


packages/forma-36-react-components/src/components/Dropdown/Dropdown/Dropdown.tsx
```diff
@@ -197,6 +199,7 @@ export class Dropdown extends Component<DropdownProps, DropdownState> {
         {toggleElement}
         {this.state.isOpen && (
           <DropdownContainer
+            className={dropdownContainerClassName}
```
sbezludny: Do we also need to pass className to the DropdownContainer on [L176](https://github.com/contentful/forma-36/pull/234/files#diff-33e7e56e26c1eb1c22349a4e4ab6ce9fR176)?

wichniowski:
@sbezludny i also added a prop to get the dropdownContainer ref which is pretty hand to check if clicks happened inside the dropdown

packages/forma-36-react-components/src/components/Dropdown/Dropdown/Dropdown.tsx
```diff
@@ -174,6 +177,8 @@ export class Dropdown extends Component<DropdownProps, DropdownState> {
           <DropdownContainer
             anchorDimensionsAndPositon={this.state.anchorDimensionsAndPositon}
             position={this.props.position}
+            className={`${className}__container`}
```
sbezludny: that's a bit _unconventional_ I think, `className` might be smth like `f36-margin-top--m f36-margin-bottom--m`, right? do we really need it after exposing `ref`

wichniowski:
@sbezludny good point i'll change it back to containerClassName. I still think we need it since there are several cases where you still want to add custom styles to the portal container.

packages/forma-36-react-components/src/components/Dropdown/Dropdown/Dropdown.tsx
m10l: Totally didn't think about that case - good point.

PR #245, opened 2019-07-12
chore: Added alpha workbench components

suevalov:
# Purpose of PR

Alpha version of `Workbench` component based on https://f36.contentful.com/guidelines/layout/workbench/ specification.

* `Workbench` component is not exposed to a main package and can be imported separately using `@contentful/f36-react-components/dist/alpha`
* Styles for Workbench components are located in the main `styles.css` but they are not used by any other components.
* Stories for `Workbench` components are separated from stable components.

I tested it out in several views in our WebApp, but please be critical about my CSS skills 😄 

## PR Checklist

- [x] Tests are passing


packages/forma-36-react-components/src/components/Workbench/Workbench.tsx
```diff
@@ -0,0 +1,135 @@
...
+    >
+      {props.children}
+    </div>
+  );
+}
+
+interface WorkbenchContentProps {
+  children: React.ReactNode;
+  type: 'default' | 'text' | 'full';
+  testId: string;
```
sbezludny:
```suggestion
  testId?: string;
```

packages/forma-36-react-components/src/components/Workbench/Workbench.tsx
```diff
@@ -0,0 +1,135 @@
...
+  className?: string;
+  /**
+   * Child nodes to be rendered in the component
+   */
+  children?: React.ReactNode;
+  /**
+   * An ID used for testing purposes applied as a data attribute (data-test-id)
+   */
+  testId?: string;
+} & typeof defaultProps;
```
sbezludny: `& typeof defaultProps;` is harmful in this case (actually we have it across everywhere and have to change) as it makes testId required

```suggestion
}
```

`defaultProps` on the other hand, must implement Partial<WorkbenchProps>

```ts
const defaultProps: Partial<WorkbenchProps> = {
  testId: 'cf-ui-workbench',
};

```

packages/forma-36-react-components/src/components/Workbench/Workbench.css
```diff
@@ -0,0 +1,99 @@
...
+.Workbench__sidebar {
+  padding: var(--spacing-l);
+  text-align: left;
+  overflow-x: hidden;
+  overflow-y: auto;
+  z-index: 0;
+  background-color: var(--color-element-lightest);
+}
+
+.Workbench__sidebar--left {
```
guilebarbosa: Wouldn't it be better if the left and right sidebars were separate components which you can choose to use? To make things really customisable we could let people opt to use something that doesn't have this width constraint.

packages/forma-36-react-components/src/components/Workbench/Workbench.tsx
```diff
@@ -0,0 +1,135 @@
...
+    </div>
+  );
+}
+
+interface WorkbenchContentProps {
+  children: React.ReactNode;
+  type: 'default' | 'text' | 'full';
+  testId: string;
+  leftSidebar?: React.ReactNode;
+  rightSidebar?: React.ReactNode;
```
guilebarbosa: Did you consider making those not render props but rather just the `<WorkbenchSidebar>` component that can be used together with `<Workbench.Content>` You are then free to use your own styles. It should respect the grid/flexbox  regardless.

```
<Workbench>
  <WorkbenchHeader />
  <WorkbenchSidebar position="left" />
  <WorkbenchContent />
  <WorkbenchSidebar><MyCustomHugeSlideInSidebar /></WorkbenchSidebar>
</Workbench>
```
It would also be possible to have multiple sidebars on one side i.e. Jira's multi-level side nav. 

This would require the `Workbench--wrapper` to be outside of `<WorkbenchContent>`

packages/forma-36-react-components/src/components/Workbench/Workbench.tsx
suevalov: @sbezludny Hmm, I remember there were some problem with the suggested approach. Does it make `testId` required? We're using TS with Forma for several extensions and there are no problems. 

I'd rather keep it consistent for the time being and replace all at once if you think it is needed.

packages/forma-36-react-components/src/components/Workbench/Workbench.css
suevalov: Yeah, probably it makes sense to give possibility to change predefined width. @wichniowski wdyt?

packages/forma-36-react-components/src/components/Workbench/Workbench.tsx
suevalov: @guilebarbosa In this case implementation would be something like.

1. Find `WorbenchHeader` and `rest` components in `children`
2. Wrap `rest` (all sidebars and contents) with  a wrapper `div`.

right?

suevalov: I'm 👍 for making this changes if @wichniowski and @m10l  are agree.

m10l: Sounds good to me!

packages/forma-36-react-components/src/components/Workbench/Workbench.css
m10l: I guess it's actually required to make the sidebar a separate component seeing as they already use different widths? I'm a little concerned this could add more inconsistency though.

guilebarbosa: In my idea, you would always have to use the `<WorkbenchSidebar>` component, even for custom ones, but not necessarily the render props for left and right.

You can still opt in for the predefined widths by _optionally_ setting the `position` prop (which could maybe be renamed to smth like `sidebarType`)

PR #270, opened 2019-07-30
fix(Tooltip): Keep `className` prop in `otherProps`

MarkusLaut:
Otherwise it won't be added to the `TooltipContainer` and `className` won't have an effect.


wichniowski:
in this case i'd add something like containerClassName because you don't always want the container to have the same className as the tooltip itself.

MarkusLaut:
@wichniowski 
That's fair, maybe I don't completely understand the code.
I'd expect the `className` prop to end up on the top level element (a `span`), but I don't see it anywhere.

MarkusLaut:
Screenshot from 2019-07-30 17-31-28

MarkusLaut:
Looks like `targetWrapperClassName` was what I was looking for :sweat_smile:

PR #272, opened 2019-08-02
Feat/add adjust width to toggle

wichniowski:
This PR adds a prop to adjust the width of the dropdown to the width of the toggle element.


packages/forma-36-react-components/src/components/Dropdown/Dropdown/Dropdown.stories.tsx
```diff
@@ -16,6 +16,7 @@ function DefaultStory() {
     <Dropdown
       isOpen={isOpen}
       onClose={() => setOpen(false)}
+      adjustWidthToToggleElement={boolean('adjustWidthToToggleElement', false)}
```
m10l: Hmm. How about `isFullWidth`?

PR #276, opened 2019-08-06
fix(DropdownListItem): Propagate isDisabled to interactive element

wichniowski:
Propagate isDisabled to interactive element


packages/forma-36-react-components/src/components/Dropdown/DropdownListItem/__snapshots__/DropdownListItem.test.tsx.snap
```diff
@@ -41,6 +41,7 @@ exports[`renders the component with a href 1`] = `
   <a
     className="DropdownListItem__button"
     data-test-id="cf-ui-dropdown-list-item-button"
+    disabled={false}
```
m10l: The disabled attribute won't work on a link - https://stackoverflow.com/questions/13955667/disabled-href-tag/13955695

PR #284, opened 2019-08-21
fix(Tabs): Add withDivider property

wichniowski:
This PR adds a withDivider property to the Tabs component. if this is true a border will be shown underneath the tabs panel. 

image

### DIFF
--- a/packages/forma-36-react-components/src/components/Tabs/Tabs.css
+++ b/packages/forma-36-react-components/src/components/Tabs/Tabs.css
@@ -30,6 +30,10 @@
   margin-right: 0;
 }
 
+.Tabs__with-divider {
+  box-shadow: inset 0 -3px 0px 0px var(--color-element-mid);
+}
+
 .Tab__selected {
   font-weight: var(--font-weight-medium);
 }
--- a/packages/forma-36-react-components/src/components/Tabs/Tabs.stories.tsx
+++ b/packages/forma-36-react-components/src/components/Tabs/Tabs.stories.tsx
@@ -1,7 +1,7 @@
 import React, { useState } from 'react';
 import { storiesOf } from '@storybook/react';
 import { action } from '@storybook/addon-actions';
-import { text } from '@storybook/addon-knobs';
+import { text, boolean } from '@storybook/addon-knobs';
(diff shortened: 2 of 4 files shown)


packages/forma-36-react-components/src/components/Tabs/Tabs.css
```diff
@@ -5,6 +5,7 @@
 
 .Tabs {
   display: flex;
+  padding: 0 var(--spacing-l);
```
wichniowski: @m10l @joaoramos  I added some extra padding left and right to visually balance the tab panel but I'm not sure if it is needed. Any thoughts on that?

packages/forma-36-react-components/src/components/Tabs/Tabs.css
```diff
@@ -30,6 +30,10 @@
   margin-right: 0;
 }
 
+.Tabs__with-divider {
+  border-bottom: 1px solid var(--color-element-mid);
```
m10l: The following looks quite nice as an alternative, but it's a bigger change and maybe doesn't match anything else in the UI :D

```
box-shadow: 0 -3px 0 inset var(--color-element-mid);
```

Screenshot 2019-08-21 at 16 07 40

packages/forma-36-react-components/src/components/Tabs/Tabs.stories.tsx
```diff
@@ -12,7 +12,10 @@ function DefaultStory() {
 
   return (
     <div>
-      <Tabs className={text('className', '')}>
+      <Tabs
+        className={text('className', '')}
+        withDivider={boolean('withDivider', false)}
```
m10l: Nitpick but would be nice to add this to the second Tabs story too

packages/forma-36-react-components/src/components/Tabs/Tabs.css
wichniowski: ahh that is what you meant that looks nice will do so!

wichniowski: @m10l done

PR #302, opened 2019-08-30
Fix TextLinks in Notes

m10l:
# Purpose of PR

This PR does two things:

1. Ensures that `TextLinks` are rendered using the appropriate intent. Occasionally there are instances where the CSS from the `Note` component is not overriding links properly. This should prevent that from happening, however, perhaps we don't want to do this change in general.
2. Adds a warning `linkType` to the `TextLink` component

I'd like to suggest removing the intent-based styling for TextLinks and instead use `secondary` TextLinks so that there is adequate contrast for TextLinks - please discuss in the comments!

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Note/Note.css
+++ b/packages/forma-36-react-components/src/components/Note/Note.css
@@ -21,7 +21,7 @@
   border: 1px solid var(--color-ice-dark);
 }
 .Note--primary a {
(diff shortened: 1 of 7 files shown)


m10l:
I'm going to close this PR for now as it's stale. If we notice issues we can open again

PR #309, opened 2019-09-16
fix(EntityListItem): Fix ct overflow

wichniowski:
# Purpose of PR
This PR fixes the overflow of the content type. It'll hide the content type if the title is to long. The title will have a content type tooltip using the title attribute.

## PR Checklist

- [ ] Tests are passing


m10l:
Having a title attribute on an element which doesn't relate to the content of that element seems like a bad move. You could always add a title element to the React component which has both the title and content type in it. This wouldn't require any changes to the component.

It also feels a bit weird to hide it - I'm wondering if a max-width on the content type element might be a better option.

wichniowski:
@m10l ok good point we could use flex tor reserve space for the 2 elements.

PR #314, opened 2019-09-18
TextInput/Textarea - ESC to blur field

m10l:
# Purpose of PR

This PR adds the ability to blur the focus of a TextInput or Textarea by pressing the ESC key.

## PR Checklist

- [x] Tests are passing


packages/forma-36-react-components/src/components/Textarea/Textarea.tsx
```diff
@@ -45,6 +55,18 @@ export class Textarea extends Component<TextareaProps, TextareaState> {
...
   }
 
+  handleKeyDown = (e: KeyboardEvent<HTMLTextAreaElement>) => {
+    const ESC = 27;
+
+    if (this.props.onKeyDown) {
+      this.props.onKeyDown(e);
+    }
+
+    if (e.keyCode === ESC) {
```
wichniowski: @m10l maybe we should add a prop blurOnEscape to opt out of this. Maybe this is not desirable in some cases, wdyt?

m10l: @wichniowski Seems like a good idea, I think this should be true by default to ease migration as this is a convention we want to push

PR #323, opened 2019-10-18
feat(EntryCard): Add scheduled status icon

mohamedturki:
# Purpose of PR

This adds an icon to EntryCard component that indicates whether or not the entry is scheduled for publish/unpublish/<supported actions>

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Card/EntryCard/EntryCard.tsx
+++ b/packages/forma-36-react-components/src/components/Card/EntryCard/EntryCard.tsx
@@ -8,6 +8,7 @@ import EntryCardSkeleton from './EntryCardSkeleton';
 import CardDragHandle, {
   CardDragHandlePropTypes,
 } from '../CardDragHandle/CardDragHandle';
+import Icon, { IconType } from '../../Icon/Icon';
 
 const styles = require('./EntryCard.css');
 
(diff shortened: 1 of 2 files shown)


packages/forma-36-react-components/src/components/Card/EntryCard/EntryCard.tsx
```diff
@@ -34,6 +35,10 @@ export type EntryCardPropTypes = {
    * The publish status of the entry
    */
   status?: EntryCardStatus;
+  /**
+   * The entry has a scheduled action
+   */
+  isScheduled?: boolean;
```
m10l: @mohamedturki I think it might be better to have a more generic `statusIcon` prop which accepts a string representing the desired icon.

packages/forma-36-react-components/yarn.lock
```diff
@@ -1433,22 +1433,10 @@
     puppeteer "^1.2.0"
     yargs "^11.1.0"
 
-"@contentful/forma-36-fcss@^0.0.31":
```
m10l: @mohamedturki Please revert the changes in `yarn.lock` - I believe this happens if you've ran `yarn` from the components package directory rather than the root of the monorepo

packages/forma-36-react-components/src/components/Card/EntryCard/EntryCard.tsx
mohamedturki: Yeah I like the idea. I will change it 👍

packages/forma-36-react-components/yarn.lock
mohamedturki: Roger tha'

PR #363, opened 2019-12-20
Add plop templates for react component packages

wichniowski:
This PR will add a plop template to add new React components as packages:

`$ yarn add-react-component`

In detail it will:

- Create a new package using tsdx
- install additional dependencies like emotion, jest-axe and enzyme in that package
- Create component, index and test files in that package
- Lerna add the local package to forma-36-react-components
- Add a new component in the forma-36-react-components that imports the created Component package
- Create an index and story file for that new component


wichniowski:
closed since we're not going with the single package approach

Issue #365, opened 2019-12-25
`TextInput` container sizing is bigger than intended


thebesson:
<img width="247" alt="Screen Shot 2019-12-25 at 12 20 43" src="https://user-images.githubusercontent.com/12238477/71444545-64aa5300-2712-11ea-9fe0-b4db11b66f83.png">
<img width="274" alt="Screen Shot 2019-12-25 at 12 20 30" src="https://user-images.githubusercontent.com/12238477/71444546-64aa5300-2712-11ea-8fb1-c7bb2d6b41e9.png">
The container of `input` element is bigger than the element itself. This creates a `white space` between `TextInput` and the next component when they displayed in a row.

<img width="441" alt="Screen Shot 2019-12-25 at 12 48 14" src="https://user-images.githubusercontent.com/12238477/71444848-dbe0e680-2714-11ea-9a4a-0a4e421e96f2.png">
<img width="442" alt="Screen Shot 2019-12-25 at 12 48 04" src="https://user-images.githubusercontent.com/12238477/71444849-dbe0e680-2714-11ea-93b1-e1b555437549.png">
Simmilar problem is with `TextField`. The label stays default width, even if input size is `small` or `medium`. This also creates problem when we whant to display components in one row.

wichniowski:
Hey @thebesson thanks for bringing this up! Is there a chance that other custom styles are being applied to the Input? In storybook, the sizing seems to work fine.

thebesson:
thanks, @wichniowski, I'll check!

wichniowski:
@thebesson yep you're right we need to add this prop to the TextField also. I'll open a PR asap!

wichniowski:
Hey @thebesson here we go :) 
https://github.com/contentful/forma-36/pull/367

wichniowski:
closed with #367

PR #367, opened 2019-12-27
fix(TextField): Apply width to container not just input

wichniowski:
This PR will fix an issue with the Textfield to apply the width to the container not just the input.
Related Issue: https://github.com/contentful/forma-36/issues/365


wichniowski:
@m10l didn't update the stories properly should be fixed now!

PR #371, opened 2019-12-30
feat(autocomplete): port Autocomplete component from user_interface

arpl:
# Purpose of PR
Port Autocomplete component from user_interface


packages/components/forma-36-react-autocomplete/README.md
```diff
@@ -0,0 +1,3 @@
+# Forma 36 React Autocomplete component (Alpha)
+
+Autocomplete component based on Forma-36 Dropdown and TextInput.
```
m10l:
```suggestion
Autocomplete component based on Forma 36 Dropdown and TextInput.
```

arpl: done =)

packages/forma-36-react-components/.size-limit
```diff
@@ -3,7 +3,7 @@
       "path": "dist/index.js",
       "name": "gzipped js",
       "gzip": false,
-      "limit": "282 KB",
+      "limit": "289 KB",
```
arpl: I guess this increase is caused by `is-hotkey`

arpl:
@wichniowski , @m10l 

Developing with a cyclic dependency to `forma-36-react-components` caused various issues so, I proceeded with `yarn generate`, more specifically
* No straightforward way to link `Autocomplete` to `Dropdown` for debugging purposes
* No straightforward way to add a story from a package
* CI build was failing because Netlify exceeded some default build timeout (18m)
* The component can be packaged after we package(pull out) all of its dependencies out of `forma-36-react-components`

packages/forma-36-react-components/package.json
```diff
@@ -31,10 +31,13 @@
     "@storybook/addon-notes": "^5.2.1",
     "@storybook/addon-storysource": "^5.2.1",
     "@storybook/react": "^5.2.1",
+    "@testing-library/jest-dom": "^4.2.4",
+    "@testing-library/react": "^9.4.0",
     "@types/classnames": "2.2.7",
     "@types/enzyme": "3.9.1",
-    "@types/jest": "24.0.11",
-    "@types/jest-axe": "2.2.4",
+    "@types/is-hotkey": "^0.1.1",
```
arpl: `@types` version bump is necessary because `@testing-library/jest-dom` included newer `jest` type augmentations.

packages/forma-36-react-components/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,325 @@
...
+        query: action.payload,
+      };
+    case ITEM_SELECTED:
+      return { ...initialState };
+    default:
+      return state;
+  }
+};
+
+export const Autocomplete: FunctionComponent<AutocompleteProps> = ({
```
arpl: A named `export` and referencing `FunctionComponent` through it's named export, as opposed to `React.FunctionComponent` is necessary for `react-docgen-typescript-loader` to generate docs.

packages/forma-36-react-components/src/utils/isBrowser.ts
```diff
@@ -1,3 +1,3 @@
-const isBrowser = () => typeof window !== 'undefined';
+const isBrowser = typeof window !== 'undefined';
```
arpl: This was caught by the latest TS compiler which complains when it detects a function as the condition in `if (condition)`. We were using `isBrowser` without calling it as a function.

packages/forma-36-react-components/.size-limit
m10l: @arpl How do you feel about stripping it out? This is the only component that uses it and it should be pretty simple to replace, but I also see the convenience of having it if we start using it elsewhere.

arpl: @m10l Scope of this PR is to port the component from user-interface. I agree it's beneficial to keep the bundle small let me know if this is a hard dependency for a merge.

m10l: @arpl Ah okay, no worries then - something we can do after merging

packages/forma-36-react-components/src/components/Autocomplete/Autocomplete.css
```diff
@@ -0,0 +1,21 @@
...
+  -webkit-appearance: none;
+}
+
+.autocompleteInput input::-ms-clear {
+  display: none;
+}
+
+.inputIconButton {
+  position: absolute;
+  top: calc(50% - 9px);
```
m10l: I think this would work and means we're using our tokens for spacing

```suggestion
  top: calc(50% - var(--spacing-xs));
```

packages/forma-36-react-components/src/components/Autocomplete/Autocomplete.css
```diff
@@ -0,0 +1,21 @@
...
+  right: 5px;
```
m10l:
```suggestion
  right: var(--spacing-xs);
  margin: 0;
```

PR #378, opened 2020-01-07
improvement(autocomplete): allow underlying dropdown configuration

arpl:
Autocomplete client code should be able to reach/configure the underlying dropdown

# Purpose of PR: 
Reach Dropdown API parity by adding namespace `dropdownProps` which is responsible for configuring the underlying Dropdown.

### DIFF
--- a/packages/forma-36-react-components/src/components/Autocomplete/Autocomplete.tsx
+++ b/packages/forma-36-react-components/src/components/Autocomplete/Autocomplete.tsx
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import isHotKey from 'is-hotkey';
 
 import TextInput from '../TextInput';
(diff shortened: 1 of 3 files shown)


packages/forma-36-react-components/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -251,7 +251,7 @@ Autocomplete.propTypes = {
   emptyListMessage: PropTypes.string,
   noMatchesMessage: PropTypes.string,
   willClearQueryOnClose: PropTypes.bool,
-  isFullWidth: PropTypes.bool,
+  dropdownProps: PropTypes.any,
```
arpl: `{ bar: PropTypes.object }` isn't type compatible to `type fooProps = { bar?: barProps }`

PR #384, opened 2020-01-13
fix(cardactions): fix testId propagation

arpl:
# Purpose of PR
Fix `testId` propagation in `CardActions` component

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Card/CardActions/CardActions.tsx
+++ b/packages/forma-36-react-components/src/components/Card/CardActions/CardActions.tsx
@@ -71,6 +71,7 @@ export class CardActions extends Component<
         position="bottom-right"
         className={className}
         isOpen={this.state.isDropdownOpen}
+        testId={testId}
         toggleElement={
           <IconButton
             iconProps={{ icon: 'MoreHorizontal' }}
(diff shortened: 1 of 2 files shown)


packages/forma-36-react-components/src/components/Card/CardActions/CardActions.tsx
```diff
@@ -55,7 +55,6 @@ export class CardActions extends Component<
     const {
       className,
       children,
-      testId,
```
m10l: @arpl I don't think this should be removed?

m10l: Instead, we should explicitly pass it to the `Dropdown` component below.

arpl: Since `CardAction.testId` matches `Dropdown.testId` and it felt 'too' explicit ;)

arpl:
```
    const {
      className,
      children,
      testId,
      iconButtonProps,
      isDisabled,
      ...otherProps
    } = this.props;

    return (
      <Dropdown
        onClose={() => {
          this.setState({
            isDropdownOpen: false,
          });
        }}
        position="bottom-right"
        className={className}
        isOpen={this.state.isDropdownOpen}
        testId={testId}
        toggleElement={
          <IconButton
            iconProps={{ icon: 'MoreHorizontal' }}
            buttonType="secondary"
            disabled={isDisabled}
            label="Actions"
            {...iconButtonProps}
            onClick={event => {
              event.preventDefault();
              this.handleClick(event);
            }}
          />
        }
        {...otherProps}
      >

```

PR #385, opened 2020-01-15
fix(cardactions): close cardactions on blur

arpl:
## Purpose
Prevent multiple `CardActions` from being simultaneously open when navigating through the keyboard

## PR Checklist

- [x] Tests are passing


arpl:
Needs more thinking

arpl:
`onMouseDown` will cause a blur event which we suppress by calling `preventDefault`. Then, `onClick` will get a chance and only after `onClick has run to completion` a blur event will still happen.

PR #398, opened 2020-01-21
fix(Asset): Do not attempt to render images with a status of archived

freakyfelt:
# Purpose of PR

Check the status as part of rendering an `Asset`.
  * If the status is present and is archived short-circuit to render as an asset
  * Else if the type is an image attempt to render a preview using an `img` tag

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Asset/Asset.tsx
+++ b/packages/forma-36-react-components/src/components/Asset/Asset.tsx
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import cn from 'classnames';
 import Illustration, { IllustrationType } from '../Illustration/Illustration';
+import { AssetState } from '../Card/AssetCard/AssetCard';
 
 const styles = require('./Asset.css');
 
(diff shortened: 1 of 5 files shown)


packages/forma-36-react-components/src/components/Asset/Asset.tsx
```diff
@@ -25,6 +26,7 @@ export type AssetProps = {
   title: string;
   type?: AssetType;
   className?: string;
+  status?: AssetState;
```
wichniowski: I don't think we need an additional prop we can just default to render Asset if no src is present

wichniowski: or even better show a missing image icon

freakyfelt: > I don't think we need an additional prop we can just default to render Asset if no src is present

From yesterday's meeting it sounds like the src is still there, but when you request it the asset may or may not be accessible anymore. This change sets a consistent behaviour in that instance

freakyfelt: Updated the PR title to be clearer

PR #399, opened 2020-01-23
chore(cleanup): drop proptypes usage

arpl:
# Purpose of PR
Drop PropTypes dependency since we don't support them. Update all `yarn.lock`s after issuing `lerna clean && lerna bootstrap` from project root to keep them aligned.

### DIFF
--- a/packages/components/forma-36-react-datepicker/yarn.lock
+++ b/packages/components/forma-36-react-datepicker/yarn.lock
@@ -670,17 +670,17 @@
     core-js-pure "^3.0.0"
     regenerator-runtime "^0.13.2"
 
-"@babel/runtime@^7.1.2", "@babel/runtime@^7.3.4", "@babel/runtime@^7.4.5", "@babel/runtime@^7.6.3", "@babel/runtime@^7.7.2", "@babel/runtime@^7.7.4":
-  version "7.7.6"
-  resolved "https://registry.yarnpkg.com/@babel/runtime/-/runtime-7.7.6.tgz#d18c511121aff1b4f2cd1d452f1bac9601dd830f"
-  integrity sha512-BWAJxpNVa0QlE5gZdWjSxXtemZyZ9RmrmVozxt3NUXeZhVIJ5ANyqmMc0JDrivBZyxUuQvFxlvH4OWWOogGfUw==
-  dependencies:
-    regenerator-runtime "^0.13.2"
-
 "@babel/runtime@^7.4.2", "@babel/runtime@^7.5.1", "@babel/runtime@^7.6.0", "@babel/runtime@^7.6.2":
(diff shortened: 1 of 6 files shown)


packages/components/forma-36-react-datepicker/yarn.lock
```diff
@@ -734,20 +734,6 @@
     exec-sh "^0.3.2"
     minimist "^1.2.0"
 
-"@contentful/forma-36-react-components@^3.25.1":
```
m10l: I'm not sure if this should be removed

packages/components/forma-36-react-timepicker/yarn.lock
```diff
@@ -734,25 +734,6 @@
     exec-sh "^0.3.2"
     minimist "^1.2.0"
 
-"@contentful/forma-36-react-components@^3.25.1":
```
m10l: Same, not sure this should have been removed. Maybe try running from the root

packages/forma-36-react-components/yarn.lock
```diff
@@ -1405,22 +1405,10 @@
     puppeteer "^1.2.0"
     yargs "^11.1.0"
 
-"@contentful/forma-36-fcss@^0.0.34":
```
m10l: Same here

packages/forma-36-react-components/yarn.lock
```diff
@@ -1405,22 +1405,10 @@
...
-  resolved "https://registry.yarnpkg.com/@contentful/forma-36-fcss/-/forma-36-fcss-0.0.34.tgz#afc4514e9b946c67339202ac61df94dc702b3bbe"
-  integrity sha512-AhK2QJn2pQ/MWR8d10H4B6g625MU1rz4lkakkw8/0E0KFpeow53XW2lbN0vTdOXDaNM1TeMOP0F700qFIoeMkA==
-  dependencies:
-    "@contentful/forma-36-tokens" "^0.5.1"
-
 "@contentful/forma-36-tokens@0.2.0":
   version "0.2.0"
   resolved "https://registry.yarnpkg.com/@contentful/forma-36-tokens/-/forma-36-tokens-0.2.0.tgz#c4a1062e7acc2ae0ae5699e7c8606d5b5047a1c5"
 
-"@contentful/forma-36-tokens@^0.5.1":
```
m10l: And here

packages/components/forma-36-react-datepicker/yarn.lock
arpl: Running `lerna bootstrap` from root will link the locally found packages hence the difference between the `lock` files. We should commit these locks and we won't see differences when pulling and `lerna bootstrap`-ing.

PR #400, opened 2020-01-24
chore(cleanup): drop proptypes usage

arpl:
# Purpose of PR

Drop `PropTypes` dependency. Update `yarn.lock`s after issuing `lerna clean && lerna bootstrap` from the project's root to keep them aligned.

### DIFF
--- a/packages/forma-36-react-components/yarn.lock
+++ b/packages/forma-36-react-components/yarn.lock
@@ -1405,21 +1405,9 @@
     puppeteer "^1.2.0"
     yargs "^11.1.0"
 
-"@contentful/forma-36-fcss@^0.0.34":
-  version "0.0.34"
-  resolved "https://registry.yarnpkg.com/@contentful/forma-36-fcss/-/forma-36-fcss-0.0.34.tgz#afc4514e9b946c67339202ac61df94dc702b3bbe"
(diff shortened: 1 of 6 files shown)


packages/forma-36-react-components/yarn.lock
```diff
@@ -1405,22 +1405,10 @@
...
-  resolved "https://registry.yarnpkg.com/@contentful/forma-36-fcss/-/forma-36-fcss-0.0.34.tgz#afc4514e9b946c67339202ac61df94dc702b3bbe"
-  integrity sha512-AhK2QJn2pQ/MWR8d10H4B6g625MU1rz4lkakkw8/0E0KFpeow53XW2lbN0vTdOXDaNM1TeMOP0F700qFIoeMkA==
-  dependencies:
-    "@contentful/forma-36-tokens" "^0.5.1"
-
 "@contentful/forma-36-tokens@0.2.0":
   version "0.2.0"
   resolved "https://registry.yarnpkg.com/@contentful/forma-36-tokens/-/forma-36-tokens-0.2.0.tgz#c4a1062e7acc2ae0ae5699e7c8606d5b5047a1c5"
 
-"@contentful/forma-36-tokens@^0.5.1":
```
wichniowski: whats up with those lock files :D

arpl: Running lerna bootstrap from root will link the locally found packages hence the difference between the lock files. We should commit these locks and we won't see differences when pulling and lerna bootstrap-ing.

arpl:
Had to close #399 and open it as #400 to restart 'All Checks'

PR #417, opened 2020-02-14
fix(DropdownContainer): Recalculate dropdown position

freakyfelt:
# Purpose of PR

Check if the size of the dropdown list dimensions have changed and, if so, re-calculate the offset

## PR Checklist

- [x] Tests are passing


freakyfelt:
Exceeding a max KB limit of 289 KB which is breaking the build

freakyfelt:
Ah. Is there a reason https://github.com/contentful/forma-36/blob/master/packages/forma-36-react-components/.size-limit claims "gzipped js" but then looks at the non-gzipped dist/index.js?

PR #419, opened 2020-02-17
fix(SkeletonText): Allow string as proptype for percentage values

wichniowski:
This PR will add string as proptype of widths and other spacing props so we don't get warnings for percentage values.


packages/forma-36-react-components/src/components/Skeleton/SkeletonImage/SkeletonImage.tsx
```diff
@@ -1,12 +1,12 @@
 import React, { Component } from 'react';
 
 export type SkeletonImageProps = {
-  width: number;
-  height: number;
-  radiusX: number;
-  radiusY: number;
-  offsetLeft?: number;
-  offsetTop?: number;
+  width: number | string;
```
arpl: You can make a type `type T = number | string` and use it instead. Less typing.

PR #428, opened 2020-02-25
refactor(datepicker, timepicker): refactoring

thebesson:
Refactor date and time picker components

# Purpose of PR

A bit of refactoring to adopt the component in the future. WIP, please, add any feedback.

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/packages/components/forma-36-react-datepicker/src/Datepicker.tsx
+++ b/packages/components/forma-36-react-datepicker/src/Datepicker.tsx
@@ -1,7 +1,11 @@
 import React, { Component, FocusEventHandler, FocusEvent } from 'react';
 import Pikaday from 'pikaday';
-import * as dateFns from 'date-fns';
-import { TextField } from '@contentful/forma-36-react-components';
+import format from 'date-fns/format';
+import {
+  TextInput,
+  FormLabel,
+  ValidationMessage,
+} from '@contentful/forma-36-react-components';
 import { css, cx } from 'emotion';
 
 const styles = {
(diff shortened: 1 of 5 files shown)


packages/components/forma-36-react-timepicker/src/TimeZonePicker.tsx
```diff
@@ -0,0 +1,39 @@
...
+      aria-label="Select timezone"
+      testId="timezone-input"
+      value={value}
+      width="medium"
+      isDisabled={disabled}
+      onChange={(e: ChangeEvent<HTMLSelectElement>) => {
+        onChange(e.currentTarget.value);
+      }}
+    >
+      {zoneOffsets.map(offset => (
```
wichniowski: this might work for current times but not for future looking dates since you'll not be able to respect daylight saving time when scheduling something. E.G: Berlin and Cairo are on the same offset in summer but not in winter. If I'm not able to select the the TZ database name its not possible to schedule the correct future time for a certain timezone.

thebesson: Do you think it would be better to use some sort of framework for displaying timezones with daylight savings times? should this be a component at all?

wichniowski: I think we need to bring this up in the frontend/backend chapter how we want to store dates in the future. Using offsets will definitely result in wrong future looking dates thats why we chose to use the TZ Name for scheduled publishing. For now and to migrate it it should be fine to continue with the offsets.

PR #430, opened 2020-03-04
feat(Icon): Add cycle icon

wichniowski:
This PR will add a Cycle Icon that we'll use for circular references


m10l:
@wichniowski Build is failing due to going over the size-limit

PR #434, opened 2020-03-12
Allow Notifications without timeout

m10l:
# Purpose of PR

This PR allows users to set the duration of a Notification to be 0, resulting in a Notification which persists rather than dismissing after X seconds like normal.

## PR Checklist

- [x] Tests are passing


packages/forma-36-react-components/src/components/Notification/NotificationItemContainer.tsx
```diff
@@ -3,7 +3,7 @@ import AnimateHeight from 'react-animate-height';
 import NotificationItem, { NotificationItemProps } from './NotificationItem';
 
 export interface NotificationItemContainerProps extends NotificationItemProps {
-  duration: number;
+  duration: number | undefined;
```
m10l: Not sure if this is a terrible idea or not, but it got around TypeScript complaining

packages/forma-36-react-components/src/components/Notification/NotificationsManager.tsx
```diff
@@ -18,7 +18,7 @@ export interface Notification {
   id: string | number;
   text: string;
   close: Function;
-  duration: number;
+  duration: number | undefined;
```
m10l: Not sure if this is a terrible idea or not, but it got around TypeScript complaining

packages/forma-36-react-components/src/components/Notification/NotificationsManager.tsx
```diff
@@ -120,7 +120,9 @@ export class NotificationsManager extends PureComponent<
 
   show: ShowAction<Notification> = (text, settings) => {
     const duration =
-      settings && settings.duration ? settings.duration : this.state.duration;
+      settings && Number.isInteger(settings.duration)
```
m10l: Failing Typescript linting:

```
Argument of type 'number | undefined' is not assignable to parameter of type 'number'.
  Type 'undefined' is not assignable to type 'number'.

123       settings && Number.isInteger(settings.duration)
                                       ~~~~~~~~~~~~~~~~~
```

packages/forma-36-react-components/src/components/Notification/NotificationItemContainer.tsx
madtrick: @m10l what do you want to express here? that `duration` can be optional? In that case you can rewrite the interface like: 
```suggestion
  duration?: number;
```

packages/forma-36-react-components/src/components/Notification/NotificationsManager.tsx
madtrick: It's failing because the function `Number.isInteger` can only take a number as argument. Since you above said that `duration` can also be `undefined` the compiler complaints about passing it to `Number.isInteger`.  You'd have to do something like:

```typescript
// The ? after settings in the optional chaining operator
// http://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining 
const duration = Number.isInteger(settings?.duration || this.state.duration)

```

packages/forma-36-react-components/src/components/Notification/NotificationsManager.tsx
```diff
@@ -120,7 +120,9 @@ export class NotificationsManager extends PureComponent<
 
   show: ShowAction<Notification> = (text, settings) => {
     const duration =
-      settings && settings.duration ? settings.duration : this.state.duration;
+      settings && typeof settings.duration !== 'undefined' // Needed as 0 is falsy but 0 is valid to disable auto-closing a notification
```
madtrick: nit: maybe

```typescript
const duration = settings?.duration !== undefined ? settings.duration : this.state.duration
```

Issue #433, opened 2020-03-11
Table Component: showing loading state with the Skeleton Component


gui-santos:
![table_skeleton](https://user-images.githubusercontent.com/6597467/76405848-14de3e00-6389-11ea-8cbf-c17072d2dc5c.gif)

I would like to propose an extension to the Table Component (more specifically the TableBody)
The case is that, in some places, we fill tables with async data, right?
Well, one way of dealing with that would be, let's say, showing a spinner in the page and only after the data is fetched show the table.
A better pattern for the users would be mimicking the final look of the table after with the Skeleton Component that forma already have (take a look on the gif ⬆️ ).

Initially, my idea is that we could pass a bool prop to TableBody and while this is `true` we render a "SkeletonRow" instead of the TableBody's children. Something like:

```jsx
<TableBody loading={true}>
    <TableRow>
        // ... n TableCells here
    </TableRow>
</TableBody>
```
And in  /TableBody.js
```jsx
<tbody data-test-id={testId} className={className} {...otherProps}>
    {loading ? <SkeletonRow /> : children}
</tbody>
```

This would render a pretty standard skeleton row. In case the table needs a more personalised skeleton setup, you could do it without the prop by just setting the load yourself as a children (I guess... more cases will show up as this gets developed hahaha)

This is just an initial thought and I would love to know what you think about it.
And if we can already have some suggestions 😄

m10l:
Nice, looks like a good addition!

wichniowski:
@gui-santos can we close this issue?

gui-santos:
@wichniowski yes! I totally missed closing this

PR #435, opened 2020-03-13
Feat/loading state for table

gui-santos:
# Purpose of PR

This PR creates the SkeletonRow and SkeletonTableCell components

SkeletonRow props
| prop  | type | default | description
| ------------- | ------------- | ------------- | ------------- |
| rowCount  | number  | 1 | number of rows that will be rendered
| columnCount  | number  | 5 | number of columns that will be rendered

more details in the opened issue #433 

The way that this should be used is inside a TableBody component while fetching the table's data. Just render a SkeletonRow instead of the TableRows until the data is completely fetched

```ts
function MyTable() {
  const [isLoading, setIsLoading] = useState(true);

  return (
      <Table>
        <TableHead>
          <TableRow>
            <TableCell>Column 1</TableCell>
            <TableCell>Column 2</TableCell>
            <TableCell>Column 3</TableCell>
          </TableRow>
        </TableHead>
        <TableBody>
          {isLoading ? (
            <SkeletonRow
              rowCount={1}
              columnCount={3}
            />
          ) : (
            <>
              <TableRow>
                <TableCell>Row 1 Column 1</TableCell>
                <TableCell>Row 1 Column 2</TableCell>
                <TableCell>Row 1 Column 3</TableCell>
              </TableRow>
            </>
          )}
        </TableBody>
      </Table>
  );
}
```

## PR Checklist

- [x] Tests are passing


gui-santos:
I'm not quite happy with passing those two props `numberOfRows` and `numberOfCells`... One alternative that I thought is combine both in a prop called `loadingOptions` that would accept an object like:
 `{ rows: number, cells: number }`

Another thing is the name of the component... What is better `SkeletonTable` or `SkeletonRows`? or something else?

What do you think?

packages/forma-36-react-components/src/components/Table/TableBody/TableBody.tsx
```diff
@@ -1,9 +1,14 @@
 import React, { Component } from 'react';
 
+import SkeletonTable from '../../Skeleton/SkeletonTable';
+
 export type TableBodyProps = {
   className?: string;
   style?: React.CSSProperties;
   testId?: string;
+  loading?: boolean;
```
m10l: We've got some naming conventions we're trying to stick to with components - https://github.com/contentful/forma-36/tree/master/packages/forma-36-react-components#component-principles - so I recommend changing to the following:

```suggestion
  isLoading?: boolean;
```

packages/forma-36-react-components/src/components/Table/TableBody/TableBody.tsx
```diff
@@ -1,9 +1,14 @@
...
+  numberOfRows?: number;
```
m10l:
```suggestion
  loadingRowCount?: number;
```

packages/forma-36-react-components/src/components/Table/TableBody/TableBody.tsx
```diff
@@ -1,9 +1,14 @@
...
+  numberOfCells?: number;
```
m10l:
```suggestion
  loadingColumnCount?: number;
```

packages/forma-36-react-components/src/components/Table/TableBody/TableBody.tsx
m10l: Did you consider just sticking to a single row for this?

m10l:
> I'm not quite happy with passing those two props `numberOfRows` and `numberOfCells`... One alternative that I thought is combine both in a prop called `loadingOptions` that would accept an object like:
> `{ rows: number, cells: number }`
> 
> Another thing is the name of the component... What is better `SkeletonTable` or `SkeletonRows`? or something else?
> 
> What do you think?

Yeah it's a shame the columns can't be inferred by the children, I'm sure there's a way to do it but I'm presuming it'd be a bit flakey depending on which children are passed.

Naming is tricky too… Maybe SkeletonRow makes more sense if it's only rendering one row

packages/forma-36-react-components/src/components/Skeleton/SkeletonTable/SkeletonTableCell/SkeletonTableCell.tsx
```diff
@@ -0,0 +1,35 @@
...
+    const { clipId } = this.props;
+
+    return (
+      <TableCell>
+        <SkeletonContainer
+          svgHeight={42}
+          clipId={clipId}
+          gradientId={`${clipId}-gradient`}
+        >
+          <SkeletonBodyText numberOfLines={2} />
```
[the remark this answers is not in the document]
gui-santos: Cool! It's also good because a single line is the default look of SkeletonBodyText :)

Thanks!

packages/forma-36-react-components/src/components/Table/Table.stories.tsx
```diff
@@ -19,38 +93,5 @@ storiesOf('Components|Table', module)
...
-              <TableCell>July 27, 2019</TableCell>
-            </TableRow>
-          </TableBody>
-        </Table>
-      </div>
-    ),
-    { notes },
-  );
+  .add('default', () => <DefaultStory />, { notes })
+  .add('Asynchronous Data', () => <AsyncDataStory />, { notes });
```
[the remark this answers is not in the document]
gui-santos: nice! It makes sense... more descriptive

packages/forma-36-react-components/src/components/Skeleton/SkeletonTable/SkeletonTableCell/SkeletonTableCell.tsx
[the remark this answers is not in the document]
gui-santos: Screenshot 2020-03-17 at 09 44 34

Do you think we should centralize the line?

I don't want to decrease the cell height because this is the height that the cell will get with the data

packages/forma-36-react-components/src/components/Table/Table.stories.tsx
```diff
@@ -9,6 +9,80 @@ import TableBody from './TableBody';
...
+      <Table>
+        <TableHead>
+          <TableRow>
+            <TableCell>Name</TableCell>
+            <TableCell>Email</TableCell>
+            <TableCell>Organization role</TableCell>
+            <TableCell>Last activity</TableCell>
+          </TableRow>
+        </TableHead>
+        <TableBody loading={loading} numberOfRows={3} numberOfCells={4}>
```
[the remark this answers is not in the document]
gui-santos: Oh yeah sure! I forgot about the knobs
Cool! thanks

packages/forma-36-react-components/src/components/Table/TableBody/TableBody.tsx
gui-santos: Cool! Thanks for the link too 😄

packages/forma-36-react-components/src/components/Table/TableBody/TableBody.tsx
gui-santos: I started developing with a single row. But then I just experimented setting the number of rows to the limit of the pagination 

I think it's easier to explain with a real case.
Below I have a table with several users, and the pagination limit is 10.
With the loading state having one row and the final table having 10,  the table will jump from 1 to 10 rows very quickly

start with a single row...
Screenshot 2020-03-17 at 09 58 22

... ends with 10 rows
Screenshot 2020-03-17 at 09 55 47

But this is also true in the reverse case, anyways. The good thing about one row is that this prop becomes unnecessary and this means less code which means fewer bugs

I think I will change back to one then. Also, if someone needs more rows, it's possible to do it by passing more rows as children

packages/forma-36-react-components/src/components/Table/TableBody/TableBody.tsx
gui-santos: Makes sense! Make it specific to loading and not the number of cells of the table

thanks!

packages/forma-36-react-components/src/components/Skeleton/SkeletonRow/SkeletonRow.stories.tsx
```diff
@@ -0,0 +1,35 @@
...
+          <TableRow>
+            <TableCell>Name</TableCell>
+            <TableCell>Email</TableCell>
+            <TableCell>Organization role</TableCell>
+            <TableCell>Last activity</TableCell>
+            <TableCell>2FA status</TableCell>
+          </TableRow>
+        </TableHead>
+        <TableBody>
+          <SkeletonRow />
```
m10l: Be good to add the Storybook knobs here too

gui-santos:
@m10l In the case of having multiple text rows, should we create a prop to control this?

packages/forma-36-react-components/src/components/Skeleton/SkeletonRow/SkeletonTableCell/SkeletonTableCell.tsx
```diff
@@ -0,0 +1,35 @@
...
+export class SkeletonTableCell extends Component<SkeletonTableCellProps> {
+  static defaultProps = defaultProps;
+
+  render() {
+    const { clipId } = this.props;
+
+    return (
+      <TableCell>
+        <SkeletonContainer
+          svgHeight={42}
```
gui-santos: @m10l if I remove the svgHeight this is what happens

Screenshot 2020-03-19 at 17 50 24

The reason for this is that the `<rect>` in the svg gets the default height which is 100%

What do you think we should do?

PR #436, opened 2020-03-13
fix: (pill) fix overflow

dominicbonnice:
Added a fix for the overflow for the pill component, thanks @Spring3 !


suevalov:
I'm not sure if this change should be done here and not in the `field-editors` repo.

Spring3:
> I'm not sure if this change should be done here and not in the `field-editors` repo.

I don't see a Pill component there, so probably here

Spring3:
I mean it uses Pill from forma-36-react-components and then just slightly modifies the styles (like margins). But I guess that overflow is an actual bug in the component as it clearly has to handle long texts properly

PR #447, opened 2020-03-31
Update yarn lockfiles

m10l:
# Purpose of PR

Yarn lockfiles are looking a little out of sync, will test to see if there are any issues with the component library Storybook or the documentation website before merging.

## PR Checklist

- [x] I have read the relevant `readme.md` file(s)
- [x] All commits follow our [Git commit message convention](https://github.com/contentful/forma-36/tree/master/packages/forma-36-react-components#commits)
- [x] Tests are added/updated/not required
- [x] Tests are passing
- [x] Storybook stories are added/updated/not required
- [x] Usage notes are added/updated/not required
- [x] Has been tested based on [Contentful's browser support](https://www.contentful.com/faq/about-contentful/#which-browsers-does-contentful-support)
- [x] Doesn't contain any sensitive information

PR #448, opened 2020-04-01
Add search on Forma36's website

gui-santos:
Screenshot 2020-04-03 at 11 38 55

This PR implements a search input on Forma36's website. The search was implemented with Algolia's [DocSearch](https://docsearch.algolia.com/docs/what-is-docsearch) (@mshaaban0  contacted them to make this possible, thanks!)

To load docsearch scripts, I used [gatsby-plugin-algolia-docsearch](https://www.gatsbyjs.org/packages/gatsby-plugin-algolia-docsearch/). 
But because of [this issue](https://github.com/ironstar-io/gatsby-plugin-algolia-docsearch/issues/2), I also had to activate the plugin in the component using `useEffect`

**[ERRATA]**
I forgot to mention that to test this you need to create a `.env.development` like this:
```
DOCSEARCH_API_KEY={value of the key}
```
The API key is configured in Travis so you can get its value there

## PR Checklist

- [x] Tests are passing


mshaaban0:
Thanks @gui-santos 

Two points: 

1- there is a client side error caused by docSearch...

Screenshot 2020-04-02 at 11 44 12

2- If I can be bit picky about the design, it feels like we stacking all Navbar elements to the right, what do you think about these options...

search-navbar
search-sidebar

Looping @sarakalinoski @domarku, do you have thoughts?

gui-santos:
Hey, @mshaaban0 

Thanks for the comments!

> 1- there is a client side error caused by docSearch...

Wow! I will take a look into that. I thought I had fixed this before pushing the branch.

> 2- If I can be a bit picky about the design, it feels like we stacking all Navbar elements to the right, what do you think about these options...

I think I assumed that it was supposed to be in the right because of some design pattern.
I love the alternative of putting it in the sidebar. I think it will be present all the time and also will be next to the "subjects" of the search. Really cool!

One more thing, I didn't find an Input component that would accept icons, so I put the Icon outside. But I prefer the way you did, with the Icon inside the input.

Should we update the TextInput component work like that? Right now, you can only say if it has a copy button or not.

m10l:
Nice! I'll have a proper review later

+1 for search in the sidebar

mshaaban0:
@gui-santos I would just hack it for this case, and introduce the TextInput change in another PR 👍

gui-santos:
@mshaaban0 I could not reproduce the bug you found and I thought that maybe it's because you didn't pass the `DOCSEARCH_API_KEY` as an environment variable... Did you do that?

packages/forma-36-website/src/components/Docsearch.jsx
```diff
@@ -0,0 +1,55 @@
...
+import tokens from '@contentful/forma-36-tokens';
+import { TextInput, Icon } from '@contentful/forma-36-react-components';
+
+const styles = {
+  container: css`
+    position: relative;
+    display: flex;
+    align-items: center;
+    width: 100%;
+    margin-bottom: 0.5rem;
```
m10l, line 12: Please use a F36 token here

```suggestion
    margin-bottom: ${tokens.spacingXs};
```

packages/forma-36-website/src/components/Docsearch.jsx
```diff
@@ -0,0 +1,55 @@
...
+
+    & .algolia-autocomplete {
+      width: 100%;
+    }
+  `,
+  icon: css`
+    position: absolute;
+    right: 0px;
```
m10l, line 20: Nitpicking so feel free to ignore

```suggestion
    right: 0;
```

packages/forma-36-website/src/components/Docsearch.jsx
```diff
@@ -0,0 +1,55 @@
...
+    margin-right: ${tokens.spacingXs};
+    z-index: 1000;
+  `,
+};
+
+const Docsearch = () => {
+  useEffect(() => {
+    if (typeof window === 'undefined') return;
+
+    window.docsearch({
```
m10l, line 31: I think it'd be a good idea to log a message to the console if this is unavailable due to a missing API key or whatever.

packages/forma-36-website/src/components/Header.js
```diff
@@ -51,7 +51,7 @@ const Logo = () => (
     width="32px"
     height="32px"
     viewBox="0 0 90 90"
-    enable-background="new 0 0 90 90"
+    enableBackground="new 0 0 90 90"
```
m10l, line 54: Doh! Good spot

packages/forma-36-website/src/components/Docsearch.jsx
```diff
@@ -0,0 +1,55 @@
...
+        name="search"
+        type="text"
+        placeholder='Search the docs (Press "/" to focus)'
+        value=""
+      />
+    </div>
+  );
+};
+
+export default Docsearch;
```
m10l, line 55: Renaming to DocSearch would be good to keep naming consistency

packages/forma-36-website/src/components/Header.js
```diff
@@ -12,32 +12,32 @@ const styles = {
...
   logoLink: css`
     display: flex;
     text-decoration: none;
     color: #fff;
   `,
-  // TODO: Sort out vertical text alignment
   logoText: css`
     font-weight: ${tokens.fontWeightDemiBold};
     font-size: ${tokens.fontSizeXl};
+    line-height: 32px;
```
m10l, line 23: This scares me right now but I've not checked out why it's needed yet 😅

packages/forma-36-website/yarn.lock
```diff
@@ -778,6 +785,31 @@
     lodash "^4.17.13"
     to-fast-properties "^2.0.0"
 
+"@contentful/forma-36-fcss@^0.0.34":
```
m10l, line 788: What happens if you run `yarn` from the root of the repo? Wondering if these disappear from the lock file again (I removed them in https://github.com/contentful/forma-36/pull/447)

mshaaban0:
@gui-santos I was checking it on preview https://deploy-preview-448--forma-36.netlify.com maybe because we haven't configure it on Netlify yet?

packages/forma-36-website/src/components/Header.js
gui-santos: TBH, I found it following a warning in the console hahaha there some other ones, but we can fix in other PR I think 😄

packages/forma-36-website/src/components/Header.js
gui-santos: Hmmm maybe I should change this to flex

Basically there was a TODO saying something like "figure out a way to center the logo"
In this line, I put a line-height of 'Forma 36' text equal to the svg height. But I will change to flex and then we don't have magic numbers. It will be future proof too

packages/forma-36-website/yarn.lock
gui-santos: I will try it again!

gui-santos:
@mshaaban0 ohhhhhhh man.... I only did it in Travis. I think I don't have access to our Netlify, can you give it to me?

packages/forma-36-website/src/components/Docsearch.jsx
gui-santos: I passed a wrong API_key to see what happens and this error is printed in the console. It comes from DocSearch's script

Screenshot 2020-04-03 at 11 16 05

I still added a console.error in case window.docsearch is undefined

packages/forma-36-website/yarn.lock
gui-santos: Well, I rebased and got the yarn.lock from master. Then I ran `yarn` from the root but the .lock it's still here. What am I missing here?

domarku:
> search-sidebar
> 

+1 for the search in the sidebar

PR #452, opened 2020-04-03
Remove warnings in Navigation about GatsbyLink

gui-santos:
Screenshot 2020-04-03 at 14 01 39

# Purpose of PR

While working on the search input I saw that we were getting the warnings above. In this PR I refactored the code in the Navigation component to remove the warnings and also to convert the class components to function components

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-website/src/components/Navigation.js
+++ b/packages/forma-36-website/src/components/Navigation.js
@@ -58,6 +58,7 @@ const styles = {
     position: relative;
     border-radius: 4px;
     padding: ${tokens.spacingXs};
+    cursor: pointer;
   `,
 
   linkIcon: css`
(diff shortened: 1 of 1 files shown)


packages/forma-36-website/src/components/Navigation.js
```diff
@@ -1,8 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import { Link } from 'gatsby';
-/** @jsx jsx */
```
m10l, line 4: Not sure why this isn't needed anymore or why it was added in the first place. I think it was required after upgrading something.

gui-santos: yeah I removed and then nothing broke so I kept the change hahaha

PR #458, opened 2020-04-17
feat(EntityStatusTag): Added Entity Status Tag component

wichniowski:
This PR will add the EntityStatusTag which simplifies color-coding the entity status.
image

### DIFF
--- a/packages/forma-36-react-components/src/components/Tag/Tag.stories.tsx
+++ b/packages/forma-36-react-components/src/components/Tag/Tag.stories.tsx
@@ -22,6 +22,17 @@ storiesOf('Components|Tag', module)
         },
         'primary',
       )}
+      entityStatusType={select(
+        'Entity Status Type',
+        {
+          Published: 'published',
+          Archived: 'archived',
+          draft: 'draft',
+          changed: 'changed',
+          'None (default)': undefined,
+        },
+        undefined,
+      )}
       className={text('className', '')}
     >
       {text('Children', 'Published')}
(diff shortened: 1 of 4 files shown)


packages/forma-36-react-components/src/components/Tag/EntityStatusTag/EntityStatusTag.stories.tsx
```diff
@@ -0,0 +1,25 @@
...
+  .addParameters({
+    propTypes: EntityStatusTag['__docgenInfo'],
+  })
+  .add('default', () => (
+    <EntityStatusTag
+      statusLabel={select(
+        'statusLabel',
+        {
+          'Published (default)': 'published',
+          Archived: 'archived',
```
mshaaban0, line 17:
```suggestion
          archived: 'archived',
```

wichniowski: @mshaaban0 ahh ja fantastisch wunderbar.. I like the entityStatusType.. will change ... :)

artibella: Would adding the `entityStatusType` prop maybe limit the reuse of the Tag component to only appear in the context of some entities that have a content lifecycle? Could I still use the component for non-content entities (e.g. showing the status of a user invite)?

mshaaban0: @artibella you should be able to use the Tag `tagType` prop as it is now `<Tag tagType='warning' />`. the entityStatusType is just an extra prop.

wichniowski: @artibella I added this as an extra prop so it is configurable like mo described it

artibella: @mshaaban0 @wichniowski Thanks for clarifying!

PR #459, opened 2020-04-20
fix(Tab): Add disabled state for tab

wichniowski:
This PR fixes the disabled state of the tab component that had no impact visually on the component.


packages/forma-36-react-components/src/components/Tabs/Tabs.css
```diff
@@ -37,6 +37,9 @@
 .Tab__selected {
   font-weight: var(--font-weight-medium);
 }
+.Tab__disabled {
+  opacity: 0.5;
```
m10l, line 41:
```suggestion
  opacity: 0.5;
  cursor: not-allowed;
```

PR #464, opened 2020-04-28
fix(Switch): Replace checked prop with defaultChecked

backflip:
# Purpose of PR

Remove the following React warning:
```
You provided a `checked` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultChecked`. Otherwise, set either `onChange` or `readOnly`
```

## PR Checklist

- [x] Tests are passing


m10l:
Hi @backflip! Apologies for the delay, huge thanks for your contribution. I just updated the branch, I'll wait for the build to be green and then I'll merge :)

PR #465, opened 2020-04-29
fix: type type maxHeight

marcolink:
# Purpose of PR
Removes this warning:
```
Warning: Failed prop type: Invalid prop `maxHeight` of type `string` supplied to `DropdownList`, expected `number`.
```

Prop `maxHeight` is currently defined as optional `number` but is used as css property.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Dropdown/DropdownList/DropdownList.tsx
+++ b/packages/forma-36-react-components/src/components/Dropdown/DropdownList/DropdownList.tsx
@@ -8,7 +8,7 @@ export type DropdownListProps = {
   className?: string;
   testId?: string;
   border?: 'top' | 'bottom';
-  maxHeight?: number;
+  maxHeight?: 'initial' | 'inherit' | string | number;
   styles?: object;
 } & typeof defaultProps;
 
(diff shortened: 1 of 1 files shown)


packages/forma-36-react-components/src/components/Dropdown/DropdownList/DropdownList.tsx
```diff
@@ -8,7 +8,7 @@ export type DropdownListProps = {
   className?: string;
   testId?: string;
   border?: 'top' | 'bottom';
-  maxHeight?: number;
+  maxHeight?: 'initial' | 'inherit' | string | number;
```
burakukula, line 11: @marcolink would `string | number` not be enough? I am wondering about adding `'initial' | 'inherit' `...

marcolink: @burakukula 
technically it would, but this gives you a better autocomplete hint support.

burakukula: @marcolink I'm not convinced, but if that make life easier then sure thing. 🙂

mshaaban0: @marcolink @burakukula we have [csstype](https://www.npmjs.com/package/csstype), I believe we can do `number |  CSS.MaxHeightProperty<string>` similar to usage in the [grid](https://github.com/contentful/forma-36/blob/master/packages/forma-36-react-components/src/components/Grid/Grid.tsx#L2) component

marcolink: @mshaaban0 I changed it accordingly to your suggestion 👍

PR #468, opened 2020-05-12
fix: add missing button type

Jwhiles:
#Purpose of PR
Spotted a that `warning` was missing in buttons prop types - which is causing an error to be shown when developing the web app.

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Button/Button.tsx
+++ b/packages/forma-36-react-components/src/components/Button/Button.tsx
@@ -23,7 +23,7 @@ export type ButtonProps = {
   loading?: boolean;
   disabled?: boolean;
   testId?: string;
-  buttonType?: 'primary' | 'positive' | 'negative' | 'muted' | 'naked';
+  buttonType?: 'primary' | 'positive' | 'negative' | 'warning' | 'muted' | 'naked';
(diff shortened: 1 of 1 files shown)


packages/forma-36-react-components/src/components/Button/Button.tsx
```diff
@@ -23,7 +23,7 @@ export type ButtonProps = {
   loading?: boolean;
   disabled?: boolean;
   testId?: string;
-  buttonType?: 'primary' | 'positive' | 'negative' | 'muted' | 'naked';
+  buttonType?: 'primary' | 'positive' | 'negative' | 'warning' | 'muted' | 'naked';
```
davidfateh, line 26: Is `warning` even valid? Does it actually have a different styling?

Jwhiles: it is, it exists in the story file and the css :)

burakukula: Hey @davidfateh , yeah, seems like it's a valid case and has different styling https://f36-storybook.contentful.com/?path=/story/components-button--button

PR #477, opened 2020-05-25
Adding missing pages with links

burakukula:
Adding some missing components over here. Simple version for the time being  - we can update each of them slowly now.
WDYT?

Screen Recording 2020-06-04 at 08 47 05


packages/forma-36-website/.eslintrc
```diff
@@ -0,0 +1,5 @@
+{
+  "rules": {
+    "@typescript-eslint/no-explicit-any": "off"
```
gui-santos, line 3: We need to discover what is happening with this thing... I've just pulled your branch and removed this file so I this rule on basically and it worked! hahahha

I say that also because this rule is good... it forbids us from creating types like  `interface Props { prop: any }`

packages/forma-36-website/gatsby-config.js
```diff
@@ -151,30 +183,62 @@ module.exports = {
...
+            name: 'Radio button',
+            link: '/components/radio-button/',
+          },
+          {
+            name: 'Select',
+            link: '/components/select/',
+          },
+          {
+            name: 'SelectField',
+            link: '/components/select-filed/',
```
m10l, line 200:
```suggestion
            link: '/components/select-field/',
```

packages/forma-36-website/src/pages/components/card/index.mdx
```diff
@@ -0,0 +1,49 @@
...
+status: 'stable'
+github: 'https://github.com/contentful/forma-36/tree/master/packages/forma-36-react-components/src/components/Card'
+storybook: https://f36-storybook.contentful.com/?path=/story/components-card-entrycard--default'
+---
+Detailed documentation will be provided soon. For now, please have a look into the storybook for implementation details.
+
+```jsx
+<React.Fragment>
+  <div>
+    <EntryCard
```
m10l, line 13: This should just be the `<Card>` component

packages/forma-36-website/src/pages/components/card/index.mdx
```diff
@@ -0,0 +1,49 @@
+---
+title: 'Card'
+type: 'component'
+status: 'stable'
+github: 'https://github.com/contentful/forma-36/tree/master/packages/forma-36-react-components/src/components/Card'
+storybook: https://f36-storybook.contentful.com/?path=/story/components-card-entrycard--default'
```
m10l, line 6:
```suggestion
storybook: 'https://f36-storybook.contentful.com/?path=/story/components-card--default'
```

packages/forma-36-website/src/pages/components/card/index.mdx
burakukula: @m10l yeah, I thought it would just look better for now. We would probably add here documentation for all type of Cards anyway. but no prob, I can changed it to the simple card.

packages/forma-36-website/gatsby-config.js
burakukula: @m10l fixed

packages/forma-36-website/src/pages/components/card/index.mdx
burakukula: And done :)

PR #480, opened 2020-05-28
feat(autocomplete): multiselect

marcolink:
# Purpose of PR

Add a `multiselect` feature for the [Autocomplete](https://github.com/contentful/forma-36/tree/master/packages/forma-36-react-components/src/components/Autocomplete) component.

- While selecting, dropdown stays open.
- Selecting items by mouseclick OR spacebar.

Screen Recording 2020-05-28 at 14 11 41

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Autocomplete/Autocomplete.tsx
+++ b/packages/forma-36-react-components/src/components/Autocomplete/Autocomplete.tsx
@@ -47,6 +47,7 @@ export interface AutocompleteProps<T extends {}> {
   willClearQueryOnClose?: boolean;
   dropdownProps?: DropdownProps;
   renderToggleElement?: (props: RenderToggleElementProps) => React.ReactElement;
+  multiselect?: boolean;
 }
 
 interface State {
@@ -56,10 +57,14 @@ interface State {
 }
 
 type Action =
-  | { type: typeof TOGGLED_LIST; payload?: boolean }
-  | { type: typeof NAVIGATED_ITEMS; payload: number | null }
-  | { type: typeof QUERY_CHANGED; payload: string }
-  | { type: typeof ITEM_SELECTED };
+  | { type: typeof TOGGLED_LIST; payload?: boolean; multiselect?: boolean }
+  | {
+      type: typeof NAVIGATED_ITEMS;
+      payload: number | null;
+      multiselect?: boolean;
+    }
+  | { type: typeof QUERY_CHANGED; payload: string; multiselect?: boolean }
+  | { type: typeof ITEM_SELECTED; payload: number; multiselect?: boolean };
 
 enum Direction {
   DOWN = 1,
@@ -73,6 +78,9 @@ const initialState: State = {
 };
 
 const reducer = (state: State, action: Action): State => {
+  // eslint-disable-next-line no-console
+  console.debug('EVENT', action.type);
+
   switch (action.type) {
     case TOGGLED_LIST:
       return {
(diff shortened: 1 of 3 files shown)


packages/forma-36-react-components/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -73,6 +78,9 @@ const initialState: State = {
 };
 
 const reducer = (state: State, action: Action): State => {
+  // eslint-disable-next-line no-console
+  console.debug('EVENT', action.type);
```
gui-santos, line 82: Do we need this?

packages/forma-36-react-components/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -89,12 +97,18 @@ const reducer = (state: State, action: Action): State => {
     case QUERY_CHANGED:
       return {
         ...state,
-        highlightedItemIndex: null,
+        highlightedItemIndex: action.multiselect
+          ? state.highlightedItemIndex
+          : null,
```
gui-santos, line 102: It seems to me here that you want the state, more specifically the `highlightedItemIndex`, to stay the same if the multiselect is true... is that right?

If that's so, what do you think about doing something like this?

```js
{
  ...state,
   // only if multiselect is false you alter the state
  ...(!action.multiselect && { highlightedItemIndex: null }),
  isOpen: true,
  query: action.payload
}
```

packages/forma-36-react-components/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -129,17 +146,31 @@ export const Autocomplete = <T extends {}>({
...
+    }
     onChange(item);
   };
 
   const updateQuery = (value: string) => {
-    dispatch({ type: QUERY_CHANGED, payload: value });
+    dispatch({
+      type: QUERY_CHANGED,
+      payload: value,
+      multiselect: !!multiselect,
```
gui-santos, line 172: Since `multiselect` is a boolean, I think the double `!!` is unnecessary, and then this would be cleaner to do:

```js
{
  type: QUERY_CHANGED,
  payload: value,
  multiselect,
}
```

What do you think?

packages/forma-36-react-components/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -162,10 +195,17 @@ export const Autocomplete = <T extends {}>({
...
+        payload: newIndex,
+        multiselect: !!multiselect,
+      });
+    } else if (
+      (isEnter && hasUserSelection) ||
+      (isSpace && multiselect && hasUserSelection)
+    ) {
       const selected = items[highlightedItemIndex as number];
-      selectItem(selected);
+      selectItem(selected, highlightedItemIndex || 0);
```
gui-santos, line 208: Do we still need the `|| 0` since in the function declaration the second argument defaults to 0?

m10l:
Noticed a little bug - clicking on the text won't select the item.

I think there are some usability issues here too that maybe need some more design thought - by filtering it's possible to hide what has been selected, which makes it difficult to keep track of what has already been added. Perhaps selected items should always be shown in the dropdown list when filtering.

In general, this pattern feels a bit odd to me. I think something closer to https://react-select.com/home would be a better solution

burakukula:
@marcolink We will close it for now, if that's ok with you.

PR #481, opened 2020-05-28
fix(ValidationMessage): adjust size and position of icon in validation

domarku:
- Adjust size and position of icon in the validation message below the input field

### Current state

The icon is aligned to the center of the text block and resizes with every new line of text.

Screenshot 2020-05-28 at 17 29 54

### Proposed solution

Screenshot 2020-05-28 at 17 21 20

## PR Checklist

- [x] Tests are passing


packages/forma-36-react-components/src/components/ValidationMessage/ValidationMessage.css
```diff
@@ -4,12 +4,13 @@
 
 .ValidationMessage {
   display: flex;
-  align-items: center;
 }
 
 .ValidationMessage__icon {
   margin-right: calc(1rem * (7 / var(--font-base-default)));
-  margin-bottom: calc(1rem * (1 / var(--font-base-default)));
+  margin-top: calc(1rem * (2 / var(--font-base-default)));
```
mshaaban0, line 11: Nice @domarku 🔥do you want to use one of the spacing values here instead?

domarku: 100% yes, thanks for the suggestion

domarku: Actually, come to think of it, when icons are put in context with text, it's good practice to use the base font size to set the relationship between the two. If the base font size changes, the chances are greater that it's still going to work, as opposed to using an external spacing variable. @m10l any thoughts?

packages/forma-36-react-components/src/components/ValidationMessage/ValidationMessage.css
```diff
@@ -1,15 +1,17 @@
 @import 'settings/colors';
 @import 'settings/typography';
 @import 'settings/transitions';
+@import '@contentful/forma-36-tokens/dist/css/spacing.css';
```
m10l, line 4: This should work and will keep things more consistent -

```suggestion
@import 'settings/dimensions';
```

packages/forma-36-react-components/src/components/ValidationMessage/ValidationMessage.css
m10l: No strong opinion from me, think there are good arguments for both

domarku:
Ready to merge!

PR #493, opened 2020-06-11
feat: Add missing empty states

burakukula:
Hello lovely F36 people.

I am adding some missing docs over here. And missing empty states for existing pages. Fixing my own 🐛 over here 😕

PR #496, opened 2020-06-16
Chore(website): remove eslintrc and  ignore react-components/dist

gui-santos:
# Purpose of PR

Previously we ignored the files in `/dist/esm` but maybe it's not enough and we are still getting the error @burakukula posted in #493 

So this PR ignores all `/dist`

I'm not 100% sure if this is correct tough 

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-website/.eslintignore
+++ b/packages/forma-36-website/.eslintignore
@@ -1,1 +1,1 @@
-../forma-36-react-components/dist/esm/index.js
+../forma-36-react-components/dist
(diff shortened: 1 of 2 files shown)


burakukula:
Looks good! but we would need to fix types in the broken files anyway at some point, right?

gui-santos:
@burakukula yes but I thought that we have linter covering the dev files, so covering the bundles seems unnecessary(????????) but I'm not sure. I just took Johannes solution and expanded to other files hahahaa

burakukula:
@gui-santos ahhh yeah, I think you’re right!

PR #503, opened 2020-06-30
fix(button): make border color the same blue when button is a link

gui-santos:
# Purpose of PR

I noticed that we created this tiny bug after the color change in our buttons. We decided to make the border the same color as the background, but in buttons that use the `<a>` tag the border was still darker which makes them look different

Screenshot 2020-06-30 at 14 39 17
Screenshot 2020-06-30 at 14 39 43

This PR makes the border of those buttons the same color as their background
 
## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Button/Button.css
+++ b/packages/forma-36-react-components/src/components/Button/Button.css
@@ -106,7 +106,7 @@
   }
 
   &.Button:link {
-    border-color: var(--color-blue-dark);
+    border-color: var(--color-blue-mid);
   }
 }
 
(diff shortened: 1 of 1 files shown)


packages/forma-36-react-components/src/components/Button/Button.css
```diff
@@ -106,7 +106,7 @@
   }
 
   &.Button:link {
-    border-color: var(--color-blue-dark);
+    border-color: var(--color-blue-mid);
```
gui-santos, line 109: I was not sure if I could just remove the `&.Button:link` block so I just changed the color

Issue #466, opened 2020-05-05
💡 Proposal - new ProductIcon / NavigationIcon component in Forma36


burakukula:
# Forma 36 ProductIcon / NavigationIcon

## The problem
Hey there 👋

I would like to propose to extend Forma36 to contain the `<ProductIcon />` /  `<NavigationIcon />` component. This component is used by navigation and headlines of the pages in the webapp. By moving this icons into Forma36 we would improve the visibility of the product related icons, potentially it might be useful for customers that are creating their custom solutions. This would also give us a chance to present our guidelines and recommendations for icons, for developers and designers.
In the future, if we consider moving navigation elements into Forma36, this step could be also beneficial.
We would have all icons aligned, visible and be part of the design system, which would give us more consistency and transparency. 

## The proposed solution

Proposed component was already created in the Contentful web app. It is used by top navigation items and headlines of the pages:

<img width="1127" alt="Screenshot 2020-05-05 at 09 39 30" src="https://user-images.githubusercontent.com/4272331/81044337-77991500-8eb4-11ea-80c2-05685b22e41e.png">
<img width="1790" alt="Screenshot 2020-05-05 at 09 39 46" src="https://user-images.githubusercontent.com/4272331/81044343-78ca4200-8eb4-11ea-954c-9919f29a36b0.png">

1. New `<NavigationIcon />` in the web app has predefined set of sizes:
```
  small: '18px',
  medium: '24px',
  large: '32px',
  xlarge: '48px'
```
I would like to propose to align sizes of Forma36 `<Icon />` and `<NavigationIcon />` component in terms of sizes, to keep them consistent. We would end up with following set for both components:
```
  tiny: '16px',
  small: '18px',
  medium: '24px',
  large: '32px',
  xlarge: '48px'
```

2. Currently `<NavigationIcon />` in the Contentful web app takes only 2 colors: `colorGreenBase` and `colorWhite`, but we could extend it if needed. Maybe align `<Icon />` and `<NavigationIcon />` when it comes to colors?

3. I would like to also propose to have an option to define tag type to be `'div' | 'span'` within this new component.

4. We could allow this component to display custom SVG just in case it's needed, something like (pseudo code):
```
{type ? <IconComponent /> : children}
```
5. Apart of the above, we would also allow `testId` and `className` of course.

6. Alignment of the documentation for existing `<Icon />` and creating docs for new component:
[...]

PR #505, opened 2020-07-01
feat: new NavigationIcon component

burakukula:
Hey there F36!

Following the proposal of NavigationIcon I am adding this new component - https://github.com/contentful/forma-36/issues/466
It is currently used by navigation and headlines of the pages

For example:

Screenshot 2020-05-05 at 09 39 46

The set contains all the icons used by the nav right now, we will change couple of them for more intuitive from the same set (like Teams or Spaces), but this can be done later and I don't consider it a blocker for this component.

Additionally I have changed `Icon` component large size from 36px to 32px to be aligned with NavigationIcon size.

I am not sure if it supposed to be added in alpha though.. WDYT?
As soon as this PR is done:
 - I will add documentation
 - replace NavigationIcon in the web app with the this new F36 component.


packages/forma-36-react-components/src/components/NavigationIcon/NavigationIcon.tsx
```diff
@@ -0,0 +1,115 @@
...
+import Apis from './svg/Apis.svg';
+import Billing from './svg/Billing.svg';
+import Home from './svg/Home.svg';
+import Media from './svg/Media.svg';
+import ContentModel from './svg/ContentModel.svg';
+import Content from './svg/Content.svg';
+import Apps from './svg/Apps.svg';
+import Oauth from './svg/Oauth.svg';
+import OrgInfo from './svg/OrgInfo.svg';
+import Oraganizations from './svg/Oraganizations.svg';
```
gui-santos, line 15: Typo?

packages/forma-36-react-components/src/components/NavigationIcon/NavigationIcon.tsx
```diff
@@ -0,0 +1,115 @@
...
+}
+
+const defaultProps = {
+  testId: 'cf-ui-navigation-icon',
+  size: 'medium',
+  color: 'positive',
+  tagType: 'div'
+};
+
+export class NavigationIcon extends Component<NavigationIconProps & typeof defaultProps> {
```
gui-santos, line 79: All good! But what do you think about making this a functional component? I think it does not need to be a class. I know may of our components are classes, but maybe we can refactor them in small steps and new components we create as functions

But maybe this is also something we should discuss as a team... Idk

packages/forma-36-react-components/src/components/NavigationIcon/NavigationIcon.tsx
```diff
@@ -0,0 +1,115 @@
...
+    const Element = iconComponents[icon];
+    const Tag = tagType;
+
+    return (
+      <Tag
+        className={classNames}
+        data-test-id={testId}
+      >
+        <Element  className={svgFillClass} {...otherProps} />
+      </Tag>
```
gui-santos, line 110: I think I didn't understand the <Tag> here, it can be a div or a span. But do we need to create it with this wrapper? I looked into the Icon component and it does not have it

packages/forma-36-react-components/src/components/NavigationIcon/constants.ts
```diff
@@ -0,0 +1,21 @@
...
+  Apis: 'Apis',
+  Apps: 'Apps',
+  Billing: 'Billing',
+  ContentModel: 'ContentModel',
+  Content: 'Content',
+  Home: 'Home',
+  Media: 'Media',
+  Oauth: 'Oauth',
+  OrgInfo: 'OrgInfo',
+  Oraganizations: 'Oraganizations',
```
gui-santos, line 11: Just pointing the typo here :)

packages/forma-36-react-components/src/components/NavigationIcon/NavigationIcon.tsx
burakukula: Ups, on my way

packages/forma-36-react-components/src/components/NavigationIcon/NavigationIcon.tsx
burakukula: @gui-santos yeah, good point. I thought that it might be sligthly different in this case - the container controll the size of the icon, SVG itself knows only about the the color. when implementing we would need to create wrapper on the application side anyway I guess, so I figured that maybe it could live in the component itself? This is how I implemented this component in the app, but it can totally be changed if you fill like that's not the right approach. We can go with what we have in the Icon 1:1.

packages/forma-36-react-components/src/components/NavigationIcon/NavigationIcon.tsx
burakukula: totally, I will refactor it. thx!

packages/forma-36-react-components/src/components/NavigationIcon/NavigationIcon.tsx
gui-santos: Ohhh I get it now!
Thanks for the explanation. I think it makes sense, so IMO you can leave it like this :)

cool stuff

packages/forma-36-react-components/src/components/NavigationIcon/NavigationIcon.tsx
burakukula: @gui-santos fixed

packages/forma-36-react-components/src/components/NavigationIcon/NavigationIcon.tsx
burakukula: @gui-santos done 🙂

packages/forma-36-react-components/src/components/NavigationIcon/constants.ts
burakukula: @gui-santos fixed :)

packages/forma-36-react-components/src/alpha.ts
```diff
@@ -1,3 +1,4 @@
 export { Autocomplete } from './components/Autocomplete/Autocomplete';
 export { Grid } from './components/Grid';
 export { GridItem } from './components/Grid/GridItem';
+export { NavigationIcon } from './components/NavigationIcon/NavigationIcon';
```
burakukula, line 4: I am not sure if this should be alpha 🤔

gui-santos: Me neither... I would say "yes", maybe ask the other Forma team people (????)

burakukula: other Forma team people! Please help! 😄
@m10l @mshaaban0 WDYT?

m10l: I'd say yes to it being alpha :)

PR #522, opened 2020-07-20
feat(note): close button

marcolink:
# Purpose of PR

This pull request adds an optional flag to add a close icon to the Note component.

## PR Checklist

- [ ] Tests are passing


packages/forma-36-react-components/src/components/Note/Note.css
```diff
@@ -58,3 +65,19 @@
   display: flex;
   margin-top: 1px;
 }
+
+.Note__dismiss {
```
domarku, line 69: I think the button needs some space on the left to avoid clashing with the text (image 1)

Screenshot 2020-07-21 at 11 01 06

(I did it with `margin-left` but not sure if that's the right approach)

...and perhaps a `margin-bottom: auto` to avoid it having a taller click area than it should.

Before:
Screenshot 2020-07-21 at 11 01 38

After:
Screenshot 2020-07-21 at 11 02 05

packages/forma-36-react-components/src/components/Note/Note.css
```diff
@@ -58,3 +65,19 @@
...
+  cursor: pointer;
+  outline: none;
+  margin-top: 0;
+  margin-right: 0;
+  margin-left: auto;
+  pointer-events: all;
+  display: flex;
+}
+
+.Note__dismiss span {
```
m10l, line 81: If you take @domarku's suggestion for using `margin-bottom: auto` we can get rid of this bit of CSS completely

packages/forma-36-react-components/src/components/Note/Note.css
m10l: +1

marcolink: Screenshot 2020-07-21 at 22 57 50

To keep the button floating right, I had to apply the margin to the middle content block (in case we show the close button). There might be better CSS solutions that I don't know off :)

domarku: Looks good to me! 👍

packages/forma-36-react-components/src/components/Note/Note.tsx
```diff
@@ -57,6 +61,20 @@ export class Note extends Component<NoteProps> {
...
+          <IconButton
+            buttonType="secondary"
+            iconProps={{ icon: 'Close' }}
+            onClick={() => {
+              if (this.props.onClose) {
+                this.props.onClose();
+              }
+            }}
+            testId="cf-ui-note-close"
+            label="Dismiss"
```
mshaaban0, line 74: <3

marcolink:
> LGTM, please note that the clickable area should be 44x44 pixels. [Target Size](https://www.w3.org/WAI/WCAG21/Understanding/target-size.html)

@mshaaban0 
Currently, it's the same size as we have for `IconButton` and `Notification` (where we show the same close `x`) which is 18x18.

But I agree with the specs you posted. Maybe it's worth creating the close icon as its own component.  
What do you think?

mshaaban0:
@marcolink not sure about it's own component 🤔 
I agree that the spec should be part of higher level component, maybe IconButton itself as minimum width, unless specified otherwise.

burakukula:
@marcolink what's up over here? Would you like to continue working on this one? 🙂

marcolink:
@burakukula 
Yes, I already discussed with @domarku how we should proceed, cause i'm a bit unclear about the right way to go forward.

gui-santos (approved):

burakukula:
@marcolink I think it's a good point, that @mshaaban0 made and we will address this issue, but I think it's better if would go ahead with your PR and we could solve the general `IconButton` issue in a separate PR. WDYT? This would unblock you I think. 🤔

m10l (approved):
I'd say leave the icon button size issue for now and this should be done on a wider scale so it's consistent with all components

PR #525, opened 2020-07-24
feat: button readme proposal

burakukula:
Proposal of the template for README for each component. Updating REAMDE for our components would allow us to gather content for "code" part of the documentation and then we might want to reuse it on the documentation website. This way we could work in parallel with the design part of the documentation. 

Overview:
https://github.com/contentful/forma-36/tree/readme-button-proposal/packages/forma-36-react-components/src/components/Button

I was thinking that we could have this structure for each README:

1. General description of the component
2. Component types (primary, secondary, muted, etc) with general description of the usage - more detailed examples will be provided in the design part and I guess it's more relevant for designers than for developers. But it would be cool to have one sentence or so to give a general understanding regarding the difference
3. Components variants - active, disable, sizes, any particular property that is worth to describe could be mentioned here
4. Example of usage - this is where we could show some real examples from our codebase, how to use this component in a context, if needed "bad" and "good"
5. Best practices  - this part is a one that we had before, I am not sure if we should keep it. In my opinion it something that is part of the "design part". WDYT?
6. Content recommendations - recommendations from Rachel - maybe we will transfer into separate section as soon as we start improve website, but maybe for now it would be useful to create this section over here to gather content with help from Rachel
7. Accessibility - I think we could explicitly point out what we do have implemented but maybe also list things that are currently missing and what we plan to implement. This might be a way for us to be very transparent regarding our accessibility level and a way to gather things that we want to implement. dunno... WDYT?

PS. I changed  Button.md to README.md - this way we display the content on GitHub page as well. Hope that's ok.


packages/forma-36-react-components/src/components/Button/README.md
```diff
@@ -0,0 +1,117 @@
...
+## Button types
+
+- **Primary** - Used for the most important actions in any scenario. Don’t use more than one primary button in a section or screen to avoid overwhelming users.
+- **Muted** - Used for a secondary actions, the most commonly used button type.
+- **Positive** - For use when the action has a positive connotation such as creating or publishing a new entity.
+- **Negative** - For destructive actions - when something can't be undone. For example, deleting entities.
+- **Naked** - Link button
+
+
+![button overview](images/img.png)
```
burakukula, line 14: Is it a good idea to add screeshots to this docs?

packages/forma-36-react-components/src/components/Button/Button.stories.tsx
```diff
@@ -5,7 +5,7 @@ import { text, boolean, select } from '@storybook/addon-knobs';
 import { iconName } from '../Icon/constants';
 
 import Button from './Button';
-import notes from './Button.md';
+import notes from './README.md';
```
m10l, line 8: I'd recommend keeping this to match the component name for better developer experience when editing multiple readmes at the same time

packages/forma-36-react-components/src/components/Button/README.md
m10l: I think we should use code blocks so we can render them with MDX - removes the overhead of having to update images for changes in appearance

packages/forma-36-react-components/src/components/Button/README.md
```diff
@@ -0,0 +1,117 @@
...
+ - Start labels with action verbs. If the button is part of an action dialog, make sure it matches the dialog header.
+ - Use specific words, ideally ones that align with preceding content.
+ - For buttons that are used to cancel destructive actions: label them "Never mind" or similar, instead of cancel. It makes things much easier for users to understand.
+
+## Accessibility
+
+Button component is used to trigger an action or event. By default button is implemented using the native `<button>` element with the proper type="button" provided. If the href element is provided, then the element would be rendered as `<a>` element in the document.
+When a button has focus, both the Space and Enter keys will activate the button. If the button is disabled, then the disabled attribute is provided.
+Contentful buttons are checked in terms of color contrast and pass all the requirements.
+Buttons has trap focus implemented with the tabindex="-1".
```
burakukula, line 96: @m10l could you help me out to phrase this focus solution so it would be clear? something like: "users navigating using keyboard, will have the same focus experience"? ayayayayay

packages/forma-36-react-components/src/components/Button/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Examples of usage
+
+```jsx
+import {Button} from '@contentful/forma-36-react-components';
+
+<Button buttonType="primary" data-test-id="create-content-type-empty-state">Add content type</Button>
+
+```
+Contentful react button will be renderd as:
```
m10l, line 36: I'd recommend leaving this out and making use of MDX to show a HTML preview.

Since we currently provide our components only as React components it probably doesn't make too much sense to show the rendered HTML markup anyway

packages/forma-36-react-components/src/components/Button/README.md
```diff
@@ -0,0 +1,117 @@
...
+<button data-test-id="create-content-type-empty-state" class="Button__Button___1ZfFj a11y__focus-border--default___60AXp Button__Button--primary___JImeO" type="button">
+  <span tabindex="-1" class="TabFocusTrap__TabFocusTrap___39Vty Button__Button__inner-wrapper___3qrNC">
+   <span class="Button__Button__label___3tcOj">
+    Add content type
+   </span>
+  </span>
+</button>
+```
+
+Buttons  can be used in different contexts in your project. You might want to use it for dropdowns or display icons inside of the button.
```
m10l, line 48: I think this information should live with the Dropdown component as it's a configuration option of the the Dropdown

packages/forma-36-react-components/src/components/Button/README.md
```diff
@@ -0,0 +1,117 @@
...
+## Accessibility
+
+Button component is used to trigger an action or event. By default button is implemented using the native `<button>` element with the proper type="button" provided. If the href element is provided, then the element would be rendered as `<a>` element in the document.
+When a button has focus, both the Space and Enter keys will activate the button. If the button is disabled, then the disabled attribute is provided.
+Contentful buttons are checked in terms of color contrast and pass all the requirements.
+Buttons has trap focus implemented with the tabindex="-1".
+
+## Properies overview:
+
+|Prop|Description|
```
m10l, line 100: I don't think we should add props here, they should be generated from the code rather than manually added

packages/forma-36-react-components/src/components/Button/README.md
m10l:
```suggestion
Buttons have a focus state when using keyboard navigation
```

Not sure it's necessary to give details about the tab focus trap  or the tab index here

packages/forma-36-react-components/src/components/Button/README.md
```diff
@@ -0,0 +1,117 @@
+# Button
```
m10l, line 1: Adding a table of contents would be great, but maybe this can be autogenerated when we consume the readme in the Forma 36 docs website

packages/forma-36-react-components/src/components/Button/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+
+![button overview](images/img.png)
+
+Buttons can be used in couple of different variations. Worth mentioning is `indicateDropdown` property in the component, which enables chevron icon on the right side of the button. Using `icon` property in the Button on the other hand, enables choosen icon on the left side of the button.
+
+![Button variations](images/img-1.png)
+
+Contentful buttons are available in 3 different sizes: 
+ - large - 46px height
```
m10l, line 21: I'd list only the name of the sizes, or just showcase using code blocks, but I wouldn't add values here

packages/forma-36-react-components/src/components/Button/Button.stories.tsx
burakukula: @m10l sure, I thought that it would be useful to have it displayed in here https://github.com/contentful/forma-36/tree/readme-button-proposal/packages/forma-36-react-components/src/components/Button. If it's a Button.md it won’t be displayed, I think

packages/forma-36-react-components/src/components/Button/README.md
burakukula: @m10l I think it might be interesting to explain it. I didn't know it myself and I guess it's a cool info to share with other engineerings.

packages/forma-36-react-components/src/components/Button/README.md
burakukula: hmm, true. Maybe it's better to show something simple, like Button with icon?

packages/forma-36-react-components/src/components/Button/README.md
burakukula: Ok, I get it. I thought it might be useful what html markup component will generate, but yeah, maybe it's an overkill.

packages/forma-36-react-components/src/components/Button/README.md
burakukula: @m10l I think we might want to create some sort of secondary menu in the website? something that is sticky and user is able to use it when scrolling down? But it would be some sort of table of contents I guess...

packages/forma-36-react-components/src/components/Button/README.md
burakukula: @m10l yeah, you’re right. I removed them.

packages/forma-36-react-components/src/components/Button/README.md
burakukula: @m10l hmm, yeah. good point, but I like the fact that we have small description of the props available in the table. I guess from maintenance perspective it would be good to have them generated, but I guess we want change API that often, so I figured it would be added manually only once. Maybe it's worth having this overview? or overkill?

packages/forma-36-react-components/src/components/Button/Button.stories.tsx
gui-santos: @burakukula I also like to have the README displayed in github
I vote for keeping it README

packages/forma-36-react-components/src/components/Button/README.md
m10l: I don't think it's worth the effort and would be more harmful to keep as it could easily contain the wrong information in the future.

burakukula: @m10l ok, I will remove it. Maybe there is a way to generate it from code 🤔

burakukula: @m10l oki! removed, I will try to figure out how to autogenerate those. thx a lot for feeback

m10l (approved):

PR #526, opened 2020-07-24
Revert "chore(deps): bump lodash from 4.17.15 to 4.17.19 in /packages/forma-36-react-components"

mshaaban0:
Reverts contentful/forma-36#514

Attempt to fix failing builds


mshaaban0:
not the problem

PR #531, opened 2020-08-03
Accordion Component

### LINKED ISSUE #513: 💡 Proposal - Accordion Component

## The problem

In the feature that I'm working on we had to create a FAQ section for a page. But the content of around 6 questions is quite long to be always on the page, also a user may want to know about one thing only so scrolling over the other 5 questions would be annoying or make the user miss what they a looking for. 

## The proposed solution

Let's create an Accordion component. It would also need an `<AcordionHeader />` and an `<AcordionPanel />`

[prototype](https://www.figma.com/proto/q73ZvGDXXliJTyMklNDmJq/Accordion?node-id=16%3A0&viewport=-245%2C-150%2C0.39369869232177734&scaling=min-zoom)

Anatomy

> **Accordion Header:**
> Label for or thumbnail representing a section of content that also serves as a control for showing, and in some implementations, hiding the section of content.
> **Accordion Panel:**
> Section of content associated with an accordion header.
> (taken from [WAI-ARIA](https://www.w3.org/TR/wai-aria-practices-1.2/#accordion))

### Chevron variation

We could have a prop to define two different alignments for the chevron

Chevron Orientation

### Hover/Focus state

For hover and focus state we can use the same styles we use in the `naked` button

Focus state

Considering the case where the accordion has a gray background, this is how it would look
Focus state GRAY

### Usage

The component allows other components to be passed as a title for the accordion, but it is better to avoid abusing this feature. In most cases, a Heading or Subheading component will be enough.
It's worth to point out that there is a case in Contentful's webapp where we have a `...` button in an accordion header. I propose that his pattern should not be supported because it harms the accessibility of the component. The reason for this can be checked in the [W3 docs for accordions](https://www.w3.org/TR/wai-aria-practices-1.2/#wai-aria-roles-states-and-properties)

> - The title of each accordion header is contained in an element with role button.
> - Each accordion header button is wrapped in an element with role heading that has a value set for aria-level that is appropriate for the information architecture of the page.
>   - If the native host language has an element with an implicit heading and aria-level, such as an HTML heading tag, a native host language element may be used.
>   - The button element is the only element inside the heading element. That is, if there are other visually persistent elements, they are not included inside the heading element.

DO and DONTS

## Important requirements
- Accessibility following [WAI-ARIA](https://www.w3.org/TR/wai-aria-practices-1.2/#accordion)
- Possibility to pass other components to the Accordion Header

## Breaking changes

Uhmmm I don't think there is any breaking change, but maybe we would release it as an alpha component first

gui-santos:
# Purpose of PR

This PR implements two main components: `Accordion` and `AccordionItem`

The first one is a component to group all the accordions. It has only one prop, the `align` prop that serves to tell where to render the chevron icon.
The second one handles all the content of each accordion and is formed by an `AccordionHeader` and `AccordionPanel`. Its only prop is `title` that is required and accepts React nodes

Anatomy

More details can be found in the Proposal #513 and in the README of the component

Closes #513 

## PR Checklist

- [x] Tests are passing


packages/forma-36-react-components/src/components/Accordion/Accordion.css
```diff
@@ -0,0 +1,73 @@
...
+@import 'settings/transitions';
+@import 'settings/dimensions';
+
+.Accordion {
+  box-sizing: border-box;
+  padding: 0;
+  margin: 0;
+  list-style: none;
+
+  & li {
```
m10l, line 12: Please use a class here

packages/forma-36-react-components/src/components/Accordion/Accordion.css
```diff
@@ -0,0 +1,73 @@
...
+
+    &:first-child {
+      border-top: 1px solid var(--color-element-mid);
+    }
+  }
+}
+
+.Accordion--start {
+  & .AccordionHeader svg {
+    min-width: 9px; /* necessary to algin the chevron properly */
```
m10l, line 25: <3 for commenting weird-looking code

packages/forma-36-react-components/src/components/Accordion/Accordion.css
```diff
@@ -0,0 +1,73 @@
...
+    justify-content: space-between;
+  }
+}
+
+.AccordionHeader {
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  border: 0;
+  padding: 1rem;
```
m10l, line 42: Please use tokens here

packages/forma-36-react-components/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,51 @@
...
+   */
+  children?: React.ReactNode;
+  /**
+   * An ID used for testing purposes applied as a data attribute (data-test-id)
+   */
+  testId?: string;
+  /**
+   * Specify the alignment of the chevron inside the accordion header
+   */
+  align?: AlignTypes;
```
m10l, line 24: I think I'd prefer something more explicit/descriptive here like `chevronPosition`

packages/forma-36-react-components/src/components/Accordion/AccordionHeader/AccordionHeader.tsx
```diff
@@ -0,0 +1,66 @@
...
+  handleClick,
+  isExpanded,
+  ariaId,
+}: AccordionHeaderProps) => {
+  const classNames = cn(styles.AccordionHeader, {
+    [styles['AccordionHeader--expanded']]: isExpanded,
+  });
+
+  return (
+    <Subheading>
```
m10l, line 43: Should be able to pass an `element` prop to the Subheading component as by default it'll use a `h2`, can imagine wanting control over this. Alternatively, could use `details`/`summary` elements https://caniuse.com/#feat=details

packages/forma-36-react-components/src/components/Accordion/README.md
```diff
@@ -0,0 +1,125 @@
...
+
+- The title should be a short message that summarize the content of the accordion.
+- Anything can be passed as the content of the accordion, but often organizing it with Paragraphs and Tables would be enough.
+- When using headings, be mindful about the heading levels. The header is a Subheading with `h2` tag.
+
+## Best practices
+
+- The title prop accepts React nodes, but we recommend that only Typography components are used.
+- Since the whole header is a button, do not use Buttons or Links in the title prop. The only click event that should exist is the header's event.
+
```
burakukula, line 107: @gui-santos Good stuff! 💪 We could add accessibility section at some point over here, I think. (Maybe it should be part of our README template 🤔 ?). But we can totally do it later.

packages/forma-36-react-components/src/components/Accordion/Accordion.css
gui-santos: Oops! I will fix it

packages/forma-36-react-components/src/components/Accordion/Accordion.tsx
gui-santos: Agreed!

packages/forma-36-react-components/src/components/Accordion/AccordionHeader/AccordionHeader.tsx
gui-santos: @mohamedturki also suggested using details/summary. I will take a look on that

packages/forma-36-react-components/src/components/Accordion/README.md
gui-santos: yeah! I didn't know if a11y would be part of the design doc, but I put this small thing here hahaha

gui-santos:
@m10l just to comment on the padding, I know what you mean and I agree 👍

gui-santos:
Thanks @burakukula and @m10l for the comments

packages/forma-36-react-components/src/components/Accordion/AccordionHeader/AccordionHeader.tsx
gui-santos: I discovered something about using details/summary and now I don't know what is the best option

When testing with Mac's VoiceOver tool I couldn't navigate using headings because the heading now needed to be inside the summary (the first child of details needs to be summary) and clicking on the heading won't expand the accordion panel since the clickable element is the summary, not the heading

I got around this by using the same aria properties I used for the button solution. You can navigate to the heading correctly, but the VoiceOver does not let you go to the panel to read its content and I think the reason is something to do with [the way summary deals with headings](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/summary#Summaries_as_headings)

I'm committing the version with details/summary so you all can test and maybe find a solution

In case we can't solve the heading navigation I would suggest keeping the button/heading solution because it offers a better a11y experience

(in any case, we create the `element` prop)

gui-santos: Also, I'm getting this on Travis but not locally

Screenshot 2020-08-04 at 16 47 43

`jsx-a11y/click-events-have-key-events`: summary tag works with keyboard by default, I don't understand why we would need to add key events
`jsx-a11y/interactive-supports-focus`: summary and details are tabbable but they are not in the [rules list](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/interactive-supports-focus.md) apparently 

☝️  I would add up these to the cons of using details/summary for now

mohamedturki: @gui-santos I guess you meant @mshaaban0 :)

gui-santos: 😆  yes I did, @mohamedturki! I wrote "@moe" and selected the first name but that's you... sorry

@mshaaban0 I swear I know your name 🙏

mshaaban0: @gui-santos although you just mentioned a third person "moe" 😅 but thank you for implementing this.

Regarding, the details/summary, I'm wondering what is the use case for headings inside the summary tag? headings are meant to be identifiers for sections, and big blocks, I think the behaviour is normal as headings inside buttons isn't valid html5 in the first place. 

Regarding the a11y errors, maybe removing the custom aria roles on summary would fix it?

gui-santos: > @gui-santos although you just mentioned a third person "moe" 😅  

🤦 

The case for the headings was more trying to follow [these directives from WAI-ARIA](https://www.w3.org/TR/wai-aria-practices-1.2/#wai-aria-roles-states-and-properties) which assume we are using buttons and headings. 
But if heading navigation is not something that the accordion should worry about then that's solved I guess hahaha

Uhmm I will remove the aria roles later, but I think I had a different problem when I didn't have them. I will need to test again to remember, but I think I was having a problem going to the accordion content using the VoiceOver. Let me check again and I post it here

gui-santos: Forget everything I just said! It worked great with details and summary

Thanks you all!

gui-santos:
@m10l @mshaaban0 
Ok I found a new thing and I decided to go back to the `button` solution (It's the last time I promise)
Using `details` affected the animation of the component, look:

Animation using button
animation with button

Animation using details
Animation with details

The version with the details tag lost the closing animation of the accordion content, it disappears instead of animating its height. I researched a little bit and I think it's just how details work. Also at mozilla's documentation, it says that:

> Unfortunately, at this time there's no built-in way to animate the transition between open and closed.

The open animation I managed to hack something, but the closing one I don't see a workaround

dalach:
Hi Forma Friends, Gui is out this week, and I'd like to merge and close his PRs. Are there any Forma-specific steps I should undertake?

m10l:
> Hi Forma Friends, Gui is out this week, and I'd like to merge and close his PRs. Are there any Forma-specific steps I should undertake?

Hi @dalach! A design review from @joaoramos or @domarku would be great but other than that we're all good. I've just updated the branch against master and once everything is green merging the PR will automatically create a new release

dalach:
Thank you, @m10l 🤗 I can see a lot of checks failed. Wonder how it happened? 🤔 
@joaoramos or @domarku I'd appreciate if you could look and give a 🟢

domarku:
Hello there 👋 I can't take a proper look without a preview link, unfortunately.

m10l:
@dalach Ahh looks like the Netlify build got paused - I've just retriggered it.

m10l:
> Hello there 👋 Unfortunately, I don't think I can take a proper look without a preview link. Or is there another way to do it?

@domarku Should be working now: https://deploy-preview-531--forma-36-react-components.netlify.app/

domarku:
Looks good to me 🚢  Great work on the a11y front

Two things, but not blockers:
- I don't see any open/close animations on the content and I like it that way. Animated chevrons are cool!
- I would rotate the "end" chevrons to point down when closed and up when opened, indicating the user goal ("expand down"), as mentioned [in the proposal](https://github.com/contentful/forma-36/issues/513#issuecomment-657571957), and to distinguish from mobile app patterns — when the chevron points right to open something in a new screen

dalach:
Thank you @domarku. I'll merge this PR, and let @gui-santos address your comments when he's back.

gui-santos:
> Looks good to me 🚢 Great work on the a11y front! I don't see any open/close animations on the content and I like it that way. Animated chevrons are cool!
> 
> One comment, but not a blocker:
> 
> * I would rotate the "end" chevrons to point down when closed and up when opened, indicating the user goal ("expand down"), as mentioned [in the proposal](https://github.com/contentful/forma-36/issues/513#issuecomment-657571957), like Github does it (check the status panel below with accordion for approvals and reviewers) and to distinguish from mobile app patterns — when the chevron points right to open something in a new screen

Hey @domarku I'm back!

The animations were implemented in another PR #532 that is merged already. I did this way because the PR would get to long and too hard to review

But since it's merged already you can check the animation in storybook
https://f36-storybook.contentful.com/?path=/story/alpha-accordion--default

Please tell me what you think! 🙏

PR #532, opened 2020-08-03
Implement animation in the Accordion component

gui-santos:
# Purpose of PR

This PR adds animation to the Accordion component currently in review at PR #531 

There are actually two implementations hahaha
In the first commit I used WAAPI, but look at the support for it:
Screenshot 2020-08-03 at 16 18 29

(Also, I think enzyme or jest do not understand `.animate()`)

So I replaced with requestAnimationFrame magic
Screenshot 2020-08-03 at 16 18 05

I followed some ideas from [CSS-Tricks](https://css-tricks.com/using-css-transitions-auto-dimensions/) on how to animate `height: auto`
We need it to be auto because the content of the accordion varies

It would be nice if we can make a component (or a custom Hook) out of the animations to make it modular, but I wanted to check with you first. WDYT?

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Accordion/AccordionPanel/AccordionPanel.css
+++ b/packages/forma-36-react-components/src/components/Accordion/AccordionPanel/AccordionPanel.css
@@ -2,20 +2,16 @@
 @import 'settings/transitions';
 
 .AccordionPanel {
-  display: none;
   box-sizing: border-box;
   overflow: hidden;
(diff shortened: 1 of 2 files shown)


burakukula:
@gui-santos good stuff! looking at the support and the solution +1 for requestAnimationFrame! Seems like a right approach over here. 👍

PR #538, opened 2020-08-17
feat(navigation-icon): add new purchase icon to lib

gui-santos:
# Purpose of PR

My team needs to include this new NavigationIcon

Screenshot 2020-08-17 at 16 42 42

It's going to be used in Contentful's purchase flow redesign

## PR Checklist

- [x] Tests are passing


burakukula:
@gui-santos I think we need to replace Subscription icon with something else in that case, they are very similar. We can follow up with Cassie regarding this topic, she was up for helping out to replace those vague icons with something from this set. WDYT?

gui-santos:
> @gui-santos I think we need to replace Subscription icon with something else in that case, they are very similar. We can follow up with Cassie regarding this topic, she was up for helping out to replace those vague icons with something from this set. WDYT?

I agree 100%
I will hold this PR a little bit more then :)

Thanks

Gracegross:
The reason i chose this icon was because it's related to the subscription ... but happy to change it to something else. The new icons aren't coming in time for this, so going to have to use this set

Gracegross:
Suggesting this alternative
shopping-bag-2
 :)

burakukula:
@Gracegross @gui-santos sure thing! No worries, I think we don't have to block this. The subscription icon was considered one of the icons that was vague for users and we wanted to change that. I think in this context, the icon that you have chosen totally make sense. Let's go with it and meantime we can come back to the question if we want to change subscription icon. But it's not a blocker in my opinion.

PR #542, opened 2020-08-20
fix: use FieldGroup classname property

djagya:
# Purpose of PR

Fixes `FieldGroup` which wasn't using a passed `className`.


burakukula (approved):

m10l (changes requested):
Nice spot! Travis is currently failing because there are some snapshot tests that need to be updated. Good to go after that :)

djagya:
Thanks for pointing me! Fixed

m10l (approved):
🚀

PR #545, opened 2020-08-21
feat(entityList): add forwardRef to component

### LINKED ISSUE #534: 💡 Proposal - EntityList ref prop

## The problem

When scrolling through the EntityList, there is currently no way to securely attach an event handler to the HtmlElement.
There is a way to do so by querying the DOM and searching by id, but there is no guarantee, that the item, that will be fetched, is the correct copy of the component.

## The proposed solution

I propose to allow passing a ref to the EntityList.

```js
   const entityListRef = useRef();
   return (
      <EntityList
         ref={entityListRef}
      />
   );
```

This will allow to then attach event handlers, like let's say detecting if we hit the bottom of the list

```js
   entityListRef.current.addEventListener('scroll', onScroll):
```

## Breaking changes

No breaking changes

gui-santos:
# Purpose of PR

This closes #534 

Following the proposal in #534, this PR implements forwardRef to EntityList. It also refactors the component to a functional component

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/EntityList/EntityList.tsx
+++ b/packages/forma-36-react-components/src/components/EntityList/EntityList.tsx
@@ -1,9 +1,9 @@
-import React, { Component } from 'react';
+import React, { forwardRef } from 'react';
 import cn from 'classnames';
 
 import styles from './EntityList.css';
 
-export type EntityListProps = {
+export interface EntityListProps {
   /**
    * Class names to be appended to the className prop of the component
    */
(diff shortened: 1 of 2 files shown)


packages/forma-36-react-components/src/components/EntityList/EntityList.tsx
```diff
@@ -16,25 +16,28 @@ export type EntityListProps = {
...
+        {...otherProps}
+      >
         {children}
       </ul>
     );
-  }
-}
+  },
+);
+EntityList.displayName = 'EntityList';
```
mshaaban0, line 38: do we need this? react [mentions](https://reactjs.org/docs/react-component.html#displayname) it's usually inferred from the class name.

gui-santos: It's not a class anymore and because of the forwardRef eslint can't infer this. This is what I found about it in [eslint-plugin-rect](https://github.com/yannickcr/eslint-plugin-react/issues/2269)

And this is what I get if I remove `displayName`

Screenshot 2020-08-24 at 13 19 48

mshaaban0: Ah, I see, `forwardRef` strips the class/function name. Cool, thanks for explaining.

mshaaban0 (approved):

packages/forma-36-react-components/src/components/EntityList/EntityList.tsx
gui-santos: Genau! No problem, thanks for the review

PR #546, opened 2020-08-21
feat(entity-list-item): add withThumbnail prop

### LINKED ISSUE #535: 💡 Proposal - EntityListItem optional thumbnail

## The problem

Currently if an entity doesn't have a link to the image file, or a link to an asset, that doesn't have an image file, a thumbnail renders a generic image on a grey background. This doesn't always help differentiate between the EntityListItems and in fact makes each look the same and the focus shifts towards the thumbnail, instead of the title or the description of the EntityListItem, since most humans read from left to right and top to bottom, the first thing they see is a generic thumbnail.

## The proposed solution

It would be great, if EntityListItem allowed to make such thumbnail optional. Just like EntityCard does.

I think, a good way to do it without introducing a breaking change, would be to support a boolean property `withThumbnail` or similar, that, by default, may be `true`, but if `false` is given, should not render the thumbnail at all.

## Breaking changes

No breaking changes

gui-santos:
# Purpose of PR

Closes #535 

Adds `withThumbnail` prop to control whether or not the Thumbnail should render

## PR Checklist

- [x] Tests are passing


mshaaban0 (approved):
Looks Good ✨

PR #567, opened 2020-09-21
Refactor Tooltip component to use Popper.js

gui-santos:
# Purpose of PR

Screenshot 2020-09-21 at 14 49 58

Following up with some requests this PR is implementing [Popper.js ](https://popper.js.org/) to our Tooltip component so we don't need to deal with it's positioning

I feel disclaimers:

1. We needed to update storybook to an alpha version because apparently `usePopper` does not work with the latest stable version of storybook, more on this [here](https://github.com/storybookjs/storybook/issues/12475)
2. Popper.js documentation around react-popper and usePopper is... inexistent basically. [I'm not the only one complaining about this](https://github.com/popperjs/react-popper/issues/359)
So, I used other two sources to help me on this implementation. You can check them [here](https://davidcosta.com.br/en/using-popper-as-tooltip-typeScript-and-styled-component/) and [here behind Medium paywall](https://outline.com/sf3jLE)

## Open questions to y’all

There are a couple of props that I didn't remove, but I also don't know what to do with them

- onBlur?: Function;
- onFocus?: Function;
- onMouseLeave?: Function;
- onMouseOver?: Function;
- place?: 'top' | 'bottom' | 'right' | 'left'

What do you think about them?

Also, my yarn.lock files changed once I yarned in the root directory 👀 
I committed them but I'm not sure this happened

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Tooltip/Tooltip.stories.tsx
+++ b/packages/forma-36-react-components/src/components/Tooltip/Tooltip.stories.tsx
@@ -11,7 +11,7 @@ export default {
     propTypes: Tooltip['__docgenInfo'],
   },
   argTypes: {
-    // align: { control: { type: 'select', options: ['start', 'end'] } },
+    content: { control: 'text' },
     children: { control: { disable: true } },
     className: { control: { disable: true } },
     testId: { control: { disable: true } },
--- a/packages/forma-36-react-components/src/components/Tooltip/Tooltip.stories.tsx
+++ b/packages/forma-36-react-components/src/components/Tooltip/Tooltip.stories.tsx
@@ -3,12 +3,14 @@ import React from 'react';
 import Tooltip from './Tooltip';
 import Paragraph from '../Typography/Paragraph';
 import TextLink from '../TextLink';
+import notes from './README.md';
 
 export default {
   title: 'Components/Tooltip',
   component: Tooltip,
   parameters: {
     propTypes: Tooltip['__docgenInfo'],
+    notes,
   },
   argTypes: {
     content: { control: 'text' },
@@ -26,14 +28,17 @@ export const basic = (args: { content: string }) => {
   );
 };
 basic.args = {
-  content: (
-    <>
-      I am a<br />
-      multiline <b>Tooltip</b>
-      <br />
-      🙌
-    </>
-  ),
+  content: 'I am a Tooltip 🙌',
+};
+const basicSourceCode = `<Tooltip content="I am a Tooltip 🙌">
+  <TextLink>Hover me</TextLink>
+</Tooltip>`;
+basic.parameters = {
+  docs: {
+    source: {
+      code: basicSourceCode,
+    },
+  },
 };
 
 export const withHtml = (args: { content: string }) => {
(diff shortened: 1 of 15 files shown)


packages/forma-36-react-components/src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -1,62 +1,107 @@
...
+  title: 'Components/Tooltip',
+  component: Tooltip,
+  parameters: {
     propTypes: Tooltip['__docgenInfo'],
-    component: Tooltip,
-  })
-  .add('default', () => (
-    <div>
-      <Tooltip
-        place={select(
```
mshaaban0, line 16: @gui-santos will this make it a breaking change?
What if I want the tooltip to always show on the right?

gui-santos: good point! I will take a look at how to tell popper this

m10l (commented):
Awesome <3 - couple of notes:

+1 to the placement question, we should either have 1:1 parity with the previous API, or, keep the old Tooltip component and move this new Popper-powered version to alpha components and either switch over when parity is reached, or make it a breaking change.

The CopyButton Story doesn't work as expected anymore: https://deploy-preview-567--forma-36-react-components.netlify.app/?path=/story/components-copybutton--default - previously it updated the tooltip text once the button had been clicked to notify the user that the content had been copied to their clipboard

gui-santos:
@mshaaban0 and @m10l I've just pushed the implementation of the `place` prop

I also extended it to accept poppper’s options `auto; auto-start, auto-end, top-start, top-end, left-start, left-end, etc...`

Check this example on how they look like

Screenshot 2020-09-22 at 11 53 28
left-end

Screenshot 2020-09-22 at 11 54 12
left

Screenshot 2020-09-22 at 11 53 56
left-start

tell me what you think 👌

gui-santos:
> The CopyButton Story doesn't work as expected anymore: https://deploy-preview-567--forma-36-react-components.netlify.app/?path=/story/components-copybutton--default - previously it updated the tooltip text once the button had been clicked to notify the user that the content had been copied to their clipboard

I think this is something that I wanted help to understand. We have those onSOMETHING props, but we use onMouseEnter and onMouseLeave to set the visibility of the tooltip

My question is: should those props overwrite the default events? If yes, does this mean that whoever passes a custom `onMouseEnter` also needs to pass a `isVisible` prop or it won't work, right? Or should we execute the custom functions after the default behaviour like a callback basically?

Anyway, I will check the CopyButton story to understand this better! Thanks for pointing that out I totally missed it 👌

gui-santos:
@m10l I think the story you mentioned is not working in production
https://f36-storybook.contentful.com/?path=/story/components-copybutton--default

so it's not the Tooltip’s fault that it's not changing to "copied" 

I debugged a little bit and it seems to me that the state of the CopyButton never updates 👀

mshaaban0 (approved):
Looks great to me, thank you @gui-santos 🦋

burakukula (approved):
🚢 is my favourite emoji

PR #569, opened 2020-09-21
fix: show image only there's src

suevalov:
# Purpose of PR

* `Asset` shouldn't render an image tag if `src` attribute is nullable (for example, an empty string).
* Adds padding to title div, so there not situation like this:

Screenshot 2020-09-21 at 18 37 06

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Asset/Asset.css
+++ b/packages/forma-36-react-components/src/components/Asset/Asset.css
@@ -1,5 +1,6 @@
 @import 'settings/colors';
 @import 'settings/typography';
+@import 'settings/dimensions';
 @import 'settings/transitions';
 
 .Asset {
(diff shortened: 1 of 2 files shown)


packages/forma-36-react-components/src/components/Asset/Asset.css
```diff
@@ -62,6 +63,8 @@
     1rem * (35 / var(--font-base-default))
   ); /* 35 value comes from font-size multipled by line height multipled by 2 (for limiting text to 2 lines) */
   word-wrap: break-word;
+  padding-left: var(--spacing-s);
+  padding-right: var(--spacing-s);
```
mshaaban0, line 67: I think the padding belongs to `.Asset__asset-container__title` line [#78](https://github.com/contentful/forma-36/pull/569/files#diff-751197682efe2018075d26f9d0715ba4R78)

`.Asset__title-container__title` targets another variant that already has padding

Screenshot 2020-09-21 at 22 48 12

The naming isn't so great for this one 🙃

suevalov: Updated

mshaaban0 (approved):
🍬

PR #572, opened 2020-09-23
fix: Flex grow instead of fixed-height

kdamball:
# Purpose of PR

If by chance you dont include a header - workbench has a gap as content doesn't take the full available real-estate. Let's use flex-grow and make sure it will automatically take the full available space.

## PR Checklist

- [x] Tests are passing


gui-santos (approved):
🚢

PR #574, opened 2020-09-25
fix(components) make defaultProps optional

andipaetzold:
# Purpose of PR

The exported component prop types require the default props to be set. These should be optional.

E.g. the `ButtonProps` type requires a prop `loading` which is actually optional because the `defaultProps` are set in the component.

This doesn't cause issues when you simply use `<Button />` in your code because TypeScript's JSX compiler handles the types of `defaultProps` (https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#support-for-defaultprops-in-jsx).
The described problem only causes issues when you are using `ButtonProps` or `React.ComponentProps<typeof Button>` as the special JSX handling is not running there.

At the moment the following is not possible as you have to pass all default props into `HeartButton`:
```tsx
function HeartButton(props: ButtonProps) {
    return <Button {...props}>❤️ {props.children} ❤️</Button>
}

<HeartButton /> // does not work
```

I made `defaultProps` optional in all components.

This PR might be breaking for TypeScript users that already use exported types. As they are currently broken with the required `defaultProps`, I assume that nobody will be affected.

The checks of this PR are not successful as changing the prop types to optional also makes them optional within the component using `this.props`. As you can read [here](https://github.com/microsoft/TypeScript/issues/30251#issuecomment-471328293) there doesn't seem to be a pretty and perfect solution for typing `defaultProps`. Still, I think that the exported types should to be correct in order for the components to be used within TypeScript. If the @forma-36 team agrees with this approach, I would move forward fixing the types within the components using non-null type assertions (`!`).

The alternative would be to at least fix the explicitly exported types and e.g. change the `Button` component to the following
```typescript
export type ButtonProps = {
  ...
} & Partial<typeof defaultProps>;

const defaultProps = {
  ...
};

export class Button extends Component<ButtonProps & typeof defaultProps> {
  static defaultProps = defaultProps;
```
That way we would avoid having to change the logic and loose typesafety within the component but also have a way to compose new components with the existing ones like `HeartButton`.

## PR Checklist

- [x] Tests are passing


gui-santos:
Hi @andipaetzold thanks a lot for the proposal!

I have a question because I'm not familiar with this pattern that you are proposing and I've been refactoring the typing of the props in some components when I need to change them. What I was doing is this:

```ts
export interface ComponentProps {
  required: any;
  optional?: any;
}

// ...

function Component(props: ComponentProps) {
  // ...
}
Component.defaultProps = {
  optional: 'default value'
}
```

I don't know if this solves the same problem that you are trying to solve in your proposal because I'm not familiar with it. Is it? Am I saying something completely out of topic here? hahaha

andipaetzold:
I just noticed that I assumed that `ComponentProps` and `typeof defaultProps` have distinct props. I somehow missed that the default values in `defaultProps` are already typed in `ComponentProps`. All the `& typeof defaultProps` does is changing the already existing props from optional to required. So, the effect I wanted to have with `& Partial<typeof defaultProps>` could also be achieved by fully removing the type override.

I think that is exactly approach, @gui-santos , right?

For Button this would mean:
```typescript
export type ButtonProps = {
  icon?: IconType;
  ...
  loading: boolean;
...
};

const defaultProps = {
  loading: false,
...
};

export class Button extends Component<ButtonProps> {
```

gui-santos:
> I think that is kind of your approach, @gui-santos , right?

@andipaetzold YES! I think `& typeof defaultProps` is the real villain here, but I never had the courage of migrating everything at one go hahaha but we can do it! Thx

> For Button this would mean:
> 
> ```ts
> export type ButtonProps = {
>   icon?: IconType;
>   ...
>   loading?: boolean;
> ...
> };
> 
> const defaultProps = {
>   loading: false,
> ...
> };
> 
> export class Button extends Component<ButtonProps> {
>   static defaultProps = defaultProps;
> ```

Yes, but I still believe using [interfaces](https://www.typescriptlang.org/docs/handbook/interfaces.html) are better than types for the props. And if you want your defaultProps to be typed you can `const defaultProps: ButtonProps` but I think typescript will infer those types in your Button

cool stuff 🙌

andipaetzold:
> Yes, but I still believe using [interfaces](https://www.typescriptlang.org/docs/handbook/interfaces.html) are better than types for the props.

@gui-santos I also like those more :)

I'll give it a go and update the PR. Then we'll see whether some more adjustments are required.

gui-santos:
Sweet! Thanks @andipaetzold for bringing this up and for taking care of this 🙌

andipaetzold:
@gui-santos I adjusted the PR as described. Now there is only the issue left that within the components some props are optional even though they are always defined by `defaultProps`. I'd add some non-null type assertions there. That's not really pretty for us developing the components but I think the only we can do.

gui-santos:
> @gui-santos I adjusted the PR as described. Now there is only the issue left that within the components some props are optional even though they are always defined by `defaultProps`. I'd add some non-null type assertions there. That's not really pretty for us developing the components but I think the only we can do.

Hey, @andipaetzold I think I'm missing something in this use case. Look what I did to try to replicate your HeartButton:

This is the props definition and the defaultProps in Button.js:
```jsx
export interface ButtonProps {
  // ... all props here are optional
  buttonType?:
    | 'primary'
    | 'positive'
    | 'negative'
    | 'warning'
    | 'muted'
    | 'naked';
  type?: 'button' | 'submit' | 'reset';
}

const defaultProps: ButtonProps = {
  // ...
  buttonType: 'primary',
  type: 'button',
};

export class Button extends Component<ButtonProps> {
  static defaultProps = defaultProps;
  // ...
}
```

this is the story I wrote to test it:
```jsx
() => {
    const HeartButton = ({ children, ...otherProps }: ButtonProps) => {
      return <Button {...otherProps}>❤️  {children} ❤️</Button>;
    };

    return <HeartButton>This is a HeartButton</HeartButton>;
}
```

This is what I got in storybook:
Screenshot 2020-10-01 at 10 43 30

This worked ☝️ ! There are no complaints in the linter or compiler or anywhere... 🤔 

So I think I still don't understand why we need the non-null assertions and the `const defaultProps: Partial<ComponentProps>`. I'm not a TS expert so I might be missing something. Maybe I'm also not reproducing the use case you explained before correctly, so help me! hahahaha

Thanks for the patience, bruh 🙏

taye:
@gui-santos My thinking is that when a component requires a prop to be passed, then it doesn't need a default for it. In that case it should be optional or excluded in the `defaultProps`. I guess a more accurate type definition for `defaultProps` would be something like:

```ts
type RequiredField = 'field1' | 'field2'
const defaultProps: ButtonProps & Partial<Pick<ButtonProps, RequiredField>> = {...}
```

:grimacing:

gui-santos:
> @gui-santos My thinking is that when a component requires a prop to be passed, then it doesn't need a default for it. In that case it should be optional or excluded in the `defaultProps`. I guess a more accurate type definition for `defaultProps` would be something like:
> 
> ```ts
> type RequiredField = 'field1' | 'field2'
> const defaultProps: ButtonProps & Partial<Pick<ButtonProps, RequiredField>> = {...}
> ```
> 
> 😬

OHHHHH I GET IT NOW! thanks @taye 

I found [this on typescript cheatsheet](https://github.com/typescript-cheatsheets/react#typing-defaultprops) hahaha... what do you both think? @taye and @andipaetzold

taye:
> I found this on typescript cheatsheet hahaha... what do you both think?

Ah that explains where the `& typeof defaultProps` came from :D. That makes sense. Though for values that can be multiple types, we'd have to do: 

```ts
const defaultProps = {
  width: 50 as string | number
  ...
}
```

I think just setting the types to be partial as it currently is in this PR is a good compromise of simplicity and accuracy.

And thanks for sharing that cheatsheet. It's got some really interesting/magical stuff

gui-santos:
> > I found this on typescript cheatsheet hahaha... what do you both think?
> 
> Ah that explains where the `& typeof defaultProps` came from :D. That makes sense. Though for values that can be multiple types, we'd have to do:
> 
> ```ts
> const defaultProps = {
>   width: 50 as string | number
>   ...
> }
> ```
> 
> I think just setting the types to be partial as it currently is in this PR is a good compromise of simplicity and accuracy.
> 
> And thanks for sharing that cheatsheet. It's got some really interesting/magical stuff

hey @taye, @andipaetzold and I are pairing today on this... do you want to join us?

And regarding the solution with `& typeof` I tested it in my story that I created to reproduce @andipaetzold and got this:
Screenshot 2020-10-02 at 12 12 32

But if instead of creating the defaultProps const, we just pass the default props directly in the class this error does not happen. (I also used just a interface and no typeof solution or the 'as string | number' you mentioned) 

Let's pair and find a solution that is good and easy for everyone...

andipaetzold:
Next steps after today's meeting:
* Use non-null type assertion with the components
* Add eslint rule to prevent people from using non-null type assertions

andipaetzold:
@taye @gui-santos I've added the discussed changes. In some places I replaced the noop default prop with `?.`

packages/forma-36-react-components/src/components/Checkbox/Checkbox.tsx
```diff
@@ -1,7 +1,8 @@
 import React, { Component } from 'react';
 import ControlledInput, { ControlledInputPropTypes } from '../ControlledInput';
 
-export type CheckboxProps = ControlledInputPropTypes & typeof defaultProps;
+export type CheckboxProps = ControlledInputPropTypes &
+  Partial<typeof defaultProps>;
```
taye, line 5: This still has the issue of the restrictive types from `defaultProps` overriding the declared types. Since the props types are the same as in the base component so you can do

```ts
// eslint-disable-next-line @typescript-eslint/no-empty-interface
export interface CheckboxProps extends ControlledInputPropTypes {};

const defaultProps: Partial<ControlledInputPropTypes>
```

The same for `RadioButton`, `SkeletonBodyText`, and `SkeletonDisplayText`.

taye (changes requested):

packages/forma-36-react-components/src/components/Checkbox/Checkbox.tsx
andipaetzold: Oh - seems like I missed some components...

taye (approved):
:tada:

burakukula (approved):
This is amazing! thank you very much for this great contribution and hard work on this one! 🤝🎉🙌

gui-santos (approved):
Great job! 🙌

PR #576, opened 2020-09-29
feat(EntryCard): add size=auto to EntryCard

anho:
# Purpose of PR

Introduce `size=auto` to EntryCard to achieve minimal height based on the props provided. This will maintain previous behaviour and gives us the opportunity to render long lists in a more narrow fashion.

### DIFF
--- a/packages/forma-36-react-components/src/components/Card/EntryCard/EntryCard.tsx
+++ b/packages/forma-36-react-components/src/components/Card/EntryCard/EntryCard.tsx
@@ -12,6 +12,8 @@ import styles from './EntryCard.css';
 
 export type EntryCardStatus = 'archived' | 'changed' | 'draft' | 'published';
 
+export type EntryCardSize = 'default' | 'small' | 'auto';
+
 export type EntryCardPropTypes = BaseCardProps & {
   /**
    * The title of the entry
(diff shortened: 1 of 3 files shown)


packages/forma-36-react-components/src/components/Card/EntryCard/EntryCard.tsx
```diff
@@ -234,11 +237,11 @@ export class EntryCard extends Component<EntryCardPropTypes> {
                   <div className={styles.EntryCard__body}>
                     {title && this.renderTitle(title)}
                     {description &&
-                      size === 'default' &&
+                      actualSize === 'default' &&
                       this.renderDescription(description)}
                   </div>
                   {thumbnailElement &&
-                    size === 'default' &&
+                    actualSize === 'default' &&
```
suevalov, line 244: As part of the story we should be able to show the thumbnail even if the card is small. People rely on images when they search their content.

gui-santos (approved):
🚢 

and because you added a new story,
100 points for Griffindor 🦁

burakukula (approved):
👏

anho:
@gui-santos @burakukula I provided a tiny bit more info in the opening comment due to some more changes.

suevalov (approved):

PR #577, opened 2020-09-30
fix(asset): handle overflow in asset title

anho:
# Purpose of PR

Fix issue with longer titles of an asset as seen here: 
image


suevalov (approved):

burakukula (approved):
Looks good! 🤝

PR #579, opened 2020-09-30
fix(tooltip): return children if no content provided and add z-index

mshaaban0:
# Purpose of PR

- Add tooltip `z-index` value from tokens
- Change tokens `z-index` values to `number` instead of `string` to generate the correct type

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
+++ b/packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
@@ -41,7 +41,7 @@ export interface TooltipProps {
   /**
    * It sets a max-width for the Tooltip
    */
-  maxWidth?: CSS.MaxWidthProperty<string>;
+  maxWidth?: number | CSS.MaxWidthProperty<string>;
   /**
    * Function that will be called when target gets blurred
    */
@@ -144,7 +144,7 @@ export const Tooltip = ({
   };
 
   const contentMaxWidth =
-    maxWidth && Number.isNaN(maxWidth) ? `${maxWidth}px` : maxWidth;
+    typeof maxWidth === 'string' ? maxWidth : `${maxWidth}px`;
 
   const contentStyles: CSSProperties = {
     zIndex: tokens.zIndexTooltip,
--- a/packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
+++ b/packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
@@ -4,12 +4,15 @@ import React, {
   useRef,
   MouseEvent,
   FocusEvent,
+  CSSProperties,
 } from 'react';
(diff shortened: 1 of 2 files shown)


packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
```diff
@@ -78,7 +80,7 @@ interface ArrowPositionState {
 export const Tooltip = ({
   children,
   className,
-  containerElement: ContainerElement,
+  containerElement: ContainerElement = 'span',
```
gui-santos, line 83: Why do we need to declare the default of this prop again?

packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
```diff
@@ -39,7 +41,7 @@ export interface TooltipProps {
   /**
    * It sets a max-width for the Tooltip
    */
-  maxWidth?: number | string;
+  maxWidth?: number | CSS.MaxWidthProperty<string>;
```
gui-santos, line 44: nice! 👌

gui-santos (approved):
🤝 

Noice!

packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
mshaaban0: Good point, no. I'll remove it from `defaultProps`

mshaaban0: when we declare it only in `defaultProps` typescript complains

Screenshot 2020-10-01 at 10 02 16

gui-santos: I see! I actually prefer declaring the way you did hahaha 
But I replicate the Component.defaultProps pattern that I saw around the codebase

mshaaban0: @gui-santos do we keep it in `defaultProps` so we can keep the storybook control, or no need 😬?

gui-santos: I think storybook will understand the defaults if you pass them the way you did.... but I'm not sure. I'm 80% sure hahaha

gui-santos: test it! If it explodes them we keep them in both places 🤷

mshaaban0: Hahaha, your 80% was quite enough, it worked, storybook translated it into a control, but didn't pre-fill it with `span`, I removed it anyway 👍

PR #582, opened 2020-10-02
fix: improvements in the Pill docs and new storybook format

burakukula:
Ahoy F36!

Some improvements in the Pill doc. I also transformed stories into the new format.
@domarku I was wondering if maybe we would like to extend this component to get some colors? I think I saw something like this in the webapp. Just a 💭.

## PR Checklist

- [ ] Tests are passing


packages/forma-36-react-components/src/components/Pill/README.md
```diff
@@ -0,0 +1,68 @@
...
+    <Icon
+      icon="ArrowUp"
+      color="positive"
+      style={{ padding: '0.375rem 0.625rem' }}
+    />
+  }
+/>
+
+```
+## Best practices
```
burakukula, line 58: do we have any best practices around this component?

packages/forma-36-react-components/src/components/Pill/README.md
```diff
@@ -0,0 +1,68 @@
...
+ - use Pill to represent data like contact details
+ - display data in the Pill like tagging
+ - it can be used for web content that needs to be labeled, categorized, or organized using keywords that describe them.
+
+## Content recommendations:
```
burakukula, line 63: or maybe some cool content recommendations?

packages/forma-36-react-components/src/components/Pill/README.md
```diff
@@ -0,0 +1,68 @@
...
+ - Be sure the pill text is clear and concise.
+
+## Accessibility
```
burakukula, line 66: or some awesome accessibility guideline?

packages/forma-36-react-components/src/components/Pill/Pill.tsx
```diff
@@ -56,6 +56,7 @@ export class Pill extends Component<PillProps> {
         {onClose && (
           <button
             type="button"
+            aria-label="button"
```
burakukula, line 59: catched thx to storybook accessibility tab. 🎉

mshaaban0, line 59: `aria-label` must describe what the button does, which I believe is "close" in this case, right?

packages/forma-36-react-components/src/components/Pill/Pill.stories.tsx
```diff
@@ -4,53 +4,77 @@ import { action } from '@storybook/addon-actions';
...
+basic.args = { label: 'mike.mitchell@contentful.com' };
+
+onDragAndOnClose.args = {
+  label: 'guilherme.santos@contentful.com ',
+  onClose: action('clicked'),
+  onDrag: action('dragged'),
+};
+
+CustomHandleComponent.args = {
+  label: 'moe.shaaban@contentful.com',
```
mshaaban0, line 75: Should we replace these with dummy emails? GDPR 🚔

mshaaban0 (approved):
🍬

packages/forma-36-react-components/src/components/Pill/Pill.stories.tsx
burakukula: @mshaaban0 hehe, yeah, this is what I thought. It was Johannes email before. I will change it to something else. on my way 🚀

domarku:
Hey @burakukula you membered that correct! We use pills in Contentful for tags and we've added color to indicate change and status of the tag:
tags

So, aside from the default gray one, we could add two more states:
- Active (`--color-orange-mid`, text color dark)
- Disabled (opacity at 0.5)

burakukula:
> Hey @burakukula you membered that correct! We use pills in Contentful for tags and we've added color to indicate change and status of the tag:
> tags
> 
> So, aside from the default gray one, we could add two more states:
> 
> * Active (`--color-orange-mid`, text color dark)
> * Disabled (opacity at 0.5)

Thanks a lot @domarku! I will add this to our Pill component. 🎉

domarku:
I'm not sure if we should just go ahead and add it to the library, considering it's only used like that once in the product. Also, I'm not sure if the color should be orange mid or if it should be customizable. Perhaps better to move this to an issue and just carry on with what you intended to do in this PR without that addition. Wdyt?

burakukula:
@domarku your right, that's a better approach. Maybe we could bring this into designers somehow to get the feedback if that might be useful in some other parts of the product?

PR #584, opened 2020-10-05
feat(react-components): add DateTime and RelativeDate components

freakyfelt:
# Purpose of PR

Adds new DateTime and RelativeDate components that follow the Forma36 Date Time guidelines

* add new dayjs dependency (2kb)
  * this is a much more minimal package that has a similar interface to moment with explicit extensions to pick and choose from, keeping the footprint small and easily tree shakable.
* add DateTime that renders an absolute DateTime
* add RelativeDate that renders a time relative to a baseDate

Forma36 Date Time guidelines: https://f36.contentful.com/guidelines/date-and-time/

**NOTE**: might need to add an interval to the RelativeDate component to keep the value fresh over time

## PR Checklist

- [x] Tests are passing


freakyfelt:
Not sure why it claims that `date` is not a valid knob. I verified locally that it shows up?

packages/forma-36-react-components/src/components/DateTime/DateTime/DateTime.stories.tsx
```diff
@@ -0,0 +1,41 @@
...
+const formatOptions: DateTimeFormat[] = [
+  'FULL',
+  'DATE_ONLY',
+  'TIME_ONLY',
+  'WEEKDAY_DATE',
+];
+
+const defaultDate = new Date('2019-08-12T08:00:00Z');
+
+storiesOf('Components/DateTime', module)
```
burakukula, line 17: @freakyfelt thx a lot for this contribution! 🎉
As we recently bumped storybook to the new version, it is possible now to use new format for the story. We would slowly refactor stories into that. Maybe you would like to use the new one?
More info around this https://storybook.js.org/docs/react/api/csf/
We have couple of components in the new format, if you would like to have a look https://github.com/contentful/forma-36/blob/master/packages/forma-36-react-components/src/components/Accordion/Accordion.stories.tsx

packages/forma-36-react-components/src/components/DateTime/index.ts
```diff
@@ -0,0 +1,8 @@
+import { DateTime } from './DateTime/DateTime';
+import { RelativeDate } from './RelativeDate/RelativeDate';
+import * as datetime from './dateUtils';
+export {
+  datetime,
```
freakyfelt, line 5: Not sure the best way to group all of these utility functions, so feedback is welcome

packages/forma-36-react-components/src/components/DateTime/DateTime/DateTime.stories.tsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+
+import DateTime, { DateTimeProps } from './DateTime';
+
+const exampleDate = '2020-08-13T13:45:56.0123Z'
+
+import notes from '../README.md';
+
+export default {
```
burakukula, line 9: @freakyfelt thank you! 🎉

packages/forma-36-react-components/src/components/DateTime/README.md
```diff
@@ -0,0 +1,73 @@
...
+| `FULL` | `12 Aug 2020 at 8:00 am` |
+| `DATE_ONLY` | `12 Aug 2020` |
+| `TIME_ONLY` | `8:00 am` |
+| `WEEKDAY_DATE` | `Wed, 12 Aug` |
+
+**Example Usage**
+
+```ts
+import React from 'react';
+import { DateTime } from '@contentful/forma36-react-components';
```
burakukula, line 22: @freakyfelt is that example complete? 🤔

freakyfelt: lolwups

mshaaban0 (approved):
@freakyfelt thank you for implementing this! Functionality looks great to me.

I see that we have multiple styles for displaying dates in the web app, @domarku @dominicbonnice @joaoramos maybe this is a good opportunity to unify the styles of how we display dates

image

image

packages/forma-36-react-components/src/components/DateTime/DateTime/DateTime.tsx
```diff
@@ -0,0 +1,50 @@
...
+   **/
+  format?: DateTimeFormat;
+  className?: string;
+  testId?: string;
+}
+
+/**
+ * Formats a JS Date object to an absolute date and time inside of a <time /> tag
+ */
+export const DateTime: React.FC<DateTimeProps> = ({
```
freakyfelt, line 27: @mshaaban0 starting a threaded conversation here re:styles on these components (thanks weird Github PR commenting scheme)

As we discussed in Slack I feel like these two components shouldn't have any styles applied to them, instead receiving whatever styles parent components define. In a sense these are plumbing components that can have porcelain applied by whatever context they exist in (Card, etc). Thoughts?

Regardless I'd rather get the issue of style or no style sorted out before merging as this will get ugly if assumptions about style are made that then get violated

packages/forma-36-react-components/src/components/DateTime/DateTime/DateTime.tsx
```diff
@@ -0,0 +1,50 @@
...
+  date,
+  format,
+  className,
+  testId,
+}) => {
+  format = format ?? 'FULL';
```
gui-santos, line 33: I think you could pass 'FULL' as default prop instead of checking if it exist or not. Like:

```jsx
export const DateTime: React.FC<DateTimeProps> = ({
  date,
  format = 'FULL',
  className,
  testId,
}) => {
// ...
}
```

gui-santos (approved):
looks good to me! I left a tiny comment tough :) 

thanks for the contribution 🙌

packages/forma-36-react-components/src/components/DateTime/index.ts
burakukula: This looks good, in my opinion 🤝

burakukula (approved):

packages/forma-36-react-components/src/components/DateTime/DateTime/DateTime.tsx
domarku: > As we discussed in Slack I feel like these two components shouldn't have any styles applied to them, instead receiving whatever styles parent components define. In a sense these are plumbing components that can have porcelain applied by whatever context they exist in (Card, etc). Thoughts?

I would agree!

PR #587, opened 2020-10-08
fix: return target wrapper when tooltip has no content

gui-santos:
My bad! I thought we could return the children of the tooltip in case the prop `content` was undefined, but some people rely on the `targetWrapperClassName` prop

So this makes us return the target and its container to fix that

I'm also proposing to change the default value of `place` to 'auto', because I  had to change this in user_interface more than once. Initially, I did the default value 'bottom' to preserve the old API, but it is messing up positioning in some cases where the tooltip does not have room to be placed. And 'auto' is always positioning the tooltip the best way

... or maybe we update this to use portals but that's a bigger task that we can do it in another PR


mshaaban0 (approved):
👍

PR #588, opened 2020-10-09
fix: radio button example

andipaetzold:
# Purpose of PR

There is an issue in the current radio button example:

- `checked` is a boolean and not a string
- we should be consistent with the quotes in jsx

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-website/src/pages/components/radio-button/index.mdx
+++ b/packages/forma-36-website/src/pages/components/radio-button/index.mdx
@@ -5,15 +5,16 @@ status: 'stable'
 github: 'https://github.com/contentful/forma-36/tree/master/packages/forma-36-react-components/src/components/RadioButton'
 storybook: 'https://f36-storybook.contentful.com/?path=/story/components-radiobutton--default'
 ---
+
 Detailed documentation will be provided soon. For now, please have a look into the storybook for implementation details.
 
 ```jsx
 <>
   <RadioButtonField
-    labelText='Radio button label'
+    labelText="Radio button label"
     helpText="This is a helptext"
     name="someOption"
-    checked="true"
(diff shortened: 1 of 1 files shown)


packages/forma-36-website/src/pages/components/radio-button/index.mdx
```diff
@@ -5,17 +5,18 @@ status: 'stable'
...
 
 ```jsx
 <>
   <RadioButtonField
-    labelText='Radio button label'
+    labelText="Radio button label"
     helpText="This is a helptext"
     name="someOption"
-    checked="true"
+    checked={true}
```
gui-santos, line 17: What do you think about going one step further and do this?

```jsx
<RadioButtonField
  // ...
  checked
```

gui-santos: Since the value is true, we only need to pass a value when it's false

gui-santos (approved):
LGTM! And you are right, we should be consistent with the quotes

I made one suggestion about the boolean, not a blocker

I just think boolean props don't need the `boolean={true}` when they are true, just `boolean`

packages/forma-36-website/src/pages/components/radio-button/index.mdx
andipaetzold: The very related checkbox for example has `checked={true}` in the example instead of just `checked`.
https://f36.contentful.com/components/checkbox/

That's why I didn't shorten it.

gui-santos: Ok cool! No problem

we can shorten all of them later if we want, maybe :)

PR #589, opened 2020-10-09
fix(select): specify element type for event handlers

andipaetzold:
# Purpose of PR

Specify element type of change, blur and focus event handlers as `HTMLSelectElement`.

This currently throws an error in TypeScript because `Property 'value' does not exist on type 'EventTarget & Element'.`:
```jsx
<Select
  value={value}
  onChange={(e) => {
    console.log(e.target.value);
  }
>
  <Option />
</Select>
```
At the moment you have to write the following to not get the error:
```jsx
<Select
  value={value}
  onChange={(e: ChangeEvent<HTMLSelectElement>) => {
    console.log(e.target.value);
  }
>
  <Option />
</Select>
```
With this PR you don't have to explicitly type the parameter of the `onChange` method.

This problem probably exists for other components as well.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Select/Select.tsx
+++ b/packages/forma-36-react-components/src/components/Select/Select.tsx
@@ -15,9 +15,9 @@ export interface SelectProps {
   hasError?: boolean;
   value?: string;
   isDisabled?: boolean;
-  onChange?: ChangeEventHandler;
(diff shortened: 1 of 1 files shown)


gui-santos (approved):
Beautiful!

PR #597, opened 2020-10-20
feat(dropdown): use Popper.js for a proper positioning

### LINKED ISSUE #361: Proposal: Use tether.io for attaching absolutely positioned elements to a target
Hi all,

## Description
There is an easy win out there called [tether](http://tether.io/docs/welcome/) see [docs](http://tether.io/) and [source](https://github.com/shipshapecode/tether). The library focuses on the very specific problem of attaching an absolutely positioned element to another. 

### Pros
* Solves one problem and it does it [well](http://tether.io/overview/why_you_should_use_tether/)
* Well maintained since major [components](http://tether.io/examples/projects_using_tether/) depend on it
* Great API and docs

### Cons
* Yet another dependency
* Adds 5kb gzipped

Since I'm working on porting the Autocomplete we can have an example using tether which might help towards a decision.

Your awesome point of view is an absolute must ;)
Aris

denkristoffer:
# Purpose of PR

This PR refactors the Dropdown component to use [Popper.js](https://popper.js.org/) instead of the current, custom positioning calculation.

Resolves #361 

# Todo

- [x] Dynamic content
- [x] Don't re-open when toggling on the `toggleElement`
- [ ] ~~What to do about the `isAutoalignmentEnabled` prop? It doesn't serve much of a purpose with Popper doing the positioning automatically. Is it considered a breaking change to remove it?~~

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Dropdown/Dropdown.stories.tsx
+++ b/packages/forma-36-react-components/src/components/Dropdown/Dropdown.stories.tsx
@@ -83,6 +83,7 @@ function ScrollableStory() {
     <Dropdown
       isOpen={isOpen}
       onClose={() => setOpen(false)}
+      isAutoalignmentEnabled={boolean('isAutoalignmentEnabled', true)}
       position={select(
         'position',
         {
--- a/packages/forma-36-react-components/src/components/Dropdown/DropdownContainer/DropdownContainer.tsx
+++ b/packages/forma-36-react-components/src/components/Dropdown/DropdownContainer/DropdownContainer.tsx
@@ -1,252 +1,131 @@
-import React, { Component } from 'react';
+import React, { forwardRef, useCallback, useEffect, useRef } from 'react';
 import cn from 'classnames';
 import ReactDOM from 'react-dom';
-import InViewport from '../../InViewport';
-import { positionType, AnchorDimensionsAndPositonType } from '../Dropdown';
 
+import { positionType } from '../Dropdown';
 import styles from './DropdownContainer.css';
 
-export interface DropdownContainerProps {
-  onClose?: Function;
-  dropdownAnchor?: HTMLElement | null;
-  className?: string;
+export interface DropdownContainerProps
+  extends React.HTMLAttributes<HTMLElement> {
   children?: React.ReactNode;
-  testId?: string;
-  openSubmenu?: (value: boolean) => void;
-  anchorDimensionsAndPositon?: AnchorDimensionsAndPositonType;
-  position: positionType;
+  className?: string;
   getRef?: (ref: HTMLElement | null) => void;
+  isOpen: boolean;
+  onClose?: Function;
+  openSubmenu?: (value: boolean) => void;
+  position?: positionType;
   submenu?: boolean;
-  width?: number | false;
-  isAutoalignmentEnabled?: boolean;
+  testId?: string;
 }
 
-export interface DropdownState {
-  dropdownDimensions: {
-    width: number;
-    height: number;
-  };
-  position: positionType;
-}
-
-const defaultProps: Partial<DropdownContainerProps> = {
-  testId: 'cf-ui-dropdown-portal',
-  position: 'bottom-left',
-  submenu: false,
-  isAutoalignmentEnabled: true,
-};
-
-class DropdownContainer extends Component<
-  DropdownContainerProps,
-  DropdownState
-> {
-  static defaultProps = defaultProps;
-
-  portalTarget = document.createElement('div');
-  dropdown: HTMLElement | null = null;
-  lastOverflowAt: string | null = null;
-
-  state = {
-    position: this.props.position,
-    dropdownDimensions: {
-      width: 0,
-      height: 0,
-    },
-  };
-
-  componentDidMount() {
-    document.body.appendChild(this.portalTarget);
-    if (this.dropdown) {
-      const dropdownRect = this.dropdown.getBoundingClientRect();
-      this.setState({
-        dropdownDimensions: {
-          width: dropdownRect.width,
-          height: dropdownRect.height,
-        },
-      });
-    }
-    document.addEventListener('mousedown', this.trackOutsideClick, true);
-    this.props.getRef?.(this.dropdown);
-  }
-
-  componentDidUpdate(
-    prevProps: DropdownContainerProps,
-    prevState: DropdownState,
-  ) {
-    if (!this.dropdown) {
-      return;
-    }
-
-    const dropdownRect = this.dropdown.getBoundingClientRect();
-
-    if (
-      dropdownRect.width !== prevState.dropdownDimensions.width ||
-      dropdownRect.height !== prevState.dropdownDimensions.height
-    ) {
-      this.setState({
-        dropdownDimensions: {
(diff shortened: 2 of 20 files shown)


m10l:
Awesome 😻, looks like there are some issues with the `react-timepicker` package though due to React mismatch?

```
lerna ERR! yarn run test exited 1 in '@contentful/[secure]-react-timepicker'
lerna ERR! yarn run test stdout:
$ tsdx test --env=jsdom
  console.error node_modules/jsdom/lib/jsdom/virtual-console.js:29
    Error: Uncaught [Error: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
    1. You might have mismatching versions of React and the renderer (such as React DOM)
    2. You might be breaking the Rules of Hooks
    3. You might have more than one copy of React in the same app
    See https://fb.me/react-invalid-hook-call for tips about how to debug and fix this problem.]
```

denkristoffer:
@m10l Yes, and weirdly, the error happens on Node 10 but not on Node 12 😩

m10l (approved):
Awesome work

packages/forma-36-react-components/src/components/Dropdown/DropdownContainer/DropdownContainer.tsx
```diff
@@ -1,252 +1,114 @@
...
-    }
-  };
-
-  getSubmenuClassNames = () =>
-    cn(
-      styles['DropdownContainer__submenu'],
-      styles[`DropdownContainer__container-position--${this.state.position}`],
-    );
+      return () => {
+        document.body.removeChild(portalTarget.current);
```
gui-santos, line 60: Maybe we should refactor the Tooltip to use Portals as well. But In another PR of course 😄

gui-santos (approved):
🚀

packages/forma-36-react-components/src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -18,7 +18,6 @@ function DefaultStory() {
       onClose={() => setOpen(false)}
       isFullWidth={boolean('isFullWidth', false)}
       key={Date.now()} // Force Reinit
-      isAutoalignmentEnabled={boolean('isAutoalignmentEnabled', true)}
```
burakukula, line 21: We might want to refactor it to the new format, but we can totally do it later.

burakukula (commented):
🎉 Great work! 
As you are removing `isAutoalignmentEnabled` would this mean minor change in the API?
On a separate PR we could do refactor to the story to new storybook format and add README for the component. But we can totally do it later.
Overall great! 🙌

DanweDE (commented):
> As you are removing isAutoalignmentEnabled would this mean minor change in the API?

This is kind of a breaking change in the sense that previously you could disable auto-alignment explicitly via `isAutoalignmentEnabled={false}`. If you rely on this behavior for some reason it's breaking for you.

@denkristoffer Searching for "isAutoalignmentEnabled" I noticed that we're relying on this in the web app in two places. If you really want to remove this prop, please make sure it's not strictly necessary there and that the UI around it still works fine without it.

burakukula:
@denkristoffer if this means breaking change, I think we should w8 and group this with other changes to release bigger major version. We have couple of things - We could move some of the alpha components to main build I think and I'm sure that we have couple of more things that might do into the next major release. Maybe a catch up regarding this topic would be a good thing to do? WDYT?

denkristoffer:
I've been digging into the uses of `isAutoAlignmentEnabled={false}` in the web app. Basically, what it does right now looks like this:

Screenshot 2020-10-23 at 15 48 13

And this is what it looks like with Popper. I'm not sure there's a way to force popper to ignore the viewport when positioning the element, but I'm looking into it.

Screenshot 2020-10-23 at 15 48 28

I think you can argue for and against whether this is a breaking change or not, but yeah, otherwise we can make some sort of 4.0 alpha?

denkristoffer:
@burakukula @DanweDE I think I managed to replicate disabling positioning with the `isAutoAlignmentEnabled` prop. I have un-deprecated it, so we can get this in without breaking changes. I think it's ready.

packages/forma-36-react-components/src/components/Dropdown/DropdownContainer/DropdownContainer.tsx
```diff
@@ -1,252 +1,131 @@
...
-      styles['DropdownContainer__submenu'],
-      styles[`DropdownContainer__container-position--${this.state.position}`],
+      [isOpen, onClose, dropdown.current],
     );
 
-  render() {
-    const { submenu, className, width, testId } = this.props;
+    useEffect(() => {
+      if (isOpen) {
+        document.body.appendChild(portalTarget.current);
```
mshaaban0, line 66: @denkristoffer if you can find a way to not to use portal it would be awesome. the problem with portal is keyboard focus order; upon open the dropdown content is at the end of the document, onClose the next keyboard element to be focused is the first element in the document. I hope that makes any sense, let's talk about it if not :)

denkristoffer: @mshaaban0 I agree, the portal should be removed by default. I think we should handle this separately in https://github.com/contentful/forma-36/issues/601 though, also to make sure this PR doesn't break any existing uses of the dropdown. Is that okay with you?

burakukula:
@denkristoffer ok, cool. 😎 
Sorry, maybe it's a silly question. I think the prop `isAutoalignmentEnabled` is still removed from the API in the PR? Or I am missing something? It's very possible that I am missing something. Hope you don't mind me asking. 🙃

denkristoffer:
@burakukula Of course not 🙂  The prop is being used in the Dropdown component where Popper is used, instead of the DropdownContainer where it was used before: https://github.com/contentful/forma-36/blob/dropdown-popper/packages/forma-36-react-components/src/components/Dropdown/Dropdown.tsx#L149-L164

I've also added it back to the story.

burakukula:
@denkristoffer super, thx! (the file was folded from me 😑 and I didn't see this change. sorry for confusion)

burakukula (approved):

packages/forma-36-react-components/src/components/Dropdown/DropdownContainer/DropdownContainer.tsx
```diff
@@ -1,252 +1,131 @@
...
-  };
-  position: positionType;
+  testId?: string;
 }
 
-const defaultProps: Partial<DropdownContainerProps> = {
-  testId: 'cf-ui-dropdown-portal',
-  position: 'bottom-left',
-  submenu: false,
-  isAutoalignmentEnabled: true,
```
DanweDE, line 36: @denkristoffer Since you made the prop work again, I think you'd want to keep the default value of `true` for this prop and also add it back to the `DropdownContainerProps` interface above, no?

denkristoffer: The functionality has been moved to `Dropdown.tsx`, so you shouldn't ever want/have to pass it to `DropdownContainer`.

PR #604, opened 2020-10-22
fix(AssetCard): dont force width onto small AssetCard

anho:
# Purpose of PR

This one is a bit of a questionable change. Small asset cards were given a fixed width but the default sized ones are not. To have a more consistent view/behaviour I dropped that. Looking for opinions.

### DIFF
--- a/packages/forma-36-react-components/src/components/Card/AssetCard/AssetCard.css
+++ b/packages/forma-36-react-components/src/components/Card/AssetCard/AssetCard.css
@@ -23,7 +23,6 @@
 
 .AssetCard--size-small .AssetCard__asset {
   height: calc(1rem * (142 / var(--font-base-default)));
-  width: calc(1rem * (142 / var(--font-base-default)));
 }
 
 .AssetCard--drag-active {


anho:
Another idea to make this more beautiful is to show the handle during loading:

image

Again, any opinions?

denkristoffer:
I think it looks good with the handle, but I don't know if it makes sense to indicate/allow action on an item you don't know what is yet. Is there a disabled handle? 😅

gui-santos:
I agree with @denkristoffer, it would be better if the handle also shows up disabled while loading

Maybe @domarku has also some opinion about this?

domarku:
I'm a bit out of context here and don't quite get the issue, but I would agree with @gui-santos that you should disable usually available actions (like dragging) on a loading element

denkristoffer:
@domarku The context here is that loading assets don't have a handle, so if you have a group of assets that have already loaded, and you try to add new ones, the two groups won't be the same width while loading:
fa562639-2311-49b0-822d-cb85f1bf1f88

One way to fix it is to add a handle to loading assets as well.

anho:
It's not a huge problem as of now, so closing it in favour of doing more important work ;) Thanks for the feedback so far everyone.

PR #607, opened 2020-10-23
fix: Grid shorthand conflicts

andipaetzold:
Happy Friday! It's me again 😄 

# Purpose of PR

As you can see in the [storybook](https://f36-storybook.contentful.com/?path=/story/alpha-grid-griditem--default), the GridItem is not displayed correctly. The default knobs are `columnStart=1; columnEnd=3; rowStart=1; rowEnd=4`.

Screenshot 2020-10-23 at 17 41 37

When you look at the rendered code, you see that the `style` prop is empty. When manually changing the `columnStart` and `rowStart` react prop with the react dev tools to another value and then back to the original, the GridItem is displayed correctly. The knobs are still the same.

Screenshot 2020-10-23 at 17 43 48

When doing this in dev mode locally, you get the following error message

> Updating a style property during rerender (gridColumn) when a conflicting property is set (gridArea) can lead to styling bugs. To avoid this, don't mix shorthand and non-shorthand properties for the same value; instead, replace the shorthand with separate values.

Issue is that `grid-area` is actually a shorthand for `gridColumn` and `gridRow` and it's not allowed to set the shorthand and the actual attribute at the same time.

With this PR, we only set a single attribute: `grid-area`. The props `columnStart`, `columnEnd`, `rowStart` and `rowEnd` are always combined and set as `grid-area`. The values fall back to 'auto'. This resolves the issue and the rendering is correct on page load. Now when passing in an `area`, the other grid related props are ignored. This was previously not the case and broken attributes were set.

I understand that this issue happens in the storybook as the `area` is an empty string by default and this causes conflicts. The strange thing is that I run into the identical issue when using the library. I assumed `gridArea: undefined` would be equal to not setting the `gridArea`, but looks like React just checks which style attributes exist at the same time and not whether they are `undefined` or not.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Grid/GridItem/GridItem.test.tsx
+++ b/packages/forma-36-react-components/src/components/Grid/GridItem/GridItem.test.tsx
@@ -28,30 +28,30 @@ describe('should have correct styles', () => {
   it('should result to gridColumn', () => {
     const output = shallow(<GridItem columnStart={3} columnEnd={6} />);
     const outputStyles = output.get(0).props.style;
-    expect(outputStyles).toMatchObject({ gridColumn: '3 / 6' });
+    expect(outputStyles).toMatchObject({ gridArea: 'auto / 3 / auto / 6' });
   });
 
   it('columnStart should have the correct value', () => {
     const output = shallow(<GridItem columnStart={3} />);
     const outputStyles = output.get(0).props.style;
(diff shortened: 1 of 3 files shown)


mshaaban0 (approved):
Very nice, thank you 🍫

burakukula (approved):

denkristoffer (approved):

PR #609, opened 2020-10-28
fix(dropdown): only add keydown event handler when dropdown is open

denkristoffer:
# Purpose of PR

This should prevent capturing escape press when we don't intend to.

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Dropdown/Dropdown.tsx
+++ b/packages/forma-36-react-components/src/components/Dropdown/Dropdown.tsx
@@ -176,7 +176,7 @@ export function Dropdown({
     if (forceUpdate) {
       forceUpdate();
     }
-  }, [children]);
+  }, [children, forceUpdate]);
 
   const openSubmenu = (isOpen: boolean) => {
     if (submenuToggleLabel) {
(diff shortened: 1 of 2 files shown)


packages/forma-36-react-components/src/components/Dropdown/Dropdown.tsx
```diff
@@ -204,12 +204,16 @@ export function Dropdown({
   );
 
   useEffect(() => {
-    document.addEventListener('keydown', handleEscapeKey, true);
+    if (isOpen) {
```
suevalov, line 207:
```suggestion
  if (isOpen) {
      document.addEventListener('keydown', handleEscapeKey, true);
      return () => {
         document.removeEventListener('keydown', handleEscapeKey, true);
      }
    }
    // do nohing
```

suevalov: I don't think we need to do anything if it's not opened

suevalov (approved):

burakukula (approved):
🤝

PR #610, opened 2020-10-28
Chore: implement yarn workspaces

gui-santos:
# Purpose of PR

This PR implements [yarn workspaces](https://classic.yarnpkg.com/en/package/workspaces) to our monorepo!
Basically, it hoists our dependencies to one node_modules folder at the root of the repo with only one yarn.lock and then it creates symlinks to those dependencies in our packages. This will make us avoid errors like the one on the website about having two versions of react

Please take a careful look at this PR since it changes a lot of stuff and it might explode something

I recommend doing this before checking it locally:

1. `nvm use`
2. `yarn lerna clean` to remove node_modules from packages
3. `yarn rimraf node_modules` to remove node_modules from root (I added rimraf as a dev dep but we can remove it)
4. `yarn`

## PR Checklist

- [x] Tests are passing


packages/components/forma-36-react-timepicker/src/Timepicker.tsx
```diff
@@ -79,7 +79,7 @@ function createHours() {
     hours.push(addMinutes(addHours(startOfDay(new Date()), hour), 30));
   }
   return orderBy(hours, (time: Date) => time, 'asc').map((m: Date) =>
-    format(m, DATEFNS_12H_FORMAT)
+    format(m, DATEFNS_12H_FORMAT),
```
gui-santos, line 82: The changes in this file were done by prettier

`tsdx lint` was complaining about its formatting

packages/forma-36-fcss/scss/styles.scss
```diff
@@ -1,4 +1,4 @@
-@import '../node_modules/@contentful/forma-36-tokens/dist/scss/index.scss';
+@import '../../../node_modules/@contentful/forma-36-tokens/dist/scss/index.scss';
```
gui-santos, line 1: Because yarn workspaces moves our dependencies to the root I had to do this

I tried to do something like `@contentful/forma-36-tokens/dist/scss/index.scss`, but for this to work something is still missing
we could do it in another PR I think

packages/forma-36-website/src/components/Layout.css
```diff
@@ -1,4 +1,4 @@
-@import './../../node_modules/@contentful/forma-36-tokens/dist/css/index.css';
+@import '../../../../node_modules/@contentful/forma-36-tokens/dist/css/index.css';
```
gui-santos, line 1: Same comment as the other import

denkristoffer (approved):

burakukula (approved):
Looks good to me 🤝

.prettierignore
```diff
@@ -1,3 +1,2 @@
-package.json
```
mshaaban0, line 1: 👍

packages/components/forma-36-react-timepicker/package.json
```diff
@@ -16,19 +16,14 @@
     "lint": "tsdx lint"
   },
   "peerDependencies": {
-    "react": ">=16"
+    "react": "^16.8.6",
```
mshaaban0, line 19: Do we have to pin the peer version here? isn't that what is causing the duplicate react versions?

gui-santos: I thought the duplication was caused by having it in devDependencies and here
but this was happening only on the website, no?

packages/components/forma-36-react-datepicker/tsconfig.json
```diff
@@ -20,8 +20,9 @@
     "noImplicitReturns": true,
     "noFallthroughCasesInSwitch": true,
     "moduleResolution": "node",
+    "allowJs": true,
     "baseUrl": "./",
-    "typeRoots": ["node_modules/@types", "./typings.d.ts"],
+    "typeRoots": ["../../../node_modules/@types", "./typings.d.ts"],
```
mshaaban0, line 25: So this now listens to the global types only? what if someone adds a package only for this component, will the types be available?

packages/components/forma-36-react-timepicker/package.json
mshaaban0: yes, but `peerDependencies` is different, it's used only for consumer packages, it tells the consumer package i.e. "user_interface" which version of react it needs. it's not used for compilation

gui-santos: So you are suggesting to keep devDep and removing peerDependencies totally, is that it?
Don't this mean that someone consuming react-timepicker that has no react will get an error?

mshaaban0: `peerDep` should stay, but with `>=16` just without pinning the version to avoid duplicate installation if consumer has `16.9.0` for example.
As for `devDep` i think we should keep it removed yes, should be able to get the react version from root I think.

packages/components/forma-36-react-datepicker/tsconfig.json
gui-santos: I'm not sure why we need this tbh, but I need this to get the types because the compiler was complaining and I want to focus on one refactor at a time
I even thought about unifying the tsconfigs in a later PR

What I can say to you is that when adding a dependency for only one package this dependency will be added in the local node_modules, but a dependency used by more than a package is hoisted to the root by yarn-workspaces

if know more about this `typeRoots`, pls tell me :) I can use help

packages/components/forma-36-react-timepicker/package.json
gui-santos: Ahhh I got it! I will make the change

gui-santos: Thanks a lot 🖤

packages/components/forma-36-react-datepicker/tsconfig.json
mshaaban0: Having a base `tsconfig` file is ideal, and then we can extend it for packages if needed. Let's pair on this tomorrow, and get it cleaned up 👍

packages/components/forma-36-react-timepicker/package.json
denkristoffer: It should still be `>=16.8` because of hooks.

packages/forma-36-fcss/scss/styles.scss
m10l: Hmm, worried this change will break things when using the package in another project

denkristoffer: I think you're right @m10l, but this was brittle before anyway because of the relative path. We should find out how to get the absolute import working.

gui-santos: I will try more!
I suspect that we need to tell webpack or whatever is building the packages that `@contentful/forma-36-tokens` is a real dependency in node_modules in the root

And again, my change is just updating the relative path. The version in production is using a relative path already

packages/components/forma-36-react-datepicker/tsconfig.json
gui-santos: I totally removed this typeRoots and everything seems to be fine actually 👀 

or it will explode something that I'm missing

packages/forma-36-fcss/scss/styles.scss
gui-santos: Hey! I managed to make the relative path work, but I needed to add this dependency
https://www.npmjs.com/package/node-sass-import

And to ask yarn to not hoist it because it's not working properly when hoisted

And then we use it in the npm-script `build` like this:
`node-sass scss/styles.scss -o dist --output-style compressed --importer node_modules/node-sass-import`

packages/components/forma-36-react-datepicker/tsconfig.json
```diff
@@ -20,8 +20,8 @@
     "noImplicitReturns": true,
     "noFallthroughCasesInSwitch": true,
     "moduleResolution": "node",
+    "allowJs": true,
```
denkristoffer, line 23: Is there a reason to add this?

packages/components/forma-36-react-timepicker/jest.config.js
```diff
@@ -1,5 +0,0 @@
-module.exports = {
```
denkristoffer, line 1: Nice to get rid of this again 😍

packages/components/forma-36-react-timepicker/package.json
```diff
@@ -39,9 +34,8 @@
     "@types/lodash.orderby": "^4.6.6",
     "@types/react": "^16.9.11",
     "@types/react-dom": "^16.9.3",
+    "eslint-plugin-prettier": "^3.1.4",
```
denkristoffer, line 37: Is this needed?

packages/components/forma-36-react-timepicker/tsconfig.json
```diff
@@ -20,6 +20,7 @@
     "noImplicitReturns": true,
     "noFallthroughCasesInSwitch": true,
     "moduleResolution": "node",
+    "allowJs": true,
```
denkristoffer, line 23: As the other comment, is there a reason to allow JS now?

packages/forma-36-react-components/package.json
```diff
@@ -30,14 +30,14 @@
     "@storybook/addon-essentials": "^6.0.21",
     "@storybook/addon-knobs": "^6.0.21",
     "@storybook/addon-storysource": "^6.0.21",
-    "@storybook/react": "6.1.0-alpha.10",
+    "@storybook/react": "6.0.27",
```
denkristoffer, line 33: Why the downgrade?

packages/forma-36-react-components/package.json
```diff
@@ -97,7 +97,7 @@
     "storybook-prop-types-addon": "1.0.2",
     "style-loader": "^0.23.1",
     "travis-deploy-once": "^4.4.1",
-    "typescript": "3.7.5",
+    "typescript": "3.9.7",
```
denkristoffer, line 100: 😍

packages/forma-36-website/src/components/Layout.css
```diff
@@ -1,4 +1,4 @@
-@import './../../node_modules/@contentful/forma-36-tokens/dist/css/index.css';
+@import '~@contentful/forma-36-tokens/dist/css/index.css';
```
denkristoffer, line 1: I guess no one else actually uses this package, but in theory, if they did, would this import work for them? Shouldn't it be absolute?

denkristoffer (changes requested):
Okay, I added some annoying questions, sorry 😬

packages/forma-36-website/src/components/Layout.css
mshaaban0: unfortunately, that works with webpack, we're using `node-sass` directly. we should revisit the build system for `fcss` packages later, or even deprecate it, I remember Johannes mentions multiple times it's not being used much.

packages/components/forma-36-react-datepicker/package.json
```diff
@@ -40,8 +41,6 @@
     "@types/react-dom": "^16.9.3",
```
mshaaban0, line 41: Should we remove the `@types` as well?

packages/components/forma-36-react-timepicker/package.json
```diff
@@ -16,19 +16,14 @@
     "lint": "tsdx lint"
   },
   "peerDependencies": {
-    "react": ">=16"
+    "react": ">=16.8",
+    "react-dom": ">=16.8"
```
mshaaban0, line 20: Do we only support `16.8` or anything above `16`?

packages/forma-36-fcss/scss/styles.scss
```diff
@@ -1,4 +1,4 @@
-@import '../node_modules/@contentful/forma-36-tokens/dist/scss/index.scss';
+@import '@contentful/forma-36-tokens/dist/scss/index.scss';
```
mshaaban0, line 1: Aha so you used the `node-sass-importer` Noice 👍

packages/forma-36-react-components/package.json
```diff
@@ -170,7 +170,9 @@
     "extend": "3.0.2",
     "@types/react": "16.8.6",
     "set-value": "2.0.1",
-    "js-yaml": "3.13.1"
+    "js-yaml": "3.13.1",
+    "react": "^16.8.6",
+    "react-dom": "^16.8.6"
```
mshaaban0, line 175: Do we need this to be in resolutions, [when to use resolutions](https://classic.yarnpkg.com/en/docs/selective-version-resolutions/#toc-why-would-you-want-to-do-this)?

packages/components/forma-36-react-datepicker/tsconfig.json
mshaaban0: @gui-santos sounds good, if the build is passing then it's reading it properly 👍 
We will look into unifying the tsconfig in another PR <3

packages/forma-36-react-components/package.json
denkristoffer: IMO we should avoid resolutions as much as possible, as they can make it difficult to understand the actual version of a package that's used. The version in a `package.json` might not actually be the version used. Also, as far as I remember they don't get updated by tools like Dependabot.

packages/components/forma-36-react-datepicker/tsconfig.json
gui-santos: Oh I forgot now, but I got an error when building the datepicker and the error message suggested to set `allowJs` to true
I will remove this and get the error again so we can analyze this better because I forgot hahahah

packages/components/forma-36-react-timepicker/package.json
gui-santos: When I was linting (or building) the time picker `tsdx` telling me that `eslint-plugin-prettier` could not be found, but this was before I learned about the `nohoist` config so maybe we can use that instead

What I can say is that apparently `tsdx` does not understand workspaces well right now

I will post the error we get here so we can all see. And I will try to find the open issue in `tsdx` related to the error
maybe you all can help me with alternatives 🙌

packages/components/forma-36-react-timepicker/tsconfig.json
gui-santos: The same thing that happened to date picker happened here too, let me get the error and post it here

packages/forma-36-react-components/package.json
gui-santos: I think when I was implementing popper.js for the first time I had to upgrade to the alpha version for some reason

And then now I downgraded to the last stable version to check if we really needed alpha and everything seems to work fine! So I thought we would prefer to have the stable version of storybook

Wdyt? We can go back to alpha

packages/forma-36-react-components/package.json
gui-santos: I tried to update to v4 but I got too many errors so let's keep it for the next time 😄

packages/forma-36-website/src/components/Layout.css
gui-santos: Uhmm I don't know now
`@contentful/forma-36-tokens` is a dependency in this package (website)
and I thought that `~` just tells webpack (well gatsby using webpack) to get this in node_modules
isn't it how this works?

packages/components/forma-36-react-datepicker/package.json
gui-santos: Hmm I don't know, I will remove it and if it does not break we remove it for real hahahahah

packages/components/forma-36-react-timepicker/package.json
gui-santos: I thought because we use hooks we want 16.8

gui-santos: ahhh wait these components don't use hooks, right?

packages/forma-36-react-components/package.json
gui-santos: That's why we have code reviews!
I think this is a leftover from some stuff that I was trying to do
I will remove it, thanks for spotting it 🙌

packages/forma-36-website/src/components/Layout.css
mshaaban0: Ah yeah, it's the website, it should be fine. I thought it's the `fcss` package, sorry 🤦

packages/forma-36-react-components/package.json
mshaaban0: for me stable is good if it works 👍

packages/forma-36-website/src/components/Layout.css
gui-santos: @mshaaban0 your point about the `fcss` is still a valid point tough
I think `tokens` pkg is using it, but maybe we could avoid it

packages/components/forma-36-react-timepicker/package.json
mshaaban0: Good point about the hooks, I thought it was much earlier than `16.8` 
we can keep it for consistency or fallback to `16` if we want to be accurate, up to you.

packages/components/forma-36-react-datepicker/tsconfig.json
gui-santos: Good news! I removed this and no errors! hahahaha

packages/components/forma-36-react-timepicker/package.json
gui-santos: So this is what we get when I remove this and run `tsdx lint`:
```
Error: Failed to load plugin 'prettier' declared in 'BaseConfig': Cannot find module 'eslint-plugin-prettier'
Require stack:
- /Users/gui.santos/Contentful/forma-36/packages/components/forma-36-react-timepicker/__placeholder__.js
Referenced from: BaseConfig
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:636:15)
    at Function.resolve (internal/modules/cjs/helpers.js:33:19)
    at Object.resolve (/Users/gui.santos/Contentful/forma-36/node_modules/eslint/lib/shared/relative-module-resolver.js:44:50)
    at ConfigArrayFactory._loadPlugin (/Users/gui.santos/Contentful/forma-36/node_modules/eslint/lib/cli-engine/config-array-factory.js:959:39)
    at ConfigArrayFactory._loadExtendedPluginConfig (/Users/gui.santos/Contentful/forma-36/node_modules/eslint/lib/cli-engine/config-array-factory.js:779:29)
    at ConfigArrayFactory._loadExtends (/Users/gui.santos/Contentful/forma-36/node_modules/eslint/lib/cliengine/config-array-factory.js:725:29)
    at ConfigArrayFactory._normalizeObjectConfigDataBody (/Users/gui.santos/Contentful/forma36/node_modules/eslint/lib/cli-engine/config-array-factory.js:660:25)
    at _normalizeObjectConfigDataBody.next (<anonymous>)
    at ConfigArrayFactory._normalizeObjectConfigData (/Users/gui.santos/Contentful/forma36/node_modules/eslint/lib/cli-engine/config-array-factory.js:596:20)
    at _normalizeObjectConfigData.next (<anonymous>)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Does anyone have any clue?

denkristoffer: tsdx should install `eslint-plugin-prettier` for us. Is it being hoisted or did you add it to `nohoist`?

gui-santos: I'm getting it even when I do this:

```
"nohoist": [
      "@contentful/forma-36-fcss/node-sass-import",
      "**/eslint-plugin-prettier",
      "**/eslint-plugin-prettier/**"
    ]
```

Should I put tsdx in this list instead? It's only used in timepicker and date picker right now

gui-santos: yep! I think not hoisting tsdx worked

I have to say, this is a con for tsdx. It should work with workspaces

denkristoffer: There's this, but doesn't seem like there's much actual information there: https://github.com/formium/tsdx/issues/275

gui-santos: Yep! I saw this issue before. I think it's ok to not hoist tsdx for now. It's used only in date and timepicker

wdyt?

denkristoffer: I agree, but I wish I understood why it's giving us problems 🙂 [Formik is a monorepo](https://github.com/formium/formik) using tsdx as well and they don't have to jump through hoops to get it working. Maybe it'll get better once `react-components` also use tsdx.

denkristoffer (approved):
LGTM!

packages/components/forma-36-react-timepicker/package.json
gui-santos: Let's keep this in mind and try to go back to an ideal setup in the next PR's 🙌

package.json
```diff
@@ -2,14 +2,26 @@
...
   "private": true,
+  "workspaces": {
+    "packages": [
+      "packages/*",
+      "packages/components/*"
+    ],
+    "nohoist": [
+      "@contentful/forma-36-fcss/node-sass-import",
+      "**/tsdx",
+      "**/tsdx/**"
```
mshaaban0, line 13: Just a question, why we need this to be here?

mshaaban0 (approved):
Nice work Gui ✨👏

package.json
gui-santos: Because of @denkristoffer about the eslint-prettier pkg

packages/components/forma-36-react-timepicker/package.json
gui-santos: @mshaaban0 this is why we need tsdx to not be hoisted

package.json
mshaaban0: Cool, thank you

PR #611, opened 2020-10-29
build(forma-36-react-components): build with tsdx

denkristoffer:
# Purpose of PR

This PR moves `@contentful/forma-36-react-components` to a [tsdx](https://tsdx.io) build setup. To allow this I had to do some subtasks:

- Change to SVG components instead of importing through Babel (This can be supported by Rollup as well, but then we have a difference with Storybook, which uses Webpack. Let's discuss this later).
- Lint fixes all over the place as `tsdx` bring its own eslint config, which we cannot really extend.
- Changing some directory structures, like moving `forma-36-react-components` test config into `src/` to have it included in the TS `rootDir`.
- We cannot extend the tsdx babel config outside of tsdx, which means that we cannot get jest to use it. I decided the best solution was then to use tsdx's default config as is, because at least it will be used the same everywhere. **This means that we cannot choose our own babel modules and have them work in tests**!
- This currently doesn't build a file per component like the old setup, so imports like `import { CardActions } from '@contentful/forma-36-react-components/dist/components/Card/CardActions/CardActions';` don't work. Do we want to support this?

## Some issues

eslint doesn't allow shared configs to bring their own modules, who knew! See https://github.com/eslint/eslint/issues/3458 – this is crazy to me. Anyway, that means we'll have either to "reinstall" the modules used by tsdx's eslint config, like `eslint-config-prettier`, in each package or we can avoid using hoisting. I prefer the first option because it means less duplication.

## PR Checklist

- [x] Tests are passing


denkristoffer:
Phew, lots of lint errors 😅

denkristoffer:
This currently doesn't build a file per component like the old setup, so imports like
```
import { CardActions } from '@contentful/forma-36-react-components/dist/components/Card/CardActions/CardActions';
```
don't work. Do we want to support this?

packages/forma-36-react-components/tools/.storybook/main.js
```diff
@@ -12,10 +12,82 @@ module.exports = {
...
+    // // TypeScript
+    // config.resolve.extensions.push('.ts', '.tsx', '.js', '.json', '.css');
+    // config.module.rules.push({
+    //   test: /\.(js|ts)x?$/,
+    //   exclude: /node_modules/,
+    //   use: [
+    //     require.resolve('babel-loader'),
+    //     require.resolve('react-docgen-typescript-loader'),
+    //   ],
+    // });
```
gui-santos, line 37: Is this a leftover?

gui-santos (approved):
LGTM! Should we merge this before #610 ?

denkristoffer:
@gui-santos Sorry, it's not ready yet 😬 The tests aren't running, I think it's because of tsdx issues that I hope will work better after Yarn workspaces are in.

gui-santos:
> @gui-santos Sorry, it's not ready yet 😬 The tests aren't running, I think it's because of tsdx issues that I hope will work better after Yarn workspaces are in.

Hmmmm I see!

Should we wait to merge the workspaces because of the frozen deployment? (I'm thinking on the case that someone needs to push a hotfix in the webapp and then it would try to get forma in the pipeline)

denkristoffer:
I got the tests running, but they're currently failing because SVGs seem to not be loaded correctly. Same issue as [here](https://github.com/formium/tsdx/issues/662#issuecomment-707418421) – will look into this more later, but help is welcome.

packages/forma-36-react-components/src/components/Accordion/__snapshots__/Accordion.test.tsx.snap
```diff
@@ -73,8 +74,9 @@ exports[`renders the component with an additional class name 1`] = `
         <svg
           class="Icon Icon--small Icon--secondary Icon--trimmed AccordionHeader__icon"
           data-test-id="cf-ui-icon"
+          height="1em"
           viewBox="0 0 9 18"
-          xmlns="http://www.w3.org/2000/svg"
+          width="1em"
```
suevalov, line 79: Are these `width/height` changes expected? I see a lot of them in different svgs and snapshots

denkristoffer: They're from the way SVGR CLI generates components based on SVG files. I don't know if it would be preferable to keep the sizes locked with px?

packages/forma-36-react-components/src/components/Icon/svg/ArrowDown.tsx
```diff
@@ -0,0 +1,12 @@
...
+function SvgArrowDown(props: React.SVGProps<SVGSVGElement>) {
+  return (
+    <svg width="1em" height="1em" viewBox="0 0 24 24" {...props}>
+      <path d="M7 10l5 5 5-5z" />
+      <path d="M0 0h24v24H0z" fill="none" />
+    </svg>
+  );
+}
+
+export default SvgArrowDown;
```
gui-santos, line 12: @burakukula look!
This is basically what we need for an icon package

gui-santos (approved):
This is great!

gui-santos:
> * Change to SVG components instead of importing through Babel (This can be supported by Rollup as well, but then we have a difference with Storybook, which uses Webpack. Let's discuss this later).

This is perfect! Since we are dating with the idea of making a separate package for icons. Moving them to components is 50% the work already hahahaha

> * Lint fixes all over the place as `tsdx` bring its own eslint config, which we cannot really extend.

If it's a good eslint config, I'm ok with that

> * This currently doesn't build a file per component like the old setup, so imports like `import { CardActions } from '@contentful/forma-36-react-components/dist/components/Card/CardActions/CardActions';` don't work. Do we want to support this?

I don't this will be a problem 😄 

> eslint doesn't allow shared configs to bring their own modules, who knew! See [eslint/eslint#3458](https://github.com/eslint/eslint/issues/3458) – this is crazy to me. Anyway, that means we'll have either to "reinstall" the modules used by tsdx's eslint config, like `eslint-config-prettier`, in each package or we can avoid using hoisting. I prefer the first option because it means less duplication.

I prefer the first option too 👍

packages/forma-36-react-components/src/components/Icon/svg/ArrowDown.tsx
burakukula: yeah! exactly! 🙂

burakukula (approved):
Looks amazing! 🎉 
Thanks for that contribution and yeah, most of the work for the icon package is done 🤯 I will try take it and move it to the package in the PR proposal in the next couple of days.

m10l (approved):

packages/forma-36-react-components/package.json
```diff
@@ -3,35 +3,68 @@
   "description": "Contentful UI Component Library",
   "version": "3.67.4",
   "main": "dist/index.js",
-  "module": "dist/esm/index.js",
+  "module": "dist/forma-36-react-components.esm.js",
```
mshaaban0, line 6: Wouldn't this break consumer builds who explicitly require an absolute path? if we can avoid by keeping the old convention would be good, if not 🤷‍♂️

packages/forma-36-react-components/package.json
```diff
@@ -3,35 +3,68 @@
...
   "style": "dist/styles.css",
   "license": "MIT",
   "types": "dist/index.d.ts",
   "files": [
-    "package.json",
-    "dist"
+    "dist",
+    "src"
```
mshaaban0, line 12: Nice 👍

mshaaban0 (approved):
Awesome work 🍪

I'm not sure about `width="1em" height="1em"` on svgs especially since the `viewbox` is still in pixels as well. If there's a way to keep them in pixels it would be awesome.

I'm going to hold of #641 until this is merged, and hope conflicts would be nice to me 😏

denkristoffer:
@mshaaban0 (and /cc @suevalov): SVGR automatically adds missing `width` and `height` attributes when the output mode is `icon`, which I used in this case. The [svgr PR here](https://github.com/gregberge/svgr/pull/40) doesn't have a lot of reasoning for this, but I trust them to know what they're doing. What do you think, can we keep them or should I remove the attributes?

packages/forma-36-react-components/package.json
denkristoffer: It would if they require `index`. Have we ever supported this? I'm not sure that we can modify tsdx enough to change such things.

suevalov:
@denkristoffer are `width` and `height` attributes present in the output `dist` of the main branch? If they're the same then I have no objections

packages/forma-36-react-components/package.json
suevalov: Do we want to have `src` files in the npm package?

denkristoffer: Tbh I had the same thought. It's the tsdx default, but I'm not 100% sure why they add it.

suevalov: None of our experience components have it: https://github.com/contentful/experience-packages/blob/master/packages/experience-components/package.json, so I'd not include it here too.

suevalov (approved):
Let's remove `src` folder from the built package, but otherwise is 🎖️

PR #612, opened 2020-10-29
chore: lint fixes

denkristoffer:
# Purpose of PR

Lint fixes. This PR should make it easier to land some other changes we have coming 🙂

We should decide what we want to do with the eslint rule `@typescript-eslint/no-explicit-any`, which is producing the last warnings. I'm thinking we can turn it off and trust people to know when to use `any`. I did the same with `@typescript-eslint/no-empty-function`.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/.eslintrc.js
+++ b/packages/forma-36-react-components/.eslintrc.js
@@ -15,7 +15,7 @@ module.exports = {
   ],
   parserOptions: {
     ecmaVersion: 7,
-    project: 'tsconfig.json',
+    project: './tsconfig.json',
     sourceType: 'module',
   },
   env: {
@@ -25,19 +25,19 @@ module.exports = {
   },
   settings: {
     react: {
-      version: '16.3.0',
+      version: 'detect',
     },
   },
   rules: {
     'react/prop-types': 'off',
-    'react/prefer-stateless-function': [0, { ignorePureComponents: true }],
-    'react/jsx-filename-extension': { extensions: ['.js', '.jsx'] },
-    'react/forbid-prop-types': 0,
-    'react/destructuring-assignment': 0,
-    'react/jsx-one-expression-per-line': 0,
-    'react/sort-comp': 0,
-    'react/jsx-wrap-multilines': 0,
-    'function-paren-newline': 0,
+    'react/prefer-stateless-function': ['off', { ignorePureComponents: true }],
+    'react/jsx-filename-extension': ['off', { extensions: ['.js', '.tsx'] }],
+    'react/forbid-prop-types': 'off',
+    'react/destructuring-assignment': 'off',
+    'react/jsx-one-expression-per-line': 'off',
+    'react/sort-comp': 'off',
+    'react/jsx-wrap-multilines': 'off',
+    'function-paren-newline': 'off',
     'react/button-has-type': 'error',
     'react/jsx-no-target-blank': 'error',
     'react/jsx-no-bind': ['error', { allowArrowFunctions: true }],
@@ -57,6 +57,7 @@ module.exports = {
     '@typescript-eslint/no-var-requires': 'off',
     '@typescript-eslint/no-use-before-define': 'off',
     '@typescript-eslint/no-non-null-assertion': 'error',
+    '@typescript-eslint/no-empty-function': 'off',
     'react-hooks/rules-of-hooks': 'error',
     'react-hooks/exhaustive-deps': 'warn'
   },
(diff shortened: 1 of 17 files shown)


denkristoffer:
Why is this not being run on Travis? 🤔

gui-santos (approved):
Great stuff!

Regarding `no-explicit-any` I'm not sure
If someone really needs to use `any` can use `eslint-disable next line` or something
Although I think the right thing to do is to create a type or if you don't know the type you can use `unknown` 

🙌

m10l:
> Great stuff!
> 
> Regarding `no-explicit-any` I'm not sure
> If someone really needs to use `any` can use `eslint-disable next line` or something
> Although I think the right thing to do is to create a type or if you don't know the type you can use `unknown`
> 
> 🙌

I have a feeling `no-explicit-any` was used because of an external library that was getting linted, but I might be wrong. I'd be up for trying Gui's comment if there aren't any issues when building once removed.

denkristoffer:
@gui-santos  agreed, that's a better way going forward. Maybe it should be an error instead of a warning then, to force people to think about it?

packages/forma-36-react-components/.eslintrc.js
```diff
@@ -31,7 +31,7 @@ module.exports = {
   rules: {
     'react/prop-types': 'off',
     'react/prefer-stateless-function': [0, { ignorePureComponents: true }],
-    'react/jsx-filename-extension': { extensions: ['.js', '.jsx'] },
+    'react/jsx-filename-extension': [0, { extensions: ['.js', '.jsx'] }],
```
mshaaban0, line 34: Question: Why the `[0]` thingy?

mshaaban0 (approved):
Good stuff 🍬

I'd second Gui, and vote to have `@typescript-eslint/no-explicit-any`

I honestly don't get the difference between `unknown` and `any`, they say in their [RC](https://devblogs.microsoft.com/typescript/announcing-typescript-3-0-rc-2/#the-unknown-type)...
> unknown is the type-safe counterpart of any.

If I don't know the type how is it TypeSafe ™️ 🤷🤷‍♀️🤷‍♂️

packages/forma-36-react-components/.eslintrc.js
denkristoffer: To make it an error. The old setting actually didn't work because it was missing the array and error level. I'll change it (and the other 0 values) to a string to make it more obvious.

denkristoffer:
> I honestly don't get the difference between unknown and any, they say in their RC...
>
> >unknown is the type-safe counterpart of any.
>
> If I don't know the type how is it TypeSafe ™️ 🤷🤷‍♀️🤷‍♂️

`unknown` forces you to check if a property you want to access actually exists on the object, while `any` lets you try to access it even if it's not there.

denkristoffer:
Can anyone tell why this line doesn't work? https://github.com/contentful/forma-36/blob/lint/packages/forma-36-react-components/src/components/Notification/index.tsx#L40

denkristoffer:
I think switching from `any` to `unknown` counts as a breaking change for consumers, maybe we should hold off on that until v4? @m10l @gui-santos

gui-santos:
> I think switching from `any` to `unknown` counts as a breaking change for consumers, maybe we should hold off on that until v4? @m10l @gui-santos

Does it break even if it's just a warning? @denkristoffer 
If yes, then you are right we should hold it for now

denkristoffer:
> Does it break even if it's just a warning? @denkristoffer

It's a type error, not a linting thing, so I think it would break. I'll change it back to any for now. But it's a good change to get in later, I think.

gui-santos:
> Can anyone tell why this line doesn't work? https://github.com/contentful/forma-36/blob/lint/packages/forma-36-react-components/src/components/Notification/index.tsx#L40

@denkristoffer 
I think it's because the any is in the argument of the function that is being returned. I did this and the disable worked
```js
const afterInit = (fn: Function) => {
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  return (...args: any[]) => {
    if (!initiated) {
      initiated = true;
      return new Promise((resolve) => {
        createRoot(() => {
          resolve(fn(...args));
        });
      });
    } else {
      return Promise.resolve(fn(...args));
    }
  };
};

It's not pretty tough :/
```

denkristoffer:
Okay this is ready to go I think. One last thought that we can decide on later is whether it's worth linting test and story files.

PR #619, opened 2020-11-03
fix(dropdownlistitem): always apply button classname and test-id

denkristoffer:
# Purpose of PR

`DropdownListItem` class and test-id should not be overwritten by props. Also adds a React display name to the `forwardRef`.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Dropdown/DropdownListItem/DropdownListItem.tsx
+++ b/packages/forma-36-react-components/src/components/Dropdown/DropdownListItem/DropdownListItem.tsx
@@ -79,6 +79,8 @@ export const DropdownListItem = forwardRef<HTMLElement, DropdownListItemProps>(
 
         return (
           <Element
+            className={styles['DropdownListItem__button']}
+            data-test-id="cf-ui-dropdown-list-item-button"
             onClick={!isDisabled && onClick !== undefined ? onClick : undefined}
             onMouseDown={
               !isDisabled && onMouseDown !== undefined ? onMouseDown : undefined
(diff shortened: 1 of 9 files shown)


packages/forma-36-react-components/src/components/Dropdown/DropdownListItem/DropdownListItem.tsx
```diff
@@ -79,21 +79,15 @@ export const DropdownListItem = forwardRef<HTMLElement, DropdownListItemProps>(
...
+            onClick={!isDisabled && onClick !== undefined ? onClick : undefined}
+            onMouseDown={
+              !isDisabled && onMouseDown !== undefined ? onMouseDown : undefined
+            }
             type="button"
             {...(href ? linkProps : buttonProps)}
-            {...otherProps}
+            {...props}
+            className={styles['DropdownListItem__button']}
+            data-test-id="cf-ui-dropdown-list-item-button"
```
suevalov, line 90: We probably override this default test-id in many places already, so I'd be careful changing this.

denkristoffer: You're right, I intended to restore [previous functionality](https://github.com/contentful/forma-36/blob/271861a49f288bc0b1f510d9c97f8ef97bb099d7/packages/forma-36-react-components/src/components/Dropdown/DropdownListItem/DropdownListItem.tsx) but it turns out that it actually let you overwrite the submenu toggle class.

gui-santos (approved):

PR #620, opened 2020-11-04
fix(ListItem): specify list-style-type initial

### LINKED ISSUE #549: 🐛 Bug - ListItem doesn't enforce list-style-type
## Summary

When using `ListItem` the `li` style-type gets overridden by higher level definitions. e.g. the webapp legacy css contain the following:
```css
ul li {
  list-style-type: none;
}
```

This overrides the ListItem styles, as there's no `style-type` defined for it within the component.

## Expected results

`List` and `ListItem` should retain the `list-style-type` default styles, and potentially provide an api to choose between different `list-style-type` values

denkristoffer:
# Purpose of PR

Adds a `list-style-type` to the `ListItem` component, which should resolve #549 

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/List/ListItem/ListItem.css
+++ b/packages/forma-36-react-components/src/components/List/ListItem/ListItem.css
@@ -6,7 +6,7 @@
   font-size: var(--font-size-m);
   font-family: var(--font-stack-primary);
   line-height: var(--line-height-default);
-  list-style-type: initial;
+  list-style-type: disc;
 }
 
 .ListItem--nested-list {
--- a/packages/forma-36-react-components/src/components/List/ListItem/ListItem.css
+++ b/packages/forma-36-react-components/src/components/List/ListItem/ListItem.css
@@ -6,6 +6,7 @@
   font-size: var(--font-size-m);
   font-family: var(--font-stack-primary);
   line-height: var(--line-height-default);
+  list-style-type: initial;
 }
 
 .ListItem--nested-list {
--- a/packages/components/forma-36-react-timepicker/yarn.lock
+++ b/packages/components/forma-36-react-timepicker/yarn.lock
@@ -4887,25 +4887,25 @@ qs@~6.5.2:
   resolved "https://registry.yarnpkg.com/qs/-/qs-6.5.2.tgz#cb3ae806e8740444584ef154ce8ee98d403f3e36"
   integrity sha512-N5ZAX4/LxJmF+7wN74pUD6qAh9/wnvdQcjq9TZjevvXzSUo7bfmw91saqMjzGS2xq91/odN2dW/WOl7qQHNDGA==
 
-react-dom@16.10.2:
-  version "16.10.2"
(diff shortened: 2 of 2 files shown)


m10l (approved):
Do we want the `yarn.lock` change there? Something that was missed out on a previous update, or is it coming from something else?

denkristoffer:
They must've been left out of something else, they keep coming up when I make changes based on master. I figured I'd just add them this time 🤷

packages/forma-36-react-components/src/components/List/ListItem/ListItem.css
```diff
@@ -6,6 +6,7 @@
   font-size: var(--font-size-m);
   font-family: var(--font-stack-primary);
   line-height: var(--line-height-default);
+  list-style-type: initial;
```
mshaaban0, line 9: Should it be `initial` or `none`? isn't `initial` basically same as default behaviour?

denkristoffer: Yes it just enforces browser default. The way I understand #549, the problem is that the list style is being overriden to `none` by the web app because of specificity?

mshaaban0: Ah, sorry if it wasn't too clear. I meant people are rewriting `list-style-type: none` many times in the web app. So we need to add an option to reset the default `ul` and `li` styles, maybe a `reset={true}` prop? where we remove the style-type, margin and padding?

gui-santos (approved):

packages/forma-36-react-components/src/components/List/ListItem/ListItem.css
m10l: There's no point in using the 36 `List` components in these cases though? The point of the component is to provide consistently styled lists.

I think ideally, we should default to a specific list style, and perhaps allow the user to override it using a `listStyle` prop.

denkristoffer: All right I've changed to disc, which is also the browser default 😉

mshaaban0: Sorry, I think my comment is confusing everyone, what's in the ticket is what's needed. 👍 
'disc' or 'initial' will work, I'd vote for what mike said though, about `listStyle` prop.

As for the `reset` prop to remove padding and margin, I ran into cases in planner and experience components where I needed to reset the styles specifically for building navigation lists like "user-profile", "main navigation", "app-switcher", ..etc. So it may be a good option to consider later.

mshaaban0 (approved):

packages/forma-36-react-components/src/components/List/ListItem/ListItem.css
denkristoffer: Sounds good, such a prop makes sense, but I think it should go on `<List />` and not on `<ListItem />` component. If you agree I think we can merge this PR for now.

PR #622, opened 2020-11-04
feat: add support for status icon for an asset card

Spring3:
# Purpose of PR

The purpose is to be able to set an icon to the AssetCard. Since Assets can now be scheduled, it would be nice to be able to set an indicator as an icon there.

This replicates the behaviour of EntryCard component

## PR Checklist

- [x] Tests are passing

Screenshot 2020-11-04 at 18 19 42
Screenshot 2020-11-04 at 18 19 49


packages/forma-36-react-components/src/components/Card/AssetCard/AssetCard.css
```diff
@@ -44,6 +44,8 @@
   padding: 0 calc(1rem * (14 / var(--font-base-default)));
   height: calc(1rem * (45 / var(--font-base-default)));
   display: flex;
+  justify-content: flex-end;
+  gap: .5rem;
```
mshaaban0, line 48: It doesn't look like `gap` is working on safari, I'd create a separate class for icon similar to `AssetCard__actions`, and use `margin-right` with tokens

Screenshot 2020-11-06 at 14 46 01

packages/forma-36-react-components/src/components/Card/AssetCard/AssetCard.tsx
```diff
@@ -93,9 +98,20 @@ export class AssetCard extends Component<AssetCardProps> {
...
+      <>
+        {statusIcon && typeof statusIcon === 'string' ? (
+          <Icon
+            icon={statusIcon as IconType}
+            color="muted"
+            className="f36-margin-right--xs"
+          />
+        ) : (
+          statusIcon
+        )}
```
mshaaban0, line 110: I think we should just expose the `statusIcon` as IconType, and maybe `statusIconColor` as IconColor, so we can avoid the string check, and ensure spacing consistency

packages/forma-36-react-components/yarn.lock
```diff
@@ -15850,16 +15850,6 @@ react-docgen@^5.0.0:
     node-dir "^0.1.10"
     strip-indent "^3.0.0"
 
-react-dom@16.8:
```
mshaaban0, line 15853: I think after rebase, this should go away :)

mshaaban0 (commented):
Thanks for the contributions, left a couple of comments and then it looks good to me.

packages/forma-36-react-components/src/components/Card/AssetCard/AssetCard.tsx
Spring3: With such implementation, we are able to pass a component. For entries, scheduled icons have a tooltip, that we won't be able to add if this property will accept only a string, while with this implementation, it is possible to manually wrap an icon into another component (Tooltip) and pass it here as a prop

domarku (approved):

dominicbonnice (approved):

packages/forma-36-react-components/src/components/Card/AssetCard/AssetCard.tsx
gui-santos: I agree with @mshaaban0
I think making statusIcon accept both a string and a React component makes the AssetCard to open for inconsistency in design

From what I understood, you want an icon for each `status`, right? Which means, for example, that draf status will always have a clock icon and "Draf" label. I suggested using the `status` props that already exist to control both the label and the icon like this:

```jsx
let label;
let type;
let icon;

switch(status) {
  case 'published':
    label = 'Published'
    type = 'positive'
    icon = 'calendar''
    // ... other cases
}

<>
  <Icon icon={icon} />
  <Tag tagType={type}>{label}</Tag>
</>
```

This way we don't increase the complexity of the components and its API and we keep design consistency

Wdyt?

gui-santos (changes requested):

packages/forma-36-react-components/src/components/Card/AssetCard/AssetCard.css
```diff
@@ -44,11 +44,16 @@
   padding: 0 calc(1rem * (14 / var(--font-base-default)));
   height: calc(1rem * (45 / var(--font-base-default)));
   display: flex;
+  justify-content: flex-end;
```
mshaaban0, line 47: do we still need this?

packages/forma-36-react-components/src/components/Card/AssetCard/AssetCard.css
```diff
@@ -44,11 +44,16 @@
...
   align-items: center;
   border-bottom: calc(1rem * (1 / var(--font-base-default))) solid
     var(--color-element-mid);
 }
 
+.AssetCard__status {
+  margin-left: .5rem;
```
mshaaban0, line 54: can we use a spacing token here?

packages/forma-36-react-components/src/components/Card/AssetCard/AssetCard.tsx
Spring3: I think we are thinking of different use cases. I need to support both: standard set of icons and also allow to insert custom component.

ScheduledIcon, that we need to support, is neither a string, nor an Icon. It's a component on it's own, that has some wrappers around the icon.

Spring3: Here is an example of usage i am trying to explain

https://github.com/contentful/field-editors/blob/master/packages/reference/src/entries/WrappedEntryCard/WrappedEntryCard.tsx#L113-L125

Since we support scheduling assets, we need to be able to inject this component as an icon [into AssetCard ](https://github.com/contentful/field-editors/blob/master/packages/reference/src/assets/WrappedAssetCard/WrappedAssetCard.tsx#L96-L128)(just like EntryCard does)

Do you have an idea of other way to support it?

Spring3: cc @gui-santos @mshaaban0

packages/forma-36-react-components/src/components/Card/AssetCard/AssetCard.tsx
```diff
@@ -32,6 +33,10 @@ export interface AssetCardProps extends BaseCardProps {
    * The publish status of the asset
    */
   status?: AssetState;
+  /**
+   * An icon for the status of the entry
+   */
+  statusIcon?: React.ReactNode;
```
gui-santos, line 39: Since you mentioned you need to support both 'string' and a custom React component, I think this prop should be something like this, right?

`statusIcon?: React.ReactNode | IconProps['icon']`

P.S.: IMO, using `IconProps['icon']` is better than `IconTypes` because if we change the `icon` prop in the future to use something else than IconTypes, we won't need to update this component

packages/forma-36-react-components/src/components/Card/AssetCard/AssetCard.tsx
gui-santos: Well, in this case, I think that's the only way then

I just comment something about the `statusIcon` prop. I think you will need to update it

gui-santos (changes requested):

packages/forma-36-react-components/src/components/Card/AssetCard/AssetCard.css
Spring3: Without it:
Screenshot 2020-11-24 at 11 24 49

With it:
Screenshot 2020-11-24 at 11 24 57


Which way do we want to go @mshaaban0 ?

packages/forma-36-react-components/src/components/Card/AssetCard/AssetCard.tsx
Spring3: AssetCard doesn't have an `icon` prop. So I can only add `IconType` there as type definition for `statusIcon`

gui-santos (approved):

packages/forma-36-react-components/src/components/Card/AssetCard/AssetCard.css
mshaaban0: Ah ok, I actually like the one on the left side, but let's keep the same convention 👍

mshaaban0 (approved):

PR #630, opened 2020-11-06
chore(deps): add support for React 17 peer dependency

denkristoffer:
# Purpose of PR

[React 17 has no relevant breaking changes](https://reactjs.org/blog/2020/08/10/react-v17-rc.html#other-breaking-changes), we should add it to Forma 36's allowed `peerDependencies` 😎 

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/package.json
+++ b/packages/forma-36-react-components/package.json
@@ -112,8 +112,8 @@
     "truncate": "^2.0.1"
   },
   "peerDependencies": {
-    "react": "^16.8.6",
-    "react-dom": "^16.8.6"
+    "react": "^16.8.6 || ^17.0.0",
+    "react-dom": "^16.8.6 || ^17.0.0"
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


packages/forma-36-react-components/package.json
```diff
@@ -114,8 +114,8 @@
     "truncate": "^2.0.1"
   },
   "peerDependencies": {
-    "react": "^16.8.6",
-    "react-dom": "^16.8.6"
+    "react": "^16.8.6 || ^17.0.0",
+    "react-dom": "^16.8.6 || ^17.0.0"
```
mshaaban0, line 118: I think this could just be `">= 16.8"` no?

denkristoffer: I think the range is more future proof, but there's not a big difference either way in my opinion.

suevalov (approved):

PR #631, opened 2020-11-06
fix: build components packages before publish

mshaaban0:
# Purpose of PR

Previously there was a separate `bin/build` job to build all packages before release, and `prepublishOnly` script that repeats the build process.

Currently `release` job on circleci relies on `prepublishOnly` only to avoid repetition. 

This PR adds `prepublishOnly` to the components packages. Later we should look into having `prepublishOnly` on the root and see if it builds on all packages.

### DIFF
--- a/packages/forma-36-react-components/src/components/Grid/Grid.tsx
+++ b/packages/forma-36-react-components/src/components/Grid/Grid.tsx
@@ -5,7 +5,17 @@ import cn from 'classnames';
 
 import styles from './Grid.css';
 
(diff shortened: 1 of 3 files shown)


gui-santos (approved):

packages/forma-36-react-components/src/components/Grid/Grid.tsx
```diff
@@ -5,7 +5,17 @@ import cn from 'classnames';
 
 import styles from './Grid.css';
 
-export type GapTypes = 'none' | 'spacing2xs' | 'spacingXs' | 'spacingS' | 'spacingM' | 'spacingL' | 'spacingXl' | 'spacing2Xl' | 'spacing3Xl' | 'spacing4Xl';
+export type GapTypes =
```
mshaaban0, line 8: 👆fix prettier to add changes so lerna can detect package changes and publish

PR #633, opened 2020-11-10
refactor(dropdown): add comment about onClose function

gui-santos:
# Purpose of PR

After some debugging and talking to @denkristoffer , I think the current solution to solve the problem should be kept as the official solution until we refactor the Dropdown and the Autocomplete in a breaking change version

I just added a comment explaining why we are doing that and refactored Autocomplete.stories.tsx to use storybook CSF

## Why?

The list of options is getting closed as soon as it opens and this happens because the list is a Dropdown. The Dropdown component is using the `useOnClickOutside` hook which makes it possible for the user to click outside the dropdown and close it. Well, the input in the autocomplete is outside, so when you click on the input it focuses it and opens the list, but immediately closes it

I think the hook is doing the right job in adding the event listener, so the components using it should tell the hook when not to call the handler. That's why the current solution is good, and we can improve how the dropdown works in the future to avoid this since it needs some refactoring anyway

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Autocomplete/Autocomplete.stories.tsx
+++ b/packages/forma-36-react-components/src/components/Autocomplete/Autocomplete.stories.tsx
@@ -1,24 +1,30 @@
 import React, { useState, useCallback } from 'react';
-import { storiesOf } from '@storybook/react';
 import { action } from '@storybook/addon-actions';
-import { boolean, number, object, text, select } from '@storybook/addon-knobs';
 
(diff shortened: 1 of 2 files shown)


denkristoffer (approved):

anho:
@gui-santos maybe we can hold back on this PR, I'm currently working on something in regards to the dropdown hook you are referring to that makes your comment obsolete.

mshaaban0 (approved):

PR #634, opened 2020-11-11
feat(dropdown): ability to exclude elements from closing the dropdow

anho:
# Purpose of PR

We have other components upstream that need to prevent the dropdown from closing (e.g. a filter value on the entity searchbar) so here is a way to do that by enhancing the `useOnClickOutside` hook to check additional references.

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Autocomplete/Autocomplete.tsx
+++ b/packages/forma-36-react-components/src/components/Autocomplete/Autocomplete.tsx
@@ -118,10 +118,8 @@ export const Autocomplete = <T extends {}>({
   dropdownProps,
   renderToggleElement,
 }: AutocompleteProps<T>) => {
-  const listRef: React.MutableRefObject<HTMLDivElement | undefined> = useRef();
-  const inputRef: React.MutableRefObject<
-    HTMLInputElement | undefined
-  > = useRef();
+  const listRef = useRef<HTMLDivElement | null>(null);
+  const inputRef = useRef<HTMLInputElement | null>(null);
 
   const [{ isOpen, query, highlightedItemIndex }, dispatch] = useReducer(
     reducer,
@@ -236,13 +234,12 @@ export const Autocomplete = <T extends {}>({
 
   return (
     <Dropdown
+      nonClosingRefs={[inputRef]}
       className={dropdownClassNames}
       isOpen={isOpen}
       onClose={() => {
-        if (inputRef.current !== document.activeElement) {
-          willClearQueryOnClose && updateQuery('');
-          dispatch({ type: TOGGLED_LIST, payload: false });
-        }
+        willClearQueryOnClose && updateQuery('');
+        dispatch({ type: TOGGLED_LIST, payload: false });
       }}
       toggleElement={renderToggleElementFunction(toggleProps)}
       {...dropdownProps}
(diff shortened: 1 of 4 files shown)


packages/forma-36-react-components/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -118,10 +118,8 @@ export const Autocomplete = <T extends {}>({
   dropdownProps,
   renderToggleElement,
 }: AutocompleteProps<T>) => {
-  const listRef: React.MutableRefObject<HTMLDivElement | undefined> = useRef();
-  const inputRef: React.MutableRefObject<
-    HTMLInputElement | undefined
-  > = useRef();
+  const listRef = useRef<HTMLDivElement | null>(null);
```
anho, line 121: The change is to match the types with Dropdown.

packages/forma-36-react-components/src/components/Dropdown/Dropdown.tsx
```diff
@@ -125,6 +125,13 @@ export interface DropdownProps {
    * Defaults to `true`
    */
   usePortal?: boolean;
+  /**
+   * By passing down references from upstream components as part of this array,
+   * you will be able to disable automatic closing of the dropdown if you click
+   * somewhere else in the document besides the dropdown itself. See the
+   * Autocomplete component for an actual example of this usage.
+   */
+  nonClosingRefs?: RefObject<HTMLElement>[];
```
gui-santos, line 134: Looking at this solution made me realize something
Our issue right now is that when we click outside the dropdow it will close it and the `toggleElement` is outside the dropdown.
So, I think we should pass a reference to the `toggleElement`. And only the `toggleElement`
Why? Because we make the API more restricted and simple... if in the future, for some reason, we need the dropdown to accept other elements that will not trigger closing it we can discuss this and extend

And now that I mentioned this alternative, I wonder if we really need a new prop or if we can just use the `toggleElement` prop to tell the hook what to do

Wdyt?

gui-santos (changes requested):

packages/forma-36-react-components/src/components/Dropdown/Dropdown.tsx
denkristoffer: Sounds great if we can make it work automatically, much easier for the developers 👍

gui-santos (approved):

packages/forma-36-react-components/src/components/Dropdown/Dropdown.tsx
anho: @denkristoffer small misunderstanding here. We still need to behaviour for element that trigger the dropdown programmatically from outside. But I think the point from @gui-santos is still a valid one to be considered for the toggle element itself.

gui-santos:
After talking to @anho , he showed me a case where a feature needs more than one reference to be passed to the useOutsideClick hook

I still think the UX is not great and should be redesigned in the future, but so we don't block users that are having a problem with that case I'm approving the PR 

🙏

packages/forma-36-react-components/src/components/Dropdown/Dropdown.tsx
denkristoffer: @anho But isn't the input used as `toggleElement`? 

> We still need to behaviour for element that trigger the dropdown programmatically from outside.

I don't quite follow you here. Could you explain when you would need this?

PR #635, opened 2020-11-11
chore(deps): migrate from enzyme to @testing-library/react

denkristoffer:
# Purpose of PR

Migrate from `enzyme` to `@testing-library/react`. This is to align testing libraries used across components, to more quickly allow for major React upgrades, and to unblock #611. 

There's still a single component on enzyme, `InViewport.test.tsx`:

These tests use enzyme's `instance()` which testing-library [doesn't support](https://testing-library.com/docs/react-testing-library/migrate-from-enzyme#triggering-class-methods-in-tests-wrapperinstance):

> As we already discussed, we are against testing implementation details and things that uses are not aware of it, and we aim to test and interact with the component like how our users would.

I'm not sure how to test it without rewriting the component completely, which defeats the purpose.

## PR Checklist

- [x] Tests are passing


mshaaban0 (approved):
Very nice, well done 👏

m10l (approved):
Awesome. We'll be able to deprecate InViewport soon, I wonder what peoples thoughts are on just removing the test so that we can completely get rid of Enzyme?

denkristoffer:
@m10l I'm for it, but I don't really have an idea of how much `InViewport` is used outside of Forma.

gui-santos (approved):
GREAT!

I vote for removing the enzyme test as @m10l suggested

denkristoffer:
I've deleted the test and removed enzyme completely 👍

PR #636, opened 2020-11-12
feat(tooltip): make the tooltip stay open when hovering on the content

Miretxu:
# Purpose of PR

This PR adds the ability for the tooltip to stay open while the user interacts with its content.
i.e. Useful when implementing the below design (see the pic) 
Screenshot 2020-11-11 at 15 03 12

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
+++ b/packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
@@ -39,6 +39,10 @@ export interface TooltipProps {
    */
   isVisible?: boolean;
   /**
+   * It controls whether or not the tooltip stays open when hovering on the content
+   */
+  closeOnMouseLeave?: boolean;
+  /**
    * It sets a max-width for the Tooltip
    */
   maxWidth?: number | CSS.MaxWidthProperty<string>;
@@ -84,6 +88,7 @@ export const Tooltip = ({
   content,
   id,
   isVisible,
+  closeOnMouseLeave,
   maxWidth,
   onBlur,
   onFocus,
@@ -135,6 +140,18 @@ export const Tooltip = ({
     }
   }, [attributes.popper]);
 
+  const [isHoveringTarget, setIsHoveringTarget] = useState(false)
+  const [isHoveringContent, setIsHoveringContent] = useState(false)
+  useEffect(() => {
+    if (closeOnMouseLeave) {
+      setShow(isHoveringTarget)
+    } else {
+      setShow(isHoveringContent || isHoveringTarget)
+    }
+  }, [closeOnMouseLeave, isHoveringTarget, isHoveringContent])
+
+  const delay = closeOnMouseLeave ? 0 : 1000;
+
   const arrowStyles = {
     ...popperStyles.arrow,
     ...arrowPosition,
@@ -164,19 +181,19 @@ export const Tooltip = ({
         ref={elementRef}
         className={targetWrapperClassName}
         onMouseEnter={(evt: MouseEvent) => {
-          setShow(true);
+          setIsHoveringTarget(true);
           if (onMouseOver) onMouseOver(evt);
         }}
         onMouseLeave={(evt: MouseEvent) => {
-          setShow(false);
+          setTimeout(() => setIsHoveringTarget(false), delay);
           if (onMouseLeave) onMouseLeave(evt);
-        }}
+      }}
         onFocus={(evt: FocusEvent) => {
-          setShow(true);
+          setIsHoveringTarget(true);
           if (onFocus) onFocus(evt);
         }}
         onBlur={(evt: FocusEvent) => {
-          setShow(false);
+          setTimeout(() => setIsHoveringTarget(false), delay);
           if (onBlur) onBlur(evt);
         }}
         {...otherProps}
@@ -195,6 +212,12 @@ export const Tooltip = ({
             [styles['Tooltip--hidden']]: !show,
           })}
           data-test-id={testId}
+          onMouseEnter={() => {
+            setIsHoveringContent(true)
+          }}
+          onMouseLeave={() => {
+            setIsHoveringContent(false)
+          }}
           {...attributes.popper}
         >
           {content}
@@ -211,6 +234,7 @@ export const Tooltip = ({
 Tooltip.defaultProps = {
   containerElement: 'span',
   isVisible: false,
+  closeOnMouseLeave: true,
   maxWidth: 360,
   testId: 'cf-ui-tooltip',
   place: 'auto',
--- a/packages/forma-36-react-components/src/components/Tooltip/Tooltip.stories.tsx
+++ b/packages/forma-36-react-components/src/components/Tooltip/Tooltip.stories.tsx
@@ -44,7 +44,7 @@ basic.parameters = {
 export const withHtml = (args: { content: string }) => {
   return (
     <Paragraph>
-      Lorem Ipsum dolor sit amet &nbsp;
+      Lorem Ipsum dolor sit amet&nbsp;
       <Tooltip {...args}>
         <TextLink>Hover me</TextLink>.
       </Tooltip>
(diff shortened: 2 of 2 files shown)


packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
```diff
@@ -38,6 +38,10 @@ export interface TooltipProps {
    * It controls the initial visibility of the Tooltip
    */
   isVisible?: boolean;
+  /**
+   * It controls whether or not the tooltip stays open when hovering on the content
+   */
+  canInteract?: boolean;
```
Miretxu, line 44: Is this prop name clear?

packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
```diff
@@ -164,19 +172,23 @@ export const Tooltip = ({
...
-          setShow(true);
+          if (!insidePopover) setShow(true);
           if (onMouseOver) onMouseOver(evt);
         }}
         onMouseLeave={(evt: MouseEvent) => {
-          setShow(false);
+          setTimeout(() => { 
+            if (!insidePopover) setShow(false);
+          }, delay)
           if (onMouseLeave) onMouseLeave(evt);
```
Miretxu, line 182: I'm not sure if `onMouseLeave()` , `onMouseOver()`, etc,  should be called inside the `setTimeout()`, after the delay? 

Also whether it should be called when the tooltip stays open (as the cursor is now on the content) but the cursor has already left the container.

Miretxu:
Experimenting with this hack
https://github.com/popperjs/popper-core/issues/667

denkristoffer:
Cool!

This is a small change that I think can have a big impact, for instance it lets users copy the text from the tooltip 👍 I'm not sure about the UX pattern of having actionable items in the tooltip itself, but I like this either way.

A quick thought from my side is whether the previous behaviour is worth keeping around at all. Maybe we can just make this the new default and avoid adding a new prop?

packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
gui-santos: Uhmmm wdyt about renaming it to `closeOnMouseLeave`?

And then the default value is true, the user that does not want it to close onMouseLeave would need to pass

```jsx
<Tooltip closeOnMouseLeave={false}>
  {/** ... */}
</Tooltip
```

packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
gui-santos: Yep I'm not sure as well, but looking into the code makes me think if we should do something with an effect

What I mean is something like we need to track if the mouse is over the target and/or the tooltip. And then depending on the prop you created we hide the tooltip when the mouse is not over any of them

```jsx
const [isHoveringTarget, setIsHoveringTarget] = useState(false)
const [isHoveringContent, setIsHoveringContent] = useState(false)

useEffect(() => {
  if (closeOnMouseLeave) {
    // change show value considering only target hover
    setShow(isHoveringTarget)
  } else {
    // change show value considering hover on both elements
    setShow(isHoveringContent || isHoveringTarget)
  }
}, [closeOnMouseLeave, isHoveringTarget, isHoveringContent])


<Container
  onMouseEnter={() => setIsHoveringTarget(true)}
  onMouseLeave={() => setIsHoveringTarget(false)}
/>
{show && (
  <span
    onMouseEnter={() => setIsHoveringContent(true)}
    onMouseLeave={() => setIsHoveringContent(false)}
  >
    {content}
  </span>
)}

```

I did not test this btw, but I think something like this might work

You are already doing something similar by using that `insidePopover` variable
But I think if you make the logic as a side effect of changing the state then you won't need the timeouts

wdyt?

gui-santos (changes requested):
Thanks for the contribution! This is going to be a cool addition

I made a suggestion that I think might work and make the code a bit easier to mantain

packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
mshaaban0: @gui-santos tbh, I'm not sure if we need all that magic 🧙🏻‍♀️I'd just keep it visible on hover at all times, do we really need the switch prop? 
so what I mean is move the content inside the container, and then everything is controlled by one `onMouseEnter` and one `onMouseLeave` ✨wdyt?

gui-santos: We can do that, but then we won't be able to use React Portals with the Tooltip in the future which means the tooltip will always calculate its position within its parent

Also I don't think there is any magic, it's just react hahahaha 
you have a state, when it changes it re-renders. If you want to do something related with the new state, it's a side effect
https://reactjs.org/docs/hooks-effect.html

I was just suggesting something to remove the necessity of the timeouts inside the mouse events
And using the react state system seemed pretty fair, it's already what the variable `insidePopover` does with the disadvantages of being just a variable and not a state

mshaaban0: ok cool, why do we need the portal though? I think for accessibility and keyboard focus order, It's always better to avoid using portal and keep everything encapsulated in one container

gui-santos: The dropdown is using portals now for example

Without it, this is what popper.js does
Screenshot 2020-11-13 at 16 30 43

You see, in relation to the window, we have enough space for the tooltip to show at the bottom side of its target. But popper checks the target's parent and it sees that there is no room for the tooltip so it shows the tooltip over the left side of the target

packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
Miretxu: Sounds good!

packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
Miretxu: @gui-santos I will try the suggested way using the effects and see if I can get rid of the setTimteout() 👍

The reason I added `setTImeout()` is that there is a gap between the Container and the Tooltip content, and as soon as the cursor leaves the Container and before reaching the Content, the `mouseLeave()` on Container is triggered. 
This then hides the Content immediately, hence the `mouseEnter()` on the Content (`<span >...`) never gets triggered. I wanted to have some delay before the tooltip gets hidden, so that we could actually detect the `mouseEnter()` over the Content.

Miretxu:
> Cool!
> 
> This is a small change that I think can have a big impact, for instance it lets users copy the text from the tooltip 👍 I'm not sure about the UX pattern of having actionable items in the tooltip itself, but I like this either way.
> 
> A quick thought from my side is whether the previous behaviour is worth keeping around at all. Maybe we can just make this the new default and avoid adding a new prop?

@denkristoffer thank you for the feedback! If I can get rid of the delay/setTimeout(), then I think we can make this the default behaviour. 
I had a plan to increase the delay for `hide` up to 3000ms and I thought that could be annoying if someone was using the tooltip with a very small content and the page had multiple tooltips. In this case, the desired behaviour would be to make it disappear right away.

packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
```diff
@@ -164,19 +181,19 @@ export const Tooltip = ({
         ref={elementRef}
         className={targetWrapperClassName}
         onMouseEnter={(evt: MouseEvent) => {
-          setShow(true);
+          setIsHoveringTarget(true);
           if (onMouseOver) onMouseOver(evt);
         }}
         onMouseLeave={(evt: MouseEvent) => {
-          setShow(false);
+          setTimeout(() => setIsHoveringTarget(false), delay);
```
Miretxu, line 188: @gui-santos I couldn't get it to work with no `setTimeout()`. The tooltip content gets hidden too quickly before I can move the cursor on to it. Do you see any drawbacks to using setTimeout() for this purpose?

packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
```diff
@@ -164,19 +181,19 @@ export const Tooltip = ({
...
-        }}
+      }}
         onFocus={(evt: FocusEvent) => {
-          setShow(true);
+          setIsHoveringTarget(true);
           if (onFocus) onFocus(evt);
         }}
         onBlur={(evt: FocusEvent) => {
-          setShow(false);
+          setTimeout(() => setIsHoveringTarget(false), delay);
```
Miretxu, line 196: I think we also need setTimeout here for the keyboard short cut to work properly i.e. to focus on the button inside the tooltip box

packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
gui-santos: Ahhhh yes you told me this before! I totally forgot about this. Sorry!
Yes I guess the timeout needs to exist

packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
gui-santos: Nop! I forgot that it's impossible to move your mouse quickly enough to keep the tooltip open, sorry

gui-santos (approved):
🙌 

this is great! Thanks for the contribution

mshaaban0 (approved):
Looks good Mire, Thank you 🍬

PR #638, opened 2020-11-13
fix(deps): update React types and fix csstype errors

denkristoffer:
# Purpose of PR

This PR ensures that the same React types are used in all components and fixes the `csstype` errors this lead to. It also adds `skipLibCheck: true` to the `tsconfig.json` in `forma-36-react-components` to avoid errors from `@types/react` itself.

Is this a `fix` or is it bigger than that? 🤔 

## PR Checklist

- [x] Tests are passing


gui-santos (commented):
Uh I don't know
It seems ok to me, curious about what the others think 👀

mshaaban0 (commented):
I think it's just the `css-type` syntax that is different, the type values remains the same.

Not so sure about `skipLibCheck` though, does that mean `@contentful/forma-36-tokens` and other packages are skipped if they contain type errors as well?

packages/components/forma-36-react-timepicker/package.json
```diff
@@ -33,8 +33,8 @@
     "@types/is-hotkey": "^0.1.1",
     "@types/jest": "^24.0.22",
     "@types/lodash.orderby": "^4.6.6",
-    "@types/react": "^16.9.11",
-    "@types/react-dom": "^16.9.3",
+    "@types/react": "^16.9.56",
+    "@types/react-dom": "^16.9.9",
```
mshaaban0, line 37: do you think we should add `csstype` as a dependency, so we control the version and avoid relying on it from react?

denkristoffer: Yeah good point, we should when we're using it directly anyway.

denkristoffer:
> Not so sure about skipLibCheck though, does that mean @contentful/forma-36-tokens and other packages are skipped if they contain type errors as well?

I think so, but we're building them separately and type checking them in that process, no?

gui-santos (approved):
LGTM

Issue #304, opened 2019-09-04
Issue with code splitting when creating extension


sohara:
I'm not sure if this is the best repo to file this issue, but when I created a new Contentful extension using the `@contentful/create-contentful-extension` package I was surprised to see how big the included component library made the final build. Below is the output I got when I ran `npm run deploy` on my extension. I am only using one component (TextField) so I'm surprised how expensive it is to use `forma-36-react-components`. I though Parcel is usually pretty good at this stuff so I wonder if the package itself needs to be changed to better support code splitting?

```sh
> publish-path@0.1.0 build /Users/sohara/code/soulcycle/publish-uri
> contentful-extension-scripts build

✨  Built in 7.52s.

build/src.f69400ca.js                                                       391.96 KB    6.96s
├── node_modules/@contentful/forma-36-react-components/dist/esm/index.js    241.46 KB    5.43s
├── node_modules/react-dom/cjs/react-dom.production.min.js                  109.72 KB    2.06s
├── node_modules/contentful-ui-extensions-sdk/dist/cf-extension-api.js       24.27 KB    530ms
├── node_modules/react/cjs/react.production.min.js                            6.41 KB    704ms
├── node_modules/scheduler/cjs/scheduler.production.min.js                    5.41 KB     66ms
├── src/index.tsx                                                             1.98 KB    969ms
├── node_modules/object-assign/index.js                                       1.95 KB    4.86s
├── node_modules/react-dom/index.js                                             374 B     89ms
├── node_modules/scheduler/index.js                                             165 B    3.51s
└── node_modules/react/index.js                                                 157 B     79ms

build/src.f69400ca.css                                                       57.56 KB    5.67s
├── src/index.css                                                                94 B    5.66s
└── node_modules/@contentful/forma-36-react-components/dist/styles.css           50 B    5.66s

build/index.html                                                                209 B    504ms

Extension size: 433 KB from 512 KB allowed.


✨  Successfully created extension:

Space: u860i9c783wb

Environment: dev-test-faq

Your extension: https://app.contentful.com/spaces/u860i9c783wb/environments/dev-test-faq/settings/extensions/publish-path
```

m10l:
Hi @sohara, thanks for adding an issue. This is the right place for it. We're aware of this pain point and we're working on a better solution so that it is less expensive to use a single component from the library.

We'll keep you updated when there's something to announce.

dmitri-suvorov-sumup:
[...]

PR #641, opened 2020-11-13
refactor(form-controls): use functional components and react hooks

mshaaban0:
# Purpose of PR

Refactor form controls to use functional components and react hooks, why:
* https://reactjs.org/docs/hooks-intro.html#motivation
* Remove `UNSAFE` lifecycle methods
* Cleaner code
* Should help with tree shaking and less bundle size, see #304

**Note:** This shouldn't introduce any breaking changes, so if you see anything please flag it.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/ControlledInput/ControlledInput.tsx
+++ b/packages/forma-36-react-components/src/components/ControlledInput/ControlledInput.tsx
@@ -3,6 +3,7 @@ import React, {
   ChangeEvent,
   FocusEvent,
   KeyboardEvent,
+  useCallback,
 } from 'react';
 import cn from 'classnames';
 
@@ -56,13 +57,17 @@ export const ControlledInput = (props: ControlledInputPropTypes) => {
     [styles['ControlledInput--disabled']]: disabled,
   });
 
-  const handleKeyDown = (e: KeyboardEvent<HTMLInputElement>) => {
-    const ESC = 27;
+  const handleKeyDown = useCallback(
+    (e: KeyboardEvent<HTMLInputElement>) => {
+      e.persist();
+      const ESC = 27;
 
-    if (e.keyCode === ESC && willBlurOnEsc) {
-      e.currentTarget.blur();
-    }
-  };
+      if (e.keyCode === ESC && willBlurOnEsc) {
+        e.currentTarget.blur();
+      }
+    },
+    [willBlurOnEsc],
+  );
 
   return (
     <input
--- a/packages/forma-36-react-components/src/components/ControlledInput/ControlledInput.tsx
+++ b/packages/forma-36-react-components/src/components/ControlledInput/ControlledInput.tsx
@@ -60,9 +60,9 @@ export const ControlledInput = (props: ControlledInputPropTypes) => {
   const handleKeyDown = useCallback(
     (e: KeyboardEvent<HTMLInputElement>) => {
       e.persist();
-      const ESC = 27;
+      const ESC = '27';
 
-      if (e.keyCode === ESC && willBlurOnEsc) {
+      if (e.nativeEvent.code === ESC && willBlurOnEsc) {
         e.currentTarget.blur();
       }
     },
--- a/packages/forma-36-react-components/src/components/ControlledInput/ControlledInput.tsx
+++ b/packages/forma-36-react-components/src/components/ControlledInput/ControlledInput.tsx
@@ -60,9 +60,8 @@ export const ControlledInput = (props: ControlledInputPropTypes) => {
   const handleKeyDown = useCallback(
     (e: KeyboardEvent<HTMLInputElement>) => {
       e.persist();
-      const ESC = '27';
 
-      if (e.nativeEvent.code === ESC && willBlurOnEsc) {
+      if (e.nativeEvent.code === 'Escape' && willBlurOnEsc) {
         e.currentTarget.blur();
       }
     },
--- a/packages/forma-36-react-components/src/components/ControlledInput/ControlledInput.tsx
+++ b/packages/forma-36-react-components/src/components/ControlledInput/ControlledInput.tsx
@@ -1,5 +1,4 @@
 import React, {
-  Component,
   EventHandler,
   ChangeEvent,
   FocusEvent,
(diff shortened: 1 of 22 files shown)


packages/forma-36-react-components/src/components/TextField/TextField.tsx
```diff
@@ -47,118 +48,104 @@ const defaultProps: Partial<TextFieldProps> = {
...
-    this.setState({ value: (evt.target as HTMLInputElement).value });
-    if (this.props.onChange) this.props.onChange(evt);
+  const handleOnChange = (
+    e: ChangeEvent<HTMLTextAreaElement | HTMLInputElement>,
+  ) => {
+    setValueState(e.target.value);
+    if (onChange) onChange(e);
   };
 
-  static getDerivedStateFromProps(
```
mshaaban0, line 63: please review this one carefully, should this be replaces with `useEffect`? see https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops

gui-santos (approved):
Wow! Thanks a lot @mshaaban0

packages/forma-36-react-components/src/components/TextField/TextField.tsx
gui-santos: Do we need something to replace it?

It looks like it was setting an initial value to the input, but I believe your useState is already doing that 👌 
no?

mshaaban0: I honestly don't know what exactly it was doing, to me the component works as expected, not sure if there's some magic behind it though

packages/forma-36-react-components/src/components/Textarea/Textarea.tsx
```diff
@@ -44,87 +46,81 @@ const defaultProps: Partial<TextareaProps> = {
...
-        value: nextProps.value,
-      });
-    }
-  }
+  useEffect(() => {
+    setValueState(value);
+  }, [value]);
 
-  handleKeyDown = (e: KeyboardEvent<HTMLTextAreaElement>) => {
+  const handleKeyDown = (e: KeyboardEvent<HTMLTextAreaElement>) => {
```
denkristoffer, line 76: I think you should wrap this in a `useCallback` to avoid rebinding it on every keystroke 🙂

denkristoffer (changes requested):

packages/forma-36-react-components/src/components/ControlledInput/ControlledInput.tsx
```diff
@@ -34,72 +34,74 @@ const defaultProps: Partial<ControlledInputPropTypes> = {
...
+    [styles['ControlledInput--disabled']]: disabled,
+  });
 
-    if (e.keyCode === ESC && this.props.willBlurOnEsc) {
-      e.currentTarget.blur();
-    }
-  };
+  const handleKeyDown = useCallback(
+    (e: KeyboardEvent<HTMLInputElement>) => {
+      e.persist();
```
mshaaban0, line 62: Is `e.persist()` a good idea here? react throws many warnings without it.
React peeps say they will get rid of [pooling](https://reactjs.org/docs/events.html) and `e.persist()` will no longer be needed in react `17.x`

denkristoffer: Yes it's fine here, and even though it doesn't do anything in React 17 it shouldn't complain, so we can keep it and stay compatible with both 16 and 17 👍

packages/forma-36-react-components/src/components/Select/Select.tsx
```diff
@@ -38,95 +36,84 @@ const defaultProps: Partial<SelectProps> = {
...
-  };
+  const handleKeyDown = (e: KeyboardEvent<HTMLSelectElement>) => {
+    const ESC = 27;
 
-  UNSAFE_componentWillReceiveProps(nextProps: SelectProps) {
-    if (this.props.value !== nextProps.value) {
-      this.setState({
-        value: nextProps.value,
-      });
+    if (e.keyCode === ESC && willBlurOnEsc) {
```
denkristoffer, line 62: Same thoughts here as the other key handler.

packages/forma-36-react-components/src/components/Select/Select.tsx
```diff
@@ -38,95 +36,84 @@ const defaultProps: Partial<SelectProps> = {
...
+      e.currentTarget.blur();
     }
-  }
+  };
 
-  handleKeyDown = (e: KeyboardEvent<HTMLSelectElement>) => {
-    const ESC = 27;
+  const handleChange = (e: ChangeEvent<HTMLSelectElement>) => {
```
denkristoffer, line 67: I think we should wrap all event handlers in `useCallback` just for good measure.

packages/forma-36-react-components/src/components/SelectField/SelectField.tsx
```diff
@@ -39,101 +40,94 @@ const defaultProps: Partial<SelectFieldProps> = {
...
-      onChange,
-      onBlur,
-      value,
-      name,
-      id,
-      ...otherProps
-    } = this.props;
+  useEffect(() => {
+    setValueState(value);
+  }, [value]);
```
denkristoffer, line 76: I'm not sure I get what this is necessary for, are there cases where you're not setting the state value to the value it already has?

packages/forma-36-react-components/src/components/SelectField/SelectField.tsx
```diff
@@ -39,101 +40,94 @@ const defaultProps: Partial<SelectFieldProps> = {
...
+    <div className={classNames} {...otherProps} data-test-id={testId}>
+      <div className={styles['SelectField__label-wrapper']}>
+        <FormLabel {...{ ...formLabelProps, htmlFor: id, required }}>
+          {labelText}
+        </FormLabel>
+        {textLinkProps && (
+          <TextLink
+            {...{
+              ...textLinkProps,
+              className: styles['SelectField__label-link'],
```
denkristoffer, line 90: Is there a reason to spread like this? I've never seen it done before 🤔

packages/forma-36-react-components/src/components/TextField/TextField.tsx
```diff
@@ -47,118 +48,104 @@ const defaultProps: Partial<TextFieldProps> = {
...
+  } = props;
+  const [valueState, setValueState] = useState<string | undefined>(value);
 
   // Store a copy of the value in state.
   // This is used by this component when the `countCharacters`
   // option is on
-  handleOnChange = (evt: ChangeEvent) => {
-    this.setState({ value: (evt.target as HTMLInputElement).value });
-    if (this.props.onChange) this.props.onChange(evt);
+  const handleOnChange = (
```
denkristoffer, line 78: `useCallback` would be nice here, I imagine performance can drop quite a bit with quick typing otherwise?

denkristoffer (commented):
Thank you for the extensive work ❤️ I've added a few questions and thoughts. The biggest thing is that I think we should make sure to wrap event handlers in `useCallback` for good measure.

mshaaban0:
@denkristoffer Thanks for all the useful comments, addressed 👍

packages/forma-36-react-components/src/components/ControlledInput/ControlledInput.tsx
```diff
@@ -34,72 +34,74 @@ const defaultProps: Partial<ControlledInputPropTypes> = {
...
-      name,
-      onChange,
-      checked,
-      value,
-      type,
-      labelText,
-      willBlurOnEsc,
-      ...otherProps
-    } = this.props;
+      if (e.nativeEvent.code === ESC && willBlurOnEsc) {
```
denkristoffer, line 65: My bad, `code` works a bit differently than `keyCode`, as it checks against the key name instead of a number. See https://github.com/contentful/forma-36/blob/986eabd3d981818d8a3e35fb03b04132bcbd6033/packages/forma-36-react-components/src/components/Dropdown/Dropdown.tsx#L215

That's actually why I like it, since the condition is easier to read 🙂

mshaaban0: Ah, I should have tested it, cool though, it's clean. 👍

denkristoffer (approved):

denkristoffer:
Great!

packages/forma-36-react-components/src/components/SelectField/SelectField.tsx
mshaaban0: @denkristoffer I had to put this back, TimePicker tests were failing because `value` prop changes doesn't get propagated properly without it. 

it's like `componentWillReceiveProps`, However this is kinda confusing; Shouldn't react functional components re-render automatically on prop value changes? 🤔

denkristoffer: Yeah I did the same with the Dropdown, but I share your thoughts. It seems like a code smell. Anyway it's fine for now, maybe we can rethink this pattern at some point.

burakukula (approved):
Amazing! 🎉

PR #645, opened 2020-11-18
refactor(card): make it a functional component and add mdx stories

gui-santos:
# Purpose of PR

I'm currently working on something in the Card component, but before getting into that I refactored it to a functional component and also tested MDX stories for storybook

I liked it! Wdyt?

One of the pros is that we write documentation and stories in the same place and storybook generate both interactive stories(where we can play with controls) and a documentation page

And we use all the features Markdown documents have and all the features storybook has

On the other side, because the stories are in MDX we lose some features from coding TS in vscode. There is a way to mix MDX and CSF stories, but I found it very confusing to make it work. We also lose the README.md file which is pretty convenient when navigating through the repo in GitHub

## PR Checklist

- [ ] Tests are passing


packages/forma-36-react-components/src/components/Card/Card.tsx
```diff
@@ -55,48 +50,44 @@ const defaultProps: Partial<CardPropTypes> = {
   selected: false,
 };
 
-export class Card extends Component<CardPropTypes> {
-  static defaultProps = defaultProps;
+export const Card: React.FC<CardPropTypes> = ({
```
denkristoffer, line 53: I'm not a big fan of using this `FC` type as it leads to implicit props being included on the component that we might not want in all cases, for instance `children`. But if we do it, we should agree to do it for all our components 🙂 What do you think?

denkristoffer (commented):
I really like the idea of writing documentation and stories in the same place, if we can make reusing it everywhere work somehow.

> We also lose the README.md file which is pretty convenient when navigating through the repo in GitHub

This is a shame though 😞

burakukula:
@gui-santos the idea is great! It would be awesome to just write stories, but I think it's a petty to loose README. Having docs in README would gives us more flexibility I think - we could display README on the website (I guess, we still need to check if that's possible though). Does it make sense to have both?

packages/forma-36-react-components/src/components/Card/Card.tsx
gui-santos: Oohhhhh I never thought about that... you are right!

I'm not attached to `FC` tbh. It's just the way I'm used to writing components by now hahaha
And now that you mentioned the implicit props, I agree we should use another type as a convention
Should we declare `children` in the Props interface every time a component uses it?

gui-santos:
> @gui-santos the idea is great! It would be awesome to just write stories, but I think it's a petty to loose README. Having docs in README would give us more flexibility I think - we could display README on the website (I guess, we still need to check if that's possible though). Does it make sense to have both?

Yep, good questions. I guess we would need something else to generate pages for the website

But let's check other options and choose the easiest to maintain and give some documentation to users ✌️

packages/forma-36-react-components/src/components/Card/Card.tsx
denkristoffer: > Should we declare children in the Props interface every time a component uses it?

That would be my preference at least. I'd rather add a `children` prop when children are used than delete it from the implicit props when they aren't used 😅

gui-santos:
Closing this PR since its only purpose was testing MDX stories

PR #647, opened 2020-11-19
refactor: function components with readme

denkristoffer:
# Purpose of PR

Refactors the `<Asset />` component to a function component and adds a README file.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Asset/Asset.tsx
+++ b/packages/forma-36-react-components/src/components/Asset/Asset.tsx
@@ -1,5 +1,6 @@
-import React, { Component } from 'react';
+import React from 'react';
 import cn from 'classnames';
+
 import { AssetState } from '../Card/AssetCard/AssetCard';
 import { AssetIcon } from './AssetIcon/AssetIcon';
 
@@ -23,6 +24,36 @@ export function isAssetType(type: string): type is AssetType {
   return Object.keys(types).includes(type);
 }
 
+const renderImage = (src: string, title: string) => (
+  <React.Fragment>
+    <div className={styles['Asset__image-container']}>
+      <img
+        className={styles['Asset__image-container__image']}
+        src={src}
+        alt={title}
+      />
+    </div>
+    {title && (
+      <div className={styles['Asset__title-container']}>
+        <span className={styles['Asset__title-container__title']}>{title}</span>
+      </div>
+    )}
+  </React.Fragment>
+);
+
+const renderAsset = (type: AssetType, title: string) => {
+  return (
+    <div className={styles['Asset__asset-container']}>
+      <div className={styles['Asset__illustration-container']}>
+        <AssetIcon type={type} />
+      </div>
+      {title && (
+        <span className={styles['Asset__asset-container__title']}>{title}</span>
+      )}
+    </div>
+  );
+};
+
 export type AssetType = keyof typeof types;
 
 export interface AssetProps {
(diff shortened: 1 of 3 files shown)


packages/forma-36-react-components/src/components/Asset/Asset.tsx
```diff
@@ -23,6 +24,36 @@ export function isAssetType(type: string): type is AssetType {
   return Object.keys(types).includes(type);
 }
 
+const renderImage = (src: string, title: string) => (
+  <React.Fragment>
```
gui-santos, line 28: Is there a difference between `<>` and `<React.Fragment>`? Or is a matter of personal preference?

If it's just preference, should we have a code styling convention?

packages/forma-36-react-components/src/components/Asset/Asset.tsx
```diff
@@ -23,6 +24,36 @@ export function isAssetType(type: string): type is AssetType {
...
+    <div className={styles['Asset__asset-container']}>
+      <div className={styles['Asset__illustration-container']}>
+        <AssetIcon type={type} />
+      </div>
+      {title && (
+        <span className={styles['Asset__asset-container__title']}>{title}</span>
+      )}
+    </div>
+  );
+};
```
gui-santos, line 55: Should we convert these render functions into functional components? They are just receiving two arguments and returning jsx so... wdyt?

packages/forma-36-react-components/src/components/Asset/Asset.tsx
```diff
@@ -34,75 +65,33 @@ export interface AssetProps {
...
+  type,
+  ...otherProps
+}: AssetProps): React.ReactElement {
+  const classNames = cn(styles.Asset, className);
 
-export class Asset extends Component<AssetProps> {
-  static defaultProps = defaultProps;
+  // Archived images will not have a preview available
+  const asImage =
+    type && type === 'image' && (!status || status !== 'archived') && src;
```
gui-santos, line 81: This condition is a bit hard to read in my opinion. Could we make it simpler?

If I understood correctly, we show a preview for when the Asset is an image. The exception is if `status === 'archived'`, right?

```jsx
// check if asset is of type image and has a src
const isImage = type === 'image' && !!src

// if status is archived, do not show preview
const showPreview = isImage && status !== 'archived'
```

packages/forma-36-react-components/src/components/Asset/AssetIcon/AssetIcon.tsx
```diff
@@ -14,32 +14,26 @@ export interface AssetIconProps
...
-    }
-
-    return (
-      <Illustration
-        illustration={illustrationName as IllustrationType}
-        {...otherProps}
-      />
-    );
+export function AssetIcon({ type, ...otherProps }: AssetIconProps) {
+  let illustrationName = type!.charAt(0).toUpperCase() + type!.slice(1); // eslint-disable-line @typescript-eslint/no-non-null-assertion
```
gui-santos, line 21: Why do we need to disable @typescript-eslint/no-non-null-assertion?

I'm actually not very familiar with  non-null assertions, but can't we refactor it instead of disabling the rule?

Something from these example, maybe?
https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-non-null-assertion.md
https://palantir.github.io/tslint/rules/no-non-null-assertion/

packages/forma-36-react-components/src/components/Asset/README.mdx
```diff
@@ -0,0 +1,107 @@
+import { Asset } from './Asset';
+
+# Asset
+
+[Source code](https://github.com/contentful/forma-36/blob/master/packages/forma-36-react-components/src/components/Asset/Asset.tsx)
```
gui-santos, line 5: I don't think we need this link in the README, do we?

I mean if the person is reading this, then they are in github already

packages/forma-36-react-components/src/components/Asset/README.mdx
```diff
@@ -0,0 +1,107 @@
...
+The asset component is a visual representation of an asset file such as a video or a PDF document.
+
+<Asset title="Archive of files" type="archive" />
+
+## Table of contents
+
+- [Usage](#usage)
+  - [With an image](#with-an-image)
+  - [With link and target](#with-link-and-target)
+- [Props](#props)
```
gui-santos, line 16: I think this Table of contents is a leftover from the Card MDX, isn't it? hahaha

gui-santos (changes requested):
Looks good so far, I just have some questions 🙏

gui-santos:
Oops I think this was closed while I was reviewing it

denkristoffer:
@gui-santos I decided to make this a part of a larger PR. I will use your thoughts in the new one 😄

packages/forma-36-react-components/src/components/Asset/AssetIcon/AssetIcon.tsx
denkristoffer: The problem is the default props are used currently. IMO this would be solved by moving from `defaultProps` to "normal" JS with default values. So we go from 

```jsx
export function Asset({
  className,
  src,
  status,
  testId,
  title,
  type,
  ...otherProps
}: AssetProps): React.ReactElement { ... }

Asset.defaultProps = {
  type: 'image',
  testId: 'cf-ui-asset',
};
```

to

```jsx
export function Asset({
  className,
  src,
  status,
  testId = 'cf-ui-asset',
  title,
  type = 'image',
  ...otherProps
}: AssetProps): React.ReactElement { ... }
```

packages/forma-36-react-components/src/components/Asset/Asset.tsx
denkristoffer: There's no difference. Most people think `<>` look better, but `<Fragment />` can lead to less issues from switching JSX pragma.

packages/forma-36-react-components/src/components/Asset/AssetIcon/AssetIcon.tsx
gui-santos: Ohhhhh I prefer that too, but a while ago we had this proposal
https://github.com/contentful/forma-36/pull/574

And because of that, it is done the way it is now

gui-santos: I don't remember exactly why we could not use the defaults in the arguments, but maybe the proposal has some explanation

packages/forma-36-react-components/src/components/Asset/Asset.tsx
gui-santos: I had to google JSX pragma
But at a first glance, it does not look like something we need to worry about
In any case, it would be nice to have a lint rule or something to force us to use only one way in the codebase

I like only `<>` because it's less typing honestly 🤷

packages/forma-36-react-components/src/components/Asset/AssetIcon/AssetIcon.tsx
gui-santos: maybe @andipaetzold has a better memory

denkristoffer: Yeah I remember seeing that. I think it can be kind of a first step towards what I mentioned, since the default values are a pretty big change that only work with function components. That's why I'm working on refactoring all our classes to functions first, then after that we can look into changing from `defaultProps` to default values if we want to 🙂 Then once that is done we can get rid of non-null assertions.

andipaetzold: Yes, using the "normal" function components default values instead of `defaultProps` should solve this issue and the types would be correct. Especially as ["defaultProps on functions will eventually get deprecated."](https://twitter.com/dan_abramov/status/1133878326358171650), I think we should try to slowly get away from `defaultProps`.

In the mentioned PR I didn't get rid of `defaultProps` because I would have to convert all class components to functional components and wanted to do have a simple and consistent migration. If all components were functional components, I would have dropped `defaultProps`.

I have something in my head that the storybook might break if we remove the `defaultProps`, but I do not fully remember.

packages/forma-36-react-components/src/components/Asset/Asset.tsx
denkristoffer: It would mostly be relevant if we used something like Emotion. Agreed it would be nice if we could lint it.

PR #648, opened 2020-11-19
Feat/card selected state

gui-santos:
# Purpose of PR

selectable_cards

We are introducing a new pattern for cards in the web app where we will be able to use cards as selectable elements
@Gracegross has more context about it

This PR updates the selected state and implement keyboard navigation to the card

So, a Card has these states now
Screenshot 2020-11-19 at 14 46 46

🌮  => selected/active
🍕 => focused and hover
🥦  => idle

## Open question

@Gracegross , should we update the look of the states in the ToggleButton component as well?

## PR Checklist

- [x] Tests are passing


packages/forma-36-react-components/src/components/Card/Card.tsx
```diff
@@ -27,13 +23,17 @@ export interface BaseCardProps {
    * Class names to be appended to the className prop of the component
    */
   className?: string;
+  /**
+   * Child nodes to be rendered in the component
+   */
+  children?: React.ReactNode;
```
mshaaban0, line 29: Nit: import `ReactNode` from react like MouseEvent, etc.

packages/forma-36-react-components/src/components/Card/Card.tsx
```diff
@@ -46,57 +46,55 @@ interface CardPropTypes extends BaseCardProps {
...
 
-    this.props.onClick(event);
-  };
-
-  render() {
-    const {
-      className,
-      testId,
-      children,
+  return React.createElement(
```
mshaaban0, line 76: should we replace this with the common `return (...)`?

packages/forma-36-react-components/src/components/Card/Card.tsx
```diff
@@ -46,57 +46,55 @@ interface CardPropTypes extends BaseCardProps {
...
-      },
-      children,
-    );
-  }
-}
+      target: href && target,
+      'data-test-id': testId,
+      className: classNames,
+      ...(!!onClick && {
+        tabindex: '0', // necessary to make the div accessible with tab navigation
```
mshaaban0, line 84: cool, let's have an `aria-label` here then?

mshaaban0 (approved):
Nice Gui 👏

packages/forma-36-react-components/src/components/Card/Card.tsx
```diff
@@ -46,57 +46,55 @@ interface CardPropTypes extends BaseCardProps {
...
+        onClick: (event: MouseEvent<HTMLElement>) => onClick(event),
+        onKeyDown: (event: KeyboardEvent<HTMLElement>) => {
+          if (event.keyCode === 13) {
```
denkristoffer, line 87: [`event.keyCode` is deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) and key code numbers are hard to understand. What about using [event.code](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) instead? `event.code === "Enter"`

denkristoffer: Also, something tells me usual a11y practices would say the card should be selectable with the space key.

packages/forma-36-react-components/src/components/Card/Card.tsx
gui-santos: Ah ok! I will change it
thanks!

packages/forma-36-react-components/src/components/Card/Card.tsx
gui-santos: I prefer also hahaha
I didn't replace it because I thought there was a reason to use the createElement

I will refactor it! thanks :)

packages/forma-36-react-components/src/components/Card/Card.tsx
gui-santos: what should I put in the label? `aria-label="card"` 👀

packages/forma-36-react-components/src/components/Card/Card.tsx
gui-santos: Oh I didn't know it was going to be deprecated!

I will change that and implement the space key as well
thanks 🙌

packages/forma-36-react-components/src/components/Card/Card.tsx
mshaaban0: If there's a card title you can point to it using `aria-labelledby="idOfTheTitle"`, if no title, then maybe a `label` prop like the one we have in [IconButton](https://github.com/contentful/forma-36/blob/master/packages/forma-36-react-components/src/components/IconButton/IconButton.tsx#L9), also don't make it optional 😈

gui-santos: got it!
Yeah I will make it mandatory, maybe even a mean message in the prop description like:
"Yep this is mandatory, if you think it should be optional you can't sit with us during lunch!"

mshaaban0: Hahaha, great, thanks Gui 💟

Gracegross (commented):
This is sick!!! I think the thicker "selected" border looks great here. Makes it really clear.

Toggles are usually smaller than cards so we'd have to make sure it doesn't look crazy if we matched the style.

WDYT?

For the record I select 🌮

gui-santos:
> Toggles are usually smaller than cards so we'd have to make sure it doesn't look crazy if we matched the style.
> 
> WDYT?

Let's compare!

this is toggle-button in idle state
Screenshot 2020-11-23 at 10 57 03

this is the current active/selected state
Screenshot 2020-11-23 at 10 55 15

this is the same active/selected state we will be using in the Card component
Screenshot 2020-11-23 at 10 56 50

I think the current one looks too much like the hover state of the card, but the new one is too violent hahahah. Maybe something in between? (I will test 2px later)


> For the record I select 🌮

Yeah! Fun-fact: on the day I was doing this thing, I had to order 🌮  for lunch

Gracegross:
> > Toggles are usually smaller than cards so we'd have to make sure it doesn't look crazy if we matched the style.
> > WDYT?
> 
> @Gracegross Let's compare!
> 
> this is toggle-button in idle state
> Screenshot 2020-11-23 at 10 57 03
> 
> this is the current active/selected state
> Screenshot 2020-11-23 at 10 55 15
> 
> this is the same active/selected state we will be using in the Card component
> Screenshot 2020-11-23 at 10 56 50
> 
> I think the current one looks too much like the hover state of the card, but the new one is too violent hahahah. Maybe something in between? (I will test 2px later)
> 
> > For the record I select 🌮
> 
> Yeah! Fun-fact: on the day I was doing this thing, I had to order 🌮 for lunch

I agree the current toggle isn't different enough from hover! I think we should match them

packages/forma-36-react-components/src/components/Card/Card.tsx
```diff
@@ -27,76 +23,79 @@ export interface BaseCardProps {
...
-      onClick,
-      padding,
-      selected,
-      ...otherProps
-    } = this.props;
+  if (href) {
+    Element = 'a';
+  } else if (onClick) {
+    Element = 'button';
+  }
```
gui-santos, line 79: @mshaaban0 and @denkristoffer 
I implemented the changes about the keyDown event, but then I started reading the ARIA links Moe sent to me and they kinda made me think "Wait a minute... I think we should use a `<button>` when the card has onClick instead of `div`"

so I changed it to do exactly that... and because it's a button, we do not need onKeyDown or tabindex

we still need aria-label and a aria-pressed

WDYT?
P.S.: this makes the selectable card very similar to the ToggleButton, but with a11y features working
I will update ToggleButton in another PR, but maybe in Forma36 v4 we will have to think if we need both 👀

mshaaban0: even better 👏
> we still need aria-label and a aria-pressed
+1

denkristoffer: Yes absolutely, huge +1! 😄

PR #650, opened 2020-11-19
style(iconbutton.css): increase clickable area for icon button

### LINKED ISSUE #416: Increase clickable area of action button in Card components
Screenshot 2020-02-11 at 10 11 55

I would be nice to increase a clickable area of this button to improve usability and a11y. As for me, that click at right right corner should always trigger an action. Right now I miss almost every second time.

joneskj55:
fix #416

# Purpose of PR
This PR addresses issue #416 increasing the clickable area of the .IconButton.

## PR Checklist

- [x] Tests are not required
- [x] Tests are passing
- [x] Storybook stories are not required
- [x] Usage notes are not required


packages/forma-36-react-components/src/components/IconButton/IconButton.css
```diff
@@ -7,7 +7,7 @@
   display: inline-block;
   cursor: pointer;
   border: 0;
-  padding: 0;
+  padding: 10px;
```
mshaaban0, line 10: Thanks a lot for the contribution @joneskj55 

Can you please use a spacing token here, maybe `spacing-xs`?
Usage would be similar to [Accordion.css](https://github.com/contentful/forma-36/blob/master/packages/forma-36-react-components/src/components/Accordion/Accordion.css#L26)

joneskj55: Will do.

joneskj55: I ended up using `var(--spacing-xs)` because it was closest in rem to the 10px of spacing. Please let me know if this will suffice. If not, I can use `xs` as you suggested.

mshaaban0: Great, thank you. will have a look, and test the component with its consumers.

burakukula:
Hey @joneskj55! Thank you for your contribution. 🎉 
👍  for using tokens. It does not work for me right now. I think, In order for this to work, you would need to import the dimensions from settings at the top of this css file, like follows: `@import 'settings/dimensions';`. 🙂

joneskj55:
@burakukula Aaah, you're right! I'll add another commit to the same branch. Thanks!

packages/forma-36-react-components/src/components/IconButton/IconButton.css
```diff
@@ -1,13 +1,14 @@
...
 @import 'settings/transitions';
+@import 'settings/dimensions';
 
 .IconButton {
   composes: focus-outline--default from './../../styles/settings/a11y.css';
   display: inline-block;
   cursor: pointer;
   border: 0;
-  padding: 0;
+  padding: var(--spacing-s);
```
mshaaban0, line 11: Thanks again for the PR @joneskj55 
I just did some quick visual review, I believe we need to fix the following components before we get this in.

I am not sure if we should treat this as a fix, It's more like a breaking change, I'm afraid consumers would run into visual issues upon upgrading.

cc:// @gui-santos @burakukula @denkristoffer @m10l 

**EntryCard**
Screenshot 2020-11-23 at 11 09 44

**EditorToolbarButton**
Screenshot 2020-11-23 at 11 11 30

**EntityListItem**
Screenshot 2020-11-23 at 11 12 41

**ModalHeader**
Screenshot 2020-11-23 at 11 13 38

**InlineEntryCard**
Screenshot 2020-11-23 at 11 17 58

**CardHeader**
Screenshot 2020-11-23 at 11 19 17

**AutoComplete**
Screenshot 2020-11-23 at 11 24 26

**NotificationItem**
Screenshot 2020-11-23 at 11 25 57

burakukula: @mshaaban0 yeah, I agree. I guess all of the examples above needs to be adjusted if we would like to move forward.
And yeah, +1 for including this change into the next major release.

gui-santos: Hmmm yeahh I think it's a breaking change

but @joneskj55 if you really need this change. You can have it by passing a class to Button using the className prop
https://github.com/contentful/forma-36/blob/master/packages/forma-36-react-components/src/components/Button/Button.tsx#L37

joneskj55: Thanks @gui-santos!

gui-santos:
I will close this now since we found an alternative

But if I'm wrong, please re-open this @joneskj55 😄 
and ping us if you need any help always 🙌

Issue #651, opened 2020-11-20
💡 Proposal - Refactor to function components 


denkristoffer:
# Forma 36 contribution proposal

## The problem

Forma 36 currently uses a mix of class and function components. There are no difference between the two for consumers of the package, but function components are preferable for us Forma 36 developers for a few reasons:

- Hopefully building all components in the same way makes it easier to understand the codebase and easier to contribute to F36. I think this is especially true for newcomers to React, who might not even have learned how to build class components.
- They allow for the use of hooks.
- They allow us to migrate away from defaultProps to normal JS default values.

## The proposed solution

- [x] ~~Accordion: Is already a function component~~
- [x] ~~Asset: Under way in #652~~
- [x] ~~Autocomplete: Is already a function component~~
- [x] ~~Button: Under way in #641~~
- [x] ~~Card: Under way in #648~~
- [x] ~~Checkbox: Under way in #641~~
- [x] ~~CheckboxField: Under way in #641~~
- [x] ~~ControlledInput: Under way in #641~~
- [x] ~~ControlledInputField: Under way in #641~~
- [x] ~~CopyButton: Under way in #652~~
- [x] ~~DateTime: Is already a function component~~
- [x] ~~RelativeDate: Is already a function component~~
- [x] ~~Dropdown: Is already a function component~~
- [x] ~~EditorToolbar: Under way in #652~~
- [x] ~~EmptyState: Under way in #717~~
- [x] ~~EntityList: Is already a function component~~
- [x] ~~FieldGroup: Under way in #641~~
- [x] ~~Flex: Is already a function component~~
- [x] ~~Form: Under way in #641~~
- [x] ~~FormLabel: Under way in #641~~
- [x] ~~Grid: Is already a function component~~
- [x] ~~HelpText: Under way in #717~~
- [x] ~~Icon: Under way in #717~~
- [x] ~~IconButton: Under way in #641~~
- [x] ~~Illustration: Under way in #717~~
- [ ] InViewport: This is going to be bothersome to changerefactor
- [x] ~~List: Under way in #717~~
- [x] ~~Modal: Under way in #717~~ 
- [x] ~~Note: Under way in #717~~ 
- [x] Notification components: Under way in #720
- [x] ~~Option: Under way in #641~~
- [x] ~~Pill: Is already a function component.~~
- [x] ~~Portal: Is already a function component.~~
- [x] ~~ProductIcon: Is already a function component.~~
- [x] ~~RadioButton: Under way in #641~~
[...]

PR #652, opened 2020-11-20
refactor: function components with MDX readme

denkristoffer:
# Purpose of PR

- Refactors `Asset`, `CopyButton`, `EditorToolbar`, `EditorToolbarButton`, and `EditorToolbarDivider` to function components as part of #651.
- Updates and adds README files as necessary.
- Updates Storybook files to CSF.
- Updates the Plop template to function components.
- Disables the rule `@typescript-eslint/no-object-literal-type-assertion` in Storybook story files, which lets us assert types for default exported objects without using an unnecessary `const`.

## PR Checklist

- [x] Tests are passing


packages/forma-36-react-components/src/components/Asset/AssetIcon/README.mdx
```diff
@@ -0,0 +1,19 @@
+import { AssetIcon } from './AssetIcon';
```
burakukula, line 1: @denkristoffer awesome! I think README could be based on the following structure:

1. Description
2. Variations/Overview
3. Examples of usage
4. Best practices
5. Content recommendations (if applying)
6. Accessibility (if applying)

This is what we were trying to do in other components, like Button or Accordion. I will try to document it somehow, maybe create template for it or something.

packages/forma-36-react-components/src/components/Asset/AssetIcon/AssetIcon.tsx
```diff
@@ -14,32 +14,26 @@ export interface AssetIconProps
...
-    }
-
-    return (
-      <Illustration
-        illustration={illustrationName as IllustrationType}
-        {...otherProps}
-      />
-    );
+export function AssetIcon({ type, ...otherProps }: AssetIconProps) {
+  let illustrationName = type!.charAt(0).toUpperCase() + type!.slice(1); // eslint-disable-line @typescript-eslint/no-non-null-assertion
```
gui-santos, line 21: I thought we were going to pass the defaultProps in the arguments of the function so we would not need this disable-line, or is this something else?

packages/forma-36-react-components/src/components/CopyButton/CopyButton.stories.tsx
```diff
@@ -1,31 +1,33 @@
...
 
-storiesOf('Components/CopyButton', module)
-  .addParameters({
+export default {
+  argTypes: {
+    children: { control: { disable: true } },
+    className: { control: { disable: true } },
+    testId: { control: { disable: true } },
+    tooltipText: { control: 'text' },
+    tooltipCopiedText: { control: 'text' },
```
gui-santos, line 14: Do we need this? I thought storybook would automatically guess the controls for these props based on their types

gui-santos (changes requested):
Hey this is great!

I have only questions about two things:

1. Are we moving the default props to the argument of the function in all components?

2. I think you don't need to define some controls, only if you need the control to work differently from what storybook can guess with the type. For example, disabling className is something we need to do, but if a prop is a type 'string' sb will create a text control for it. Or if its type is `'a' | 'b' | 'c'` sb will create a select.... it should create
so maybe you can remove some control declarations, idk

packages/forma-36-react-components/src/components/Asset/AssetIcon/AssetIcon.stories.tsx
```diff
@@ -1,19 +1,30 @@
...
-  .add('with type image', () => (
-    <AssetIcon
-      className={text('className', '')}
-      type={select('type', types, 'image') as AssetType}
-    />
-  ));
+    notes,
+  },
+  title: 'Components/Asset/AssetIcon',
+} as Meta;
```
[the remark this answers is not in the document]
denkristoffer: It's just to have the default export typed in case we need it at some point. I saw it in the Storybook docs somewhere but I can't find it now 🤔

packages/forma-36-react-components/src/components/Asset/AssetIcon/AssetIcon.tsx
denkristoffer: I wanted to wait and move all components from `defaultProps` to default arguments in a separate PR, because I was unsure if it was a breaking change.

packages/forma-36-react-components/src/components/Asset/Asset.stories.tsx
```diff
@@ -1,36 +1,55 @@
...
+export default {
+  argTypes: {
+    className: { control: { disable: true } },
+    src: { control: 'text', defaultValue: 'https://placekitten.com/200/300' },
+    status: {
+      control: {
+        type: 'select',
+        options: ['archived', 'changed', 'draft', 'published'],
+      },
+    },
```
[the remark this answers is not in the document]
denkristoffer: No, the problem is that the AssetStatus is only a type, it doesn't have string values.

packages/forma-36-react-components/src/components/Asset/AssetIcon/AssetIcon.stories.tsx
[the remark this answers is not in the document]
gui-santos: Cool! I will start doing it as well then
Thanks

packages/forma-36-react-components/src/components/Asset/AssetIcon/AssetIcon.tsx
gui-santos: Makes sense!

gui-santos (approved):

packages/forma-36-react-components/src/components/Asset/AssetIcon/AssetIcon.stories.tsx
```diff
@@ -1,19 +1,30 @@
 import React from 'react';
-import { storiesOf } from '@storybook/react';
-import { text, select } from '@storybook/addon-knobs';
+import type { Meta, Story } from '@storybook/react/types-6-0';
 
-import { types, AssetType } from '../Asset';
 import { AssetIcon } from './AssetIcon';
+import type { AssetIconProps } from './AssetIcon';
+import { types } from '../Asset';
+import notes from './README.mdx';
```
burakukula, line 7: @denkristoffer is that working for you?

packages/forma-36-react-components/src/components/Asset/AssetIcon/README.mdx
denkristoffer: I tried to follow this structure now, although I don't really know about best practices for these components. Do you think there's anything to say there?

packages/forma-36-react-components/src/components/Asset/AssetIcon/README.mdx
```diff
@@ -0,0 +1,79 @@
...
+
+# AssetIcon
+
+The AssetIcon component is a visual representation of an asset type such as an archive or a video.
```
burakukula, line 5: @denkristoffer What do you think if we mention where this component is used, by which components, and that it is a subcomponent? It is a subcomponent, right?

packages/forma-36-react-components/src/components/Asset/README.mdx
```diff
@@ -0,0 +1,144 @@
...
+  - [Plaintext](#plaintext)
+  - [Presentation](#presentation)
+  - [Rich text](#rich-text)
+  - [Spreadsheet](#spreadsheet)
+  - [Video](#video)
+- [Usage](#usage)
+  - [An image asset and an archived image asset](#an-image-asset-and-an-archived-image-asset)
+- [An archive asset](#an-archive-asset)
+
+## Variations
```
burakukula, line 27: I would just remove it if we don't have any variations, just for visual reasons.

packages/forma-36-react-components/src/components/Asset/AssetIcon/AssetIcon.stories.tsx
denkristoffer: Actually it's not, I still see the old Storybook docs page. Maybe we can loop @gui-santos in here, any ideas why these docs aren't being displayed?

packages/forma-36-react-components/src/components/Asset/README.mdx
```diff
@@ -0,0 +1,144 @@
...
+<Asset
+  src="https://placekitten.com/200/300"
+  title="An archive asset"
+  type="archive"
+/>;
+```
+
+### An image asset and an archived image asset
+
+Image assets are rendered directly unless they have the prop `status="archived"`.
```
burakukula, line 123: @denkristoffer I find it a bit unclear. It would find it a bit more useful if it would more descriptive. Something around the fact that if the status is archived instead of the image the asset icon would be displayed. What do you think about it?

packages/forma-36-react-components/src/components/Asset/README.mdx
```diff
@@ -0,0 +1,144 @@
+import { Asset } from './Asset';
+
+# Asset
+
+The Asset component is a visual representation of an asset file such as a video or a PDF document.
```
burakukula, line 5: Maybe we could come up with some examples where this component can be used? Or we consider it as a subcomponent only - if so, maybe it's worth mentioning which component uses it?

packages/forma-36-react-components/src/components/CopyButton/README.mdx
```diff
@@ -0,0 +1,34 @@
+import { CopyButton } from './CopyButton';
+
+# CopyButton
+
+The CopyButton is a styled button that copies text into the user's clipboard.
```
burakukula, line 5: @denkristoffer I'm not sure how we use this component right now, but maybe it's worth mentioning how it might be used?

packages/forma-36-react-components/src/components/Asset/AssetIcon/AssetIcon.stories.tsx
burakukula: @denkristoffer yeah, we talked about it with @gui-santos. I tried to fix it, but I failed for now. We decided that it's not the most important thing right now. There is a small hack, that Gui showed me, done in the `preview.js` in the storybook config, where we potentially could fix it (hack it more 😀).

packages/forma-36-react-components/src/components/EditorToolbar/EditorToolbar/README.mdx
```diff
@@ -0,0 +1,57 @@
+import { Button } from '../../Button';
+import { EditorToolbar } from './EditorToolbar';
+import { EditorToolbarButton } from '../EditorToolbarButton';
+
+# EditorToolbar
+
+The EditorToolbar is a toolbar component for rich text editors. It provides a styled shell that toolbar actions such as buttons can be wrapped in.
+
+<EditorToolbar style={{ justifyContent: 'space-between' }}>
```
burakukula, line 9: This part should be displayed below ## Example of usage section?

packages/forma-36-react-components/src/components/Asset/AssetIcon/README.mdx
```diff
@@ -0,0 +1,79 @@
...
+
+### Spreadsheet
+
+<AssetIcon type="spreadsheet" />
+
+### Video
+
+<AssetIcon type="video" />
+
+## Usage
```
burakukula, line 73: We could call this `Example of usage`, just to keep it consistent with other README's. WDYT?

packages/forma-36-react-components/src/components/Asset/README.mdx
```diff
@@ -0,0 +1,144 @@
...
+
+<Asset title="Archive of files" type="archive" />
+
+## Table of contents
```
burakukula, line 9: I will add `Table of contents` to other README's if we want to have it (I think it's a good idea). And I will update the template as well.

packages/forma-36-react-components/src/components/EditorToolbar/EditorToolbarButton/README.mdx
```diff
@@ -0,0 +1,62 @@
+import { Button } from '../../Button';
+import { EditorToolbar } from '../EditorToolbar';
+import { EditorToolbarButton } from './EditorToolbarButton';
+
+# EditorToolbarButton
+
+The EditorToolbarButton is a styled button for actions that should be shown on a rich text editor toolbar. It composes the IconButton component, which means it can show any icon that the IconButton can.
+
+<EditorToolbar>
```
burakukula, line 9: @denkristoffer same here - I would move it below the Examples of usage (and adjust the title of the section)

burakukula:
@denkristoffer amazing work! 🎉 Just a couple of comments from my side regarding README files. 
1. I guess for subcomponents it's worth mentioning that we treat them like so and maybe include info about which components use those? 
2. Maybe it would be possible to give some examples where components might be used, which are our recommendations, some use cases if the component is not a subcomponent from our perspective.
3. I would recommend naming sections consistently across all README files. (I know that probably existing ones probably needs reviewing, I will do it and rise PR)

I will add the `Table of contents` section into the template and to existing README files. It's a very good point to have it.

packages/forma-36-react-components/src/components/EditorToolbar/EditorToolbar/README.mdx
denkristoffer: Do we want to take advantage of MDX and render the component? That was my thinking here, and then show actual code in examples of usage.

packages/forma-36-react-components/src/components/Asset/AssetIcon/AssetIcon.stories.tsx
gui-santos: ☝️ exactly

more on how the DocsPage work [here](https://storybook.js.org/docs/react/writing-docs/docs-page) <- we can change how it renders with a bit of work

and what we were doing with the `README.md` were just what storybook was recommending for the migration to v6 for people using addon-info/notes. Because addon-info/notes are binge deprecated and now it's addon-docs (comes in v6 with addon-essentials) that does that job

more on the deprecation of the mentioned addon here:
https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#60-deprecations

denkristoffer:
I deleted the AssetIcon README and story. I'm thinking there's no need to document subcomponents that are only meant to be used by another component. What do you say?

PR #654, opened 2020-11-23
Feat: change Spinner animation

gui-santos:
# Purpose of PR

Hey! @Gracegross has this proposal for updating the loading message at Contentful’s webapp and it will need a change in the animation of the Spinner for these three dots that were animated by @domarku 🙌 
new_spinner_1

this is how it looks in the loading state of a button
button

This PR makes the Spinner a functional component, adds the SVG and CSS for the animation, and 'primary' as an option for the color

Wdyt?

## PR Checklist

- [x] Tests are passing


denkristoffer (commented):
Wow looks great! Nice work @domarku and @gui-santos 😄 It's much cooler with a custom animation like this. One thought, can we check how this works in all the places the old spinner is used? Is it only buttons, or are there other special cases?

m10l:
Nice animation. I have a small concern that it's too similar to the context menu trigger though. I don't think it'd be mistaken for a context menu, but there's something about having two visually similar things used for two very different meanings which feels a bit odd to me.

Gracegross:
> context menu trigger

@m10l Do you mean the button loading state? I see your point. Given the loading state is only shown for ~1second maybe it's not enough time to get confused? I mean, my brain doesn't work that fast anyway.

Gracegross (approved):
Love it! This is going to make the web app much more fun! Can't wait. 😆

Danke @domarku & @gui-santos 👏

packages/forma-36-react-components/src/components/Spinner/README.mdx
```diff
@@ -0,0 +1,21 @@
+# Spinner
+
+A Spinner is a graphic element that helps communicating to users that an action is in progress and it will take some time to be completed.
+
+```jsx
+<Spinner />
+```
+
+## Examples of usage
```
burakukula, line 9: @gui-santos maybe some examples of usage could be added here?

gui-santos:
> ... One thought, can we check how this works in all the places the old spinner is used? Is it only buttons, or are there other special cases?

@denkristoffer in F36, I only found the button loading state. In Contentful’s web app there are more uses, but they generally are similar to this:

new_spinner_1
(Spinner on top of some text)

or this...
loading
(some text followed by the spinner. I saw a few cases where the spinner is before the text, but I'm changing them in the webapp)

and of course, there are cases where it shows only the spinner in some empty space. I think the button is the only place I saw it inside another component

👌

burakukula (approved):
Looks good! 🎉

burakukula:
Oh! @gui-santos it is actually not a spinner any more? 🤔

packages/forma-36-react-components/.size-limit
```diff
@@ -12,7 +12,7 @@
     {
       "path": "dist/styles.css",
       "name": "gzipped css",
-      "limit": "75 KB",
+      "limit": "85 KB",
```
mshaaban0, line 15: Is the animation actually increasing the library 10kb gzipped?

mshaaban0 (approved):
🤩

packages/forma-36-react-components/.size-limit
gui-santos: It's increasing in 5.something KB

I can make the limit `80.5 KB`

gui-santos: Is it still too much 5kb? What can we do to make it lighter? 🤔

denkristoffer: It must compress very poorly. I'm not sure why.

denkristoffer:
Any chance we can get the gzipped size down? I'm happy to pair on it if you want to, although I don't really know where to start 😅

domarku:
cool! hey @gui-santos are we sure the animation is going to perform the same all the time (like will it load good on slow internet) and could we tweak the speed of the animation?

I think it should be a bit faster, feels better when looking at it

comparing the SVG one above with the CSS original below:

Nov-30-2020 13-08-39

gui-santos:
@domarku 

> cool! hey @gui-santos are we sure the animation is going to perform the same all the time (like will it load good on slow internet) and could we tweak the speed of the animation?

We can't be sure. But I think that's a UI issue in general right?
Different browsers with different engines, different computers, and different internet connections make the experience slower or faster. With lags or no lags
Even if we make this a gif, we can't guarantee that all browsers on all computers will render the same thing. There are too many variables out of our control
But I think once HTML and CSS are fully loaded the majority of people using the browsers we support with average connection will see the three dots jumping

> I think it should be a bit faster, feels better when looking at it
> 
> comparing the SVG one above with the CSS original below:

We can make it faster for sure! I was copying the animation just by looking at it and I may have made it slower
I will increase the speed and share a gif here with you

gui-santos:
spinner_fast

@domarku I set the animation duration to 1s, wdyt?

(my bad I was setting it to 1.5s before)

domarku:
oh, cool. 1s looks better 👍 

> But I think once HTML and CSS are fully loaded the majority of people using the browsers we support with average connection will see the three dots jumping

you're right, it's a bit too unpredictable. I guess it's more about reducing it's size.

domarku (approved):

gui-santos:
> Any chance we can get the gzipped size down? I'm happy to help or pair on it if you want to, although I don't really know where to start 😅

@denkristoffer can we work on this in another PR? Since we both don't know what to do

PR #660, opened 2020-11-25
feat: refactor table components to functional components [PUL-1298]

mshaaban0:
# Purpose of PR

Refactor table components to functional components, and extend prop types properly. Part of #651

Although we are spreading `{...otherProps}` but ts doesn't really recognise any props the consumer passes that doesn't exist in the props definition. with extending the actual html type it recognises `otherProps` and tight it to the supported attributes

e.g. 
```tsx
export interface TableRowProps extends HTMLProps<HTMLTableRowElement> {
  className?: string;
  testId?: string;
  children: React.ReactNode;
}

// TS works, and recognises onClick
<TableRow onClick={...} /> 


export interface TableRowProps {
  className?: string;
  style?: React.CSSProperties;
  testId?: string;
  children: React.ReactNode;
}
// Won't work
<TableRow onClick={...} /> 
```

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Table/Table.tsx
+++ b/packages/forma-36-react-components/src/components/Table/Table.tsx
@@ -1,11 +1,10 @@
-import React, { Component } from 'react';
+import React, { HTMLProps } from 'react';
 import cn from 'classnames';
 import styles from './Table.css';
 
-export interface TableProps {
+export interface TableProps extends HTMLProps<HTMLTableElement> {
   testId?: string;
   className?: string;
-  style?: React.CSSProperties;
   children: React.ReactNode;
 }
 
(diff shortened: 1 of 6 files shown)


denkristoffer (approved):
⚒️

burakukula (approved):
Looks good! 🥇

packages/forma-36-react-components/src/components/Table/Table.tsx
```diff
@@ -1,36 +1,33 @@
...
 }
 
 const defaultProps: Partial<TableProps> = {
   testId: 'cf-ui-table',
 };
 
-export class Table extends Component<TableProps> {
-  static defaultProps = defaultProps;
+export const Table = (props: TableProps) => {
+  const { className, children, testId, ...otherProps } = props;
```
gui-santos, line 16: Wdyt about putting the default props here during this refactor?

gui-santos (approved):
Amazing 🙌

packages/forma-36-react-components/src/components/Table/Table.tsx
mshaaban0: tbh, I don't know the difference, all form controls refactor uses `defaultProps` I'm happy to change if we want it always to be defined inline

gui-santos: One reason to make them inline is that we won't need to do `Partial<TableProps>` for the `defaultProps` anymore

More info here https://github.com/contentful/forma-36/pull/647#discussion_r526932324

gui-santos: It's not urgent tough!

PR #661, opened 2020-11-26
Proposal - separate icons package

burakukula:
Hey!

So just a draft for now - some parts are still missing:
 - all the icons would need to end up in this package
 - product icons also
 - docs around icons
 - improvements in storybook

But for now I would like to get your opinion around the topic.
At the beginning I thought that maybe this package should be framework agnostic - I think this is what originally @mshaaban0 had in mind in his proposal. 

```
import { search } from '@contentful/forma-36-icons`
import { Icon } from '@contentful/forma-36-react-components`

<Icon color="primary" size="m" icon={search} />
```

But then, I thought that maybe the developer experience would be better if we would have just one import while using the package. And after all, we support only react when it comes to components, so I figured I will try to do react icons package. 

So now the fun part - I'm trying to figure it out 😁 What I would like to support are 2 things:

`import {Icon} from '@contentful/forma-36-icons'` -> **this imports all icons**
`import {ArrowUp} from '@contentful/forma-36-icons/ArrowUp'` -> **this would import only one icon.**

I managed to build the first way and it seems like it's working. I am having trouble to figure out the way to build the second way. 
If we would support the import of single icon, it would be necessary to create some sort of icon layer to manage size and color - I think I have some idea how to do it. But the thing I'm struggling with is the building part. I think rollup supports building separate modules, but I can not make it work with tsdx. I got to the point that I'm thinking about dropping tsdx...

But maybe it's possible to make it work somehow. Do you have any ideas? Thx for your help!!!


denkristoffer:
> But then, I thought that maybe the developer experience would be better if we would have just one import while using the package. And after all, we support only react when it comes to components, so I figured I will try to do react icons package.

SVGR and similar can allow us to do something like this, which is perhaps worth considering as an in between?

```jsx
import svgUrl, { ReactComponent } from '@contentful/forma-36-icons'
```


> But maybe it's possible to make it work somehow. Do you have any ideas? Thx for your help!!!

I think *maybe* the rollup alias plugin is something that could help us here, just guessing: https://github.com/rollup/plugins/tree/master/packages/alias#readme

Also, I'm of the initial thought that treeshaking with named exports is good enough here. Do you expect we have a lot of consumers of the library who don't use either Webpack, Rollup, or Parcel? I'm not yet convinced that it's worth having and supporting a custom build setup to achieve this 🙂

gui-santos:
> `import {Icon} from '@contentful/forma-36-icons'` -> **this imports all icons**
> `import {ArrowUp} from '@contentful/forma-36-icons/ArrowUp'` -> **this would import only one icon.**

Have you checked [this](https://www.robinwieruch.de/webpack-code-splitting-library)? the guy is using Webpack to code split his library, Idk how that would look with tsdx

If we achieve that would be amazing because people keep complaining about the bundle size of forma36 so splitting the icons would be awesome

Also, I have a question. I know this is a draft, but I saw some eslint and tsconfig files, do we need them? I thought we unified all of that with tsdx

denkristoffer:
My understanding is that tsdx should let us be able to do `import { ArrowUp } from '@contentful/forma-36-icons'` and if the consumer then uses tree shaking, it should only import that one icon.

burakukula:
> > `import {Icon} from '@contentful/forma-36-icons'` -> **this imports all icons**
> > `import {ArrowUp} from '@contentful/forma-36-icons/ArrowUp'` -> **this would import only one icon.**
> 
> Have you checked [this](https://www.robinwieruch.de/webpack-code-splitting-library)? the guy is using Webpack to code split his library, Idk how that would look with tsdx
> 
> If we achieve that would be amazing because people keep complaining about the bundle size of forma36 so splitting the icons would be awesome
> 
> Also, I have a question. I know this is a draft, but I saw some eslint and tsconfig files, do we need them? I thought we unified all of that with tsdx

Thx @gui-santos ! I didn't try this approach. I will try to set up this with webpack. If I make it work, maybe someone else can do it using tsdx. I think maybe with webpack would be easier for me to try it out - even though I am not an webpack expert at all. But I have even less experience with tsdx.

denkristoffer:
@burakukula I think having both is the best but I don't think it's a must. It's better because it gives consumers choice, but I don't think it's worth it for us to spend days of effort on. I think 99% of consumers of a React library already have a build step set up with Webpack/Rollup/Parcel. My own experience with React libraries are that a lot of them don't support absolute paths either. But if we can get it working we should.

burakukula:
Closing as we have the Icons package in v4 already solved.

PR #664, opened 2020-11-26
fix(text input): remove webkit search styling

m10l:
# Purpose of PR

This PR removes some webkit specific styling that was showing in Safari

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/TextInput/TextInput.css
+++ b/packages/forma-36-react-components/src/components/TextInput/TextInput.css
@@ -28,6 +28,13 @@
     color: var(--color-ice-dark);
   }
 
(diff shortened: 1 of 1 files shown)


burakukula (approved):

denkristoffer (approved):
I don't know if this is still the case after Edge changed motor but they used to have their own input elements as well that we might want to make sure aren't a problem, things like `input::-ms-clear`.

m10l:
@denkristoffer Looks like this is no longer needed since version 79 of Edge https://developer.mozilla.org/en-US/docs/Archive/Web/CSS/::-ms-clear - I'll go ahead and merge, thanks for raising it though

PR #667, opened 2020-11-26
fix(tooltip): the isVisible prop is not respected

Miretxu:
# Purpose of PR
This PR fixes the issue where `isVisible` is not respected. It is being overwritten to `false` always by the default false `isHoveringTarget` and `isHoveringContent`.

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
+++ b/packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
@@ -99,7 +99,7 @@ export const Tooltip = ({
   testId,
   ...otherProps
 }: TooltipProps) => {
-  const [show, setShow] = useState(isVisible);
+  const [show, setShow] = useState(false);
   const [arrowPosition, setArrowPosition] = useState<ArrowPositionState>(
     getArrowPosition('bottom'),
   );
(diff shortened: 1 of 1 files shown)


packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
```diff
@@ -150,6 +150,10 @@ export const Tooltip = ({
     }
   }, [closeOnMouseLeave, isHoveringTarget, isHoveringContent]);
 
+  useEffect(() => {
```
Miretxu, line 153: Run this once to set the initial visibility of the Tooltip to True if applicable

gui-santos: Did you try to make the default value of isVisible to be true?

It looks to me that this effect should be doing the same as setting the initial state to true, but because isVisible does not have a value than it does not work, but idk tbh

gui-santos (approved):

packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
Miretxu: @gui-santos the initial state `const [show, setShow] = useState(isVisible);` alone doesn't work because `show` state is set again [here](https://github.com/contentful/forma-36/blob/42f6e513330980afa8fc1dc8ac7c6fe542a641bd/packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx#L147) and [here](https://github.com/contentful/forma-36/blob/42f6e513330980afa8fc1dc8ac7c6fe542a641bd/packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx#L149) to `false`. 
So I added another useEffect() block here, just to set the initial `show` state to the `isVisiable` value

gui-santos: I see! Thanks for explaining it ✌️

Issue #663, opened 2020-11-26
🐛 Bug - `v3.63.3` breaks Button and Select styles


gregoralbrecht:
# Forma 36 bug report

## Summary

You are probably already aware of it, but `@contentful/forma-36-react-components@3.63.3` breaks some styles for me. I didn't do anything differently with `v3.63.2` and it works.

I started working on a UI extension yesterday which looks like this:

```jsx
import {
  Button,
  SelectField,
  Option,
} from '@contentful/forma-36-react-components'

const Sidebar = () => {
  return (
    <div className={styles.container}>
      <div className={styles.select}>
        <SelectField
          id="selectEnvironment"
          name="selectEnvironment"
          labelText="Select environment"
          onChange={event => setEnv(event.target.value)}
          width="full"
          willBlurOnEsc
        >
          <Option value="next">Next</Option>
          <Option value="production">Production</Option>
        </SelectField>
      </div>
      <Button
        buttonType="primary"
        isFullWidth
        onClick={() => console.info(`Deploying to ${env}...`)}
      >
        Deploy content
      </Button>
    </div>
  )
}
```

After running `yarn install` again today, I saw broken styles:

## `v3.63.3`
![image](https://user-images.githubusercontent.com/6179211/100341424-d5ba5080-2fdc-11eb-8b37-1d1f10c13d50.png)

## `v3.63.2`
![image](https://user-images.githubusercontent.com/6179211/100341508-ee2a6b00-2fdc-11eb-9a55-7d0f74143a19.png)

denkristoffer:
Hi @gregoralbrecht, thanks for reporting this. Just to be sure, are you importing both the styles and tokens in the project?

gregoralbrecht:
@denkristoffer Yes, I imported them in the root component. The only thing that changed between these two screenshots was that I ran `yarn add @contentful/forma-36-react-components@3.63.2` which fixed the missing styles. It's weird, though, because the `<Tabs>`, `<Tab>`, and `<TabPanel>` still worked.

denkristoffer:
[3.63.3](https://github.com/contentful/forma-36/releases/tag/%40contentful%2Fforma-36-react-components%403.63.3) doesn't contain any changes that should affect this, so my initial thought is that there's something else going on as well, perhaps an update of some underlying dependency. Would it be possible for you to share the two lock files before and after updating?

gregoralbrecht:
[...]

PR #668, opened 2020-11-27
fix: proper styles.css build

suevalov:
# Purpose of PR

Fix missing CSS variables

Fixes https://github.com/contentful/forma-36/issues/663

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/tools/postcss.config.js
+++ b/packages/forma-36-react-components/tools/postcss.config.js
@@ -11,7 +11,7 @@ module.exports = {
     postCssUrl,
     postcssPresetEnv({
       stage: 0, // Allow experimental CSS features
-      browsers: 'last 2 versions, ie >= 11',
+      browsers: ['Chrome >= 75', 'Edge >= 17', 'Firefox >= 70', 'Safari >= 12'],
       preserve: false, // Remove the pre-polyfilled CSS
     }),
   ],
--- a/packages/forma-36-react-components/tsdx.config.js
+++ b/packages/forma-36-react-components/tsdx.config.js
@@ -13,6 +13,7 @@ module.exports = {
         config: {
           path: path.resolve(__dirname, './tools/postcss.config.js'),
         },
+        minimize: true,
         modules: true,
         extract: 'styles.css',
       }),


packages/forma-36-react-components/tools/postcss.config.js
```diff
@@ -11,8 +11,11 @@ module.exports = {
     postCssUrl,
     postcssPresetEnv({
       stage: 0, // Allow experimental CSS features
-      browsers: 'last 2 versions, ie >= 11',
+      browsers: ['Chrome >= 75', 'Edge >= 17', 'Firefox >= 70', 'Safari >= 12'],
       preserve: false, // Remove the pre-polyfilled CSS
     }),
+    // require('cssnano')({
```
m10l, line 17: I guess we can remove this?

packages/forma-36-react-components/package.json
```diff
@@ -93,6 +93,7 @@
     "postcss-import": "^12.0.1",
     "postcss-preset-env": "^6.7.0",
     "postcss-url": "^8.0.0",
+    "cssnano": "^4.1.10",
```
m10l, line 96: And we can remove this?

m10l (approved):
@suevalov Whilst looking into this issue I noticed that if I remove uses of `compose` (like `composes: focus-outline--default from './../../styles/settings/a11y.css';`) the CSS is built correctly (variables are removed from the built and replaced with the hardcoded values instead as expected).

Happy to accept that PR just to get shit moving again, but it might not really be solving the root cause.

packages/forma-36-react-components/tsdx.config.js
```diff
@@ -13,6 +13,7 @@ module.exports = {
         config: {
           path: path.resolve(__dirname, './tools/postcss.config.js'),
         },
+        minimize: true,
```
m10l, line 16: 🙇

denkristoffer:
Would be cool if we could synchronise browser targets automatically in the future. Is that even possible? 😄

suevalov:
@denkristoffer https://github.com/browserslist/browserslist#shareable-configs

Feel free to create one and rely on it in our repos

PR #671, opened 2020-11-30
fix: checkbox field docs

burakukula:
Hello,

Just some docs for CheckboxField component.

One note: We are exposing the Checkbox component itself. It just naked checkbox. I'm not sure about the usecases here. Maybe it could be just a subcomponent? WDYT?

## PR Checklist

- [ ] Tests are passing


packages/forma-36-react-components/src/components/CheckboxField/README.mdx
```diff
@@ -0,0 +1,57 @@
...
+  helpText="Helper text displayed below"
+  labelIsLight
+/>
+```
+
+## Content recommendations
+
+ - Start every option from a capital letter
+ - Do not use commas at the end of each line
+ - If the checkbox is used to agree to terms and conditions, we recommend to use first person in the sentence
```
burakukula, line 40: @restiles what do you think about those recommendations? Does it make sense from your perspective?

packages/forma-36-react-components/src/components/CheckboxField/README.mdx
```diff
@@ -0,0 +1,57 @@
+# CheckboxField
+
+CheckboxField is a component that gives an possiblity to easy setup proper checkbox with it's label. Component contains also an options that allows to set helper text displayed below the label. You might see checkboxes in forms, they might be also used for filtering data on the page or terms and conditions. Everywhere where user have multiple choice you might want to use this component.
```
denkristoffer, line 3: typo: 

>it's

"its"

denkristoffer: > contains also 

"also contains"

denkristoffer: > allows to set
"allows setting"

burakukula: @denkristoffer thx! good catch. Fixed 🙇‍♀️

packages/forma-36-react-components/src/components/CheckboxField/README.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+<CheckboxField
+  checked={}
+  onChange={hadnleOnChange}
+  name="checkbox name"
+  id="randomId"
+  labelText="Label text"
+  helpText="Helper text displayed below"
+/>
+```
```
gui-santos, line 18: Shouldn't we start every doc with the simplest way of using the component? Is this the simplest way? I mean, do I need to pass all these props to see a checkbox on my screen?

packages/forma-36-react-components/src/components/CheckboxField/README.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+Light label example
```
gui-santos, line 20: What is a light label?

packages/forma-36-react-components/src/components/CheckboxField/README.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+## Content recommendations
+
+ - Start every option from a capital letter
+ - Do not use commas at the end of each line
+ - If the checkbox is used to agree to terms and conditions, we recommend to use first person in the sentence
+
+## Best practices
+
+ - Checkboxes should work independently, checking one should not affect any other checkbox in the interface.
```
gui-santos, line 44: I suggest linking the Radio component here since it's the proper component for lists where the user can choose only one option

gui-santos (changes requested):
Nice! Just some small suggestions :)

packages/forma-36-react-components/src/components/CheckboxField/README.mdx
burakukula: @gui-santos yeah, sounds good, but I think we don't have a README for radio component yet - which probably would be ideal. Maybe I will update as soon as we create docs for it?

packages/forma-36-react-components/src/components/CheckboxField/README.mdx
restiles: @burakukula These make total sense to me. :)

packages/forma-36-react-components/src/components/CheckboxField/README.mdx
burakukula: I mention it in the Best practices and at the beginning - you can specify that the label has regular font weight. I will add this explanation over here as well.

packages/forma-36-react-components/src/components/CheckboxField/README.mdx
burakukula: yeah, make sense. On my way. Thx!

packages/forma-36-react-components/src/components/CheckboxField/README.mdx
gui-santos: your call

burakukula: Done. I linked to the tsx for now. I will change it as soon as we have docs for RadioButton. ✅

packages/forma-36-react-components/src/components/CheckboxField/README.mdx
burakukula: @restiles thank you!

packages/forma-36-react-components/src/components/CheckboxField/README.mdx
burakukula: Fixed, thx for feedback 🙇‍♀️

burakukula:
@gui-santos I this it's ready, if you can have a look again. thx!

gui-santos (approved):

packages/forma-36-react-components/src/components/CheckboxField/README.mdx
```diff
@@ -0,0 +1,64 @@
+# CheckboxField
+
+CheckboxField is a component that gives an possiblity to easy setup proper checkbox with its label. Component also contains an options that allows setting helper text displayed below the label. You might see checkboxes in forms, they might be also used for filtering data on the page or terms and conditions. Everywhere where user have multiple choice you might want to use this component.
```
domarku, line 3: @burakukula What do you think about writing the intro paragraph in the form of **when to use** the component?
(instead of describing the component parts and where users might find it)

So, something like:

_Use a CheckboxField to give users a way to select items from a list, when they could select zero, one or more items._

domarku: - without mentioning variants (helper text in this case)

domarku: - focusing on **when** to use the component, not where, I would omit the part with forms, but maybe keep "e.g. when users need to agree to Terms and conditions."

burakukula: @domarku I made some changes, I still think forms are a valid example of usage. Please have a look if this is close to what you have in mind?

packages/forma-36-react-components/src/components/CheckboxField/README.mdx
```diff
@@ -0,0 +1,64 @@
+# CheckboxField
+
+CheckboxField is containing a checkbox with its label. Use this component to give users the possibility to select multiple elements, zero, one or more items. It can be used in forms or terms and conditions. For a single choice, please consider using RadioButton component.
```
domarku, line 3: I'm still in favour of this order:

1. "when to use"
2. what it consists of
3. (optional) when to not use it, but rather use the radio button

burakukula: @domarku oki! 😁 WDYT if we do this:
"Use this component to give users the possibility to select multiple elements, zero, one or more items. It can be used in forms or terms and conditions. CheckboxField is containing a checkbox with its label. For a single choice, please consider using RadioButton component."

domarku (approved):

PR #674, opened 2020-11-30
fix(component): Add delay property to Toolbox component

0math:
# Purpose of PR

This adds `delay` property to `Toolbox` component.

- Previous delay of 1000ms was too long
- We think it’d be beneficial to be able to set custom delay via props, separately for `show` and `hide` (so that there will be no request just to change the delay time etc)
 - We aren’t sure if `show` delay is useful as usually we want to show the tooltip right away. But lots of library like https://material.angular.io/components/tooltip/overview and bootstrap has `show` and `hide` delay as props. For the sake of completeness, we can have it. Otherwise, just keep the delay for `hide`.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
+++ b/packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
@@ -47,6 +47,10 @@ export interface TooltipProps {
    */
   maxWidth?: number | CSS.Property.MaxWidth;
   /**
+   * It sets a delay period for the Tooltip
+   */
+  delay?: Partial<DelayProps>;
+  /**
    * Function that will be called when target gets blurred
    */
   onBlur?: (evt: FocusEvent) => void;
@@ -80,6 +84,10 @@ interface ArrowPositionState {
   top: string;
   left: string;
 }
+export interface DelayProps {
+  hide: number;
+  show: number;
+}
 
 export const Tooltip = ({
   children,
@@ -90,6 +98,7 @@ export const Tooltip = ({
   isVisible,
   closeOnMouseLeave,
   maxWidth,
+  delay,
   onBlur,
   onFocus,
   onMouseLeave,
@@ -100,6 +109,7 @@ export const Tooltip = ({
   ...otherProps
 }: TooltipProps) => {
   const [show, setShow] = useState(false);
+  const [delaySettings, setDelaySettings] = useState(delay);
   const [arrowPosition, setArrowPosition] = useState<ArrowPositionState>(
     getArrowPosition('bottom'),
   );
@@ -153,9 +163,15 @@ export const Tooltip = ({
   useEffect(() => {
     if (isVisible) setShow(true);
     // eslint-disable-next-line react-hooks/exhaustive-deps
+    if (!delay) {
+      if (closeOnMouseLeave) {
+        setDelaySettings({ hide: 0, show: 0 });
+      } else {
+        setDelaySettings({ hide: 500, show: 0 });
+      }
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
-
-  const delay = closeOnMouseLeave ? 0 : 1000;
 
   const arrowStyles = {
     ...popperStyles.arrow,
@@ -186,11 +202,17 @@ export const Tooltip = ({
         ref={elementRef}
         className={targetWrapperClassName}
         onMouseEnter={(evt: MouseEvent) => {
-          setIsHoveringTarget(true);
+          setTimeout(
+            () => setIsHoveringTarget(true),
+            delaySettings && delaySettings.show,
+          );
           if (onMouseOver) onMouseOver(evt);
         }}
         onMouseLeave={(evt: MouseEvent) => {
-          setTimeout(() => setIsHoveringTarget(false), delay);
+          setTimeout(
+            () => setIsHoveringTarget(false),
+            delaySettings && delaySettings.hide,
+          );
           if (onMouseLeave) onMouseLeave(evt);
         }}
         onFocus={(evt: FocusEvent) => {
@@ -198,7 +220,10 @@ export const Tooltip = ({
           if (onFocus) onFocus(evt);
         }}
         onBlur={(evt: FocusEvent) => {
-          setTimeout(() => setIsHoveringTarget(false), delay);
+          setTimeout(
+            () => setIsHoveringTarget(false),
+            delaySettings && delaySettings.hide,
+          );
           if (onBlur) onBlur(evt);
         }}
         {...otherProps}
--- a/packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
+++ b/packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
@@ -39,17 +39,13 @@ export interface TooltipProps {
    */
   isVisible?: boolean;
   /**
-   * It controls whether or not the tooltip stays open when hovering on the content
-   */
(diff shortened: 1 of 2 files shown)


0math:
Hey @gui-santos, may I kindly ask you to look at this at let me know what do you think (as you're the author of the component)? We're aiming to fix an issue on the usage page with this PR. Thank you!

packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
```diff
@@ -46,6 +46,10 @@ export interface TooltipProps {
    * It sets a max-width for the Tooltip
    */
   maxWidth?: number | CSS.Property.MaxWidth;
+  /**
+   * It sets a delay period for the Tooltip
+   */
+  delay?: Partial<DelayProps>;
```
gui-santos, line 52: IMO, having one prop for each delay is a better API than passing an object to a prop. For two reasons:

1. Comparing object is a bit weird
2. Easier to use if I only want to set one of the delays
`<Tooltip showDelay={100} />` 
`<Tooltip hideDelay={100} />` 
`<Tooltip showDelay={100} hideDelay={100} />`

packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
```diff
@@ -186,19 +202,28 @@ export const Tooltip = ({
         ref={elementRef}
         className={targetWrapperClassName}
         onMouseEnter={(evt: MouseEvent) => {
-          setIsHoveringTarget(true);
+          setTimeout(
+            () => setIsHoveringTarget(true),
+            delaySettings && delaySettings.show,
+          );
```
gui-santos, line 208: Separating the delay props will help here. This will become:

```jsx
setTimeout(
  () => {},
  showDelay
)
```

Considering that we set the default value for the delays to 0 👌

packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
```diff
@@ -100,6 +109,7 @@ export const Tooltip = ({
   ...otherProps
 }: TooltipProps) => {
   const [show, setShow] = useState(false);
+  const [delaySettings, setDelaySettings] = useState(delay);
```
gui-santos, line 112: Why do we need to save this in a state?
Do we want to support delays being changed on the fly? Even if we want that, do we need to save it in a state

packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
```diff
@@ -153,10 +163,16 @@ export const Tooltip = ({
...
     if (isVisible) setShow(true);
     // eslint-disable-next-line react-hooks/exhaustive-deps
+    if (!delay) {
+      if (closeOnMouseLeave) {
+        setDelaySettings({ hide: 0, show: 0 });
+      } else {
+        setDelaySettings({ hide: 500, show: 0 });
+      }
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
gui-santos, line 173: I didn't understand why we need to do this, sorry :(

Also, there is two eslint-disable in this useEffect. Do we need two?

gui-santos (changes requested):
Hey! Thanks for the contribution

I think I understood the use case, but I have some suggestions to improve the API of the component and some doubts about how the delay was implemented

packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
0math: You're right, we could've avoid using states from the beginning. Just refactored it without states.

0math:
Hey @gui-santos, thanks for the feedback! Addressed it, could you please have a look?

packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
```diff
@@ -180,6 +184,8 @@ export const Tooltip = ({
     );
   }
 
+  const delay = !hideDelay ? (closeOnMouseLeave ? 0 : 300) : hideDelay;
+
```
gui-santos, line 188: When @Miretxu introduced the behavior for a user to keep the tooltip open, we didn't think about allowing them to set different delays and that's why we have the `closeOnMouseLeave` prop

now that we have the custom delay functionality, it looks to me that we could deprecate that prop and simply pass hideDelay to the setTimeout like this: 
 
```jsx
function Tooltip (
  hideDelay = 0 // setting the default to 0 because the default behavior is not having any delay, right?
  // ...
)

// inside the return...
<ContainerElement
  onMouseLeave={(evt: MouseEvent) => {
    setTimeout(() => setIsHoveringTarget(false), hideDelay);
    if (onMouseLeave) onMouseLeave(evt);
   }
>
// rest of the jsx...
```

so a user that wants the tooltip to close immediately after not hovering the target would do:
`<Tooltip />`

and a user that wants the Tooltip to stay open would do this:
`<Tooltip hideDelay={300} />`

I think this is simpler than having all the logic mixing two props that are essentially doing the same. My question to you and @Miretxu now is: are you ok with deprecating `closeMouseOnLeave`? (I believe your team is the only ones using it right now)

gui-santos: Sorry for keep commenting on your PR, we are trying to keep components easy and simple to use. But sometimes solutions take time to show up 🙏 

If you want to pair on this ping me :)

gui-santos (changes requested):
Hey! This is looking great

While reviewing for the second time this PR, I had an epiphany hahahah so I left one more comment to you and @Miretxu 

Cheers 🙌

packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
gui-santos: On a BIG sidenote here

I suggested in the comment above to set default value for the prop in the argument of the function, but this component is still using the old way of setting default props in line 245 (Component.defaultProps = { bla bla bla })

The forma36 team decided to refactor this, so it would be a HUGE help if you use this PR to move all the defaults to the argument of the function, but not necessary

Miretxu: @gui-santos happy to deprecate it to make the component simpler and cleaner 👍
We can keep an example in the storybook to demonstrate how to use hideDelay to achieve the content hover-over.

If we remove `closeMouseOnLeave` props, we can set the hideDelay default to 0 in `Component.defaultProps = { }`.
 Previously, we had to calculate it either 0 or 300 depending on the `closeMouseOnLeave` value, and the prop value wasn't available in `Component.defaultProps = { }`, that's why we defined in the argument of the function. But we won't have that problem now 🙂

Miretxu: @gui-santos I wonder if we should the default `hideDelay` to 0 or 300? 300 will enable the content hover over but not sure if it will introduce "drags" to the UI for a small tooltip, like a copy-paste button.

0math: @gui-santos Thank you very much for your suggestions! Me and @Miretxu just addressed all the comments and refactored the component as recommended. Could you please have a look again?

gui-santos (approved):
WOW! This is great now @Miretxu and @0math 

Thanks for adding the example in storybook as well

Ship it

PR #702, opened 2020-12-03
feat: gatsby page consumes REAMDE files from react components

burakukula:
👋 F36!

We would like to have the possibility to display README files on our documentation website. In order to do it, we pull MDX files from GitHub. 

**How it would work:**
 - we would display README content for each component right after the top part of the page, but we are still able to display something else from the `index.mdx` of each component below.
- in order to not get strange errors about imports (the components are already there, maybe that's why it's complaining about it) we would need to comment out imports in examples in the README files
- examples with hooks should we wrapped in function

Before this PR, I would need to fix the REAMDE files and comment out imports (probably I will also adjust them to the new structure). I would like to get your feedback before I start working on it.

**In the next PR I will move existing docs from website to README files in react-components.**
## PR Checklist

- [ ] Tests are passing


burakukula:
I created the PR with adjustments necessary in existing README files - https://github.com/contentful/forma-36/pull/706. I would need to merge it first and then adjust this PR.

packages/forma-36-website/src/components/ComponentSource.js
```diff
@@ -37,7 +37,7 @@ class ComponentSource extends React.Component {
   render() {
     return (
       <div className="f36-margin-bottom--m">
-        <LiveProvider code={this.props.children} scope={{ ...f36Components, Grid, GridItem, Autocomplete }}>
+        <LiveProvider code={this.props.children.trim()} scope={{ ...f36Components, Grid, GridItem, Autocomplete, Flex, ProductIcon }}>
```
burakukula, line 40: Currently, when we have anything in the examples that is declared or imported, we are getting errors from `LiveProvider`. Does anyone have an idea on how to solve this issue?

packages/forma-36-website/src/components/ComponentSource.js
```diff
@@ -26,18 +34,30 @@ const styles = {
 };
 
 class ComponentSource extends React.Component {
-  state = { isOpen: false };
+  state = { isOpen: true };
```
burakukula, line 37: I think we could keep it open by default? I'm not insisting though.

packages/forma-36-website/src/components/Layout.js
```diff
@@ -43,67 +45,95 @@ const Layout = props => (
...
+      const arrFromSlug = item.node.childMdx.slug.split('/');
+      const trimmedTitle = props.pageContext.frontmatter.title.replace(
+        /\s/g,
+        '',
       );
-    }}
-  />
-);
+
+      if (arrFromSlug.indexOf(trimmedTitle) > -1) {
```
burakukula, line 79: Probably there is a more elegant way to do it?

denkristoffer:
Preview can be seen at https://deploy-preview-702--forma-36.netlify.app/

packages/forma-36-website/src/components/Layout.js
denkristoffer: `Array.includes` is probably more readable.

denkristoffer:
This is great, thank you so much for looking into this! A few questions:

This builds docs from `master` on deploy, so we need to trigger a website build with every commit to `react-components`, is this set up already?

And similarly, is pulling the docs from `master` necessary or is it possible to use the current branch instead, so that we can also preview docs changes on the website?

packages/forma-36-website/src/components/ComponentSource.js
```diff
@@ -3,7 +3,15 @@ import { css } from '@emotion/core';
...
-import { Grid, GridItem, Autocomplete } from '@contentful/forma-36-react-components/dist/alpha';
+import {
+  Grid,
+  GridItem,
+  Autocomplete,
+  Flex,
+  ProductIcon,
+  Accordion,
+  AccordionItem,
+} from '@contentful/forma-36-react-components/dist/alpha';
```
gui-santos, line 14: Wdyt about doing the same as the line above?

```js
import * as f36Components from '@contentful/forma-36-react-components';
import * as f36AlphaComponents from '@contentful/forma-36-react-components/dist/alpha';
```

gui-santos (commented):
This looks great!

Apart form my suggestion which is not a blocker and @denkristoffer comments I have only one question now, I checked the preview and apparently when the examples use hooks we got some errors in the Preview component. Are we addressing them now?

I think we could just remove hooks from the examples and address this later. Wdyt?

burakukula:
@gui-santos in order to make hooks work with the preview we would wrap them in `function` - you can see those changes in `Card` component for example. I resolved issues in the same PR (I hope all of them), that why they ar not reflected on the website deployed from this branch - README's are taken from master.

@denkristoffer it is possible to pull README for the current branch - I did that during development. I am not sure what we would like to do, it's not super comfortable to get it from the master branch while working on something, as it will not be visible. 
I didn't think about triggering a website build with every commit to react-components, I guess that's need to happen, you are right. Would you like to help me out with this?

m10l (approved):
Nice work!

A few issues I've found with documentation in general, some might be related but I imagine the majority are not:

https://deploy-preview-702--forma-36.netlify.app/components/autocomplete/
- ReferenceError: items is not defined

https://deploy-preview-702--forma-36.netlify.app/components/card/
- SyntaxError: Unexpected token (1:8) 1 : return (const [show, setShow] = useState(false); ^

https://deploy-preview-702--forma-36.netlify.app/components/dropdown/
- "Accessibility Missing" - maybe it's best not to have this section if it's empty

https://deploy-preview-702--forma-36.netlify.app/components/grid/
- SyntaxError: Adjacent JSX elements must be wrapped in an enclosing tag (6:0) 2 : // import { Grid, GridItem } from '@contenful/forma-36-react-components/dist/alpha'; 3 : 4 : <Grid columns={6}></Grid> 5 : // Or 6 : <Grid columns={'auto 1fr'}></Grid>) ^
- broken heading for "Code examples"

https://deploy-preview-702--forma-36.netlify.app/components/pill/
- SyntaxError: Unexpected token (3:0) 1 : return (// import { Pill } from '@contentful/forma-36-react-components'; 2 : 3 : const renderTags = (tags) => { ^

https://deploy-preview-702--forma-36.netlify.app/components/product-icon/
- ReferenceError: NavigationIcon is not defined

https://deploy-preview-702--forma-36.netlify.app/components/tabs/
- Example has a href going to Contentful for every tab, which means you can't interact with the tabs in the playground without being immediately taken away

https://deploy-preview-702--forma-36.netlify.app/components/checkbox/
- Has no information at all. Should have at least a rendering of the component along with the text other non-documented components have.

General
- Alpha documentation should use the Note component for displaying warnings - maybe something we can add programmatically via frontmatter as a "status" or something
- Scrollbar resets on page refresh
- Need to fix the link font size in the documentation
- Weird spacing in a lot of code examples (check Tooltip for example)

burakukula:
Hey @m10l thanks a lot for feedback - ReferenceErrors and syntax errors will be resolved as soon as we merge this PR - I solved issues on the same branch, but they are not reflected in the deployed version as README's are taken from master. 
The rest maybe I could address on a separate PR?

m10l:
@burakukula @denkristoffer Confused about this part - the whole point of using Lerna is that it handles these situations for us - when developing locally you should see local changes to components in the website, and likewise when we publish changes to the components, the website should pick up that this has been updated and bump the version of components being used for us which should trigger a site deploy

packages/forma-36-website/gatsby-config.js
```diff
@@ -296,6 +304,15 @@ module.exports = {
   plugins: [
     'gatsby-plugin-react-helmet',
     'gatsby-plugin-svgr',
+    {
+      resolve: `gatsby-source-git`,
+      options: {
+        name: `forma-36-react-components`,
+        remote: `https://github.com/contentful/forma-36.git`,
+        branch: `master`,
```
m10l, line 312: @burakukula Ah I see now, because we're using git as a source

denkristoffer:
> I didn't think about triggering a website build with every commit to react-components, I guess that's need to happen, you are right. Would you like to help me out with this?

Yes! I've set up Netlify to comment with a preview of the website for every PR, lets see if it's enough 😄

packages/forma-36-website/gatsby-config.js
burakukula: @m10l yeah, exactly :)

packages/forma-36-website/src/components/ComponentSource.js
burakukula: 🤦‍♀️✅

packages/forma-36-website/gatsby-config.js
gui-santos: Do we need to do this? Can't we use the code in the branch always using relative paths or something? (so when creating a PR the preview would use the current code of that branch)

burakukula: @gui-santos it's an option as well, but I just thought it's better to rely on GtiHub, where files are verified already. But I am open to change the approach. Maybe it would actually be more comfortable during development. Maybe I would do it on a separate branch? I would like to move files from website to README's first.

gui-santos: IMO, it would be better to be able to preview by change made in the repo right way. For example, you fixed the error messages Mike and I commented before but we can't see if they are really fixed since the changes are in this branch and not in master

But I guess we can do it in another branch of course (maybe the website will have some errors idk)

gui-santos (approved):

PR #721, opened 2020-12-18
fix: refactor Button storybook

burakukula:
Refactor Button to the new story format and add the overview of component variants for testing purposes

## PR Checklist

- [ ] Tests are passing


packages/forma-36-react-components/src/components/Button/Button.stories.tsx
```diff
@@ -1,47 +1,79 @@
...
+      small
+    </Button>
+    <Button buttonType="primary">default size</Button>
+    <Button buttonType="primary" size="large">
+      large
+    </Button>
+  </>
+);
+
+export const statesOverview = (args: ButtonProps) => (
```
burakukula, line 59: Or maybe we would prefer to have just one story with the overview of all the variants and states?

denkristoffer: Don't many stories work best with Chromatic?

burakukula: Maybe let's try to merge this one and I will create PR with some changes so we would be able to see it in Chromatic? WDYT?

m10l (changes requested):
I think splitting up the states to show more variants would be better and help us catch any issues

So something like this:
- isActive - primary, positive, warning, negative, muted, naked
- disabled - primary, positive, warning, negative, muted, naked
- indicateDropdown - primary, positive, warning, negative, muted, naked
- loading - primary, positive, warning, negative, muted, naked

packages/forma-36-react-components/src/components/Button/Button.stories.tsx
m10l: You can already see this in Chromatic here: https://www.chromatic.com/pullrequest?appId=5fd1dda724cc620021ace8c5&number=721&view=changes (click details on UI Review GitHub check).

I think having a single story would be nicer for Storybook users. It shouldn't matter too much as Chromatic will still pick up everything, I just guess the visual diffs might be a bit noisier.

burakukula: Sorry, didn't see your comment. Ok, sure, I will merge them into one story.

burakukula:
@m10l ✅ please have a look, if you have a moment. 🙂

m10l (approved):

packages/forma-36-react-components/src/components/Button/Button.stories.tsx
denkristoffer: I think I'm misunderstanding something, I only see a single button in the Basic story, is that right? How can Chromatic detect changes in the other types then?

PR #723, opened 2020-12-21
docs: move README files vol.2

burakukula:
Hello!

Docs vol 2:

- fix Assets docs complaining about imports
- small visual changes - topbar and top section in the dark blue color, navigation on the left side fixed to pass the accessibility requirements
- link adjusted to paragraph - I am not sure if we really want to do that though.
- Form elements grouped together
- Chevron Icons slightly bigger
- more docs in README files.

## PR Checklist

- [ ] Tests are passing


packages/forma-36-website/gatsby-config.js
```diff
@@ -162,12 +158,50 @@ module.exports = {
             link: '/components/flex/',
           },
           {
-            name: 'Form',
-            link: '/components/form/',
-          },
-          {
-            name: 'Form label',
-            link: '/components/form-label/',
+            name: 'Form Elements',
```
burakukula, line 161: Form elements grouped together. WDYT?

packages/forma-36-website/src/components/Container.js
```diff
@@ -49,34 +49,34 @@ const styles = {
...
-  ul: props => <List className="f36-margin-bottom--m" {...props} />,
-  li: props => (
+  h1: (props) => <DisplayText {...props} />,
+  h2: (props) => <Heading element="h2" {...props} />,
+  h3: (props) => <Subheading element="h3" {...props} />,
+  h4: (props) => <Subheading element="h4" {...props} />,
+  h5: (props) => <Subheading element="h5" {...props} />,
+  h6: (props) => <Subheading element="h6" {...props} />,
+  p: (props) => <Paragraph className="f36-font-size--l" {...props} />,
+  a: (props) => <TextLink style={{ fontSize: tokens.fontSizeL }} {...props} />,
```
burakukula, line 59: Link the same size as Paragraph, but I think we need to do something global maybe if we really want to overwrite font-size. I am not sure... Maybe we would just keep 14px, so the default for Paragraph?

packages/forma-36-website/src/components/DocFormatter.js
```diff
@@ -144,6 +146,15 @@ const DocFormatter = ({ frontmatter, dataFromReadme, children }) => {
       <main>
         {activeSection === 'design' && hasDesignDocs && designDocs}
         {activeSection === 'develop' && hasDevelopDocs && developDocs}
+        {data.status === 'alpha' && (
+          <Flex marginBottom="spacingM">
+            <Note noteType="warning">
+              This component is in an alpha state. Breaking changes to the API
```
burakukula, line 152: Note for alpha components. But this would mean, I would remove this info from README files. Does this makes sense?

packages/forma-36-website/src/components/Header.js
```diff
@@ -6,7 +6,7 @@ import { Link } from 'gatsby';
 const styles = {
   header: css`
     display: flex;
-    background-color: ${tokens.colorBlueDark};
+    background-color: ${tokens.colorBlueMid};
```
burakukula, line 9: Color adjusted with the top side of the page on the main view? It looks strange to me, that those 2 elements have different shades of blue.

packages/forma-36-website/src/components/Navigation.js
```diff
@@ -45,16 +45,17 @@ const styles = {
...
     &:hover {
-      background-color: ${tokens.colorElementLight};
+      background-color: ${tokens.colorElementLightest};
     }
   `,
 
   linkActive: css`
-    background-color: ${tokens.colorBlueLight};
-    color: ${tokens.colorWhite};
+    background-color: ${tokens.colorElementLight};
```
burakukula, line 53: I am not sure how to handle the navigation when it comes to colors, we could get some support from designers here I think. For now, I wanted the navigation state to pass the accessibility requirements. (Currently, the active item does not pass it).
I've decided to have an active light gray background and text to be colorTextDark and bold. WDYT?

packages/forma-36-website/src/components/Container.js
```diff
@@ -49,34 +49,34 @@ const styles = {
...
-  ul: props => <List className="f36-margin-bottom--m" {...props} />,
-  li: props => (
+  h1: (props) => <DisplayText {...props} />,
+  h2: (props) => <Heading element="h2" {...props} />,
+  h3: (props) => <Subheading element="h3" {...props} />,
+  h4: (props) => <Subheading element="h4" {...props} />,
+  h5: (props) => <Subheading element="h5" {...props} />,
+  h6: (props) => <Subheading element="h6" {...props} />,
+  p: (props) => <Paragraph className="f36-font-size--l" {...props} />,
+  a: (props) => <TextLink className="f36-font-size--l" {...props} />,
```
burakukula, line 59: Link the same size as Paragraph, but I think we need to do something global maybe if we really want to overwrite font-size, to be sure that it is consistent. WDYT?

m10l: Yeah it's a bit messy but I think this is all that's needed.

We want these markdown elements to use a larger base font size for readability, but we want the actual components in the code playgrounds to use our default F36 base font size. 

Because we're using REMs we can't target one part unless I've forgotten how to CSS, so this seems like the best approach

packages/forma-36-website/src/components/DocFormatter.js
m10l: Do all alpha components have the note in the readme at the moment? If so maybe we can remove this. I'd use this wording though.

packages/forma-36-website/src/components/Navigation.js
m10l: Can you add screenshots for the lazy (like me :D)

m10l: I think we should just use the regular active blue on hover - so basically not giving active items a hover state

packages/forma-36-website/src/components/DocFormatter.js
denkristoffer: I think I'd like to keep the note in the README somehow, so that it can be understood without reading the website.

burakukula: ok, I will roll back. Maybe at some point, we will get this data in a different way from README and generate it in the Note component on the website. I think it's a nice touch.

packages/forma-36-website/src/components/Navigation.js
burakukula: So this is what I had in mind. (I don't like the blue active state, apart of the fact that now it's not passing the accessibility requirements):

Screenshot 2020-12-22 at 13 17 57
Screenshot 2020-12-22 at 13 18 08

m10l: I think blue is more consistent - from what I remember it was grey a while back and there was a deliberate decision to make it blue, so changing it back to grey seems odd (but I totally get the accessibility part)

So I think just using `#2e75d4` / `colorPrimary` as the background makes the most sense to me (as in removing that hover state and just using `colorPrimary`

burakukula: hmm, ok, I see. Ok, I will make it `colorPrimary` then. Although I think it is not the best choice for navigation. (in general, I think we just have a too strong blue appearance on the website.)

burakukula: @m10l done, I rolled back to blue, hover the same as it was, blue on active and no active hover. ✅

packages/forma-36-website/src/components/Container.js
m10l, line 59: Poop, this doesn't work, font size remains unchanged. Seems to need `!important` in order to trump the styles

burakukula: damn, I will use `style` to force it then

burakukula: @m10l checked, now it is working.

m10l (approved):

PR #728, opened 2020-12-22
refactor: JS default parameters

denkristoffer:
# Purpose of PR

Refactors function components to default parameters, which is standard JS instead of using React specific `defaultProps`.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Button/Button.tsx
+++ b/packages/forma-36-react-components/src/components/Button/Button.tsx
@@ -1,4 +1,5 @@
-import React, {
+import React from 'react';
+import type {
   CSSProperties,
   FocusEvent,
   MouseEvent as ReactMouseEvent,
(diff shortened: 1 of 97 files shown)


suevalov:
Will it be picked up by typedocs and storybook?

denkristoffer:
@suevalov No it's not going to work with the way you're using it in https://github.com/contentful/forma-36/blob/97e1d3185e6bcaf6bbbd21281be674abd9918fa4/packages/forma-36-react-components/src/components/Modal/Modal.stories.tsx#L63-L66

I guess we could extract to an object but it seems like a bit of a weird approach 🤔

suevalov:
@denkristoffer this one is an edge case so we can probably extract it to a constant

denkristoffer:
@suevalov Now that I think about it, what's the value in passing the default props if they're default anyway? 😅

denkristoffer:
@suevalov I don't think we need it actually, I rebased with a few changes.

packages/forma-36-react-components/src/components/Button/Button.tsx
```diff
@@ -48,32 +49,22 @@ export interface ButtonProps {
...
     onClick,
-    loading,
-    disabled,
-    indicateDropdown,
+    loading = false,
+    disabled = false,
+    indicateDropdown = false,
     href,
-    type,
+    type = 'button',
```
gui-santos, line 67: I think we need to destructure the props in the argument of the function otherwise Storybook does not pick the default values so the documentation looks like this:

Screenshot 2021-01-06 at 10 02 09

The Accordion, for example, is coded like this and the default values show up in the prop table:
```jsx
({ optionalProp: 'defaultValue' }: Props) => { ... }
```

gui-santos: but this is something we can do every time we need to change something in some component!

gui-santos (approved):
Great stuff!

suevalov (approved):

PR #734, opened 2021-01-04
Feat/button v3 changes

gui-santos:
# Purpose of PR

Adapts the Button to the new look and feel

Screenshot 2021-01-04 at 17 20 38

Screenshot 2021-01-04 at 17 21 46
_hover and active states_

Screenshot 2021-01-04 at 17 22 49
_focus state (In the image "Warning" and "Naked" are wrong, but I added the focus in the code)_

### Open questions

**Warning and Naked**
I added a "deprecated" tag in the storybook stories. Should I do something else?

**Muted button**
The hover state does not have a lot of contrast with the idle state. Wdyt?
@fabe  Should we use a darker grey or let's change when we change the colors?

**Spinner**
Screenshot 2021-01-04 at 17 47 38
This is how the loading of the button looks in the new look and feel
Should we revert the change we made on the Spinner? (I can make the change in another PR based on master)
I think we can implement the three dots animation in the webapp instead of forma36
Wdyt? cc/ @domarku and @Gracegross 

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Button/Button.css
+++ b/packages/forma-36-react-components/src/components/Button/Button.css
@@ -38,6 +38,7 @@
 
 .Button__icon {
   min-width: calc(1rem * (18 / var(--font-base-default)));
+  margin-right: calc(1rem * (8 / var(--font-base-default)));
 }
 
 .Button--full-width {
@@ -196,13 +197,12 @@
   display: flex;
   height: 100%;
   z-index: var(--z-index-default);
-  padding: 0 calc(1rem * (14 / var(--font-base-default)));
+  padding: 0 calc(1rem * (16 / var(--font-base-default)));
   align-items: center;
   justify-content: center;
 }
 
 .Button__label {
-  margin: 0 calc(1rem * (4 / var(--font-base-default)));
   color: var(--color-contrast-mid);
   line-height: 2;
   white-space: nowrap;
--- a/packages/forma-36-react-components/src/components/Button/Button.css
+++ b/packages/forma-36-react-components/src/components/Button/Button.css
@@ -69,15 +69,14 @@
   background-color: transparent;
   border: none;
 
-  &:hover:not(.Button--disabled),
-  &:focus:not(.Button--disabled) {
+  &.Button--is-active:not(.Button--disabled),
+  &:active:not(.Button--disabled),
+  &:hover:not(.Button--disabled) {
     background-color: var(--color-element-light);
   }
 
-  &:active:not(.Button--disabled),
-  &.Button--is-active:not(.Button--disabled) {
-    transition: none;
-    background-color: var(--color-element-mid);
+  &:focus:not(.Button--disabled) {
+    box-shadow: 0px 0px 0px 3px var(--color-element-light);
   }
 
   &.Button:link {
@@ -165,24 +164,16 @@
   background-color: var(--color-orange-mid);
   background-size: 100% 200%;
 
-  &:hover:not(.Button--disabled),
-  &:focus:not(.Button--disabled) {
+  &.Button--is-active:not(.Button--disabled),
+  &:active:not(.Button--disabled),
+  &:hover:not(.Button--disabled) {
     background-position: 0 100%;
     border-color: var(--color-orange-base);
     background-color: var(--color-orange-base);
   }
 
   &:focus:not(.Button--disabled) {
-    border-color: var(--color-orange-base);
-    background-color: var(--color-orange-base);
-  }
-
-  &:active:not(.Button--disabled),
-  &.Button--is-active:not(.Button--disabled) {
-    transition: none;
-    background-image: none;
-    border-color: var(--color-orange-base);
-    background-color: var(--color-orange-base);
+    box-shadow: 0px 0px 0px 3px var(--color-orange-light);
   }
 
   &.Button:link {
--- a/packages/forma-36-react-components/src/components/Button/Button.css
+++ b/packages/forma-36-react-components/src/components/Button/Button.css
@@ -29,6 +29,7 @@
       var(--transition-easing-default);
   vertical-align: middle;
   text-decoration: none;
+  box-shadow: 0px 1px 0px rgb(25 37 50 / 10%);
 
   &:hover:not(.Button--disabled) {
     cursor: pointer;
@@ -44,17 +45,21 @@
 }
 
 .Button--muted {
-  background-color: var(--color-element-light);
+  background-color: var(--color-white);
+  border-color: var(--color-element-light);
+  box-shadow: 0px 1px 0px rgb(25 37 50 / 4%);
 
   &:hover:not(.Button--disabled),
   &:focus:not(.Button--disabled) {
-    background-color: var(--color-element-mid);
+    background-color: var(--color-element-lightest);
+    border-color: var(--color-element-mid);
   }
 
   &:active:not(.Button--disabled),
   &.Button--is-active:not(.Button--disabled) {
     transition: none;
-    background-color: var(--color-element-dark);
+    background-color: var(--color-element-light);
+    border-color: var(--color-element-mid);
   }
 }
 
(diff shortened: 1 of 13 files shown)


denkristoffer:
>  I added a "deprecated" tag in the storybook stories. Should I do something else?

Good question, I was thinking about this just earlier today. We need a process to deprecate components. Maybe a `console.warn` is enough, if we can somehow ensure it's only printed once?

>  The hover state does not have a lot of contrast with the idle state. Wdyt?

Does it pass a11y tests?

packages/forma-36-react-components/src/components/Button/Button.css
```diff
@@ -29,48 +25,58 @@
...
+  margin-left: calc(1rem * (8 / var(--font-base-default)));
 }
 
 .Button--full-width {
   width: 100%;
 }
 
 .Button--muted {
-  background-color: var(--color-element-light);
+  background-color: var(--color-white);
```
burakukula, line 53: @gui-santos was there the idea of introducing this change in v4? From my perspective, it would be good to test it now to check the impact, so all good from my side, just something that came to my mind.

packages/forma-36-react-components/src/components/Button/Button.css
```diff
@@ -29,48 +25,58 @@
...
+  &:hover:not(.Button--disabled) {
+    background-color: var(--color-element-lightest);
   }
 
-  &:active:not(.Button--disabled),
-  &.Button--is-active:not(.Button--disabled) {
-    transition: none;
-    background-color: var(--color-element-dark);
+  &:focus:not(.Button--disabled) {
+    box-shadow: 0px 0px 0px 3px var(--color-element-light);
```
burakukula, line 64: @gui-santos so here it would be possible to use glow from tokens?

Gracegross:
LOVE the new designs! Big fan.

I think the loading spinner makes most sense, the dots are better for full screen loading. @gui-santos

packages/forma-36-react-components/src/components/Button/Button.css
gui-santos: Oops! I forgot to replace the values
I will do that now

packages/forma-36-react-components/src/components/Button/Button.css
gui-santos: uhmmm I thought all color changes were v3
I will test in the webapp today, but if anyone else wants to check it too pls do 🙏

gui-santos:
@denkristoffer 
> > I added a "deprecated" tag in the storybook stories. Should I do something else?
> 
> Good question, I was thinking about this just earlier today. We need a process to deprecate components. Maybe a `console.warn` is enough, if we can somehow ensure it's only printed once?

Ahhh do you mean like when the component is mounted in the app consuming the component?
I think it's a good idea
Like when react warns us "Stop using ComponentWillMount! 🤦 "

> > The hover state does not have a lot of contrast with the idle state. Wdyt?
> 
> Does it pass a11y tests?

It does not
Screenshot 2021-01-05 at 10 24 29

fabe:
Looking great! Could we speed up the transition for the active state a little? `0.1s` should work a little better, otherwise it's barely visible on click. 

Regarding the loading spinner, I actually didn't realize we added the dots. Happy to keep it as is 👍

domarku:
Looks good ✌️ 

I agree with @fabe about speeding up the transitions:
> Could we speed up the transition for the active state a little? `0.1s` should work a little better, otherwise it's barely visible on click.

I would go so far as to suggest no transitions on buttons at all! We don't want anything that would decrease the perceived speed of interactions with the app.

gui-santos:
@fabe @domarku sure! I will make some gifs with the transition change so we can compare and send them to you

I will also make the change regarding the Chevron spacing

burakukula (approved):

packages/forma-36-react-components/src/components/Button/Button.css
```diff
@@ -20,57 +16,75 @@
   border-radius: var(--border-radius-medium);
   font-family: var(--font-stack-primary);
   font-size: calc(1rem * (14 / var(--font-base-default)));
+  font-weight: 500;
```
mshaaban0, line 19: do we need to adjust the normal font weight token for that?

packages/forma-36-react-components/src/components/Button/Button.css
```diff
@@ -20,57 +16,75 @@
...
   background-size: 100% 200%;
-  transition: background var(--transition-duration-default)
+  transition: background var(--transition-duration-short)
       var(--transition-easing-default),
     opacity var(--transition-duration-default) var(--transition-easing-default),
     border-color var(--transition-duration-default)
       var(--transition-easing-default);
   vertical-align: middle;
   text-decoration: none;
+  box-shadow: 0px 1px 0px rgb(25, 37, 50, 0.1);
```
mshaaban0, line 29: I think it's better to put these values in tokens? maybe we will use them later for focusable notifications? `box-shadow-positive`, ..etc? wdyt

packages/forma-36-react-components/src/components/Button/Button.css
```diff
@@ -240,27 +239,23 @@
 
 .Button--small,
 .Button--small:link {
-  height: calc(1rem * (30 / var(--font-base-default)));
+  height: calc(1rem * (32 / var(--font-base-default)));
```
mshaaban0, line 242: I wonder how the fixed height would behave with two lines button!
Should we just rely on the typography line height and padding with spacing? 
that way we get a flexible 32px, and 40px height buttons, no magic as well :)

packages/forma-36-react-components/src/components/Button/Button.css
gui-santos: Yes, I can create a token for it!

packages/forma-36-react-components/src/components/Button/Button.css
gui-santos: for what I understood from my conversation with @fabe 
we still don't know if we want to change the medium token to 500

This might get clearer when we work on Typography maybe, but Idk

What should I do?

packages/forma-36-react-components/src/components/Button/Button.css
gui-santos: Do we have two lines button?

packages/forma-36-react-components/src/components/Button/Button.css
mshaaban0: Since we're using system fonts probably `Segoe UI` and `Helvatica` are the most rendered fonts as one targets windows and the other targets mac osx.

both support font weights range from **100** to **900**

Screenshot 2021-01-06 at 15 21 38

This is fine for now, as we're not sure about the naming, and could be a breaking change. but I'll create a ticket for v4

packages/forma-36-react-components/src/components/Button/Button.css
mshaaban0: I'm not sure. but better be flexible, this is fine for now, just sharing thoughts.

packages/forma-36-react-components/src/components/Button/Button.css
gui-santos: Thank you!

mshaaban0 (approved):

fabe:
- For the negative button focus state, could we darken the border to be `color-red-dark`?
- I also think we can completely remove the `background` transition to and from the `:active` state. Other transitions should stay IMO.

packages/forma-36-react-components/src/components/Button/Button.tsx
```diff
@@ -51,11 +42,11 @@ export interface ButtonProps {
 
 export const Button = (props: ButtonProps) => {
   const {
```
denkristoffer, line 44: Do you want to move these to the function parameters now?

denkristoffer (approved):

packages/forma-36-react-components/src/components/Button/Button.tsx
gui-santos: Ohhhhh yes! Totally forgot about it

thanks

PR #735, opened 2021-01-05
refactor: revert spinner to previous animation

gui-santos:
# Purpose of PR

spinner

It reverts the spinner to the old animation because it makes more sense with the new look and feel
I think this can be merged in master already instead of v3 branch, wdyt?

## PR Checklist

- [x] Tests are passing


burakukula (approved):
Maybe it would make sense to extend the component to keep both animations?

gui-santos:
> Maybe it would make sense to extend the component to keep both animations?

@burakukula  I'm actually thinking now that since there is a significant decrease in the CSS bundle size (5kb) when removing the three dots, we could create a separate three dots animation component in the future (v4) when we will also split all components so users don't need to download the whole bundle to use f36

And for now, we add the three dots in the webapp repo

PR #743, opened 2021-01-06
ci: more optimized ci workflow

suevalov:
Faster builds and proper cache reset between commits


mshaaban0 (approved):
very nice, CI now finishes in 4 minute

PR #750, opened 2021-01-07
ci: don't run ci after publish and ignore non-code changes

suevalov:
# Purpose of PR

* Don't run CI after publishing for no reason. 
* Don't publish new package if the change is only in story, tests, or documentation


gui-santos (approved):

mshaaban0 (approved):
Thanks for `[ci skip]` we needed that

PR #761, opened 2021-01-08
fix: small visual fixes in card components

suevalov:
# Purpose of PR

Small fixes with the alignment of new tags and new borders


burakukula (approved):
Looks good to me 🙂
We would like to have a look at the clickable cards separately or is that all the visual changes needed for cards components?

suevalov:
@burakukula we'll need to take a look at clickable cards separately, there's a task in our backlog for v3 changes.

PR #762, opened 2021-01-08
fix: V3 notification visual changes

burakukula:
# Purpose of PR

## PR Checklist

- [ ] Tests are passing


suevalov (approved):

packages/forma-36-react-components/src/components/Notification/Notification.stories.tsx
```diff
@@ -1,11 +1,8 @@
 import React from 'react';
-import { storiesOf } from '@storybook/react';
-import { action } from '@storybook/addon-actions';
-import { button, text, number, boolean, select } from '@storybook/addon-knobs';
 
 import Button from '../Button';
-import Notification from './index';
-import notes from './Notification.md';
+import Notification, { NotificationsAPI } from './index';
```
gui-santos, line 4: I know it's not a change that you made but... Could you rename NotificationAPI to NotificationProps in this PR, pls?
Just so it follows the naming convention we are using in the other components 

🙏

gui-santos (approved):
🚢

packages/forma-36-react-components/src/components/Notification/Notification.stories.tsx
burakukula: @gui-santos as we export this name (most likely is not used anywhere), would it be in theory, a breaking change if we change the namespace?

gui-santos: I think it's only used here, but I see your point 👍 
let's do it later then

fabe (approved):

PR #766, opened 2021-01-08
fix: align Card focus, hover and selected states with the rest of new focused styles.

suevalov:
## PR Checklist

Small adjustment for Card selection.


gui-santos:
I think we need the selected and the `:active` states to be different than the `:focus`

look at this example, you can't see the cursor in the gif but I'm selecting/unselecting cards and I can't say when the card gets unselected because the `:focus` and the `:active` states are the same as the selected 
cards

It's the same issue we had with the button this wee @fabe and @domarku

gui-santos (approved):
Cool stuff! 🙌

PR #769, opened 2021-01-11
feat(tabs): visual changes

denkristoffer:
# Purpose of PR

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Tabs/Tabs.css
+++ b/packages/forma-36-react-components/src/components/Tabs/Tabs.css
@@ -7,21 +7,18 @@
   display: flex;
 }
 
-.Tabs .Tab {
-  margin-right: var(--spacing-l);
-}
-
 .Tab {
+  box-sizing: border-box;
   white-space: nowrap;
-  color: var(--color-text-dark);
+  color: var(--color-text-lightest);
   position: relative;
   cursor: pointer;
-  padding: 0 var(--spacing-s);
-  height: 56px;
-  line-height: 56px;
+  padding: 0 var(--spacing-m);
+  height: var(--spacing-2xl);
+  line-height: var(--spacing-2xl);
   font-size: var(--font-size-m);
   font-family: var(--font-stack-primary);
-  font-weight: var(--font-weight-normal);
+  font-weight: var(--font-weight-medium);
   outline: none;
   text-decoration: none;
 }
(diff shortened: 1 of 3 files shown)


suevalov:
I'm not sure about this change. As for me, `Tabs` by themselves should be having a border around them and the border appears when you're using Tabs inside a `Module` component

suevalov:
I'd rather introduce a new property that allows you to use vertical dividers instead of a horizonal one.

`dividerType="vertical"` and by default it will be "horizontal" in v3, so we don't need to make any changes in our main web app.

m10l:
+1 for what Alex said, I think the default should be closer to the existing appearance. Would also be cool if there was an option to remove the first and last borders for the new appearance, I can imagine there being issues in some layouts, but maybe that's something for if/when we run into any issues.

denkristoffer:
> I'm not sure about this change. As for me, Tabs by themselves should be having a border around them and the border appears when you're using Tabs inside a Module component

I see your point, but I think Tabs can be sized differently than the tabs you want to show inside it. I think of Tabs as block level, Tab more like inline. You would also have to do some overrides for disabled tabs to position them on top of the Tabs border.

> dividerType="vertical" and by default it will be "horizontal" in v3, so we don't need to make any changes in our main web app.

V3 seems to divice tabs from content, you are proposing dividing tabs from each other if I understand you correctly?

suevalov:
Screenshot 2021-01-11 at 14 01 04

After a second though it seems that we have several components that have a new "embedded" concept - like Tables, Tabs.

So, I wouldn't change the default look and feel of the V3 for sure, but rather introduce vertical dividers only. And we'll be using them only in new applications without changing the main web app.

suevalov (approved):

mshaaban0:
> After a second though it seems that we have several components that have a new "embedded" concept - like Tables, Tabs.
> So, I wouldn't change the default look and feel of the V3 for sure, but rather introduce vertical dividers only. And we'll be using them only in new applications without changing the main web app.

+1 to this, tabs and table are considered "Modular Block Components" with two style `embedded` and `inline`.

we need to create a common prop name for these modular components to switch between both, and default it to the current styles to avoid breaking changes

m10l (approved):
Agree with the comments about inline/embedded - we can come up with something consistent and meaningful before merging the V3 branch

PR #772, opened 2021-01-11
feat: entry card header

suevalov:
# Purpose of PR

* Design only changes for entry cards in v3

### DIFF
--- a/packages/forma-36-react-components/src/components/Card/EntryCard/EntryCard.css
+++ b/packages/forma-36-react-components/src/components/Card/EntryCard/EntryCard.css
@@ -19,7 +19,7 @@
 }
 
 .EntryCard__header {
-  height: calc(1rem * (45 / var(--font-base-default)));
+  height: calc(1rem * (44 / var(--font-base-default)));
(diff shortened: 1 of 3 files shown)


packages/forma-36-react-components/src/components/Card/EntryCard/EntryCard.css
```diff
@@ -18,8 +18,19 @@
       var(--transition-easing-default);
 }
 
+.EntryCard__header {
+  height: calc(1rem * (45 / var(--font-base-default)));
```
mshaaban0, line 22: can we do `44px` instead of `45px` to align with the 4px grid?

mshaaban0 (approved):

packages/forma-36-react-components/src/components/Card/EntryCard/EntryCard.css
suevalov: I think we can

suevalov: updated

burakukula (approved):

PR #773, opened 2021-01-12
feat(table): visual changes

denkristoffer:
# Purpose of PR

V3 visual changes.

A thing to note, Webkit (Safari) doesn't support `box-shadow` on `tr` elements, which is used to show the blue line at the end of the row. I tried with an absolutely positioned pseudo element instead but `position: relative` on table elements is actually against the CSS spec, so this doesn't work at all. I think we can drop the behaviour in Safari and hope they choose to implement `box-shadow` at some point.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Table/Table.css
+++ b/packages/forma-36-react-components/src/components/Table/Table.css
@@ -1,6 +1,19 @@
 @import 'settings/colors';
+@import 'settings/borders';
 
 .Table {
   width: 100%;
-  border: 1px solid var(--color-element-light);
+}
+
+.Table--inline {
+  border-radius: var(--border-radius-medium);
+  box-shadow: 0 0 0 1px var(--color-element-light);
+}
+
+.Table--inline th:first-child {
+  border-top-left-radius: var(--border-radius-medium);
+}
+
+.Table--inline th:last-child {
+  border-top-right-radius: var(--border-radius-medium);
 }
--- a/packages/forma-36-react-components/src/components/Table/TableRow/TableRow.css
+++ b/packages/forma-36-react-components/src/components/Table/TableRow/TableRow.css
@@ -1,5 +1,6 @@
 @import 'settings/colors';
 
-.TableRow:hover {
-  background-color: var(--color-element-lightest);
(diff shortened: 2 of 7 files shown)


denkristoffer:
Do we want to keep the current option of a "compact" version?

suevalov:
@denkristoffer we have to keep it now to avoid breaking changes. we'll figure out what to do with it in v4

suevalov (approved):

mshaaban0 (approved):
Looking good ❤️

mshaaban0:
@denkristoffer Is `selected` state styles included?

packages/forma-36-react-components/src/components/Table/Table.css
```diff
@@ -1,6 +1,13 @@
...
-.Table {
+.Table,
+.TableContainer {
   width: 100%;
+}
+
+.TableContainer--inline {
   border: 1px solid var(--color-element-light);
+  border-radius: var(--border-radius-medium);
+  overflow: hidden;
```
m10l, line 12: Needed for the border clipping? Hopefully this doesn't cause any issues

packages/forma-36-react-components/src/components/Table/TableRow/TableRow.css
```diff
@@ -1,5 +1,6 @@
 @import 'settings/colors';
 
-.TableRow:hover {
-  background-color: var(--color-element-lightest);
+tbody .TableRow:hover {
```
m10l, line 3: Any way to not use element selectors?

m10l (approved):
Couple of comments, nothing blocking

packages/forma-36-react-components/src/components/Table/Table.css
denkristoffer: Yes it seems to be necessary for the border-radius to take effect in all browsers.

denkristoffer: I've pushed an alternative.

burakukula (approved):
🎉

PR #774, opened 2021-01-12
fix: entry card and asset card sizes

suevalov:
# Purpose of PR

Fixes EntryCard and AssetCard size issues that are found when testing latest `v3-visual-changes` in our main web application.

### DIFF
--- a/packages/forma-36-react-components/src/components/Card/AssetCard/AssetCard.css
+++ b/packages/forma-36-react-components/src/components/Card/AssetCard/AssetCard.css
@@ -24,7 +24,7 @@
 }
 
 .AssetCard--size-small .AssetCard__asset {
-  height: calc(1rem * (142 / var(--font-base-default)));
+  height: calc(1rem * (151 / var(--font-base-default)));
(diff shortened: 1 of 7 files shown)


m10l (approved):

mshaaban0 (approved):

packages/forma-36-react-components/src/components/Card/AssetCard/AssetCard.css
```diff
@@ -24,7 +24,7 @@
 }
 
 .AssetCard--size-small .AssetCard__asset {
-  height: calc(1rem * (142 / var(--font-base-default)));
+  height: calc(1rem * (151 / var(--font-base-default)));
```
denkristoffer, line 27: No chance we can get rid of these magic numbers? Isn't this just a weird way to hardcode 151px?

suevalov: I don't like it as well, but I didn't want to change the way it works now. I'd be up for a revamp in v4

PR #775, opened 2021-01-12
fix: V3 Switch visual update

burakukula:
Hey!

Visual alignment of the Switch component, not a perfect solution in my opinion. I need to overwrite icons sizes in order to make it work like required. This should be fixed in v4 when we will deal with icons - we would have 12px size icon. I'm also adding the size prop.

As an alternative, we could use both icons in the background for `:before` and `:after` elements, but we the icons could run out of date. 🤷‍♀️ 

## PR Checklist

- [ ] Tests are passing


packages/forma-36-react-components/src/components/Switch/Switch.css
```diff
@@ -1,26 +1,67 @@
...
+  }
+}
+
+.Switch--small {
+  width: calc(1rem * (28 / var(--font-base-default)));
+  height: calc(1rem * (12 / var(--font-base-default)));
+
+  & + .Switch__label {
+    & .Switch__icon--done {
+      top: 3px;
```
burakukula, line 42: Not pretty, unfortunately.

denkristoffer: 🤯

denkristoffer (approved):

suevalov:
@burakukula looks like we need to add `pointer-events: none` to icons in the component. `close` icon steals hover cursor from the container and `cursor: not-allowed` is fickering.

otherwise - looks beatiful!

suevalov (approved):

packages/forma-36-react-components/src/components/Switch/Switch.css
mshaaban0: for icon alignment?

mshaaban0 (approved):

fabe (approved):

PR #776, opened 2021-01-12
fix: proper exports from alpha datepicker and timepicker packages

suevalov:
# Purpose of PR

Proper exports from `datepicker`, `timepicker` alpha packages


burakukula (approved):
🤝

denkristoffer:
Would you mind giving some context for the reasoning behind this? I'm personally all for named exports but some libraries and methods (Like React's own `lazy`) don't play well with them.

suevalov:
@denkristoffer basically right now we just need to fix the broken build, but in general, we tend to prefer to have named exports everywhere, we even have ESLint rules that disallow default exports

PR #778, opened 2021-01-12
feat(icon-button): new visual styles

mshaaban0:
# Purpose of PR

New icon button styles

Screenshot 2021-01-12 at 18 34 35

## Todo
- [ ] Visual review

### DIFF
--- a/packages/forma-36-react-components/src/components/Button/Button.tsx
+++ b/packages/forma-36-react-components/src/components/Button/Button.tsx
@@ -9,7 +9,7 @@ import type {
 } from 'react';
 import cn from 'classnames';
 import { CSSTransition } from 'react-transition-group';
-import Icon, { IconType } from '../Icon';
+import Icon, { IconType, IconProps } from '../Icon';
 import TabFocusTrap from '../TabFocusTrap';
 import Spinner from '../Spinner';
 
@@ -17,6 +17,7 @@ import styles from './Button.css';
 
 export interface ButtonProps {
   icon?: IconType;
+  iconProps?: IconProps;
   indicateDropdown?: boolean;
   onClick?: MouseEventHandler;
   isFullWidth?: boolean;
@@ -32,10 +33,12 @@ export interface ButtonProps {
     | 'muted'
     | 'naked';
   type?: 'button' | 'submit' | 'reset';
-  size?: 'small' | 'medium' | 'large';
+  /* 'trimmed' to be removed in v4, don't use */
+  size?: 'small' | 'medium' | 'large' | 'trimmed';
   href?: string;
   style?: CSSProperties;
   className?: string;
+  innerWrapperClassName?: string;
   children?: React.ReactNode;
   isActive?: boolean;
 }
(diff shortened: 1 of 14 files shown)


packages/forma-36-react-components/src/components/IconButton/IconButton.tsx
```diff
@@ -72,38 +83,24 @@ export const IconButton = ({
...
-  }
-
   return (
-    <button {...elementProps} type="button" disabled={disabled}>
-      {content}
-    </button>
+    <Button
+      {...elementProps}
+      className={classNames}
+      innerWrapperClassName={styles[`IconButton--${size}`]}
```
denkristoffer, line 90: Is there a big reason we do funky things like this instead of just composing an `<Icon />` inside `<Button />`? 🙂

denkristoffer (approved):

packages/forma-36-react-components/src/components/IconButton/IconButton.tsx
mshaaban0: `IconButton` was basically duplicate of `Button` they both have a very similar api with `Icon` inside. The new iconButton styles are basically "muted" button.

In order to use the button and adjust some padding I had to do the `innerWrapperClassName` magic, as our button dimensions defined on a child wrapper and not the main wrapper. This led me to refactor and simplify the button component but for V4 https://github.com/contentful/forma-36/pull/779

packages/forma-36-react-components/src/components/Button/Button.tsx
```diff
@@ -32,10 +33,12 @@ export interface ButtonProps {
     | 'muted'
     | 'naked';
   type?: 'button' | 'submit' | 'reset';
-  size?: 'small' | 'medium' | 'large';
+  /* 'trimmed' to be removed in v4, don't use */
```
burakukula, line 36: 🙌

packages/forma-36-react-components/src/components/IconButton/IconButton.tsx
```diff
@@ -1,11 +1,16 @@
...
+import Button from '../Button';
 
 import styles from './IconButton.css';
 
 export interface IconButtonProps extends React.HTMLAttributes<HTMLElement> {
+  /**
+   * It controls the size button
+   * 'trimmed' to be removed in v4, don't use
+   */
+  size?: 'small' | 'large' | 'trimmed';
```
burakukula, line 13: @mshaaban0 I'm wondering, for the future if we should come up with the same sizes shared between Icon and IconButton.

mshaaban0: maybe, as per design the icon size is the same, the button size changes.
But for the size naming, I agree we should align them yes.

burakukula: Exactly, cool cool cool.

burakukula (approved):

PR #780, opened 2021-01-13
fix: update IconButton testId regression and update tests

suevalov:
# Purpose of PR

Proper test id should be passed down from `IconButton` component

### DIFF
--- a/packages/forma-36-react-components/src/components/IconButton/IconButton.test.tsx
+++ b/packages/forma-36-react-components/src/components/IconButton/IconButton.test.tsx
@@ -1,6 +1,8 @@
 import React from 'react';
-import { render, screen } from '@testing-library/react';
+import { render, screen, configure } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
+
+configure({ testIdAttribute: 'data-test-id' });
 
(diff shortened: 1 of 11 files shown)


mshaaban0 (approved):

packages/forma-36-react-components/src/components/IconButton/IconButton.test.tsx
```diff
@@ -1,7 +1,9 @@
 import React from 'react';
-import { render, screen } from '@testing-library/react';
+import { render, screen, configure } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 
+configure({ testIdAttribute: 'data-test-id' });
```
denkristoffer, line 5: We should probably start using this everywhere then.

denkristoffer (approved):

burakukula (approved):

packages/forma-36-react-components/src/components/IconButton/IconButton.test.tsx
suevalov: yeah, let's clean this up as a follow up

PR #789, opened 2021-01-14
fix(table-row): selected border regression on safari

mshaaban0:
# Purpose of PR

Found [solution](https://stackoverflow.com/questions/10040842/add-border-bottom-to-table-row-tr#10040881) for table row selected border on safari

Use border instead of box-shadow


packages/forma-36-react-components/src/components/Table/TableRow/TableRow.css
```diff
@@ -6,5 +6,5 @@ tbody .TableRow:hover {
 
 .TableRow--selected {
   background-color: var(--color-blue-lightest);
-  box-shadow: inset -2px 0 0 var(--color-blue-mid);
+  border-right: 2px solid var(--color-blue-mid);
```
denkristoffer, line 9: @mshaaban0 I can't check out this branch because of the name 🙃 Did you check if this makes the content move around?

mshaaban0: image

It looks alright, sorry about the branch, I think you can checkout `"` i.e. `git checkout 'fix(F36-476)/table'`

denkristoffer: I got it working. Check this out:

Screenshot 2021-01-14 at 14 04 32

Try `box-sizing: border-box;`

mshaaban0: Can you point out what is the issue? I tried `box-sizing` on `tr` and couldn't notice any difference :)

denkristoffer: The border on the right extends outside the table width. But yes, now that you mention it `box-sizing` obviously won't work on `tr` 🙁

mshaaban0: yes you're right, it could causes a horizontal scroll as well

PR #798, opened 2021-01-15
feat(tooltip): use Popper for arrow placement

denkristoffer:
# Purpose of PR

Uses Popper to position the Tooltip arrow. Also uses some new tokens.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
+++ b/packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
@@ -86,11 +86,6 @@ export interface TooltipProps {
   usePortal?: boolean;
 }
 
-interface ArrowPositionState {
-  top: string;
-  left: string;
-}
-
 export const Tooltip = ({
   children,
   className,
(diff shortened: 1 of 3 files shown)


packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
```diff
@@ -206,11 +184,15 @@ export const Tooltip = ({
       }}
       {...attributes.popper}
     >
-      {content}
+      <span>{content}</span>
+      {console.dir(attributes)}
```
gui-santos, line 188: Oops! I think we can remove this is a debugging leftover (???)

gui-santos (approved):
🙌 

Everything looks good, maybe the arrow needs more margin when the tooltip is left-start and right-start?

Screenshot 2021-01-15 at 17 22 01

but not a blocker I think :)

denkristoffer:
> Everything looks good, maybe the arrow needs more margin when the tooltip is left-start and right-start?

Yeah you mean to towards the top? I was actually wondering about that, I've set a padding which means Popper should add some extra space automatically but it doesn't seem to be working 🤔

gui-santos:
> Yeah you mean to towards the top? I was actually wondering about that, I've set a padding which means Popper should add some extra space automatically but it doesn't seem to be working 🤔

@denkristoffer uhmmm I think we can check that later, the tooltip is working! I don't think we need to spend time on this detail now :)

good job, man!

PR #807, opened 2021-01-25
feat(button): add additional optional properties for using Buttons as external links

VaguelySerious:
# Purpose of PR

~~Adds the optional "isExternalLink" boolean property to the Button component, which, if enabled
together with the "href" property, sets 'target="_blank"' and 'rel="noopener noreferrer"' on the
resulting html tag.~~

~~I have previously worked around this by using an onclick={() => window.open(url, '_blank');}, which isn't as nice for accessibility.~~

**Edit:**
Adds the optional `target` and `rel` props for the Button component, which only get rendered if the component gets rendered as an anchor tag.
As Kristoffer mentioned, you're already able to use these but the types are complaining, which is alleviated by this PR.

## PR Checklist

- [x] Tests are added
- [x] Tests are passing
- [x] Storybook stories are not required
- [x] Has been tested based on [Contentful's browser support](https://www.contentful.com/faq/about-contentful/#which-browsers-does-contentful-support) (not required, am only adding optional properties without UI)

### DIFF
--- a/packages/forma-36-react-components/src/components/Button/Button.tsx
+++ b/packages/forma-36-react-components/src/components/Button/Button.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { HTMLProps } from 'react';
 import type {
   CSSProperties,
   FocusEvent,
@@ -15,7 +15,19 @@ import Spinner from '../Spinner';
 
 import styles from './Button.css';
 
-export interface ButtonProps {
+type AnchorProps =
+  | {
+      href?: undefined;
+      rel?: never;
+      target?: never;
+    }
+  | {
+      href: string;
+      rel?: HTMLProps<HTMLAnchorElement>['rel'];
+      target?: HTMLProps<HTMLAnchorElement>['target'];
+    };
+
+export type ButtonProps = {
   icon?: IconType;
   indicateDropdown?: boolean;
   onClick?: MouseEventHandler;
@@ -33,12 +45,11 @@ export interface ButtonProps {
     | 'naked';
   type?: 'button' | 'submit' | 'reset';
   size?: 'small' | 'medium' | 'large';
-  href?: string;
   style?: CSSProperties;
   className?: string;
   children?: React.ReactNode;
   isActive?: boolean;
-}
+} & AnchorProps;
 
 export const Button = ({
   buttonType = 'primary',
@@ -46,6 +57,8 @@ export const Button = ({
   className,
   disabled = false,
   href,
+  rel,
+  target,
   icon,
   indicateDropdown = false,
   isActive,
@@ -96,6 +109,8 @@ export const Button = ({
       className={classNames}
       disabled={disabled}
       href={!disabled ? href : undefined}
+      rel={rel}
+      target={target}
       type={type}
       {...otherProps}
     >
--- a/packages/forma-36-react-components/src/components/Button/Button.stories.tsx
+++ b/packages/forma-36-react-components/src/components/Button/Button.stories.tsx
@@ -24,9 +24,9 @@ export default {
   },
 };
 
-interface Args extends ButtonProps {
+type Args = ButtonProps & {
   label?: string;
-}
+};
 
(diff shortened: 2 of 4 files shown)


denkristoffer:
Thanks for the PR @VaguelySerious!

I'm not sure if I'm forgetting something here, but is there anything that currently stops you from passing `rel` and `target` yourself? `<Button href="https://contentful.com" target="_blank" rel="noopener noreferrer" />`, shouldn't this work?

This might not be allowed by types but maybe we can extend `HTMLProps<HTMLAnchorElement>` if that's what's missing.

VaguelySerious:
@denkristoffer No, you're absolutely right. The types don't allow it, but if you circumvent that, it works as intended!
That already solves my use-case, but we can definitely follow up by extending the types.

Could do this by extending with `HTMLProps<HTMLAnchorElement>` or manually adding `rel`and `target` to the props.
And since this is a common combination of tags, do you think it might still make sense to have a boolean switch for those two tags in particular?

VaguelySerious:
Extending with `HTMLProps<HTMLAnchorElement>` seems to be a messier solution than just adding `rel` and `target` to props, as there are conflicting types on the interfaces. I'm modifying this to just add those props.

denkristoffer:
>  And since this is a common combination of tags, do you think it might still make sense to have a boolean switch for those two tags in particular?

I'd rather think of Forma 36 as building blocks that consumers can extend as they want to, so if there's a common use case for external link buttons one could make a simple component wrapping `<Button />` with those props.

We can look into making a union type for `rel` and `target` at some point, it's not necessary now 🙂

packages/forma-36-react-components/src/components/Button/Button.tsx
```diff
@@ -78,6 +82,13 @@ export const Button = ({
 
   const Element: ElementType = href ? 'a' : 'button';
 
+  const additionalLinkProps = href
```
denkristoffer, line 85: This doesn't seem necessary anymore.

denkristoffer (changes requested):

packages/forma-36-react-components/src/components/Button/Button.tsx
VaguelySerious: Do we want to allow rendering of (technically) invalid HTML? `rel` and `target` should only exist on `<a>` tags

mshaaban0:
Thanks a lot for the contribution @VaguelySerious.
I'm curios to know why `HTMLProps<HTMLAnchorElement | HTMLButtonElement>` didn't work?

packages/forma-36-react-components/src/components/Button/Button.tsx
denkristoffer: That's where the union type would come into play if we want to get serious 🙂 I'd do it with something like

```
type AnchorProps =
  | {
      href?: undefined;
      rel?: never;
      target?: never;
    }
  | {
      href: string;
      rel?: HTMLProps<HTMLAnchorElement>['rel'];
      target?: HTMLProps<HTMLAnchorElement>['target'];
    };
```

denkristoffer:
@mshaaban0 That would allow button props on an anchor element and vice versa.

VaguelySerious:
> I'm curios to know why `HTMLProps<HTMLAnchorElement | HTMLButtonElement>` didn't work?

@mshaaban0 In addition to what Kristoffer said, props like `size` aren't compatible, and AFAIK we would need to call a type guard internally to make sure we're using the right properties depending on control flow.

packages/forma-36-react-components/src/components/Button/Button.tsx
VaguelySerious: I recon that's a good idea. I implemented it this way now :+1:

VaguelySerious: Update: I'll need to figure out how to apply this to an interface without converting it to a type or changing the API surface...

denkristoffer (approved):

mshaaban0 (approved):

Issue #423, opened 2020-02-21
List should accept a style prop so we can have bullet lists.


vbeleuta:
![Screenshot 2020-02-21 at 16 28 17](https://user-images.githubusercontent.com/7528329/75047587-4f119980-54c7-11ea-9546-9a5a794c59b9.png)

First list works. Second list causes an error because we pass the **_style_** prop to the _**ul**_ html tag.

Ideally a _List_ with _style="disc"_ should create a bullet list by appling the value of the _style_ prop to the items as _list-style: disc;_

mshaaban0:
@vbeleuta The `style` prop exposes all css properties for the `ul`, hence, I think you can do this...
```jsx
<List style={{listStyleType: "disc"}} />
```
Is that what you're looking for?

m10l:
Actually, this component needs more work… Right now it only renders a UL (what if I want an ordered list?). It also doesn't set the list styles itself.

nikazawila:
@vbeleuta this is now addressed via https://github.com/contentful/forma-36/pull/809 - we introduced the `element` tag that lets you decide which style the list should be displayed in. 
It can either be bulleted (element="ul") or numbered (element="ol") - hope this solves your use case 🤞🏼

PR #809, opened 2021-01-26
feat: handle tag change in the List component

nikazawila:
# Purpose of PR

Addresses https://github.com/contentful/forma-36/issues/423 by introducing the tag change in the List component.

Screenshot 2021-01-26 at 08 59 52

## PR Checklist

- [x] Tests are passing

tagging @fabe for design review on the addition of the ordered list

### DIFF
--- a/packages/forma-36-react-components/src/components/List/List.tsx
+++ b/packages/forma-36-react-components/src/components/List/List.tsx
@@ -3,7 +3,10 @@ import cn from 'classnames';
 
 import styles from './List.css';
 
+type ListType = 'ul' | 'ol';
+
 export interface ListProps {
+  element?: ListType;
   className?: string;
   children: React.ReactNode;
   style?: React.CSSProperties;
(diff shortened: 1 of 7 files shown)


packages/forma-36-react-components/src/components/List/List.tsx
```diff
@@ -3,25 +3,33 @@ import cn from 'classnames';
...
 export interface ListProps {
+  as?: ListType;
   className?: string;
   children: React.ReactNode;
   style?: React.CSSProperties;
   testId?: string;
 }
 
 export function List({
+  as: Tag = 'ul',
```
burakukula, line 17: @nikazawila maybe we could follow one of the existing options for those kinds of props? We have `element` in couple of components or `htmlTag` in `<Flex />` and `<Grid />`. We should probably have just one ( a bit complicated to migrate I guess). But maybe we could avoid introducing another pattern to this family 😉 WDYT?

nikazawila: sure, consistency is king. I guess I did poor research on how it is handled atm. Two ways are already plenty 😉
Will adjust.

mshaaban0 (approved):
🚀

packages/forma-36-react-components/src/components/List/List.tsx
burakukula: Awesome! Thank you 😊

burakukula (approved):

burakukula:
@nikazawila I think it is probably `feat`, as we introduce this new prop to the component. 🤔

denkristoffer:
@all-contributors add @nikazawila for code, maintenance

Issue #805, opened 2021-01-22
🐛 Bug - Tooltip doesn't disappear when wrapping disabled elements


andipaetzold:
# Forma 36 bug report

## Summary

When wrapping `<Tooltip>` around a disabled element e.g. a `<button>`, the tooltip shows up when hovering over the button but does not disappear when leaving the button. The tooltip then just stays visible. It doesn't even disappear when clicking.

It did some first debugging and it seems like `onMouseEnter` triggers correctly but `onMouseLeave` doesn't.

This seems to be a common problem: https://github.com/facebook/react/issues/4251

A potential fix is using `onPointer(Enter|Leave)` or `onMouseOut`. `onPointer(Enter|Leave)` might not be eligible because of it's new and not supported by older browsers.

## Environment

* Chrome is broken
* Safari is broken
* Firefox works

## Steps to reproduce

https://codesandbox.io/s/relaxed-mendeleev-xl87p?file=/src/index.js

## Expected results

The tooltip should disappear when the mouse leaves the button

## Actual results

The tooltip doesn't disappear

denkristoffer:
Looking into this it seems to happen every time on a regular `<button />`, not on Forma 36's `<Button />`, `<TextLink />`, etc. 🤔

denkristoffer:
@andipaetzold Can you try this branch and see if it solves your problem? https://github.com/contentful/forma-36/tree/fix-tooltip-on-disabled-element

andipaetzold:
Hm - I also can't reproduce it in code sandbox, but it's broken in a private project. Will try that branch and let you know :)

andipaetzold:
That seems to fix it.

PR #816, opened 2021-01-27
fix(tooltip): pointer-events disabled on button toggles

denkristoffer:
A fix for #805

See https://github.com/facebook/react/issues/4251#issuecomment-267004045

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
+++ b/packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
@@ -200,7 +200,7 @@ export const Tooltip = ({
     <>
       <ContainerElement
         ref={elementRef}
-        className={targetWrapperClassName}
+        className={cn(styles.TooltipContainer, targetWrapperClassName)}
         onMouseEnter={(evt: MouseEvent) => {
           setIsHoveringTarget(true);
           if (onMouseOver) onMouseOver(evt);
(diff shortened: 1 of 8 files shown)


packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
```diff
@@ -200,7 +200,7 @@ export const Tooltip = ({
     <>
       <ContainerElement
         ref={elementRef}
-        className={targetWrapperClassName}
+        className={cn(styles.TooltipContainer, targetWrapperClassName)}
```
nikazawila, line 203: question or a possible nitpick - do we use camelCase or CapitalCamelCase in f36?

denkristoffer: With regards to CSS we seem to use a mix, of course 🙃😂 Almost all of it is like this though. I've always known it as PascalCase 🙂

nikazawila: 🙃
I'd guess we could fix it then in the whole file so it's consistent at least on the 'per file' basis?

gui-santos: I think the PascalCase one is a class from the css file, not sure why we are using PascalCase to name the classes but I think every CSS class is using that pattern

And `targetWrapperClassName` is a prop and we are always using camelCase for props (this one makes sense for me hahahaha)

One good news is that we will move all styles to emotion so we could make everything camelCase then 🙌

gui-santos (approved):

packages/forma-36-react-components/src/components/Tooltip/Tooltip.tsx
denkristoffer: Yeah, CSS file class names are PascalCase, everything else is camelCase. However, I've seen camelCase names from CSS in a few files as well 😅

PR #817, opened 2021-01-27
chore(AssetCard): align possible entity status

Spring3:
# Purpose of PR

Align possible entity statuses with [field-editor-helpers](https://github.com/contentful/field-editors/blob/master/packages/_shared/src/utils/entityHelpers.ts#L169-L186)

forma was missing the `deleted` status

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Card/AssetCard/AssetCard.tsx
+++ b/packages/forma-36-react-components/src/components/Card/AssetCard/AssetCard.tsx
@@ -11,7 +11,12 @@ import AssetCardSkeleton from './AssetCardSkeleton';
 import CardDragHandle, { CardDragHandlePropTypes } from '../CardDragHandle';
 import styles from './AssetCard.css';
 
-export type AssetState = 'archived' | 'changed' | 'draft' | 'published';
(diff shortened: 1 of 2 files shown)


denkristoffer (commented):
And just to be sure, there's not meant to be any visual changes for deleted assets?

Spring3:
I am not aware of any required changes to the Asset card in a deleted state. If that would be required, then we should be able to address it in the next PR, but here I just added the missing status because using AssetCard in typescript environment doesn't accept the "deleted" status no and then it ends with a build error due to type mismatch and we have to cherry pick that value just to match the expected set of strings

mshaaban0:
I think `deleted` and `archived` should have the same styles, similar to `Tag` component?
Also I'm not sure if there's a better way to centralise `entity-types`, maybe a separate package like browser config?
//cc @suevalov

Spring3:
Or maybe using something like contentful-types where we expose these. Cause I think its too small of a change for another package. But these are just some of my thoughts off the top of my head

denkristoffer (approved):

PR #823, opened 2021-02-04
feat/new typography styles

gui-santos:
# Purpose of PR

Screenshot 2021-02-04 at 14 11 24

It updates font-sizes, line-height, and weights in all typography
And It creates the new huge size for DisplayText
I also made the overview look like the Figma file

I got inspired by [SumUp’s CircuitUI docs](https://circuit.sumup.com/?path=/story/typography-heading--sizes) and created a new section in storybook for our typography (see image above)
but in case you prefer to leave it in Components, I can change this back

P.S: Icon size change in Note component
Screenshot 2021-02-05 at 15 40 51

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Typography/DisplayText/DisplayText.css
+++ b/packages/forma-36-react-components/src/components/Typography/DisplayText/DisplayText.css
@@ -5,17 +5,23 @@
 .DisplayText {
   display: block;
   font-family: var(--font-stack-primary);
-  font-weight: var(--font-weight-demi-bold);
+  font-weight: var(--font-weight-medium);
   color: var(--color-text-dark);
-  line-height: var(--line-height-condensed);
   text-rendering: optimizeLegibility;
   margin: 0;
 }
 
 .DisplayText--default {
   font-size: var(--font-size-2xl);
+  line-height: var(--line-height-2xl);
 }
 
 .DisplayText--large {
   font-size: var(--font-size-3xl);
+  line-height: var(--line-height-3xl);
+}
+
+.DisplayText--huge {
+  font-size: var(--font-size-4xl);
+  line-height: var(--line-height-4xl);
 }
--- a/packages/forma-36-react-components/src/components/Typography/DisplayText/DisplayText.css
+++ b/packages/forma-36-react-components/src/components/Typography/DisplayText/DisplayText.css
@@ -5,7 +5,7 @@
 .DisplayText {
   display: block;
   font-family: var(--font-stack-primary);
-  font-weight: var(--font-weight-medium);
+  font-weight: var(--font-weight-demi-bold);
   color: var(--color-text-dark);
   text-rendering: optimizeLegibility;
   margin: 0;
--- a/packages/forma-36-react-components/src/components/Typography/Heading/Heading.css
+++ b/packages/forma-36-react-components/src/components/Typography/Heading/Heading.css
@@ -5,10 +5,10 @@
 .Heading {
   display: block;
   font-family: var(--font-stack-primary);
-  font-weight: var(--font-weight-demi-bold);
+  font-weight: var(--font-weight-medium);
   color: var(--color-text-dark);
   font-size: var(--font-size-xl);
-  line-height: var(--line-height-default);
+  line-height: var(--line-height-xl);
   text-rendering: optimizeLegibility;
   margin: 0;
 }
--- a/packages/forma-36-react-components/src/components/Typography/Heading/Heading.css
+++ b/packages/forma-36-react-components/src/components/Typography/Heading/Heading.css
@@ -5,7 +5,7 @@
 .Heading {
   display: block;
   font-family: var(--font-stack-primary);
-  font-weight: var(--font-weight-medium);
+  font-weight: var(--font-weight-demi-bold);
   color: var(--color-text-dark);
   font-size: var(--font-size-xl);
   line-height: var(--line-height-xl);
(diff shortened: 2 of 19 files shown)


denkristoffer (approved):
I like the idea of splitting up the stories a bit more 👍

packages/forma-36-tokens/src/tokens/typography/font-weight.js
```diff
@@ -1,6 +1,6 @@
 const fontWeight = {
   'font-weight-normal': 400,
-  'font-weight-medium': 600,
+  'font-weight-medium': 500,
```
denkristoffer, line 3: This is a breaking change 🙀

denkristoffer (changes requested):
I like the idea of splitting up the stories a bit more 👍

packages/forma-36-tokens/src/tokens/typography/font-weight.js
gui-santos: hmmmm and that's the most important part for me hahaha

@mshaaban0 @fabe and @burakukula why didn't you say in slack that we didn't have breaking changes in Typography?

What do we do now, friends?

denkristoffer: If we change our typography it's probably not breaking, but the token change is a breaking change IMO.

But we should align with https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#common_weight_name_mapping

packages/forma-36-tokens/src/tokens/typography/line-height.js
```diff
@@ -1,6 +1,13 @@
 const lineHeight = {
   'line-height-default': '1.5',
   'line-height-condensed': '1.25',
+  'line-height-4xl': '3.5rem',
```
domarku, line 4: I see no reason to have more than two different line-heights. We're introducing unnecessary complexity.

Heading and Display should use the condensed one (1.25).
All the rest should use the default one (1.5).

packages/forma-36-tokens/src/tokens/typography/font-weight.js
burakukula: @denkristoffer @gui-santos change like it is done in here it's not a breaking change - we adjust the value of the token, not changing the namespace for the token itself. 
If we decide to change tokens namespaces, then yeah, it's v4, but I am not sure if we want to do it tbh. 
I am ok with this adjustment like it is for now.

packages/forma-36-tokens/src/tokens/typography/font-weight.js
```diff
@@ -1,6 +1,6 @@
...
   'font-weight-demi-bold': 700,
```
domarku, line 4: Change from 700 to 600

packages/forma-36-react-components/src/components/Typography/SectionHeading/SectionHeading.css
```diff
@@ -5,10 +5,10 @@
 .SectionHeading {
   display: block;
   font-family: var(--font-stack-primary);
-  font-weight: var(--font-weight-demi-bold);
+  font-weight: var(--font-weight-medium);
```
domarku, line 8: This should also be demi-bold

packages/forma-36-react-components/src/components/Typography/Subheading/Subheading.css
```diff
@@ -5,9 +5,9 @@
 .Subheading {
   display: block;
   font-family: var(--font-stack-primary);
-  font-weight: var(--font-weight-demi-bold);
+  font-weight: var(--font-weight-medium);
```
domarku, line 8: Keep it demi-bold

packages/forma-36-react-components/src/components/Typography/DisplayText/DisplayText.css
```diff
@@ -5,17 +5,23 @@
 .DisplayText {
   display: block;
   font-family: var(--font-stack-primary);
-  font-weight: var(--font-weight-demi-bold);
+  font-weight: var(--font-weight-medium);
```
domarku, line 8: Keep it demi-bold

packages/forma-36-react-components/src/components/Typography/Heading/Heading.css
```diff
@@ -5,10 +5,10 @@
 .Heading {
   display: block;
   font-family: var(--font-stack-primary);
-  font-weight: var(--font-weight-demi-bold);
+  font-weight: var(--font-weight-medium);
```
domarku, line 8: Perhaps leave the headings demi-bold

packages/forma-36-tokens/src/tokens/typography/font-weight.js
domarku: Changing the value here doesn't seem like a breaking change to me

domarku (commented):
Make sure to adjust the hacky line-height in note and notification. @burakukula might know more locations
https://github.com/contentful/forma-36/blob/master/packages/forma-36-react-components/src/components/Note/Note.css#L24

packages/forma-36-tokens/src/tokens/typography/font-weight.js
denkristoffer: After giving this a second thought I agree that this is no more breaking than all the other token changes we've done recently. I feel strongly about aligning with the normal weight names though 🙂

denkristoffer (approved):

denkristoffer:
We should publicise the token change like with did with the others recently 👍

burakukula (approved):
🏆

fabe (approved):

PR #825, opened 2021-02-05
docs(forms): create section for form components

gui-santos:
# Purpose of PR

Screenshot 2021-02-05 at 15 04 27

With the intention to organize our usage examples, I propose to create a section for the components related to Forms
Since they are usually used in the same context, grouping these components makes it easier to find them

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Switch/Switch.stories.tsx
+++ b/packages/forma-36-react-components/src/components/Switch/Switch.stories.tsx
@@ -5,7 +5,7 @@ import SectionHeading from '../Typography/SectionHeading';
 import Flex from '../Flex/Flex';
 
 export default {
-  title: 'Forms/Switch',
+  title: 'Components/Switch',
   component: Switch,
   parameters: {
     propTypes: [Switch['__docgenInfo']],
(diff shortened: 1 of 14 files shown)


scripts/.storybook/preview.js
```diff
@@ -21,7 +21,7 @@ export const parameters = {
   options: {
     storySort: {
       method: 'alphabetical',
-      order: ['Documentation', 'Components', '(alpha)'],
+      order: ['Documentation', 'Forms', 'Components', '(alpha)'],
```
denkristoffer, line 24: Nit: I'd argue for "form elements"

gui-santos: Sounds better yes! Thx

domarku (approved):

packages/forma-36-react-components/src/components/Switch/Switch.stories.tsx
```diff
@@ -5,7 +5,7 @@ import SectionHeading from '../Typography/SectionHeading';
 import Flex from '../Flex/Flex';
 
 export default {
-  title: 'Components/Switch',
+  title: 'Forms/Switch',
```
burakukula, line 8: I'm not sure about this one 🤔

burakukula (approved):
Great! Website have also this section 🙌 if we decided that `Switch` should be also part of Form elements, I will add it there as well.

packages/forma-36-react-components/src/components/Switch/Switch.stories.tsx
gui-santos: I'm not sure also
I added it because it's actually a checkbox if you think about and I was ~copying~ inspired by CircuitUI

I will remove it for now. We can revisit this later

PR #830, opened 2021-02-08
feat: primitive component

denkristoffer:
This PR will be a concrete proposal for how to address #861

## @contentful/f36-core

Implement <Primitive /> as component with polymorphic `as` behaviour.
Also exposes several types to allow type inheritance.


gui-santos:
+100

I love this idea 👌

burakukula:
Hey @denkristoffer this base alignment across all the components would be great! 🙌

denkristoffer:
I've pushed some working code. I'm not sure how best to organise this, the polymorphic part might be better to separate from the base component 🤔

packages/forma-36-react-components/src/components/Flex/Flex.tsx
```diff
@@ -172,15 +173,17 @@ export const Flex = ({
...
+      ref={forwardedRef}
+      style={{ ...inlineStyle, ...marginResult, ...paddingsResult, ...style }}
+      testId={testId}
     >
       {children}
-    </Container>
+    </Primitive>
   );
-};
+}) as Polymorphic.ForwardRefComponent<typeof DEFAULT_TAG, FlexProps>;
```
suevalov, line 187: I think there should be a way to make it automatically inferred without explicit typings.

packages/forma-36-react-components/package.json
```diff
@@ -31,6 +31,8 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.3.4",
+    "@contentful/f36-polymorphic": "^4.0.0",
+    "@contentful/f36-primitive": "^4.0.0",
```
suevalov, line 35: Can't we have both components in one package? 

I'd rather introduce `f36-core` and collocate it with several other components.

suevalov (commented):
I like the idea but typing has to be improved. Happy to pair on it.

packages/forma-36-react-components/src/index.ts
```diff
@@ -46,5 +46,6 @@ export * from './components/Tooltip';
 export * from './components/Typography';
 export * from './components/ValidationMessage';
 export * from './components/Workbench';
+export * from './components/Workbench';
```
suevalov, line 49: @denkristoffer should it be deleted?

packages/core/src/Primitive.tsx
```diff
@@ -0,0 +1,40 @@
...
+   * A selection ID used for testing purposes applied as a data attribute
+   * (data-test-id)
+   */
+  testId?: string;
+};
+
+export type PrimitiveProps<E extends React.ElementType> = PrimitiveOwnProps<E> &
+  Omit<React.ComponentProps<E>, keyof PrimitiveOwnProps>;
+
+export type PolymorphicComponentProps<E extends React.ElementType, P> = P &
```
denkristoffer, line 21: Can we use `React.PropsWithoutRef` here? It'd be awesome if we could avoid having to do `ref: typeof otherProps.ref` too.

suevalov: `ref: typeof otherProps.ref` actually helps to enforce a proper reference type based on `as` tag

.circleci/config.yml
```diff
@@ -56,6 +56,7 @@ jobs:
       - check-changed-files-or-halt:
           pattern: ^(packages)|(.js|jsx|ts|tsx|css|scss)$
       - yarn_install
+      - run: yarn run build
```
mshaaban0, line 59: This will run build again after `postinstall` right? should we build everything in `postInstall`?

packages/core/src/Primitive.tsx
```diff
@@ -0,0 +1,32 @@
...
+  as?: E;
+} & CommonProps;
+
+export type PrimitiveProps<E extends React.ElementType> = PrimitiveOwnProps<E> &
+  Omit<React.ComponentProps<E>, keyof PrimitiveOwnProps>;
+
+export type PolymorphicComponentProps<E extends React.ElementType, P> = P &
+  PrimitiveProps<E>;
+
+export type PolymorphicComponent<P, D extends React.ElementType = 'div'> = <
```
mshaaban0, line 16: Curious about `= 'div'` part, is that the default element type?

packages/forma-36-react-components/src/components/Flex/Flex.stories.tsx
```diff
@@ -51,11 +51,12 @@ const DemoBox = ({ times }: { times?: number }) => {
   return <Flex style={styles.demoBox}>Example element</Flex>;
 };
 
-interface Args extends FlexProps {
+interface Args extends FlexOwnProps {
```
mshaaban0, line 54: Do we have to use `Own` here? In the coding style guide we agreed on `ComponentNameProps`

packages/forma-36-react-components/src/components/Flex/Flex.tsx
```diff
@@ -1,6 +1,8 @@
 import React from 'react';
-import tokens from '@contentful/f36-tokens';
+import tokens from '@contentful/forma-36-tokens';
```
mshaaban0, line 2: 👁️‍🗨️ ?

packages/forma-36-react-components/src/components/Flex/Flex.tsx
```diff
@@ -97,44 +89,50 @@ export interface FlexProps {
...
-  /**
-   * style prop, for inline styles */
-  style?: React.CSSProperties;
-}
+};
+
+export type FlexProps<E extends React.ElementType> = PolymorphicComponentProps<
+  E,
+  FlexOwnProps
+>;
```
mshaaban0, line 97: hmmm, I get it now, maybe we use the same pattern as ref `_FlexProps`?

mshaaban0 (approved):
Looks good to me 👍

.circleci/config.yml
suevalov: good point, I think we should simply delete automatic post-install

packages/core/src/Primitive.tsx
suevalov: yes, it's a default tag

packages/forma-36-react-components/src/components/Flex/Flex.tsx
suevalov: 🦅

packages/forma-36-react-components/src/components/Flex/Flex.stories.tsx
suevalov: We also have `FlexProps` in the component but it includes all polymorphic props and HTML props of the element. `ComponentOwnProps` will be an internal type definition

packages/forma-36-react-components/src/components/Flex/Flex.tsx
suevalov: not sure, maybe we should state `FlexInternalProps` instead of `FlexOwnProps` to make it clear?

packages/forma-36-react-components/src/components/Flex/Flex.tsx
suevalov: updated

.circleci/config.yml
mshaaban0: but then locally, after `yarn` If the components are not built, I won't be able to use the package, `main` in package.json won't resolve because `dist` is not there, no?

suevalov: yes, we can make it clear in the documentation but I don't have a strong opinion

packages/forma-36-react-components/src/components/Flex/Flex.tsx
suevalov: updated

denkristoffer:
I'd like it if we can type the references better, can we get an element on this?

Screenshot 2021-03-15 at 10 11 31

If we could get autocompletion on the `as` prop that would be nice too.

suevalov:
@denkristoffer thanks for spotting this, it works as expected with the latest fix

packages/core/src/Primitive.tsx
```diff
@@ -0,0 +1,30 @@
...
+const defaultElement = 'div';
+
+export const Primitive: <E extends React.ElementType = typeof defaultElement>(
+  props: PrimitiveProps<E>,
+) => React.ReactElement | null = React.forwardRef(function Primitive(
+  props: PrimitiveOwnProps,
+  ref: React.Ref<Element>,
+) {
+  const { as: Element = defaultElement, testId = undefined, ...rest } = props;
+  return <Element ref={ref} data-test-id={testId} {...rest} as={undefined} />;
```
gui-santos, line 29:
```suggestion
  const { as: Element = defaultElement, testId = undefined, ...otherProps } = props;
  return <Element ref={ref} data-test-id={testId} {...otherProps} as={undefined} />;
```
Should we name this `otherProps` instead?
This is very nitpicky, but I'm with the codestyle guide that @mshaaban0 started in my head

gui-santos (approved):
Amazing! 👏  👏  👏

packages/core/src/Primitive.tsx
suevalov: updated

packages/forma-36-react-components/src/components/Flex/Flex.tsx
```diff
@@ -172,13 +170,21 @@ export const Flex = ({
...
       {children}
-    </Container>
+    </Primitive>
   );
-};
+}
+
+const _Flex: <E extends React.ElementType = typeof DEFAULT_TAG>(
+  props: FlexProps<E>,
+) => React.ReactElement | null = React.forwardRef(Flex);
```
denkristoffer, line 188: Can we wrap this in a type as well?

suevalov: yep, we can. thanks for the suggestion

denkristoffer:
Another thing that I'm not sure if it's even possible to improve, but we're losing the typed attributes in cases like this, where `viewBox` should be an error:

Screenshot 2021-03-15 at 13 21 19

suevalov:
@denkristoffer it shows up as an error to me, which Typescript version your VSCode is using?

denkristoffer:
Weird 🤔 I've tried both VS Code version (4.2.2) and our version from `node_modules` (3.9.7).

packages/core/src/Primitive.tsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import type { CommonProps } from './types';
+
+type PrimitiveOwnProps<E extends React.ElementType = React.ElementType> = {
+  as?: E;
+} & CommonProps;
+
+export type PrimitiveProps<E extends React.ElementType> = PrimitiveOwnProps<E> &
+  Omit<React.ComponentProps<E>, keyof PrimitiveOwnProps>;
```
denkristoffer, line 9: Can we use `Pick` here? It's more useful to see the list of picked properties in VS Code than the ones excluded.

denkristoffer:
I'd still like to see a more useful type here 
Screenshot 2021-03-15 at 14 11 40

denkristoffer:
Really great work @suevalov ❤️

packages/core/src/Primitive.tsx
suevalov: I don't think we can use Pick here cause we don't really know what to pick :) It's a generic type and we can only say what shouldn't be included in the type.

suevalov:
@denkristoffer `ref: typeof otherProps.ref` isn't ideal in thems how it looks but it actually provides a good type when using this component.
Screenshot 2021-03-15 at 14 21 11

denkristoffer:
@suevalov Yeah I'm okay with it since we'll only be seeing it internally, maybe there's a smart way to improve it in the future.

packages/core/src/Primitive.tsx
denkristoffer: Something like `Pick<React.ComponentProps<E>, Exclude<React.ComponentProps<E>, keyof PrimitiveOwnProps>>;` ?

PR #833, opened 2021-02-10
[WIP]docs: navigation scroll on the website

burakukula:
Hey!

Just a small fix for the sidebar navigation scroll on the website (following this https://www.gatsbyjs.com/docs/how-to/routing/scroll-restoration/)

I moved "Form elements" to separate block in the documentation, similar to what we have in the storybook. WDYT?

Screenshot 2021-02-10 at 15 09 06

## PR Checklist

- [ ] Tests are passing


denkristoffer (approved):
Cool, so this restores scroll on the navigation after changing pages?

burakukula:
@denkristoffer yeah, currently scroll position is lost between page. This should fix this issue.

m10l (changes requested):
This doesn't seem to be working on preview - if I scroll down and click workbench in the navigation, the site goes to workbench as expected but the navigation scrolls back to the top like before.

I like the idea behind grouping the form elements but I don't think this is the right way to do it - I'd rather we followed the same pattern as we have within the guidelines section (looking at copy for example). The hierarchy behind the navigation is that we have three main areas - foundation, guidelines, and components. So I think grouping within components would make the most sense.

burakukula:
@m10l oki! thx for the feedback. I will drop the navigation change then, make sense.
Interesting about the scroll. I see that in the preview is not working, but when I am testing on the localhost it looks like it is. I will try to figure it out.

Screen-Recording-2021-02-11-at-0

PR #835, opened 2021-02-10
fix: update CopyButton to the new styles

burakukula:
Hey!

Update of the `CopyButton` to the new styles, aligning with `Button` muted

BEFORE:

Screenshot 2021-02-10 at 16 43 27

Screenshot 2021-02-10 at 16 43 32

AFTER (updated):

Screenshot 2021-02-11 at 13 28 32
Screenshot 2021-02-11 at 13 28 36

Note: `TextField` with `CopyButton` was already refreshed with v3 changes.

## PR Checklist

- [ ] Tests are passing


m10l:
@burakukula Just thought I'd ask before the preview build is complete - did you check this with the TextInput component? Not sure how well it'd sit with the current styling shown in the screenshot as previously before the visual refesh it was like an addition to the TextInput field

packages/forma-36-react-components/src/components/TextInput/TextInput.css
```diff
@@ -93,7 +93,6 @@
   & button {
     border-left: none;
     height: 100%;
-    background-color: var(--color-white);
```
denkristoffer, line 96: This seems like one of those changes that might have bigger implications than we think 🙂 How does this look with the different editors and apps?

m10l (approved):

denkristoffer (commented):
+1 to Mike's question, I don't think we're actually using this without the input anywhere.

Also, do we have a max-width set on the tooltip here or what's going on? It looks funky 🤔 
Screenshot 2021-02-10 at 17 02 16

packages/forma-36-react-components/src/components/TextInput/TextInput.css
m10l: +1

burakukula:
> @burakukula Just thought I'd ask before the preview build is complete - did you check this with the TextInput component? Not sure how well it'd sit with the current styling shown in the screenshot as previously before the visual refesh it was like an addition to the TextInput field

yeah, so I tried to not affect the input field with CopyButton. We overwrite the border-radius over there. 

Screenshot 2021-02-10 at 21 14 54

@denkristoffer when it comes to usage, I thought that we use it only with TextInput. I will check tomorrow if there are any other places, where we use this component separately. Or maybe something comes to your mind?

packages/forma-36-react-components/src/components/TextInput/TextInput.css
burakukula: @denkristoffer the input with CopyButton does not change the look. It was changed to white when we did the v3 refresh. now I just remove this line, since this white background-color is now coming from `CopyButton`.

packages/forma-36-react-components/src/components/CopyButton/CopyButton.css
```diff
@@ -12,17 +13,26 @@
   height: calc(1rem * (40 / var(--font-base-default)));
   width: calc(1rem * (40 / var(--font-base-default)));
   border: 1px solid var(--color-element-mid);
-  background: var(--color-element-light);
+  border-radius: var(--border-radius-medium);
```
mshaaban0, line 16: Not sure about this, the border radius here could mess up styles for consumer components, a.k.a UI extensions

burakukula: @mshaaban0 thx for the feedback. Do you suggest to go with `border-radius-small` or to not touch it at all? @fabe maybe you would have some suggestions?

mshaaban0: a flag/prop to switch to the new version maybe?

burakukula: @mshaaban0 ok, I think let's keep the border-radius 0 for now. on our end, we almost never use this component separately from `TextInput`. Sounds good?

burakukula: @mshaaban0 great! I rolled back the change for border-radius, if you have a moment, please have a look if this would work. 🙂

mshaaban0 (approved):
Thanks, LGTM 🤝

denkristoffer (approved):

fabe (approved):

PR #836, opened 2021-02-10
feat(portal): export for external consumption

denkristoffer:
# Purpose of PR

feat(portal): export for external consumption

### DIFF
--- a/packages/forma-36-react-components/src/components/Portal/Portal.tsx
+++ b/packages/forma-36-react-components/src/components/Portal/Portal.tsx
@@ -6,7 +6,7 @@ export interface PortalProps {
   container?: Document | HTMLElement;
 }
 
-export default function Portal({
+export function Portal({
   children,
   container: containerElement,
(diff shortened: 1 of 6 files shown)


packages/forma-36-react-components/src/components/Portal/Portal.tsx
```diff
@@ -6,7 +6,7 @@ export interface PortalProps {
   container?: Document | HTMLElement;
 }
 
-export default function Portal({
+export function Portal({
```
burakukula, line 9: Looking good! 👏 Maybe we would be able to write README with examples of usage? 🙂

burakukula (approved):

packages/forma-36-react-components/src/components/Portal/Portal.tsx
denkristoffer: Ah yes, of course, thanks for reminding me 🙂 I will add this.

gui-santos (approved):

PR #838, opened 2021-02-15
feat: introducing size property in Tag component

burakukula:
Hey, Forma 36!

I would like to introduce size for the Tag component, which will cover a couple of additional use cases.

For example:
Screenshot 2021-02-15 at 18 09 35

Additionally README file for the component.

## PR Checklist

- [ ] Tests are passing


denkristoffer (approved):

domarku:
We set the smallest font size that should be used to 12px because anything lower than that becomes difficult to read. I don't think we need a smaller size but would rather replace the smaller instances with the component "in the wild".

Are there any other examples besides the NEW tag in the Apps dropdown?

packages/forma-36-react-components/src/components/Tag/Tag.css
```diff
@@ -51,3 +51,10 @@
   color: var(--color-white);
   background-color: var(--color-primary);
 }
+
+.Tag--small {
+  padding: 3px 5px;
+  font-size: 10px;
```
domarku, line 57: Seems a bit wrong to set these values in `px` anyway.

burakukula: @domarku yeah, you are right, on my way to fix it. thx!

burakukula:
> We've set the smallest font size in the tokens to 12px because anything lower than that becomes difficult to read, particularly when the text is set in all caps like here. I don't think we need a smaller size but would rather replace the smaller instances in Contentful with the component.
> 
> Are there any other examples besides the NEW tag in the Apps dropdown?

Hey @domarku!
Here is another example where we should use the small size of this Tag:

image (2)

@fabe would you be able to have a look at the font-size, we've talked about 10px, but maybe we should set it to 12, as @domarku is suggesting? WDYT?

packages/forma-36-react-components/src/components/Tag/Tag.css
```diff
@@ -51,3 +51,10 @@
   color: var(--color-white);
   background-color: var(--color-primary);
 }
+
+.Tag--small {
+  padding: 3px var(--spacing-2xs);
+  font-size: 0.625rem;
+  line-height: 0.625rem;
```
gui-santos, line 58: tokens?

gui-santos (approved):

packages/forma-36-react-components/src/components/Tag/Tag.css
burakukula: @gui-santos we don't have this token I think - 10px. I guess it might be custom, but if it's needed in other places I could add it to tokens, of course.

fabe:
@domarku @burakukula FWIW, I think using 10px here is fine, mainly because it will be used as a decorative element, almost like an icon. Don't think it's comparable to something like paragraph text or a status. Using 12px adds quite a bit of visual weight of the element that feels too extreme (like in the instance parameter definitions screenshot that was posted earlier), especially combined with 14px text.

We would of course have to document this clearly so that the small variant isn't used for regular statuses etc. This `NEW` tag in 10px has been used for lots of different releases (Tags, custom entry editors, SSO, ...), but AFAIK every team has implemented this tag custom again and again, so this could clean that up.

domarku:
@fabe yeah, you're right. Come to think of it, perhaps the smaller one could be used for statuses in the reference picker or whenever real-estate is scarce.

domarku (approved):

PR #839, opened 2021-02-16
v4 feat: align prop interface names

denkristoffer:
# Purpose of PR

Breaking change: align prop interface names so we use the same naming everywhere.


burakukula (approved):
👏

PR #843, opened 2021-02-19
feat(icon): add LinkAlternate

denkristoffer:
Taking suggestions for better names 😅


mshaaban0 (approved):

burakukula (commented):
@denkristoffer I guess the naming is ok.
I'm wondering why we are introducing a second icon for link. Are the use cases different somehow? If that's only a visual thing I guess it would be good to keep only one icon for links to have it consistent in the interface across all the places. Is it possible from your perspective to replace the existing link icon with the new one?

denkristoffer:
@burakukula No, this is a design request that's only meant to be used in one particular place of the app. They still want to keep the other link icon as well. I totally agree with what you're saying though, I see the need for this as another good argument for opening up for custom icons.

PR #846, opened 2021-02-25
feat(autocomplete): add event to onChange

damienxy:
# Purpose of PR

The purpose of this PR is to pass on the `onChange` event from the `Autocomplete` component.
The specific use case that made it necessary is the need to `preventDefault` in some cases.

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Autocomplete/Autocomplete.test.tsx
+++ b/packages/forma-36-react-components/src/components/Autocomplete/Autocomplete.test.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { KeyboardEvent, MouseEvent } from 'react';
 import { render, fireEvent, within, screen } from '@testing-library/react';
 
 import { KEY_CODE } from './utils';
@@ -18,7 +18,7 @@ const items: Item[] = [
 ];
 
 describe('Autocomplete', () => {
-  let onChangeFn: (value: Item) => void;
+  let onChangeFn: (value: Item, event: MouseEvent | KeyboardEvent) => void;
   let onQueryChangeFn: (query: string) => void;
 
   const build = ({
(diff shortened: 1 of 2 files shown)


denkristoffer:
Hi @damienxy 👋 

Can you explain a bit more about when or why you'd need to `preventDefault` for changes?

damienxy:
> Hi @damienxy 👋
> 
> Can you explain a bit more about when or why you'd need to `preventDefault` for changes?

Hey @denkristoffer,

I have a case where the selection of a dropdown option opens a modal, and if you select it via keyboard, the same event immediately sends the form that is inside the modal. So in this case `preventDefault` is needed to prevent the modal form from sending.

denkristoffer:
Thanks for explaining, I see now. In that case I'd say we should consume the key event when an item is selected instead of passing it on, so I think we might actually want to prevent default in `handleKeyDown` to solve this issue for all users. Would this work for you?

damienxy:
Yes, adding `if (isEnter) { event.preventDefault() }` [here in the code](https://github.com/contentful/forma-36/blob/master/packages/forma-36-react-components/src/components/Autocomplete/Autocomplete.tsx#L153) works for my specific case. 

Reasons I chose the above approach instead were:
- We can be sure that the change doesn't break any existing code (I don't know if we would break anything by putting it directly into `handleKeyDown`)
- Passing on the event would make it more flexible to interact with when using the component, if needed.

But I don't feel strongly about it, either way will fix my code :-)

denkristoffer:
The reason I suggest `preventDefault()` is really that I'd suggest we include the selected item in an enhanced SyntheticEvent instead of adding `event` as a second parameter in event handlers. I see this as better following the principle of least surprise as it's just an extension of the regular event API. It would be a breaking change, however, so for now `preventDefault()` could solve your issue without changing the API.

I have a few poor experiences with libraries like Microsoft's Flow UI where it used to not work out of the box with a lot of other libraries because its API is `(newValue, event)` instead `(event)`.

Anyway, let's wait and see if any other reviewers want to weigh in.

damienxy:
Ok. Should I tag anyone specific?

denkristoffer:
They are already tagged 🙂

gui-santos:
Hey! Yep I think @denkristoffer suggestion is good
+1 to that

denkristoffer:
Hey @damienxy do you want to create a PR for just the `preventDefault` change for now? Let's get you all unblocked 🙂

damienxy:
Will do, thanks both!

PR #850, opened 2021-02-25
💡 v4 Proposal - Change export structure

denkristoffer:
# Purpose of PR

This PR is a draft that proposes how we can simplify our exports and make sure they include types. I've only done this for the `Accordion` component for now.

It's mostly about removing a bunch of index files for subcomponents.


suevalov:
Looks good! Few questions and suggestions:

1. Can we make this change in v3? It doesn't look like a breaking change to me. If we can, we probably should.
2. I'd simply disallow usage of default exports (using ESLint), so we always have a consistent naming when we import code and it's easier to implement the current proposal

denkristoffer:
Yes, we probably can implement this in v3 if the structure that I've seen with Accordion is the same everywhere else. I will look into this.

> I'd simply disallow usage of default exports (using ESLint), so we always have a consistent naming when we import code and it's easier to implement the current proposal

I'm always cautious of removing default exports because React's own lazyload support [only works with default imports](https://reactjs.org/docs/code-splitting.html#named-exports), and other libraries similarly expect default imports. Maybe this is not worth keeping them around for, as far as I know we don't use this ourselves.

denkristoffer:
#852

PR #851, opened 2021-02-25
v4: remove alpha components

denkristoffer:
Can we remove the alpha setup yet?

In the future we can alpha components in their individual packages instead.


suevalov (approved):
Yes, I think we can

PR #852, opened 2021-02-26
feat: export types

denkristoffer:
# Purpose of PR

- Remove default exports and imports and use named only.
- Simplifies exports and exports all types.

Sorry for the mega PR, it touches a lot of files but the changes are simple. This builds on #850. Allow me to bring over a comment of mine from there:

> I'm always cautious of removing default exports because React's own lazyload support [only works with default imports](https://reactjs.org/docs/code-splitting.html#named-exports), and other libraries similarly expect default imports. Maybe this is not worth keeping them around for, as far as I know we don't use this ourselves.

In the end I've removed default exports.

## To do

- [x] fix `ModalConfirm` test 🤔 
- [ ] ~~Update plop templates~~


denkristoffer:
The ModalConfirm test is failing at the moment. I'm not sure I can spot why 🤔

packages/forma-36-react-components/src/index.ts
```diff
@@ -1,84 +1,50 @@
...
-export { TableBody } from './components/Table/TableBody/TableBody';
-export { TableCell } from './components/Table/TableCell/TableCell/TableCell';
-export { TableHead } from './components/Table/TableHead/TableHead';
-export { TableRow } from './components/Table/TableRow/TableRow';
-export { ToggleButton } from './components/ToggleButton/ToggleButton';
-export { AssetCard } from './components/Card/AssetCard/AssetCard';
-export { Asset, AssetIcon } from './components/Asset';
-export { Tag } from './components/Tag/Tag';
-export { Heading } from './components/Typography/Heading/Heading';
+export * from './components/Accordion';
```
suevalov, line 1: I'm fully on board with all the changes in this PR but this file. I think it's better to be specific about what you export. Otherwise, it's really easy to make a mistake and make public something that is not intended to be.

suevalov:
It would be cool to setup https://github.com/benmosher/eslint-plugin-import#static-analysis, so we can force no-default exports rule and also validate imports propertly

packages/forma-36-react-components/src/index.ts
denkristoffer: I am on board with that 🙂 But what kind of things do you expect we want to keep private? As you know we're already seeing people importing components and types that aren't public by drilling down into absolute paths.

denkristoffer:
> It would be cool to setup benmosher/eslint-plugin-import#static-analysis, so we can force no-default exports rule and also validate imports propertly

Absolutely. I was planning to tackle a greater eslint configuration change separately, but it might be a good idea to get it in here.

packages/forma-36-react-components/src/index.ts
suevalov: I can imagine some constant being exported only for the test or even some internal component being exported for the same test purpose. It would be unfortunate for them to be exported from the main entry point.

denkristoffer: I think it's easier to control from the component index file, I've moved explicit exports to there. Question is if I missed anything 😅

suevalov:
Any help needed with fixing `lint-and-test` step?

suevalov (approved):
👍🏼

.eslintrc.js
```diff
@@ -83,5 +86,18 @@ module.exports = {
...
       },
     },
+    {
+      files: [
+        '**/forma-36-website/**/*',
+        '**/forma-36-react-components/tools/**/*',
+        '**/forma-36-react-components/**/Icon/svg/*',
+        '**/forma-36-react-components/**/ProductIcon/**/*',
+        '**/forma-36-react-components/**/Illustration/**/*',
+        '**/packages/components/**/*',
```
mshaaban0, line 96: 🤔 why not just the `./src/**`?

burakukula (approved):
🎉

mshaaban0 (approved):
Looks good to me, thanks for doing the changes 🥇

.eslintrc.js
denkristoffer: That would exclude all the components we want to lint.

mshaaban0 (changes requested):
This is a breaking change for people who import components `/dist/component/.../`

denkristoffer:
🤭

suevalov:
@mshaaban0 I don't think it's a breaking change. I guess `dist` paths shouldn't be a part of public API, otherwise every file move is a breaking change.

mshaaban0:
@suevalov what I meant is users who import default `import Component from 'dist/component/'` will not work anymore, we're doing the object destructuring now, no?

PR #855, opened 2021-03-02
fix(flex): typo in fullHeight prop preventing the height to take effect

mshaaban0:
# Purpose of PR

Fix typo in Flex `fullHeight` prop


burakukula (approved):
🙇‍♀️

Issue #841, opened 2021-02-17
💡 Proposal - Tag vs Pill


burakukula:
# Tag and Pill migration in v4

## The problem
Hey everyone!

Currently, we have 2 components that might be confusing.
`<Tag />` which is mostly used for labeling of highlighting elements in the interface

Example:
<img width="126" alt="Screenshot 2021-02-18 at 15 01 33" src="https://user-images.githubusercontent.com/4446634/108368012-477bf200-71fa-11eb-82a0-35170ddada58.png">

`<Pill />` which is used for tagging elements in the interface

Example:
<img width="256" alt="Screenshot 2021-02-18 at 15 01 38" src="https://user-images.githubusercontent.com/4446634/108368068-582c6800-71fa-11eb-8b0b-539c038defa4.png">

I think in order to achieve more consistency in naming we could migrate those 2 components in v4 to align their names with their purpose

## The proposed solution
1. Migrate `<Tag />` to `<Label />`:
 - create a separate package for `<Label />` that will contain current `<Tag />`
 - I would like to introduce a set of "filled" Labels. Currently, we have only `filled-primary` and I wish we have a full set, the same as we have with the default (light) state. **Downside** - they look a bit like Buttons to me 😕
Fixing my own bug: I've got to the conclusion, that it would be better to create `solid` boolean property for the component, instead of for example `tagType="primary-filled"` it would be something like:
 
 ```
 <Label labelType="primary" size="small" solid>new</Label>
 ```
 This is how it potentially would look like (more or less):
<img width="316" alt="Screenshot 2021-02-17 at 15 19 34" src="https://user-images.githubusercontent.com/4272331/108217147-8dba4e00-7133-11eb-8fa8-4f0c0a0aa3cd.png">

2. Migrate `<Pill />` to `<Tag />`, with the existing functionalities of `Pill`. Maybe we would like to extend the color set of `Pill`? is there a usecase for having more colors for this component?

## Breaking changes
v4

## Open questions:
 - do we agree to the proposed naming or maybe we would like to migrate `<Tag />` to something like `Badge`?
 - do we want to create `solid` labels (current Tags) and do we have any objections against the `solid` boolean prop? 
 - do we want to extend the colors set for `<Pill />` (future Tag) component - more details #585?

burakukula:
Additional changes for the `<Pill />` component are mentioned in here https://github.com/contentful/forma-36/issues/585

gui-santos:
[...]

PR #857, opened 2021-03-05
feat: extract Badge to a separate package

suevalov:
# Purpose of PR

Extracts `Badge` component to a separate `@contentful/f36-badge` package.

* Refactors styles to Emotion
* Adds forward ref
* Renames `Tag` to `Badge` according to proposal: https://github.com/contentful/forma-36/issues/841
* Renames `tagType` to `variant`
* Removes `entityStatusType` from `Badge` component
* Introduces an additional component `EntityStatusBadge` which wraps `Badge` and converts entity status to badge variant under the hood.
* `Badge` and `EntityStatusBadge` (and their props) are re-exported through the main Forma36 entry point


burakukula:
Hey @suevalov! You might want to have a look on this proposal https://github.com/contentful/forma-36/issues/841

There is an idea to migrate `Tag` to `Badge` and  `Pill` to `Tag` in v4.

packages/components/tag/src/Tag.styled.tsx
```diff
@@ -0,0 +1,70 @@
+import tokens from '@contentful/f36-tokens';
+import styled from '@emotion/styled';
+import { css } from '@emotion/core';
+import type { EntityStatus, TagType } from './types';
+
+export interface TagProps {
+  tagType?: TagType;
```
burakukula, line 7: @mshaaban0 made a proposal of the properties naming convention. We might go with `variant` over here? and the `entityStatusVariant` I guess?

package.json
```diff
@@ -68,8 +68,10 @@
     "@commitlint/cli": "^11.0.0",
     "@commitlint/config-conventional": "^11.0.0",
     "@contentful/browserslist-config": "^1.0.1",
-    "@parcel/babel-plugin-transform-runtime": "2.0.0-nightly.1823",
-    "@parcel/babel-preset-env": "2.0.0-nightly.526",
+    "@parcel/babel-plugin-transform-runtime": "2.0.0-nightly.2233",
+    "@parcel/babel-preset-env": "2.0.0-nightly.612",
+    "@parcel/transformer-typescript-types": "2.0.0-nightly.612",
```
mshaaban0, line 73: 👀 `nightly`?

packages/components/tag/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "@contentful/f36-tag",
+  "version": "4.0.0",
```
mshaaban0, line 3: Will this be available instantly as version `4.0.0`? maybe `alpha`?

packages/components/tag/src/Tag.styled.tsx
mshaaban0: Yeah, I think this PR is just to validate the package separation, we might want to address broader changes in other PRs

packages/components/tag/src/Tag.styled.tsx
```diff
@@ -0,0 +1,70 @@
...
+      return css`
+        color: ${tokens.colorWhite};
+        background-color: ${tokens.colorPrimary};
+      `;
+    default:
+      return css``;
+  }
+};
+
+export const Container = styled.div<{ tagType: TagType }>`
```
mshaaban0, line 57: 🤔 why styled components?

packages/components/tag/src/Tag.tsx
```diff
@@ -0,0 +1,42 @@
...
+    testId = 'cf-ui-tag',
+    ...otherProps
+  } = props;
+
+  const calculatedTagType = entityStatusType
+    ? statusTagTypeMap[entityStatusType]
+    : tagType;
+
+  return (
+    <Container
```
mshaaban0, line 29: 💭 how would I expose the html element for `as` prop with styled `Container`?

packages/components/tag/src/Tag.tsx
```diff
@@ -0,0 +1,42 @@
...
+      tagType={calculatedTagType}
+      {...otherProps}
+    >
+      {children}
+    </Container>
+  );
+}
+
+const _Tag = React.forwardRef(Tag);
+export { _Tag as Tag };
```
mshaaban0, line 42: I like this pattern 👍

packages/components/tag/stories/Tag.stories.tsx
```diff
@@ -1,8 +1,8 @@
 import React from 'react';
 
-import Tag, { TagProps } from './Tag';
-import SectionHeading from '../Typography/SectionHeading';
-import Flex from '../Flex/Flex';
+import { Tag, TagProps } from '../src/Tag';
+import SectionHeading from '../../../forma-36-react-components/src/components/Typography/SectionHeading/SectionHeading';
+import Flex from '../../../forma-36-react-components/src/components/Flex/Flex';
```
mshaaban0, line 5: later we would be able to use `import { SectionHeading } from '@contentful/f36-typography'` correct?

mshaaban0 (commented):
Nice one 💯 added some initial thoughts

packages/components/tag/src/Tag.styled.tsx
mshaaban0: I'd agree on the naming though before merging :)

suevalov: I agree that we might want to change prop names here, but I'd rather merge this and follow-up later will all other changes.

Regarding, `Tag | Pill | Badge`:

For me, it makes sense to rename `Tag` to `Badge` but I would avoid renaming `Pill` to `Tag` and confusing people with a swap for names.

package.json
suevalov: yeah, official parcel@2 is not ready yet, but it's used by `react-spectrum` as nightly and in general works well :) As soon they are out with official parcel@2 we make a switch too.

packages/components/tag/package.json
suevalov: once merged to `forma-v4` it will released as `alpha`, according to CI setup

packages/components/tag/src/Tag.styled.tsx
suevalov: it looked like a reasonable approach for me, happy to see a discussion about it

packages/components/tag/src/Tag.tsx
suevalov: 👍🏼

packages/components/tag/stories/Tag.stories.tsx
suevalov: yep, exactly

packages/components/tag/src/Tag.tsx
suevalov: Yeah, we can do this if we want to for a11y purposes, but I'm don't have an opinion on this.

packages/components/tag/src/Tag.styled.tsx
suevalov: Left a comment in the proposal: https://github.com/contentful/forma-36/issues/841#issuecomment-791404659

suevalov: Updated according to the proposal

packages/components/badge/src/Badge.styled.tsx
```diff
@@ -1,19 +1,10 @@
...
-  style?: React.CSSProperties;
-  className?: string;
-  testId?: string;
-  children: React.ReactNode;
-  entityStatusType?: EntityStatus;
-}
-
-const tagTypeToStyles = ({ tagType }: { tagType: TagType }) => {
-  switch (tagType) {
+const tagTypeToStyles = ({ variant }: { variant: BadgeVariant }) => {
```
denkristoffer, line 6: Should the function be renamed with variant?

denkristoffer (changes requested):
Why a `Badge.styled.tsx` file and a `Badge.tsx`?

suevalov:
@denkristoffer I wanted to keep the separation of styles from the main component code, as we had with `Tag.tsx` and `Tag.modules.css`.

what's your suggestion with the styles file name?

packages/components/badge/src/Badge.styled.tsx
suevalov: renamed

denkristoffer (approved):

denkristoffer:
It's just a preference thing, I like to keep the styles in the component itself so I don't have to dig around for it. With `styled()` that becomes very verbose though.

packages/components/badge/src/types.ts
```diff
@@ -0,0 +1,16 @@
+export type BadgeVariant =
+  | 'primary'
+  | 'primary-filled'
+  | 'positive'
+  | 'negative'
+  | 'warning'
+  | 'secondary'
+  | 'muted';
+
+export type EntityStatus =
```
denkristoffer, line 10: It'd be great to make these global Contentful types at some point.

jest.config.js
```diff
@@ -120,7 +120,7 @@ module.exports = {
   setupFilesAfterEnv: ['<rootDir>/scripts/test/setupTests.js'],
 
   // A list of paths to snapshot serializer modules Jest should use for snapshot testing
-  // snapshotSerializers: [],
+  snapshotSerializers: ['jest-emotion'],
```
denkristoffer, line 123: I think it's worth challenging whether or not we want to have CSS in our snapshots. On one hand they might not be useful enough without it, but on the other hand do we want to have to update them every time we have a style change now that we also have Chromatic?

suevalov: I personally think that we shouldn't have snapshot testing as a matter of fact, especially since we have Chromatic.

Unit tests should test something meaningful: user interactions, ally tags, some logic that is applied based on props.

I'd rather rewrite these tests to something meaningful when we'll be moving them to a separate component.

packages/components/tag/src/Tag.styled.tsx
mshaaban0: I'm not aware of the differences between `css` and `styled` apis, but I think everyone is used to `css`, not a strong opinion, but worth considering consistency

mshaaban0 (approved):
LGTM, thanks for doing the changes 👍

jest.config.js
denkristoffer: I like it in places [like this](https://github.com/contentful/forma-36/pull/830/files#diff-64752a5a2d5b11e4e45b39911fff261800a48c187fa1d351e6d0f656901ccd4b) 🙂

PR #862, opened 2021-03-09
feat: extract Portal and TabFocusTrap to f36-utils package

suevalov:
# Purpose of PR

* Extracts `Portal` and `TabFocusTrap` to a separate `f36-utils` package
* Updates imports
* Updates snapshots


denkristoffer (approved):
👍

denkristoffer (changes requested):
The Chromatic changes look very different, the modal backdrop is gone etc. Is this intended?

denkristoffer:
@mshaaban0 I agree **100%** but I don't mind pushing it to a later PR that thoroughly removes it, which seems outside the scope of this one.

mshaaban0:
Removal can go later yes, but I wouldn't expose it in the new package

suevalov:
@denkristoffer we have the same diff in the v4 branch.... looks like we need to take a look there. I wish we could compare to non-main branch.

suevalov:
@mshaaban0 I agree that this component has to be removed but we need to expose it now in order to be able to proceed with refactoring. it will be only available in `4.0.0-alpha` so it shouldn't be a problem

denkristoffer (approved):

denkristoffer:
If Chromatic always compares to `master` it's not very helpful here...

suevalov:
@denkristoffer we might need to take a look at their documentation: https://www.chromatic.com/docs/branching-and-baselines

PR #867, opened 2021-03-10
v4 chore(deps): move type deps to dependencies

denkristoffer:
# Purpose of PR

Types used in our own types should be `dependencies` and not `devDependencies` to ensure that our exported types work for consumers.


suevalov (commented):
Were there any complaints about this? I don't think we've ever had problems with that.

denkristoffer:
I doubt it, we use a bunch of packages that depend on the types so I don't see a situation happening where they aren't available. One gotcha we could potentially run into though, is that the installed version could potentially be different from the version we expect.

suevalov:
@denkristoffer could you maybe add a link to an article or a documentation article that suggests adding `@types` to dependencies?

denkristoffer:
I don't have an article, but here's an answer to a GitHub issue from Daniel Rosenwasser, Program Manager of TypeScript https://github.com/microsoft/types-publisher/issues/81#issuecomment-234051345

> Using --save-dev is fine when you're writing a simple application, and it won't be used as a library. The problem comes along when you might have dependencies. If you stored your type declarations in your devDependencies, your consumers would not automatically get your type declarations. They'd then have to manually add your declaration file dependencies as their devDependencies.
> 
> Given that breaking consumers is a worse problem than slightly-larger packages, we've made --save the default in our documentation.

packages/components/timepicker/package.json
```diff
@@ -37,6 +37,7 @@
   "dependencies": {
     "@babel/runtime": "^7.6.2",
     "@contentful/forma-36-react-components": "^3.83.0",
+    "@types/react": "16.14.2",
```
suevalov, line 40: this one shouldn't be here, because `react` is a peer dependency of the package and will people will be using different react versions

packages/forma-36-react-components/package.json
```diff
@@ -36,6 +36,12 @@
     "@contentful/f36-badge": "^4.0.0",
     "@contentful/f36-utils": "^4.0.0",
     "@popperjs/core": "^2.4.4",
+    "@types/classnames": "2.2.7",
+    "@types/react": "16.14.2",
```
suevalov, line 40: Let's remove `react` and `react-dom` for the same reasons.

packages/components/timepicker/package.json
denkristoffer: I'm not sure I agree since DefinitelyTyped doesn't follow semantic versioning. We can't be sure that the user's version works with our types.

denkristoffer: It'd also require non-TS users to add `@types/react` to their `package.json` to avoid NPM complaining.

suevalov: Not-TS users shouldn't be affected

denkristoffer: But they'd still have to install the package to avoid NPM warnings on install, no?

PR #869, opened 2021-03-10
style(button spinner): reduce the size of the new spinner in buttons

domarku:
# Purpose of PR

The current size of the spinner in buttons is not related to any value in the button — padding, label or line-height.
I suggest we reduce the size so that it sits on the baseline of the label and is closer in size to an ellipsis.

**Current**
1-current

**Proposed change**
2-suggested

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Button/Button.tsx
+++ b/packages/forma-36-react-components/src/components/Button/Button.tsx
@@ -137,7 +137,7 @@ export const Button = ({
         >
           <Spinner
             className={styles.Button__spinner}
-            customSize={18}
+            customSize={12}
             color={
               buttonType === 'muted' ||
               buttonType === 'warning' ||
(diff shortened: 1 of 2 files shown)


denkristoffer:
How does this scale with different button sizes?

domarku:
@denkristoffer good question. I didn't take button sizes into account. Also, not sure how I would solve it 😳

denkristoffer:
@domarku I'm sure we can map the different button sizes to different numbers if we want to. I'd be happy to help you land that 👍

gui-santos (approved):
I think we can scale to other sizes of the button in another PR

gui-santos:
@domarku the PR was getting a green check to merge because it had some formatting issues
So I did it and pushed, let's hope it passes now 🙏

PR #877, opened 2021-03-15
feat: move Flex and Grid to `f36-core`

suevalov:
# Purpose of PR

* Moves Flex and Grid to `f36-core` package
* Uses emotion instead of CSS Modules
* No API changes => there will be a separate PR for that


mshaaban0 (commented):
🤔 hmm, isn't `f36-core` sufficient enough?

suevalov:
@mshaaban0 

as for me, it makes sense to collocate all layout component similar to what Chakra does:

Screenshot 2021-03-15 at 15 39 12

denkristoffer:
I'd say we shouldn't presuppose what kind of package grouping makes sense, I'd rather we come up with a proposal where all packages are listed and we can have a discussion.

suevalov:
@denkristoffer ok, I can move then to `core` package

denkristoffer (approved):
I'd say we should just extract to `@contentful/f36-flex` and `@contentful/f36-grid` then figure out where to go from there. Changes look good 🙏

suevalov:
@denkristoffer I would disagree with this suggestion, I'd rather merge it as an initial proposal: `f36-layout`. Creating a temporary package for each component is too much effort.

mshaaban0 (approved):
🙏

denkristoffer:
I expected `f36-layout` or whichever to just point to `f36-flex` etc.

suevalov:
@denkristoffer my point is to not create packages per component but group them logically without having packages that just re-export other packages

PR #878, opened 2021-03-16
feat: Box component

suevalov:
# Purpose of PR

* Introduces a Box component - a primitive with an ability to apply spacing properties: margins and padding
* Refactor Flex, Grid, GridItem to use Box component and extends their API to accept margin and padding properties

P.S. No breaking changes for Flex, Grid, GridItem components, it just makes them more powerful by adding margin/padding props


nikazawila:
thanks for your work @suevalov 🌻 

I am not sure if I'm a fan of margins becoming one/two consonant props. To me, they seem super vague and non-descriptive. Same with values - I'd even argue we should maybe align those across all components as well, so that we avoid sizes sometimes being `m` and sometimes `spacingM` or other... wdyt?

suevalov:
@nikazawila Basically I was inspired by https://chakra-ui.com/docs/features/style-props#margin-and-padding that its widely used in React community and also by Box component was introduced in `experience-packages` and used in new applications built by Contentful.

Ideally, I would remove dual values `m`, `spacingM` and leave only a short version, but it will be a breaking change for Flex component.

denkristoffer:
I think I've seen most design systems with a similar layout component go with the shorthand prop names so I don't think they will seem too unknown to most React developers. Personally I much prefer longer, more explicit names.

burakukula:
I would also agree that the longer version of the prop is more descriptive and it would be my preference here. When it comes to values themselves like `spacingM` - it is reflecting spacing namespaces that we use in tokens. I think we might want to keep it consistent and use the same system in both places.

suevalov:
Well, I guess I'm not the biggest fan of long property names but I will agree with the majority here.

I'll remove the shortcut versions of props and values.

suevalov:
@nikazawila @denkristoffer @burakukula update according to your feedback

packages/components/badge/stories/Badge.stories.tsx
```diff
@@ -1,15 +1,15 @@
 import React from 'react';
 import { Flex } from '@contentful/f36-core';
 
-import { Badge } from '../src/Badge';
+import { Badge, _Badge } from '../src/Badge';
```
denkristoffer, line 4: Why is this necessary if it's the same component?

packages/core/src/types.ts
```diff
@@ -13,3 +13,51 @@ export type CommonProps = {
   /** Styles */
   style?: CSSProperties;
 };
+
+export type Spacing =
```
denkristoffer, line 17: I think we should take a general decision on whether or not we want to use enums for cases like this or if we only want to support strings. If we want to do enums the value should also be able to be passed as e.g. `Spacing.2Xs`

packages/components/badge/stories/Badge.stories.tsx
suevalov: Apparently, `__docgenInfo` is not parsed properly if a component is wrapped in `forwardRef` 🤕 

As an alternative, we can do `export const BadgeDocgenInfo = _Badge['__docgenInfo']` from Badge.tsx

packages/core/src/types.ts
suevalov: We didn't use enums before and I'm not sure if we want to start this. Almost all popular React component libraries are using string literals and people don't have problems with that.

denkristoffer: It'd probably be mostly for our internal use. I don't think it should be either/or, but an addition. I like how well enums work as a user of a library but I'm fine with strings too.

packages/components/badge/stories/Badge.stories.tsx
denkristoffer: I see. I like the alternative as it's clearer why it's done 🙂

suevalov: An alternative approach didn't work :(

denkristoffer (commented):
You forgot to change to back to `_Badge`.

suevalov:
@denkristoffer Brought it back.

denkristoffer (approved):

nikazawila (approved):
thanks for aligning the props
looks good to me ✨

packages/components/badge/stories/Badge.stories.tsx
gui-santos: Oh I'm too late for this

I know you've merged this already, but I think we don't need that
```
parameters: {
    propTypes: [Badge['__docgenInfo']],
  },
```
It was used for generating the propTypes table back when we were not using storybook docs

Now that we use docs, it generates a prop table in the Docs table automagically without that parameter. Look:
Screenshot 2021-03-17 at 10 49 10

It even gets JSDoc comments as descriptions of the props ☝️ 

Long story short, I think we can remove the docgenInfo thing for every new component and then we don't need the _Badge thing 🙌

PR #745, opened 2021-01-06
feat: Forma V4

suevalov:
# Purpose of PR

This branch is an umbrella for all V4 related changes

PR #879, opened 2021-03-16
feat: icons packages

denkristoffer:
# Purpose of PR
This PR splits our current icons into two packages, `@contentful/f36-icon` and `@contentful/f36-icons`. I had to organise it like this because of the way Parcel v2 currently bundles its types.

## @contentful/f36-icons

This package is the one that will be consumed by most users. It contains all our previous icons. It's used like this:

```tsx
import { ArrowDown } from '@contentful/f36-icons'

export function Component(props) {
  return <ArrowDown variant="secondary" {...props} />
}
```

## @contentful/f36-icon

This package contains the building blocks for icons and lets you use custom icons with your own SVG or with an icon from external libraries like Material. It's used by `@contentful/f36-icons`. It has a few parts to it. The simplest is the `generateIcon` helper which does the heavy lifting for you if you give it an SVG path:

```tsx
import React, { Fragment } from 'react';
import { generateIcon } from '@contentful/f36-icon';

export const CustomIcon = generateIcon({
  name: 'CustomIcon',
  path: (
    <Fragment>
      <path d="M0 0h16.5v18H0V0z" fill="none" />
      <path d="M10.5 2.2H3.8c-.6 0-1.2.4-1.4.9L.1 8.5c-.1.1-.1.3-.1.5v1.5c0 .8.7 1.5 1.5 1.5h4.7l-.7 3.4v.2c0 .3.1.6.3.8l.8.8 4.9-4.9c.3-.3.4-.6.4-1.1V3.8c.1-.9-.6-1.6-1.4-1.6zm3 0v9h3v-9h-3z" />
    </Fragment>
  ),
  viewBox: '0 0 16.5 18',
});
```

This is really just a wrapper around an `<Icon />` component to help us avoid boilerplate. `<Icon />` is used like this:

```tsx
import React from 'react';
import { Icon } from '@contentful/f36-icon';

export const CustomIcon = (props) => {
  return (
    <Icon {...props} viewBox="0 0 16.5 18">
      <path d="M0 0h16.5v18H0V0z" fill="none" />
      <path d="M10.5 2.2H3.8c-.6 0-1.2.4-1.4.9L.1 8.5c-.1.1-.1.3-.1.5v1.5c0 .8.7 1.5 1.5 1.5h4.7l-.7 3.4v.2c0 .3.1.6.3.8l.8.8 4.9-4.9c.3-.3.4-.6.4-1.1V3.8c.1-.9-.6-1.6-1.4-1.6zm3 0v9h3v-9h-3z" />
    </Icon>
  )
}
```

`<Icon />` is also the component to use with third-party libraries:

```tsx
import React, { Fragment } from 'react';
import { Icon } from '@contentful/f36-icon';
import { MdAcUnit } from 'react-icons/md'

export const CustomIcon = (props) => {
  return <Icon {...props} as={MdAcUnit} />
}
```

## Existing components

Existing components no longer take strings as icon arguments. Instead you import and pass the icon itself. As an example, here's `<EditorToolbarButton />`:

```tsx
import React from 'react';
import { HeadingOne } from '@contentful/f36-icons';
import { EditorToolbarButton } from '@contentful/forma-36-react-components';

export function Component() {
  return <EditorToolbarButton icon={HeadingOne} label="H1" />,
}
```

## To do in a follow-up PR

- [ ] Rename all icon props to `icon`. Maybe we can extract this further and make some kind of helper that makes it easy to accept an icon in components. Then we can use it to merge `Button` and `IconButton`.
- [ ] Rename icon sizes. `tiny` to `xsmall`?


packages/components/icon/src/types.ts
```diff
@@ -0,0 +1,18 @@
...
+    color?:
+      | 'muted'
+      | 'negative'
+      | 'positive'
+      | 'primary'
+      | 'secondary'
+      | 'warning'
+      | 'white';
+    icon: string;
+    size?: 'small' | 'large' | 'medium' | 'tiny';
```
burakukula, line 16: @denkristoffer Just a call out for later, not really related to this PR. 
Something to be clarified with designers, I guess, is `size` and `color.`
`size` - I think we might need to agree about the sizes that would be shared (maybe partially with IconButton / Button - @mshaaban0 made a proposal about merging those 2)
`color` (maybe it should be `variant`?) - we might want to deprecate muted here as well.

burakukula:
@denkristoffer thx a lot for this proposal! 🙂
Looking good from my perspective. I'm wondering if we could still have an option to import the whole set of icons?

packages/components/icon/src/types.ts
denkristoffer: Yes, good point thanks, we should go with `variant`.

denkristoffer:
@burakukula You install with `npm i @contentful/f36-icon` and use it like the example in OP.

burakukula:
@denkristoffer sure, I just recall that argument when we talked about the icons package. 
oh, no more note: Just noticed that - I guess that package could be called `@contentful/f36-icons` not `@contentful/f36-icon` since we would have all the icons there?

denkristoffer:
In my opinion it's not a problem that icons are bundled in the one package as long as code splitting is possible.

suevalov:
Great proposal and a good start!

I'd suggest looking into https://chakra-ui.com/docs/media-and-icons/icon design and implementation-wise:

* so we don't have boilerplate with our icons
* can allow people using their custom icons by:

```jsx
<Icon viewBox="0 0 200 200">
  <path
    fill="currentColor"
    d="M 100, 100 m -75, 0 a 75,75 0 1,0 150,0 a 75,75 0 1,0 -150,0"
  />
</Icon>
```

```jsx
<Icon as={MdReceipt} />
```

denkristoffer:
@suevalov Nice, that looks like a good solution to building a bunch of similar components, I will work on that.

suevalov:
@denkristoffer please rebase with the latest changes in `forma-v4`, it would be nice to use `@emotion/core` and CSS prop for the styles to be consistent with other components

denkristoffer:
@suevalov emotion's CSS prop requires changing the JSX transform from `React.createElement` to emotion's, we don't do that anywhere else, do we?

packages/components/icons/src/Icon.tsx
```diff
@@ -0,0 +1,116 @@
...
+  viewBox?: `${number} ${number} ${number} ${number}`;
+} & AsOrChildren;
+
+export type IconProps = Simplify<
+  IconInternalProps & CommonProps & React.SVGAttributes<SVGSVGElement>
+>;
+
+export const Icon = forwardRef<SVGSVGElement, IconProps>(function Icon(
+  {
+    as,
```
denkristoffer, line 76: Will it be too confusing if we call this prop `as` even though it works differently than other `as` props? It doesn't take a string like `"div"`, only a component.

suevalov: I don't think it's confusing, it's merely a narrower subset of the previous type. The same will be with `Heading` component, for example, not all original values of `as` will be excepted, but only `h1`, `h2`, etc...

packages/forma-36-react-components/src/index.ts
```diff
@@ -10,6 +10,7 @@ export type {
   PaddingProps,
   Spacing,
 } from '@contentful/f36-core';
+// export * from '@contentful/f36-icons';
```
denkristoffer, line 13: Do we want to export from here? Some of the icon names clashes with component names, and I think it might be better to force people to use `import { ... } from '@contentful/f36-icons'`.

suevalov: Yeah, I'd allow importing from '@contentful/f36-icons' directly.

We should probably append `Icon` to each icon component, e.g. `Heading` => `HeadingIcon`

packages/components/icon/src/Icon.tsx
```diff
@@ -0,0 +1,142 @@
...
+type IconInternalProps = {
+  /**
+   * The name prop is used to check if "trimmed" styling should be applied,
+   * which sets the `width` to `auto`.
+   */
+  name?: string;
+  /**
+   * Determines the size of the icon
+   */
+  size?: 'large' | 'medium' | 'small' | 'tiny';
```
suevalov, line 40: Can you create `IconSize` type here?

packages/components/icon/src/Icon.tsx
```diff
@@ -0,0 +1,142 @@
...
+  /**
+   * Determines the fill color used
+   */
+  variant?: IconVariant;
+  /**
+   * SVG viewBox attribute
+   */
+  viewBox?: `${number} ${number} ${number} ${number}`;
```
suevalov, line 48: this type won't work for Typescript versions before 4.2, in this particular case I'd simply fallback to `viewBox?: string`

packages/components/icon/src/Icon.tsx
```diff
@@ -0,0 +1,142 @@
...
+      as?: never;
+      children: ReactElement | ReactElement[];
+    };
+
+type IconInternalProps = {
+  /**
+   * The name prop is used to check if "trimmed" styling should be applied,
+   * which sets the `width` to `auto`.
+   */
+  name?: string;
```
suevalov, line 36: Can we just pass `width: auto` to all icons that are have to trimmed?

I think checking if name starts with "trimmed" isn't a good API for this.

packages/components/icon/src/Icon.tsx
```diff
@@ -0,0 +1,142 @@
...
+   * Determines the fill color used
+   */
+  variant?: IconVariant;
+  /**
+   * SVG viewBox attribute
+   */
+  viewBox?: `${number} ${number} ${number} ${number}`;
+} & AsOrChildren;
+
+export type IconProps = IconInternalProps &
```
suevalov, line 51: I'd expect this component have MarginProps as Box component

packages/components/icon/src/Icon.tsx
```diff
@@ -0,0 +1,142 @@
...
+    name,
+    variant = 'primary',
+    size = 'small',
+    testId = 'cf-ui-icon',
+    viewBox = '0 0 24 24',
+    ...otherProps
+  },
+  forwardedRef,
+) {
+  const style = css`
```
suevalov, line 113: I think we should have a function that accepts all icon props and that returns a css class that needs to be applied.

`css({})` has to used only once so we don't create many rules beforehand and styles is as performant as it can be

packages/components/icon/stories/Icon.stories.tsx
```diff
@@ -0,0 +1,136 @@
...
+  </Icon>
+);
+
+export const Overview: Story = () => {
+  const sizes = ['large', 'medium', 'small', 'tiny'];
+
+  return (
+    <Fragment>
+      <Flex marginBottom="spacingS">
+        <SectionHeading as="h3">Icon sizes</SectionHeading>
```
suevalov, line 67: All `Heading`componens should have margin properties now  so you don't need to wrap them in Flex to apply margin

packages/components/icons/src/index.ts
```diff
@@ -0,0 +1,141 @@
+export type { IconComponent } from '@contentful/f36-icon';
+export { ArrowDown } from './ArrowDown';
```
suevalov, line 2: I would write a script to generate this file automatically

packages/core/src/types.ts
```diff
@@ -90,3 +90,5 @@ export type PickUnion<UnionType, Keys> = Exclude<
   UnionType,
   Exclude<UnionType, Keys>
 >;
+
+export type Simplify<T> = { [KeyType in keyof T]: T[KeyType] };
```
suevalov, line 94: Naming here is a bit confusing. Simplify what?

packages/components/icon/src/Icon.tsx
```diff
@@ -0,0 +1,142 @@
...
+import type {
+  ComponentType,
+  ExoticComponent,
+  ReactElement,
+  SVGAttributes,
+} from 'react';
+
+export type IconComponent = ExoticComponent<any> | ComponentType<any>;
+
+type IconVariant = Simplify<ComponentVariant | 'muted' | 'white'>;
```
suevalov, line 19: I really don't like hiding actual variant values with this shared types. Can't we just list all options as normal union?

packages/forma-36-react-components/src/components/Button/Button.stories.tsx
```diff
@@ -12,7 +13,7 @@ export default {
     propTypes: [Button['__docgenInfo']],
   },
   argTypes: {
-    icon: { control: { type: 'select', options: iconName } },
+    icon: { control: { disable: true } },
```
gui-santos, line 16: Why?
If you disable this control, people won't be able to try different icons in the button

packages/forma-36-react-components/src/components/Card/EntryCard/EntryCard.tsx
```diff
@@ -207,15 +206,12 @@ export function EntryCard({
...
-                    color="muted"
+                {StatusIcon ? (
+                  <StatusIcon
                     className={styles['EntryCard__icon']}
+                    variant="muted"
                   />
-                ) : (
-                  statusIcon
-                )}
+                ) : null}
```
gui-santos, line 214: Is there a reason to return null? 

I think you could do this without returning null when StatusIcon is undefined, no?

```tsx
{StatusIcon && (
  <StatusIcon
    className={styles['EntryCard__icon']}
    variant="muted"
  />
)}
```

gui-santos (changes requested):
this looks great!

I just got two questions and I support some of Alex's points 
😄

packages/components/icon/src/Icon.tsx
denkristoffer: You mean like a static list of which icons we want to trim? Most icons with a trimmed width also have a non-trimmed version.

packages/forma-36-react-components/src/components/Button/Button.stories.tsx
denkristoffer: Kind-of-but-not-really a stopgap. There's no easy way to import all icons and list them without having to update the list manually when new icons are added.

packages/components/icon/src/Icon.tsx
denkristoffer: Personally I like the certainty we get from extending the shared type, that all variants across components are the same and that we don't have to update them in multiple places. The `Simplify` type helps write out all the variants in your editor.

packages/components/icon/src/Icon.tsx
denkristoffer: As far as I know we're requiring TS v4+ for our own v4. Also, there should already be a string fallback added by the compiler itself, at least that how the type works for me.

packages/forma-36-react-components/src/components/Button/Button.stories.tsx
gui-santos: Ahhhh I see.... ok it makes sense then

packages/components/icon/src/Icon.tsx
suevalov: I think we should aim for Typescript@3.8+ for our users. And this Template syntax won't work with 4.0, 4.1, so I would rather be safe here

packages/components/icon/src/Icon.tsx
suevalov: You can pass `trimmed: true` in icon files (where `generateIcon` is used).

packages/components/icon/src/Icon.tsx
suevalov: I don't think we'll have the same list of variants for all components, they will always differ because they're contextual and I'd personally prefer to have the whole list of all options in the corresponding component file.

+ I'm not sure if <Props /> table and Storybook table will be able to parse this and show all possible options to the end user.

packages/components/icon/src/Icon.tsx
denkristoffer: It's only the 4.0 upgrade that has breaking changes. In any case, if we target 3.8 I think we should use that ourselves.

packages/components/icon/src/Icon.tsx
denkristoffer: It works with both of them without issues.

packages/core/src/types.ts
denkristoffer: It writes out the "end result" of combined types

```tsx
type ExternalUnion = 'a' | 'b'

type Local = Simplify<ExternalUnion | 'x' | 'y' | 'z'>

// Local #=> 'a' | 'b' | 'x' | 'y' | 'z'
```

Just taken from https://github.com/sindresorhus/type-fest/blob/main/source/simplify.d.ts

packages/forma-36-react-components/src/components/Card/EntryCard/EntryCard.tsx
denkristoffer: No you're right, it's the same thing. I was trying an alternative at one point and needed the ternary and never changed it back, but it should be the same now 🙂

packages/components/icon/src/Icon.tsx
denkristoffer: Shouldn't we add this to `<Primitive />` then? Are there components that shouldn't take those props?

packages/components/icon/src/generateIcon.tsx
```diff
@@ -0,0 +1,64 @@
+import React, { forwardRef } from 'react';
+import type { ReactElement } from 'react';
+
+import { Icon } from './Icon';
+import type { IconProps } from './Icon';
+
+export type GeneratedIconProps = Omit<
```
denkristoffer, line 7: Can we come up with a more descriptive name here? `BuiltInIconProps`?

packages/components/icon/src/Icon.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+const fills: { [key in IconVariant]: string } = {
+  muted: tokens.colorTextLight,
+  negative: tokens.colorNegative,
+  positive: tokens.colorPositive,
+  primary: tokens.colorBlueBase,
+  secondary: tokens.colorContrastMid,
+  warning: tokens.colorWarning,
+  white: tokens.colorWhite,
+};
```
gui-santos, line 53: Couldn't this be an enum? 🤔

gui-santos (approved):
🍰

packages/core/src/index.ts
```diff
@@ -6,7 +6,9 @@ export type {
   MarginProps,
   PaddingProps,
   PickUnion,
+  Simplify,
 } from './types';
 export * from './Flex';
 export * from './Grid';
 export * from './Box';
+export { getSpacingStyles } from './utils/getSpacingStyles';
```
suevalov, line 14: I guess that we can use the Box component in Icon directly, so we don't need to export this utility function.

packages/core/src/types.ts
suevalov: Ok if it is an original name from type-fest

packages/forma-36-react-components/src/components/TextLink/TextLink.tsx
```diff
@@ -54,8 +54,9 @@ export function TextLink({
   ...otherProps
 }: TextLinkProps): React.ReactElement {
   const renderIcon = useCallback(
-    (icon: IconType, linkType: TextLinkType) => {
-      if (!icon) return undefined;
+    (Component: TextLinkProps['icon'], linkType: TextLinkType) => {
+      console.log('Component', Component);
```
suevalov, line 58: Please, delete `console.log`

packages/forma-36-react-components/src/index.ts
```diff
@@ -30,7 +30,6 @@ export * from './components/EntityList';
 export * from './components/Form';
 export * from './components/FormLabel';
 export * from './components/HelpText';
-export * from './components/Icon';
```
suevalov, line 33: I guess that we need to add `export { Icon } from 'f36-icon';`

packages/forma-36-website/src/components/ComponentSource.js
```diff
@@ -38,7 +40,7 @@ class ComponentSource extends React.Component {
       <div className="f36-margin-bottom--m">
         <LiveProvider
           code={this.props.children.trim()}
-          scope={{ ...f36Components, tokens }}
+          scope={{ ...f36Components, tokens, ...f36icons, Icon }}
```
suevalov, line 43: I think we need to do `scope={{  tokens, ...f36icons, ...f36Components, Icon }}`, `Heading` is a Forma36 component and not an icon

packages/components/icon/src/Icon.tsx
```diff
@@ -0,0 +1,135 @@
...
+    trimmed,
+    viewBox = '0 0 24 24',
+    ...otherProps
+  },
+  forwardedRef,
+) {
+  const shared = {
+    className: cx(
+      css({
+        ...getSpacingStyles(otherProps),
```
suevalov, line 107: I'd simply use `Box` here instead of `Primitive`

packages/components/icon/src/Icon.tsx
```diff
@@ -0,0 +1,135 @@
...
+  viewBox?: SVGAttributes<SVGSVGElement>['viewBox'];
+} & AsOrChildren;
+
+export type IconProps = IconInternalProps &
+  CommonProps &
+  MarginProps &
+  PaddingProps &
+  SVGAttributes<SVGSVGElement>;
+
+export const Icon = forwardRef<SVGSVGElement, IconProps>(function Icon(
```
suevalov, line 90: ❗ I guess the last major problem to solve is to fix polymorphic typings problem:


```jsx
function MyCustomIcon(props: { foo: string }) {
   return <svg>...</svg>
}

<Icon as={MyCustomIcon} foo="bar" /> 
// this should be valid
// because actually all ...rest props are passed in <MyCustomIcon {...rest} />

```

suevalov (commented):
Amazing work 🎖️ 

Just one more major thing to improve (marked with ❗), all other comments are rather comments or small suggestions.

packages/components/icons/scripts/preBuild.js
```diff
@@ -0,0 +1,44 @@
+const path = require('path');
```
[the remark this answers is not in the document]
suevalov: thanks for writing the script

packages/components/icons/README.mdx
```diff
@@ -0,0 +1,254 @@
+---
+title: 'Icons'
```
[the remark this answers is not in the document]
denkristoffer: That'd be good, but what about the different GitHub and Storybook links etc.?

packages/components/icon/src/Icon.tsx
```diff
@@ -0,0 +1,135 @@
...
+  },
+  medium: {
+    height: '24px',
+    width: '24px',
+  },
+  small: {
+    height: '18px',
+    width: '18px',
+  },
+  tiny: {
```
burakukula, line 39: @denkristoffer I think as a follow-up, we could re-visit the sizes for Icon component, I think we should consider `xsmall` namespace instead of tiny. WDYT? (not a blocker, just a thought)

packages/components/icon/stories/Icon.stories.tsx
```diff
@@ -0,0 +1,132 @@
...
+  title: 'Components/Icon',
+} as Meta;
+
+export const WithSVGPath: Story<IconProps & { as?: never }> = (args) => (
+  <Fragment>
+    <SectionHeading as="h3" marginBottom="spacingS">
+      Icon component with SVG paths
+    </SectionHeading>
+
+    <Icon {...args}>
```
burakukula, line 30: @denkristoffer I imagine it's a bit late in the process, but I am wondering about the usage of custom icons. It seems more intuitive to me to use a custom icon somehow like this in the component `<Icon as={MyCustomIcon} size="small" color"whatever" />`, not really passing paths as children. 🤔

packages/components/icons/README.mdx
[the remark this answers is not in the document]
suevalov: Maybe we could merge storybook too, and link one readme from another?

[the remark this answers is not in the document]
suevalov: Maybe we could merge storybook too, and link one readme from another?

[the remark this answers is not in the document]
suevalov: Maybe we could merge storybook too, and link one readme from another?

packages/components/icon/stories/Icon.stories.tsx
burakukula: Right now I would need to copy and paste paths inside the `<Icon></Icon>` as children. Am I right?

denkristoffer: You can do it both ways 😄

burakukula: @denkristoffer ok, then I would extend the REAMDE file to clearly show how custom icons might be used both ways. I must say, that this flow seems not very comfortable to me - I need to open my custom SVG file, copy and paste paths inside the <Icon> component. As a user, I would probably just do my custom SVG instead.🤷‍♀️

denkristoffer: What kind of flow do you think users would like? Can we make it more comfortable?

packages/components/icon/src/Icon.tsx
denkristoffer: I think that's a good idea. The old discussion seemed unfinished to me, so I would say we leave it for a follow-up PR.

packages/components/icon/src/Icon.tsx
denkristoffer: It think it could but do you think that would make it easier to use? This is for internal use.

packages/components/icons/README.mdx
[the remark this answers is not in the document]
denkristoffer: On second thought I think this might confuse users because the two package names are so similar, `icon` and `icons`. If everything is in the same README we might mike it harder to understand what to use when.

packages/components/icon/stories/Icon.stories.tsx
burakukula: I am assuming here. For me it would be something like:

```
import myCustomIcon from 'custom-icon.svg'

<Icon as={myCustomIcon} size="small" />

or
<Icon size="small">{myCustomIcon}</Icon>

```
To me, it's a bit odd to copy and paste paths inside the `<Icon>` component. But maybe it's just me. If you can do it both ways, that's great. Let's point it out in the README - I mean, info about `as={}` option is there, but for me, it seems like we recommend using paths when you use a custom icon. It's just wording or my interpretation of it. 😁  It might be just me.
It's not a blocker. 🙂

packages/components/icon/src/Icon.tsx
gui-santos: I guess it's a tomAYto/tomAHto situation 👌 

Nevermind then

burakukula (approved):
🏆, thanks a lot for doing this!
NOTE: The custom icon topic is not a blocker from my perspective.
If there are 2 ways of doing custom icons, we might want to show them clearly in the docs (both README and storybook). Also maybe it's worth pointing out the benefits, like consistency when using our `<Icon>` component as a wrapper for `<path />`'s

packages/components/icon/src/Icon.tsx
denkristoffer: I like enums but I dropped them because we're not using them anywhere else. I don't want to rock the boat! 😅

packages/components/icon/src/Icon.tsx
mshaaban0: @denkristoffer This is aligned with design now, can we please have these [sizes](https://www.figma.com/file/mw7H4EffGQAR3FXnUu6HQS/Outdated-Campaigns-UI-Kit?node-id=1281%3A711) in place? should be mapped with spacing values as well

packages/components/icon/src/Icon.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+type IconInternalProps = {
+  /**
+   * Determines the size of the icon
+   */
+  size?: IconSize;
+  /**
+   * Whether or not to trim the icon width, i.e. set `width` to `auto`
+   */
+  trimmed?: boolean;
```
mshaaban0, line 79: Wouldn't that introduce sizes inconsistencies?

packages/components/icon/src/generateIcon.tsx
mshaaban0, line 7: Should this be `IconInternalProps` same pattern as https://github.com/contentful/forma-36/pull/745/files#diff-bbad09ab711a522cbdab27581c9bb3f575da40589810fbeab1754b2bd320d2cfR25 ?

mshaaban0 (commented):
Looks good, just a couple of comments and should be good to go 🌮

packages/components/icon/src/Icon.tsx
denkristoffer: Sure, but trimmed icons already have that.

mshaaban0: Yes, I mean should we remove it then with the new sizes, and V4?

packages/components/icon/src/generateIcon.tsx
denkristoffer: We want all the other stuff that gets added to `IconProps` like `className` and `testId`.

The difference between this case and the linked file is that the linked component extends the internal props and then uses `PolymorphicComponentProps` on its own props to add the shared props, , but in this case we're not dealing with a component with props, so I would argue that it's right to `Omit` types we don't allow instead of using `PolymorphicComponentProps`.

packages/components/icon/src/Icon.tsx
denkristoffer: Should we rename to follow the token names then? But then the smallest icon size would be called "medium" 🤔😕

denkristoffer: Nevermind the tokens wouldn't work until we introduce one for 20px. I'll just hardcode the sizes for now.

packages/components/icon/src/Icon.tsx
denkristoffer: I can't say, we should probably run it by design. I don't know where the trimmed icons are used or how much.

packages/components/icon/src/Icon.tsx
denkristoffer: @mshaaban0 I'd like to push the size changes to a later PR so that these implementation changes don't get tangled up with changes to tokens and style.

suevalov (approved):

mshaaban0 (approved):

PR #881, opened 2021-03-16
feat: Stack component

suevalov:
# Purpose of PR

The latest addition to a layout family in Forma V4 - Stack component.

It's very similar to https://chakra-ui.com/docs/layout/stack and allows to stack components vertically or horizontally with a spacing gap between them.

Internally, it's just a wrapper on top of `Grid` component that hides all complexity of Grid properties and provides a simple way to control the behavior with only two properties: `direction` and `gap`

Depends on https://github.com/contentful/forma-36/pull/878

### DIFF
--- a/packages/core/src/Stack/Stack.tsx
+++ b/packages/core/src/Stack/Stack.tsx
@@ -1,0 +1,81 @@
+import * as React from 'react';
+import {
+  PolymorphicComponentProps,
+  PolymorphicComponent,
+} from '../Primitive/Primitive';
+import { Grid } from '../Grid';
+import type { GridInternalProps } from '../Grid/Grid';
+import type { MarginProps, PaddingProps, Spacing } from '../types';
+
+function getValidChildren(children: React.ReactNode) {
+  return React.Children.toArray(children).filter((child) =>
+    React.isValidElement(child),
+  ) as React.ReactElement[];
+}
+export interface StackInternalProps extends MarginProps, PaddingProps {
+  /**
+   * Spacing between elements
+   * @default spacingM
+   */
+  gap?: Spacing;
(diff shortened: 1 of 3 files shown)


mshaaban0 (commented):
I think this component is basically Flex, per [spec](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/Relationship_of_Grid_Layout) `Flex` is designed to build one dimensional layouts (vertical **or** horizontal), and `Grid` is two dimensional (vertical **and** horizontal)

suevalov:
@mshaaban0 do you think it doesn't make sense to have this component or just to refactor it using `Flex` (Chakra is using Flex)?

I think a big difference here is that it creates a consistent `spacing` between elements and it's really convenient to wrap any components in Stack and have nice dividers between them.

mshaaban0:
@suevalov I don't think we should have a special component for controlling space between elements, maybe introducing `gap` property on Flex to control spacing between Boxes?

suevalov:
It's not super trivial to do just using Flex: https://github.com/chakra-ui/chakra-ui/blob/main/packages/layout/src/stack.tsx

I'd rather close this PR and not blow up `Flex` component

denkristoffer:
I was actually talking to @mshaaban0 about creating an abstraction for flexbox last week! I personally think it can be a good idea, but I can see Moe's argument that it might not be necessary. We agreed to take the temperature by asking the other devs in a Slack poll before moving ahead with anything, maybe that's still the way to go here.

suevalov:
I can talk about this during today's Forma Chapter.

denkristoffer:
We forgot to talk about this!

PR #884, opened 2021-03-17
docs: simplify documentation website

suevalov:
# Purpose of PR

Simplifies website documentation structure by collocating metainformation and components documentation.

There's no need to page a corresponding `website/src/pages/components/[component-name]` anymore.

One potential improvement: 

* Move all other `*.mdx` files from `packages/forma-36-website/src/content/*` to `docs/*`


denkristoffer (approved):

packages/forma-36-website/gatsby-config.js
```diff
@@ -140,7 +140,7 @@ module.exports = {
             link: '/components/card/',
           },
           {
-            name: 'Copy button',
+            name: 'CopyButton',
```
burakukula, line 143: @suevalov this is how items in the navigation would be called right as far as I remember?

suevalov: Yep, I find it more logical to have component names as they are in the navigation.

burakukula: We had a discussion about naming items in the navigation in the past. The main reason for having 'Copy button' here was that we assume that the audience of the website is not only engineers and it would be better to have navigation items written in a standard way.

burakukula: In the future, we might have more guidelines and docs for designers on how to use components for example.

suevalov: I can revert this change but I find it very confusing and I haven't seen a simple component library website doing the same.

burakukula:
Hey Alex! Awesome work! thx a lot for doing that! 🙌

One minor thing is how the README file look like in GitHub https://github.com/contentful/forma-36/tree/simplify-documentation-website/packages/forma-36-react-components/src/components/Autocomplete

But I think it's maybe not an issue. The benefits are bigger in my opinion, so maybe we should not worry about it.
When it comes to moving all other `.mdx` files, let's maybe talk about it in the team. Maybe it would be beneficial to move this content to... Contentful? 😁

packages/forma-36-website/gatsby-config.js
burakukula: hmm, I think many websites of the design systems are actually doing that - Polaris, Carbon, Primer, and probably many more. I would maybe not think about the website as a documentation of react components library, more of a documentation of the design system (which is small today, but it will be bigger) 🙂

suevalov:
@burakukula yeah, I will not be moving them right now. just an idea for a discussion later.

packages/forma-36-website/gatsby-config.js
suevalov: I check all the examples (Polaris, Carbon, Primer) and all of them are using this notation, true, but it's important to mention that all of them are not React component libraries: they all have CSS version, React, some of them Angular and Android versions. I would assume that is the only reason why they use it this way.

We have only React component library and I would stick to component names in the navigation, like all other React component libraries do:

* https://chakra-ui.com/docs/layout/aspect-ratio
* https://react-spectrum.adobe.com/react-spectrum/index.html
* https://ant.design/components/overview/
* https://react.semantic-ui.com/

burakukula: Sure, I just still think the website is not only the documentation of react components. But of course, I understand your point that this is inconsistent with how the components are called in the library. I just find a standard way of naming navigation and titles more friendly for other audiences than engineers.
Actually, Chakra example is the one where they have it inconsistent - https://chakra-ui.com/docs/layout/aspect-ratio - "Aspect Ratio" in the nav and in the title and `import { AspectRatio } from "@chakra-ui/react"` in the code example. And they have more examples of those in their website, like:
https://chakra-ui.com/docs/feedback/circular-progress
https://chakra-ui.com/docs/components/close-button
So they have it done both ways for some reason. Maybe they have the same discussion like we do 😄

suevalov: > titles more friendly for other audiences than engineers

but it makes it harder to find a component for engineers, I usually use fullsearch text on a website by using `Cmd+F` and it can't find `TextInput` component on our website right now.

I guess the main audience for React components documentations are, undoubtedly, engineers (mainly App Framework developers) and we should optimize for this.

nikazawila: I would agree with @suevalov on this one and not because others are doing it, but because that is the name of the component that you want to document, so it's worth to get the non-tech folks used to such nomenclature since, ideally, they will later be working with it and nowhere in Figma or code will they ever encounter a "Copy button". We can do the job of catering to less techy folk inside the actual documentation of the component, but names should, in my opinion, stay same.
Happy to hear other points of view and be convinced otherwise though.

nikazawila (approved):
this looks great @suevalov 
I'd wait for whatever comes out of the discussion, but otherwise great job 🚀

packages/forma-36-website/gatsby-config.js
burakukula: @nikazawila @suevalov both very good points - search on a website and shared naming convention with other users of the system. I will, of course, agree with majority here 🙂. it would be great to maybe hearing how does it look like from the designer's point of view.

gui-santos: I think the examples mentioned here are more react component libraries
If you see the documentation of whole design systems like:
https://www.carbondesignsystem.com/components/date-picker/usage/
https://material.io/components/data-tables
you will see that they use language for non-engineers as well

I agree with @burakukula , I always thought the website was documentation for a broader audience
But there was not a lot of work in creating documentation for designers for example

Maybe it's a good topic to be decided async in slack... idk

gui-santos (approved):

packages/forma-36-website/gatsby-config.js
burakukula: Yeah, seems like a good idea. I don't want to be a blocker for this PR, we can come back to this topic at some point. I added @domarku to the PR, maybe he would have some insights around this topic as well. 🙂

burakukula (approved):

packages/forma-36-website/gatsby-config.js
domarku: I prefer calling things in human language wherever possible but it seems like there's a bigger benefit in naming components consistently across the codebase, the website and in the Figma UI Kit. Everyone can then refer to them with a lesser risk of miscommunication.

PR #887, opened 2021-03-22
docs: Box & Flex on the documentation website

suevalov:
# Purpose of PR

Adds documentation about Box and Flex components to a documentation website.

* docs: improve documentation website styles, adds ability to show code blocks without a playground
* feat: adds new properties to Flex component to map other important flexbox CSS properties

### Breaking changes

* Rename properties to align with the codestyle and bring name consistency.

**Flex**: Rename `inlineFlex` to `isInline`
**Grid**: Rename `inline` to `isInline`


packages/core/src/Box/README.mdx
```diff
@@ -0,0 +1,94 @@
...
+The Box component is useful because it helps with several common cases:
+
+- Create a layout that uses Forma36 [spacing system]
+- Compose new component and allow to override element using the `as` property.
+
+## Import
+
+```js static=true
+import { Box } from '@contentful/f36-components';
+```
```
mshaaban0, line 20: `static=true` TIL 🥇

packages/core/src/Flex/README.mdx
```diff
@@ -0,0 +1,107 @@
...
+github: 'https://github.com/contentful/forma-36/tree/master/packages/forma-36-react-components/src/components/Flex'
+storybook: 'https://f36-storybook.contentful.com/?path=/story/components-flex--default'
+---
+
+Flex components is [Box] with `display: flex` and abilities to apply other flexbox CSS properties via properties of the component.
+
+The Box component is useful because it helps with several common cases:
+
+- Create a layout with uses Forma36 spacing
+- Compose new component and allow to override element using the `as` property.
```
mshaaban0, line 14: Should we mention that it's meant to be used for one direction layouts vertical or horizontal?

packages/forma-36-website/gatsby-config.js
```diff
@@ -116,23 +116,27 @@ module.exports = {
         ],
       },
       {
-        name: 'Components',
+        name: 'Layout Components',
```
burakukula, line 119: @suevalov maybe @m10l would have some insights here, we had a discussion around the structure of the navigation recently. There was the argument here, that we should keep the main structure like it is so:
**Foundation**
**Guidelines**
**Components**
and just nest the additional section inside the **Components** - the same as we do with `Form elements`. 
WDYT?

mshaaban0 (approved):
Nice one 👍

packages/core/src/Box/README.mdx
```diff
@@ -0,0 +1,94 @@
+---
+title: 'Box'
+type: 'component'
+slug: /components/box/
+github: 'https://github.com/contentful/forma-36/tree/master/packages/forma-36-react-components/src/components/Box'
+storybook: 'https://f36-storybook.contentful.com/?path=/story/components-box--default'
+---
+
+Box is the most abstract layout component that allows to apply `margin` and `padding` properties according to the [spacing system]. By default, it renders a `div` element.
```
burakukula, line 9: @suevalov thx a lot for those docs! 🎉
Small comment from my side, that maybe we could follow similar naming for sections in the documentation as we have for `Components` (of course not all of them would be relevant for Layout components, but maybe we could follow similar conversion). This is how they look like:

## Table of contents
- [How to use `compontent name`](#how-to-use-buttons)
- [Component variations](#component-variations)
- [Code examples](#code-examples)

packages/core/src/Box/README.mdx
```diff
@@ -0,0 +1,94 @@
...
+
+The Box component is useful because it helps with several common cases:
+
+- Create a layout that uses Forma36 [spacing system]
```
denkristoffer, line 13: "Forma36" -> "Forma 36" 🙂

packages/core/src/Box/README.mdx
```diff
@@ -0,0 +1,94 @@
...
+        <SmallWhiteBox />
+      </Box>
+    </div>
+  );
+};
+```
+
+### as property
+
+You can use the `as` prop to change the element render, just like styled-components.
```
denkristoffer, line 81: I'd like if we could come up with our own explanation here, maybe just two examples of rendering a span and a div or something?

packages/core/src/Flex/README.mdx
```diff
@@ -0,0 +1,107 @@
...
+slug: /components/flex/
+github: 'https://github.com/contentful/forma-36/tree/master/packages/forma-36-react-components/src/components/Flex'
+storybook: 'https://f36-storybook.contentful.com/?path=/story/components-flex--default'
+---
+
+Flex components is [Box] with `display: flex` and abilities to apply other flexbox CSS properties via properties of the component.
+
+The Box component is useful because it helps with several common cases:
+
+- Create a layout with uses Forma36 spacing
```
denkristoffer, line 13: Also "Forma 36" here.

denkristoffer (commented):
Is using the spacer different from aligning to `space-between`?

packages/forma-36-website/gatsby-config.js
suevalov: I can change it back but later I'd like to have a separate discussion about the navigation.

packages/core/src/Box/README.mdx
suevalov: What's the difference between `How to use...`, `Component variations` and `Code examples`? Aren't all of them about how to use components?

What is `Content recommendations` section?

packages/core/src/Box/README.mdx
suevalov: There's an example using it as "span" below.

packages/core/src/Flex/README.mdx
suevalov: good call, updated!

suevalov:
@denkristoffer the difference that `Flex.Spacer` takes the whole part between two elements and you can place a Box or Flex inside Spacer.

packages/core/src/Box/README.mdx
burakukula: For example `Tag` component:

## How to use:
- tags are used to inform users of the status of an element in the interface
....
## Component variations:
 - **Primary** - used for the entity that has been changed
 ...

## Content recommendations
- try to use labels with short, scannable text
....

`Component variations` was supposed to show variants and describe those. `Code example` was thought as a section with examples of usage in code. Those 2 sections not necessarily are always present.

packages/forma-36-website/gatsby-config.js
burakukula: @suevalov ok, I'm up for it, of course. Maybe we should have a brother discussion about documentation and the website.

packages/core/src/Box/README.mdx
suevalov: Updated!

packages/core/src/Box/README.mdx
denkristoffer: That's true :) I was thinking maybe in text instead of referring to styled-components. Or maybe a link to them?

denkristoffer:
@suevalov So it's like a `Flex` in `Flex`? Maybe we can make this clearer in the README. From the name I would assume a spacer would be more like those flexible spaces you might know from a browser toolbar:
Screenshot 2021-03-22 at 15 46 35

packages/core/src/Box/README.mdx
suevalov: ah, I see... yeah, I'll add our own explanation here

packages/core/src/Box/README.mdx
gui-santos: +1

packages/core/src/Flex/README.mdx
```diff
@@ -0,0 +1,107 @@
...
+```
+
+### Real-world example
+
+```jsx
+<Flex alignItems="center">
+  <Box>
+    <Subheading>User name</Subheading>
+  </Box>
+  <Flex.Spacer />
```
gui-santos, line 92: Would `Flex.Spacer` create an empty element in the DOM?
And if yes, is this ok? Is it a11y? Do we want to recommend that?

I'm asking because it seems that we can achieve the same effect of the spacer here by just passing a marginBottom to the Box

gui-santos (approved):
Great stuff!

I have only one question about `Flex.Spacer` but not a blocker

suevalov:
@denkristoffer @gui-santos Valid concerns. I'm going to delete `Flex.Spacer` for now and will bring it for the discussion later.

packages/core/src/Flex/README.mdx
suevalov: valid point, I'm going to remove Flex.Spacer for now

packages/core/src/Box/README.mdx
suevalov: updated!

denkristoffer (approved):
❤️

PR #888, opened 2021-03-22
docs: props table generation

suevalov:
# Purpose of PR

Adds the ability to generate properties table for the documentation website automatically, so we always have up-to-date documentation about components API.

Kapture 2021-03-22 at 17 23 33

### How to use

```mdx
## Props

import { Props } from '@contentful/f36-docs-utils';

<Props of="Flex" />

```


packages/components/badge/src/Badge.tsx
```diff
@@ -7,16 +7,17 @@ import { getBadgeStyles } from './getBadgeStyles';
...
   size?: BadgeSize;
   /**
-   * prop, used for setting variation of the component
+   * Determines the variation of the component
    * @default primary
    */
   variant?: BadgeVariant;
-  children: React.ReactNode;
+
+  children: React.ReactText;
```
nikazawila, line 20: this changed from ReactNode to ReactText - is it always just text?

packages/core/src/Box/README.mdx
```diff
@@ -93,4 +100,10 @@ If you want to keep all the styling you've applied to a component but just switc
 </Box>
 ```
 
+## Props
+
+
+<Props of="Box" />
```
nikazawila, line 107: this is hella cool!

packages/components/badge/src/Badge.tsx
suevalov: As far as I know - yes. It's better to narrow down the type here to prevent crazy usage of the Badge component.

nikazawila: yeah, I get it - was just thinking whether it ever happens that we, for example, throw an Icon in there...

suevalov: yeah, let's verify with folks: @burakukula @gui-santos @mshaaban0

packages/f36-docs-utils/package.json
```diff
@@ -0,0 +1,36 @@
...
+  "scripts": {
+    "build": "parcel build index.ts",
+    "watch": "parcel watch index.ts"
+  },
+  "dependencies": {
+    "constate": "^3.2.0",
+    "@babel/runtime": "^7.6.2",
+    "@contentful/f36-tokens": "^4.0.0",
+    "@contentful/f36-components": "^4.0.0",
+    "lodash.orderby": "^4.6.0"
```
denkristoffer, line 28: This can go in a different PR really, but we might as well think about it here. These separate lodash method packages aren't being updated, not even for security issues. This one is 5 years old. We should just use the main package with the webpack plugin instead.

packages/f36-docs-utils/src/Props/Props.tsx
```diff
@@ -0,0 +1,106 @@
...
+    );
+  }
+  if (name === 'children') {
+    return <>{type.raw}</>;
+  }
+  if (type.name === 'enum') {
+    if (type.value.length < 4) {
+      return <EnumPropertyType type={type} />;
+    }
+    return <LookupPropertyType type={type} />;
```
denkristoffer, line 64: Why have a lookup component instead of always just rendering the props?

packages/f36-docs-utils/package.json
suevalov: this package is used only on the documentation website, so we can use lodash as it is

packages/f36-docs-utils/src/Props/Props.tsx
suevalov: a lot of them have a lot of options... like a lot, not sure if we want to show them all the time by default.

packages/f36-docs-utils/package.json
denkristoffer: And the timepicker 🙂

denkristoffer (approved):

packages/f36-docs-utils/src/Props/Props.tsx
denkristoffer: Do they fit in the current UI?

packages/f36-docs-utils/package.json
suevalov: yeah, but this particular package is used only on the documentation website, so I'll just replace it here without touching timepicker

packages/f36-docs-utils/src/Props/Props.tsx
suevalov: yep, they fit but it can be a really long list. IMO, collapsing them in a lookup makes sense but I'm happy to discuss the potential change later.

packages/f36-docs-utils/package.json
suevalov: updated

packages/f36-docs-utils/src/Props/Props.tsx
denkristoffer: I think it's fine as is, I was just thinking based on the GIF that a long list might not work in the table at all.

packages/components/badge/src/Badge.tsx
mshaaban0: I believe it should also accept icons? we use the clock icon for scheduled entities, so maybe ReactNode is the way to go here

mshaaban0 (approved):

packages/components/badge/src/Badge.tsx
suevalov: okay, rolling it back to React.ReactNode

nikazawila (approved):
thanks @suevalov 🌻

PR #889, opened 2021-03-23
feat: typography package

suevalov:
# Purpose of PR

Migrates all Typography components to a separate `f36-typography` package.

* No breaking changes, besides `element` => `as` rename. By default, heading components no margins. When wrapped in `<Typography>` they receive a default margin-bottom.
* All typography components now have additional `Margin` spacing properties, so we can override values provided from Typography Context.

```js
<Typography>
    <Heading marginTop="spacingS" marginBottom="spacing4Xl" />
</Typography>
```

P.S. All the actual changes planned for V4 should be done in a separate PR.
P.S.S. Documentation changes will be done when be done actual changes planned for V4


packages/components/typography/src/Heading.tsx
```diff
@@ -0,0 +1,58 @@
...
+  Box,
+  CommonProps,
+  MarginProps,
+  PolymorphicComponent,
+  PolymorphicComponentProps,
+  PolymorphicComponentWithRef,
+} from '@contentful/f36-core';
+import { TypographyContext } from './Typography';
+
+export type HeadingElement = 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6' | 'p';
```
denkristoffer, line 13: What about `span` and `div`? I don't think I see in issue in opening up for all elements here.

packages/components/utils/package.json
```diff
@@ -1,7 +1,6 @@
 {
   "name": "@contentful/f36-utils",
   "version": "4.0.0",
-  "private": true,
   "description": "Forma 36: utils React Component",
```
denkristoffer, line 4: Is there a reason to publish this and core?

packages/components/typography/src/Heading.tsx
suevalov: Yeah, I guess we can do that. The current type is just a copy of the previous implementation.

packages/components/utils/package.json
suevalov: yes, I want to be able to test all the things in Codesandox

gui-santos (approved):

denkristoffer (approved):

packages/components/utils/package.json
denkristoffer: In that case we might want to change the component generation template.

suevalov: I think making it publishable should be a deliberate choice, so I'd rather generate `private: true` and change it when it ready to go live.

packages/components/typography/src/Heading.tsx
suevalov: updated

PR #890, opened 2021-03-24
refactor: revert to plain emotion

suevalov:
# Purpose of PR

It turned out that using `@emotion/core` makes it impossible to override styles with simple `emotion` which makes it a big deal and a breaking change for our internal Contentful applications.

Also, @denkristoffer found an issue https://github.com/emotion-js/emotion/issues/2275 that tells that `css=` property is causing performance regressions.


packages/components/badge/src/getBadgeStyles.ts
```diff
@@ -1,5 +1,5 @@
 import tokens from '@contentful/f36-tokens';
-import { Interpolation } from '@emotion/core';
+import type { Interpolation } from '@emotion/serialize';
```
denkristoffer, line 2: Should this be added as an explicit dependency?

mshaaban0 (approved):

packages/components/badge/src/getBadgeStyles.ts
suevalov: good point, I decided to remove explicit type here

denkristoffer (approved):
🙏

scripts/plop-templates/package/package.json.hbs
```diff
@@ -23,7 +23,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.6.2",
-    "@emotion/core": "^10.1.1",
+    "emotion": "^10.0.17",
```
denkristoffer, line 26: A pedantic nit here, if we don't sort this alphabetically yarn will do it for us and create a diff on the first run, which adds noise.

suevalov: updated 🧐

PR #893, opened 2021-03-24
chore: generate token types

suevalov:
# Purpose of PR

Automatically generate token-based typescript types.

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -1,3 +1,4 @@
+import type { F36Spacing } from '@contentful/f36-tokens';
 import type { CSSProperties } from 'react';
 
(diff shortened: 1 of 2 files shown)


denkristoffer (commented):
I think we kind of get the "F36" namespace automatically by importing from `@contentful/f36-tokens`, do you think it's worth keeping for all types?

suevalov:
@denkristoffer I guess we can rename to `SpacingTokens`, `BorderRadiusTokens`, `GlowTokens`, etc...

denkristoffer (approved):
🙏

gui-santos (approved):
Wow! That's magical

🧙

PR #901, opened 2021-03-30
feat: add customBody prop to EntryCard

nekosaur:
# Purpose of PR

Adds `contentElement` prop to EntryCard. This is useful when you want to fully customize the look of an entry beyond the structured design allowed by `title`, `description` and `thumbnailElement` props.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Card/EntryCard/EntryCard.tsx
+++ b/packages/forma-36-react-components/src/components/Card/EntryCard/EntryCard.tsx
@@ -80,9 +80,9 @@ export interface EntryCardPropTypes extends BaseCardProps {
    */
   size?: EntryCardSize;
   /**
-   * Render custom content for the entry. This will override the props title, description and thumbnailElement
+   * Render custom body for the entry. This will override the props title, description and thumbnailElement
    */
-  contentElement?: React.ReactNode;
+  customBody?: React.ReactNode;
 }
 
 export function EntryCard({
@@ -102,7 +102,7 @@ export function EntryCard({
   cardDragHandleComponent,
   cardDragHandleProps,
   withDragHandle,
-  contentElement,
+  customBody,
   ...otherProps
 }: EntryCardPropTypes): React.ReactElement {
   const renderTitle = useCallback((_size: EntryCardSize, title?: string) => {
@@ -258,7 +258,7 @@ export function EntryCard({
                 )}
               </div>
               <div className={styles.EntryCard__content}>
-                {contentElement ?? (
+                {customBody ?? (
                   <React.Fragment>
                     <div className={styles.EntryCard__body}>
                       {renderTitle(size, title)}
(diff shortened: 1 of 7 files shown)


nekosaur:
Not sure if there's anything I can do about the failing checks? It seems to be a missing token?

gui-santos (changes requested):
@nekosaur 

Thanks a lot for the contribution! It looks good
I have only one naming suggestion

I will take a look at the pipeline to check what's going on. Something is wrong with chromatic 😞

packages/forma-36-react-components/src/components/Card/EntryCard/EntryCard.tsx
```diff
@@ -79,6 +79,10 @@ export interface EntryCardPropTypes extends BaseCardProps {
    * Changes the height of the component. When small will also ensure thumbnail and description aren't rendered
    */
   size?: EntryCardSize;
+  /**
+   * Render custom content for the entry. This will override the props title, description and thumbnailElement
+   */
+  contentElement?: React.ReactNode;
```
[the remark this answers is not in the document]
nekosaur: I'd be fine with that. I chose the name `contentElement` because of the existing `thumbnailElement` and `dropdownListElements` props.

gui-santos:
@nekosaur for some reason the pipeline is not formating some CHANGELOG files and then it fails
I've just ran prettier on then and now you should be able to merge your PR 🙌

gui-santos (approved):
🚀

nekosaur:
I can't merge it since it don't have write access.

gui-santos:
> I can't merge it since it don't have write access.

@nekosaur I'm still trying to fix the chromatic problem 😭 

But as soon as I fix that I merge this and tell you
I think tomorrow we finally merge this

sorry for all these pipeline problems 🙏

packages/forma-36-react-components/src/components/Card/EntryCard/EntryCard.tsx
```diff
@@ -79,6 +79,10 @@ export interface EntryCardPropTypes extends BaseCardProps {
    * Changes the height of the component. When small will also ensure thumbnail and description aren't rendered
    */
   size?: EntryCardSize;
+  /**
+   * Render custom body for the entry. This will override the props title, description and thumbnailElement
+   */
+  customBody?: React.ReactNode;
```
suevalov, line 85: Looks like `customBody` acts like `children` here. Maybe it would be better to name it `children`?

suevalov (changes requested):

packages/forma-36-react-components/src/components/Card/EntryCard/EntryCard.tsx
nekosaur: Maybe? Naming it `children` maybe removes some semantic knowledge? I.e. if it is named `children` will someone think that it replaces everything inside the entry card? Similarly to how it works in a plain card?

gui-santos:
I removed my approval because this PR is not getting attention anymore
And I realised that I think it's possible to achieve the result wanted by using the `Card` component

https://codesandbox.io/s/angry-bash-n7lip?file=/src/App.js

The Card component accepts `children` so it's very flexible

If this does not solve your demand @nekosaur , you can re-open the PR and re start the discussion 🙌

PR #902, opened 2021-03-30
fix: equal left & right paddings in a dropdown button

Chaoste:
# Purpose of PR

A medium or large button with `indicateDropdown` being `true` has a too small right-padding. The paddings should be the same on both sides. Right now this is only the case for a small button since its CSS styles have a higher specificity than ".is--dropdown".

## PR Checklist

- [x] Tests are passing

Before:
Screenshot 2021-03-30 at 13 47 10
Screenshot 2021-03-30 at 13 47 25
Screenshot 2021-03-30 at 13 47 40

After:
Screenshot 2021-03-30 at 13 48 03
Screenshot 2021-03-30 at 13 47 49

### DIFF
--- a/packages/components/datepicker/CHANGELOG.md
+++ b/packages/components/datepicker/CHANGELOG.md
@@ -7,10 +7,6 @@ See [Conventional Commits](https://conventionalcommits.org) for commit guideline
 
 **Note:** Version bump only for package @contentful/forma-36-react-datepicker
 
-
-
-
-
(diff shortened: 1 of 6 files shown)


domarku:
It was a conscious design decision (back in the day) to make the padding on the right smaller but I don't see why we should keep it.

domarku (approved):

fabe (approved):

PR #905, opened 2021-03-31
fix(copy-button): center icon alignment

### LINKED ISSUE #904: 🐛 Bug - Icon misalignment in CopyButton
# Forma 36 bug report

## Summary

The icon in the CopyButton component is misaligned. 

image

denkristoffer:
Fix for CopyButton icon not being centered. The problem seems to be CSS specificity issues that show up in our production build but not in development, which is why the problem doesn't show in Storybook. The  `TabFocusTrap` component's default `display: inherit` value is overriding the CSS from CopyButton which sets it to `display: flex`.

I tried changing the class name order around and it didn't fix the problem 🤔 

Closes #904

### DIFF
--- a/packages/forma-36-react-components/src/components/CopyButton/CopyButton.css
+++ b/packages/forma-36-react-components/src/components/CopyButton/CopyButton.css
@@ -41,7 +41,7 @@
   composes: sr-only from './../../styles/settings/helpers.css';
 }
 
-.CopyButton__TabFocusTrap {
+.CopyButton .CopyButton__TabFocusTrap {
   width: 100%;
   height: 100%;
   display: flex;


packages/forma-36-react-components/src/components/CopyButton/CopyButton.css
```diff
@@ -44,7 +44,7 @@
 .CopyButton__TabFocusTrap {
   width: 100%;
   height: 100%;
-  display: flex;
+  display: flex !important;
```
mshaaban0, line 47: Would that work `.CopyButton .CopyButton__button`?

denkristoffer: Can you clarify your concern?

mshaaban0: `!important` is a [bad practice](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity#the_!important_exception) and I don't think we should use it to solve css specificity issues, especially if we have control over both components `FocusTrap` and `CopyButton`

denkristoffer: I fully agree with that but changing the order of class names didn't change anything in the production build, probably a tooling thing, so I'm not sure how else to go about it.

mshaaban0: Screenshot 2021-04-01 at 11 36 05

Tested this, and it fixes the issue on f36 website. I'm not sure what broke it though 💭

denkristoffer: Thanks Moe ❤️

mshaaban0 (approved):

PR #914, opened 2021-04-09
v4 feat(validation-message): refactor to v4 component

denkristoffer:
# Purpose of PR

Moves `ValidationMessage` to a separate component for v4.


packages/components/validation-message/src/ValidationMessage.tsx
```diff
@@ -0,0 +1,65 @@
...
+const DEFAULT_TAG = 'div';
+
+const styles = {
+  icon: css({
+    marginRight: tokens.spacing2Xs,
+    marginTop: `calc(${tokens.spacing2Xs} / 2)`,
+    minHeight: '18px',
+    minWidth: '18px',
+  }),
+  root: css({
```
suevalov, line 22: I guess, it's logical to use `Flex` instead of `Box` and completely remove this rule

packages/components/validation-message/src/ValidationMessage.tsx
```diff
@@ -0,0 +1,65 @@
...
+    display: 'flex',
+  }),
+  text: css({
```
suevalov, line 25: Looks like we can use `Paragraph` component and just set color and margin explicitly

denkristoffer: Much better, didn't even think of that. Thanks!

suevalov (approved):

PR #918, opened 2021-04-13
feat: extract List to a separate package

nikazawila:
This moves `List` to a separate component for v4


packages/components/list/src/List.tsx
```diff
@@ -0,0 +1,49 @@
...
+        }),
+        className,
+      )}
+      ref={ref}
+      {...otherProps}
+    >
+      {children}
+    </Tag>
+  );
+}
```
denkristoffer, line 43: I think we should use our polymorphic components here, maybe just a `Box` that can be either `ul` or `ol`? That would also give us the shared props like `testId` and `className` 🙂

denkristoffer (changes requested):
Thanks for tackling this 🙏 I think we should take advantage of our polymorphic building block components here, probably the `Box` in this case.

nikazawila:
I'd appreciate some help on figuring out what I did wrong that the deployment fails...

packages/components/list/src/ListItem/ListItem.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { List } from '../List';
+
+export interface ListItemProps {
+  // Additional className
+  className?: string;
+  // The children of the ListItem
+  children: React.ReactNode;
+  // Class used for testing purposes
+  testId?: string;
+}
```
denkristoffer, line 14: I think we should also use `CommonProps` for this component.

packages/components/list/src/List.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+export type ListProps = PolymorphicComponentProps<
+  typeof DEFAULT_TAG,
+  ListInternalProps
+>;
+
+const _List: PolymorphicComponentWithRef<
+  ListInternalProps,
+  typeof DEFAULT_TAG
+> = ({ as: Tag = 'ul', className, children, ...otherProps }, ref) => {
```
denkristoffer, line 28:
```suggestion
> = ({ as: Tag = DEFAULT_TAG, className, children, ...otherProps }, ref) => {
```

denkristoffer: One last thing 🙂

denkristoffer (approved):

PR #925, opened 2021-04-16
feat: Machine Readable Formats for DateTime with DATE_ONLY or TIME_ONLY

Chaoste:
# Purpose of PR

When one uses DateTime only showing the date or time, the `datetime` attribute should represent that, too.

I followed the respective [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time#valid_datetime_values) regarding which formats are allowed.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/DateTime/DateTime.test.tsx
+++ b/packages/forma-36-react-components/src/components/DateTime/DateTime.test.tsx
@@ -5,7 +5,7 @@ import { DateTime } from './DateTime';
 
 const exampleDate = new Date('2020-04-09T16:17:18.912Z');
 
-const monkeyPatch = (errorMessage: string) => {
+const silenceErrorsWithMessage = (errorMessage: string) => {
   // Throw expected error silently
   const originalFunction = console.error;
   const spy = jest.spyOn(console, 'error');
@@ -67,7 +67,7 @@ describe('DateTime', () => {
   it('does not allow for an unknown format', () => {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     const format = 'NOT_REAL' as any;
-    const restoreErrorPatch = monkeyPatch("Unknown date format 'NOT_REAL'");
+    const restoreErrorPatch = silenceErrorsWithMessage("Unknown date format 'NOT_REAL'");
     expect(() => {
       render(<DateTime date={exampleDate} format={format} />);
     }).toThrow(`Unknown date format 'NOT_REAL'`);
--- a/packages/forma-36-react-components/src/components/DateTime/DateTime.test.tsx
+++ b/packages/forma-36-react-components/src/components/DateTime/DateTime.test.tsx
@@ -67,7 +67,9 @@ describe('DateTime', () => {
   it('does not allow for an unknown format', () => {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     const format = 'NOT_REAL' as any;
-    const restoreErrorPatch = silenceErrorsWithMessage("Unknown date format 'NOT_REAL'");
+    const restoreErrorPatch = silenceErrorsWithMessage(
+      "Unknown date format 'NOT_REAL'",
+    );
     expect(() => {
       render(<DateTime date={exampleDate} format={format} />);
     }).toThrow(`Unknown date format 'NOT_REAL'`);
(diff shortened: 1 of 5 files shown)


packages/forma-36-react-components/src/components/DateTime/DateTime.test.tsx
```diff
@@ -51,9 +51,12 @@ describe('DateTime', () => {
...
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     const format = 'NOT_REAL' as any;
-
+    // Throw error silently
+    const spy = jest.spyOn(console, 'error');
+    spy.mockImplementation(() => {});
     expect(() => {
       render(<DateTime date={exampleDate} format={format} />);
     }).toThrow(`Unknown date format 'NOT_REAL'`);
+    spy.mockRestore();
```
denkristoffer, line 60: I'd love to silence this error, but is it jest's default behaviour to log an error that's expected, or is it our code that does it? If it's the former I'm not sure we should mess with it 🤔

Chaoste: [Here](https://github.com/facebook/jest/issues/5785) is a Github issue thread on this topic. The OP claims that he's seeing the expected errors in the console during development which makes it hard to find out if the tests actually passed. Looking at the following discussion it appears not to be expected behaviour. One guess is related to async/ sync tests.
My solution is not perfect since it solves this problem only for this specific case and if the test fails we might not see the error. Still, I believe that this kind of logging is not useful and should be omitted until required for some reason. This makes reading test results way easier.

denkristoffer: Thanks for the context! Silencing the error seems the way to go, but can we maybe add a type check here to make sure we only silence that specific error?

Chaoste: That's a good idea. Then, we will see again if something unexpected happens :)

mshaaban0 (approved):

packages/forma-36-react-components/src/components/DateTime/DateTime.test.tsx
Chaoste: I created a util function `monkeyPatch()` so we can maybe reuse that later at another point.

packages/forma-36-react-components/src/components/DateTime/DateTime.test.tsx
```diff
@@ -5,6 +5,22 @@ import { DateTime } from './DateTime';
 
 const exampleDate = new Date('2020-04-09T16:17:18.912Z');
 
+const monkeyPatch = (errorMessage: string) => {
+  // Throw expected error silently
+  const originalFunction = console.error;
+  const spy = jest.spyOn(console, 'error');
+  spy.mockImplementation((error) => {
+    if (error.toString().includes(errorMessage)) {
```
denkristoffer, line 13:
```suggestion
    if (error.message === errorMessage) {
```

Do you think it will be useful to be able to check for a substring?

Chaoste: I like that it allows more flexibility so I don't have to look out for a missing empty space or some punctuation marks. I don't see any risk in there. If someone calls `monkeyPatch('error')` it's quite obvious that this is not specific enough. On the other hand, someone is able to call monkeyPatch('') to quickly turn off all errors.

packages/forma-36-react-components/src/components/DateTime/DateTime.test.tsx
```diff
@@ -5,6 +5,22 @@ import { DateTime } from './DateTime';
 
 const exampleDate = new Date('2020-04-09T16:17:18.912Z');
 
+const monkeyPatch = (errorMessage: string) => {
```
denkristoffer, line 8: Okay then my last concern is that this should have a clearer name like `silenceErrorsWithMessage` or something.

Chaoste: I like this wording people somehow came up with but I totally agree :)

denkristoffer (approved):
🙏

denkristoffer:
@all-contributors please add @Chaoste for bug, code

PR #930, opened 2021-04-19
feat(table): update for v4

denkristoffer:
Updates table components for v4 with emotion.

This PR deletes the `TableSortingLabel` component, I can't see that it's used anywhere. Do we want to keep it around?


denkristoffer:
There's currently a small problem with the story, the styles aren't loaded for the `Button` component used to toggle the loading state. Nothing blocking.

fabe (approved):
LGTM! Tiny thing: could we set the header cells to the medium font-weight?

gui-santos (approved):
It looks great!

I just realized one thing... are we updating the components in the website?
Are the pages working now that we are moving components to their own package?

denkristoffer:
@gui-santos Yes the preview URL posted y the Netlify bot uses the new packages.

packages/forma-36-react-components/src/index.ts
```diff
@@ -19,6 +19,20 @@ export type {
 } from '@contentful/f36-core';
 export { Spinner } from '@contentful/f36-spinner';
 export type { SpinnerProps } from '@contentful/f36-spinner';
+export {
+  Table,
+  TableBody,
+  TableCell,
+  TableHead,
+  TableRow,
+} from '@contentful/f36-table';
```
mshaaban0, line 28: 💭 should we introduce a wrapper for `<tfoot>` as well, I don't know if we have any table footers we use across our apps?

mshaaban0 (approved):
Thanks for doing this, LGTM 👍🏼

packages/forma-36-react-components/src/index.ts
domarku: Not sure if we need a table footer in Contentful for now. We don't have any number-heavy tables that would benefit from it. What would be an example use?

PR #932, opened 2021-04-21
docs: update README for Textarea

nikazawila:
# Purpose of PR

Updating incomplete README for Textarea, as part of @burakukula's documentation initiative.

### DIFF
--- a/packages/forma-36-react-components/src/components/Textarea/README.mdx
+++ b/packages/forma-36-react-components/src/components/Textarea/README.mdx
@@ -4,17 +4,54 @@ type: 'component'
 status: 'stable'
 slug: /components/textarea/
 github: 'https://github.com/contentful/forma-36/tree/master/packages/forma-36-react-components/src/components/Textarea'
-storybook: 'https://f36-storybook.contentful.com/?path=/story/components-textarea--default'
+storybook: 'https://f36-storybook.contentful.com/?path=/story/form-elements-textarea--basic'
 ---
 
-Detailed documentation will be provided soon. For now, please have a look into the storybook for implementation details.
+`Textarea` is a form element that allows the user to enter a sizeable amount of multi-line plain text.
 
-```jsx
+## Table of contents
+
+- [How to use Textarea](#how-to-use-Textarea)
+- [Code examples](#code-examples)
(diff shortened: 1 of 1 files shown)


packages/forma-36-react-components/src/components/Textarea/README.mdx
```diff
@@ -4,17 +4,56 @@ type: 'component'
...
+
+```tsx
+// import { Textarea } from '@contentful/forma-36-react-components';
 
-```jsx
 <Textarea
-  name="someInput"
-  id="someInput"
+  name="textarea"
+  id="textarea"
```
denkristoffer, line 30: Maybe a more semantic value would make this even more helpful, like "description" or "story" or something else that hints at a long text?

denkristoffer (approved):
🙏

packages/forma-36-react-components/src/components/Textarea/README.mdx
```diff
@@ -4,17 +4,56 @@ type: 'component'
...
+  value="123456"
+  rows={2}
+  textareaRef={}
+/>
+```
+
+## Accessibility
+
+- To ensure accessibility, provide a `name` prop
+- Always feedback the user if the Textarea is a required field
```
burakukula, line 59: @nikazawila Good point! Maybe we would be able to provide examples for all form elements when it comes to how we recommend indicating require fields and validation patterns? WDYT? Does this make sense?
But we can totally do it as a separate task as soon as we have all form elements documented.

burakukula (approved):
Awesome! Thank you! 🎉

packages/forma-36-react-components/src/components/Textarea/README.mdx
nikazawila: I think it would make sense to have an overview Form example that we could link to from within all possible form elements- and do all the role model error handling etc there to showcase a correctly build form. Would that be okay or you'd rather do it per every component?

burakukula: yeah, that would be a nice way to show it 🙂 we can tackle it in a separate task.

PR #938, opened 2021-04-26
refactor: button component

### LINKED ISSUE #913: 💡 v4 Proposal - Combine button components

## The problem

We currently have `Button`, `CopyButton`, `IconButton`, and `ToggleButton`. That's a lot of button components that do very similar things!

## The proposed solution

We combine the above components into just `Button`, which needs new APIs to replicate the functionality of the different components:

### CopyButton

I'm not convinced `CopyButton` has a place in a design system. It's a rather simple composition of other components that could be better suited for other code bases. My suggestion would be to remove it from Forma 36 and move it to the application that uses it.

### IconButton

`Button` takes an `icon` prop and render it as a value. If no text value is passed only the icon gets rendered.

### ToggleButton

The style of `ToggleButton` can be automatically applied when `Button` is rendered as a direct child of a new `ButtonGroup` component that would be based on the `ToggleButton.Group` component. However, this might be too much magic for designers working with the UI Kit, so for that reason I'm also open to keeping this component around.

mshaaban0:
# Purpose of PR

closes #913

## Todo
- [x] Deprecate old button component code
- [x] Rename the new story title from `Component/button` to `Components/button`

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/components/button/README.mdx
+++ b/packages/components/button/README.mdx
@@ -1,3 +1,112 @@
-# @contentful/f36-button
+---
+title: 'Button'
+type: 'component'
+status: 'stable'
+slug: /components/button/
+github: 'https://github.com/contentful/forma-36/tree/master/packages/components/components/button'
+storybook: 'https://f36-storybook.contentful.com/?path=/story/components-button--button'
+typescript: ./src/Button.tsx
+---
 
-This package is part of [Forma 36](https://github.com/contentful/forma-36). See the repo for more details.
+import { Props } from '@contentful/f36-docs-utils';
+
+Buttons communicate the action that will occur when the user clicks them. They communicate calls to action to the user and allow users to interact with pages in a variety of ways. They contain a text label to describe the action, and an icon if appropriate.
+
+## Table of contents
+
+- [How to use Buttons](#how-to-use-buttons)
+- [Component variations](#component-variations)
+- [Code examples](#code-examples)
+- [Content recommendations](#content-recommendations)
+- [Accessibility](#accessibility)
+- [Props](#props)
+
+## How to use Buttons
+
+- Position buttons in consistent places in the interface
+- Use the right button for the primary action, if the button group is right-aligned
+- Use the left button for the primary action, if the button group is left- or center-aligned
+- Reduce complexity by using a small number of actions. Too many actions can create confusion when having to decide
+
+## Component variations
+
+- **Primary** - Used for the most important actions in any scenario. Don’t use more than one primary button in a section or screen to avoid overwhelming users.
+- **Secondary** - Used for a secondary actions, the most commonly used button type.
+- **Positive** - For use when the action has a positive connotation such as creating or publishing a new entity.
+- **Negative** - For destructive actions - when something can't be undone. For example, deleting entities.
+
+```jsx
+<>
+  <Button variant="primary">Primary</Button>
+  <Button variant="secondary">Secondary</Button>
+  <Button variant="negative">Negative</Button>
+  <Button variant="positive">Positive</Button>
+</>
+```
+
+Buttons can be used in couple of different variations, like active, disabled or loading. Worth mentioning is `indicateDropdown` property in the component, which enables chevron icon on the right side of the button. Using `icon` property in the Button on the other hand, enables choosen icon on the left side of the button.
+
+Contentful buttons are available in 3 different sizes:
+
+```jsx
+<>
+  <Button variant="positive" size="large">
+    Large
+  </Button>
+  <Button variant="positive">Medium (Default)</Button>
+  <Button variant="positive" size="small">
+    Small
+  </Button>
+</>
+```
+
+## Code examples
+
+```tsx static=true
+import { Button } from '@contentful/f36-button';
+
+<Button variant="primary" testId="create-content-type-empty-state">
+  Add content type
+</Button>;
+```
+
+Usage of the button with icon
+
+```tsx static=true
+import { Button } from '@contentful/f36-button';
+import { Download } from '@contentful/f36-icons';
+
+<Button variant="primary" icon={Download}>
+  Primary star
+</Button>;
+```
+
+Icon Button
+
+```tsx static=true
+import { Button } from '@contentful/f36-button';
+import { Download } from '@contentful/f36-icons';
+
+<Button variant="primary" aria-label="Download Asset" icon={Download} />;
+```
+
+## Content recommendations:
+
+- UI usage - Button labels should be no longer than 3 words
+- Webpage usage - Button labels should be no longer than 5 words
+- Start labels with action verbs. If the button is part of an action dialog, make sure it matches the dialog header.
+- Use specific words, ideally ones that align with preceding content.
+- For buttons that are used to cancel destructive actions: label them "Never mind" or similar, instead of cancel. It makes things much easier for users to understand.
+
+## Accessibility
+
+Button component is used to trigger an action or event. By default button is implemented using the native `<button>` element with the proper type="button" provided. If the href element is provided, then the element would be rendered as `<a>` element in the document.
+When a button has focus, both the Space and Enter keys will activate the button. If the button is disabled, then the disabled attribute is provided.
+Contentful buttons are checked in terms of color contrast and pass all the requirements.
+Buttons have a focus state when using keyboard navigation.
+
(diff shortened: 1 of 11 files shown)


denkristoffer (changes requested):
Although it will complicate the code a bit I think we can combine the `css` calls based on props into one call.

packages/components/button/src/Button.tsx
```diff
@@ -0,0 +1,115 @@
...
+          variant={variant === 'secondary' ? 'default' : 'white'}
+        />
+      )}
+      {isDropdown && <ChevronDown className={styles.dropdownIcon} />}
+    </Element>
+  );
+};
+
+/**
+ * @description: Buttons communicate the action that will occur when the user clicks it
```
denkristoffer, line 113: What does this `@description` tag do? 🤔

packages/components/button/src/Button.tsx
```diff
@@ -0,0 +1,115 @@
...
+      isActive,
+      isDisabled: disabled,
+      isFullWidth,
+    }),
+  );
+
+  const Element: ElementType = href ? 'a' : 'button';
+
+  return (
+    <Element
```
denkristoffer, line 86: Should we use `Box` or similar here to get the different margin props?

packages/components/button/src/Button.tsx
mshaaban0: It's a [JSDoc](https://jsdoc.app/tags-description.html) tag to document the component/function purpose, shows up in vsCode like...
Screenshot 2021-05-02 at 00 34 05

packages/components/button/src/Button.tsx
mshaaban0: hmm, I'm not sure, `Box` seems like an overhead to add another layer/component as a base for a simple button just for margin props, if we want margins maybe we could add a prop just for that

denkristoffer: Could just extend `MarginProps` but I don't think using `Box` is a problem as long as we use it. Maybe here it's best not to since we don't want the `PaddingProps` as well.

mshaaban0: Cool, are we sure about the `margin` props or should we discuss it in an issue first? I barely add margins to buttons, so I'm not sure if it's worth the code

denkristoffer: Okay let's take it to a separate issue 🙂

packages/components/button/src/Button.tsx
```diff
@@ -0,0 +1,115 @@
...
+    href,
+    disabled,
+    type = !href ? 'button' : undefined,
+    icon,
+    isActive,
+    isDropdown,
+    isLoading,
+    isFullWidth,
+    ...otherProps
+  } = props;
```
denkristoffer, line 71: Shouldn't the common props be destructured here? `className` and `testId`.

denkristoffer (changes requested):

packages/components/button/src/Button.tsx
mshaaban0: Yeah, good catch. I left them to be part of `otherProps` but `className` needs merging, and `testId` doesn't resolve to `data-test-id` automatically as I thought 👍🏼

packages/components/button/src/Button.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const Element: ElementType = href ? 'a' : 'button';
+
+  return (
+    <Element
+      ref={ref}
+      type={type}
+      className={rootClassNames}
+      href={href}
+      disabled={disabled}
+      data-test-id={testId}
```
denkristoffer, line 95: Sorry for nitpicking here but I think it would be nice to use `Primitive` here if we're not using `Box`, so that the test id prop etc. all work the same across all components.

mshaaban0: All good, that was my first [approach](https://github.com/contentful/forma-36/pull/938/commits/fb55b1af609ff9ad91277c89a6cfbe459f88d888#diff-6c783a3c20a6a5973aeb9a6d60e5308ad35ff07bf2e1d866b7d3f272151af9beR43), but unfortunately couldn't get types to work properly the current approach seemed more straight forward, If you want to pair on this, it would be great

packages/components/button/src/Button.tsx
```diff
@@ -0,0 +1,119 @@
...
+    icon,
+    isActive,
+    isDropdown,
+    isLoading,
+    isFullWidth,
+    ...otherProps
+  } = props;
+
+  const rootClassNames = cx(
+    className,
```
denkristoffer, line 76: I think this should go last in the function call to allow overriding the built-in styles?

mshaaban0: hmm does order matter in cx `parmas`?! but it feels safer, will move it

denkristoffer:
@mshaaban0 I added a commit that uses `Primitive`, but as a con it changes the component API to require people to set `as="a"` if you want to use `href`. What do you think?

packages/components/button/README.mdx
```diff
@@ -0,0 +1,112 @@
...
+
+- UI usage - Button labels should be no longer than 3 words
+- Webpage usage - Button labels should be no longer than 5 words
+- Start labels with action verbs. If the button is part of an action dialog, make sure it matches the dialog header.
+- Use specific words, ideally ones that align with preceding content.
+- For buttons that are used to cancel destructive actions: label them "Never mind" or similar, instead of cancel. It makes things much easier for users to understand.
+
+## Accessibility
+
+Button component is used to trigger an action or event. By default button is implemented using the native `<button>` element with the proper type="button" provided. If the href element is provided, then the element would be rendered as `<a>` element in the document.
```
burakukula, line 103: README 🙌 ! Just a little comment: maybe we could make a list out of it, it would be probably rendered as one block of text.

burakukula (approved):
Looks awesome! Thx a lot for this contribution! 🎉
One thing that I noticed that is not related to the PR -> the storybook is not deployed correctly I think.

mshaaban0:
@denkristoffer types doesn't seem to be working as before, tbh, I'm in favour of dropping `<Primitive />` out of this one, and maybe reconsider it's purpose.

Adding another component/layer comes with a performance cost (mount time ~2-4ms), and the benefits are questionable `CommonProps` we already need destruct in most of cases anyway i.e. `classNames` merging, `style` prop already comes with otherProps if we extend the right types, which leaves us with `testId`, and `as` which I think we can just require in common props and add it to the plop template, thoughts?

**After**
Screenshot 2021-05-04 at 12 04 41

**Before**
Screenshot 2021-05-04 at 12 29 39

packages/components/button/README.mdx
mshaaban0: @burakukula you mean a list of the outline, and add it to the plop template?

denkristoffer:
@mshaaban0 Autocomplete works for me, do you have `@contentful/f36-core` built locally so that it can load the type?

mshaaban0:
@denkristoffer yes it's built, just built it again as well, same results, it's not my main concern though, there's probably a way to fix that, my concerns here are more about what are the benefits of adding another layer, and is it really worth it?

mshaaban0:
@denkristoffer could you please unblock the PR if there are no major concerns?

denkristoffer (approved):

mshaaban0:
Since this component has visual breaking changes (sizes), I'll not replace the old ones across all components, instead I deprecated the old component and will migrate to the new one along the way of migrating other components.

Deprecation notice should show up like the following in editors with JSDoc support

Screenshot 2021-05-06 at 02 04 05

packages/components/button/README.mdx
```diff
@@ -0,0 +1,112 @@
...
+title: 'Button'
+type: 'component'
+status: 'stable'
+slug: /components/button/
+github: 'https://github.com/contentful/forma-36/tree/master/packages/components/components/button'
+storybook: 'https://f36-storybook.contentful.com/?path=/story/components-button--button'
+typescript: ./src/Button.tsx
+---
+
+import { Props } from '@contentful/f36-docs-utils';
```
Chaoste, line 11: What's the deal with this line ?

mshaaban0: This helps with displaying components props on the documentation website, see https://github.com/contentful/forma-36/pull/888#issue-598196024

PR #941, opened 2021-04-28
Migrate ControlledInput to emotion [HOMER-36]

Chaoste:
# Purpose of PR

Create new component package `f35-controlled-input` with the previous styles but using emotion library and adding a storybook.

## PR Checklist

- [x] Tests are passing


packages/components/controlled-input/src/ControlledInput.tsx
```diff
@@ -0,0 +1,135 @@
...
+  value?: string;
+  disabled?: boolean;
+  type?: 'checkbox' | 'radio';
+  className?: string;
+  testId?: string;
+  willBlurOnEsc?: boolean;
+  indeterminate?: boolean;
+}
+
+const styles = getControlledInputStyles();
```
denkristoffer, line 35: Is there a reason to use a function for this instead of just an object?

packages/components/controlled-input/src/ControlledInput.tsx
```diff
@@ -0,0 +1,135 @@
...
+    inputRef.current.indeterminate = indeterminate;
+  }, [indeterminate]);
+
+  const iconProps: IconProps = {
+    size: 'medium',
+    variant: disabled ? 'secondary' : 'white',
+  };
+
+  return (
+    <div className={wrapperClassnames} ref={ref}>
```
denkristoffer, line 89: This should probably use our `Box` component.

packages/components/controlled-input/src/ControlledInput.tsx
```diff
@@ -0,0 +1,163 @@
...
+  className?: string;
+  testId?: string;
+  willBlurOnEsc?: boolean;
+  indeterminate?: boolean;
+  ref?: React.Ref<HTMLInputElement>;
+}
+
+// Use forwarded ref with React hooks
+// Src: https://itnext.io/reusing-the-ref-from-forwardref-with-react-hooks-4ce9df693dd
+const useCombinedRefs = <T extends HTMLElement>(
```
Chaoste, line 39: Handling a `ref` which is used for `forwardRef` but also in Reak hooks requires this merging logic. Any feedback or alternative solution is welcome :)

denkristoffer: Yeah this is the way to go, but actually, in this particular case it's not necessary since we always want to forward the reference to our root element, in this case the `Box`. We've yet to decide on a good way we can expose inner element references.

packages/components/controlled-input/src/ControlledInput.tsx
```diff
@@ -32,37 +34,36 @@ export interface ControlledInputProps extends HTMLProps<HTMLInputElement> {
...
-  onFocus,
-  required = false,
-  testId = 'cf-ui-controlled-input',
-  type = 'checkbox',
-  value,
-  willBlurOnEsc = true,
-  indeterminate,
-  ...otherProps
-}: ControlledInputProps) => {
+const _ControlledInput = (
```
ronaldronson, line 37: Is the underscore the proper way to mark module scoped vars?

ronaldronson (approved):
Looks good to me!

packages/components/controlled-input/src/ControlledInput.tsx
Chaoste: I copied that style from Badge and didn't see any other. Could also call it ControlledInputWithoutRef but I like it like that :)

packages/components/controlled-input/src/ControlledInput.tsx
```diff
@@ -8,13 +8,15 @@ import React, {
...
+import { cx } from 'emotion';
 import { Done, Minus } from '@contentful/f36-icons';
 import type { IconProps } from '@contentful/f36-icons';
 
-import styles from './ControlledInput.css';
+import { styles } from './ControlledInput.styles';
+import { Box } from '@contentful/f36-core';
 
-export interface ControlledInputProps extends HTMLProps<HTMLInputElement> {
+export interface ControlledInputProps
```
denkristoffer, line 18: You will want to extend the types from the box component as well, it gives you the shared types like `className` and `testId`.

packages/components/controlled-input/src/ControlledInput.tsx
```diff
@@ -85,9 +86,14 @@ export const ControlledInput = ({
   };
 
   return (
-    <div className={wrapperClassname}>
+    <Box
+      as="div"
+      display="inline-block"
+      className={wrapperClassnames}
+      ref={ref}
+    >
```
denkristoffer, line 94: Once the props extend `BoxProps` you can spread `...otherProps` here.

packages/components/controlled-input/src/ControlledInput.tsx
Chaoste: I imagine colliding props. Should I have a separate `boxProps: BoxProps` ?

packages/components/controlled-input/README.mdx
```diff
@@ -0,0 +1,3 @@
+# @contentful/f36-controlled-input
+
+This package is part of [forma-36](https://github.com/contentful/forma-36). See the repo for more details.
```
burakukula, line 3: Looking good! Thx a lot for the contribution! 🥇 
I would encourage you to write documentation for the component if that's not too much trouble. (we could to it in the future, but it will be a bit painful later I think 😉)
For the README we have a following structure (sometimes some of the sections are not needed depends of the component):

## Table of contents

- [How to use ControlledInput](#how-to-use-controlled-input)
- [Component variations](#component-variations)
- [Code examples](#code-examples)

Chaoste: I'll prepare two follow-up PRs which will restructure this folder - it will be f36-inputs containing multiple components. If Should I create this for every component and for the containing package? If so, I would like to file this in another PR (but I'll do it definitively) :)

burakukula: sure thing! thanks a lot! ♥️

packages/components/controlled-input/src/ControlledInput.tsx
```diff
@@ -91,6 +91,7 @@ const _ControlledInput = (
       display="inline-block"
       className={wrapperClassnames}
       ref={ref}
+      data-test-id={testId}
```
denkristoffer, line 94:
```suggestion
      testId={testId}
```

packages/components/controlled-input/README.mdx
Chaoste: My pleasure and thank you :)

packages/components/controlled-input/src/ControlledInput.styles.ts
```diff
@@ -0,0 +1,119 @@
...
+    fontSize: tokens.fontSizeM,
+    opacity: '0',
+    position: 'absolute',
+    margin: '0',
+    width: tokens.spacingM,
+    height: tokens.spacingM,
+    zIndex: 1,
+    cursor: 'pointer',
+    '&:focus': {
+      '& + label': {
```
mshaaban0, line 26: Nit: this can just be `'&:focus + label': {...}`

packages/components/controlled-input/src/ControlledInput.styles.ts
```diff
@@ -0,0 +1,119 @@
...
+  }),
+  input: css({
+    display: 'inline-block',
+    fontSize: tokens.fontSizeM,
+    opacity: '0',
+    position: 'absolute',
+    margin: '0',
+    width: tokens.spacingM,
+    height: tokens.spacingM,
+    zIndex: 1,
```
mshaaban0, line 23: Can we please use a token here `zIndexDefault`

packages/components/controlled-input/src/ControlledInput.styles.ts
```diff
@@ -0,0 +1,119 @@
...
+    cursor: 'pointer',
+    '&:focus': {
+      '& + label': {
+        outline: 'none',
+        border: `2px solid ${tokens.colorPrimary}`,
+        boxShadow: tokens.glowPrimary,
+      },
+    },
+    '&:checked': {
+      '& + label': {
```
mshaaban0, line 33: Same nit: `'&:checked + label': {...}`

packages/components/controlled-input/src/ControlledInput.styles.ts
```diff
@@ -0,0 +1,119 @@
...
+        borderColor: tokens.colorBlueMid,
+        background: tokens.colorBlueMid,
+      },
+    },
+    '&:disabled': {
+      '& + label': {
```
mshaaban0, line 39: aaand one more `'&:disabled + label': {...}`

packages/components/controlled-input/src/ControlledInput.styles.ts
```diff
@@ -0,0 +1,119 @@
...
+  inputDisabled: css({
+    cursor: 'not-allowed',
+  }),
+  inputRadioButton: css({
+    '&:checked': {
+      '& + label::before': {
+        background: tokens.colorWhite,
+      },
+      '&:disabled': {
+        '& + label::before': {
```
mshaaban0, line 54: 👁️

packages/components/controlled-input/src/ControlledInput.styles.ts
```diff
@@ -0,0 +1,119 @@
...
+  }),
+  inputCheckbox: css({
+    '&:checked, &:indeterminate': {
+      '& + label': {
+        borderColor: tokens.colorBlueMid,
+        background: tokens.colorBlueMid,
+      },
+      '&:disabled': {
+        '& + label': {
+          '& svg': {
```
mshaaban0, line 73: The list goes on....

packages/components/controlled-input/src/ControlledInput.styles.ts
```diff
@@ -0,0 +1,119 @@
...
+    '&:before': {
+      content: '""',
+      borderRadius: '50%',
+      backgroundColor: tokens.colorWhite,
+      width: tokens.borderRadiusMedium,
+      height: tokens.borderRadiusMedium,
+    },
+  }),
+  ghostCheckbox: css({
+    borderRadius: '20%',
```
mshaaban0, line 111: Should we use a token here? `borderRadiusMedium`?

packages/forma-36-react-components/src/components/Checkbox/Checkbox.tsx
```diff
@@ -21,6 +24,7 @@ export const Checkbox = ({
       type={type}
       willBlurOnEsc={willBlurOnEsc}
       {...otherProps}
+      ref={undefined}
```
mshaaban0, line 27: Is this necessary?

packages/components/controlled-input/src/ControlledInput.tsx
```diff
@@ -114,22 +121,21 @@ export const ControlledInput = ({
         required={required}
         disabled={disabled}
         onKeyDown={handleKeyDown}
-        {...otherProps}
+        {...inputProps}
       />
       {type === 'radio' ? (
-        /* eslint-disable-next-line jsx-a11y/label-has-associated-control */
+        // eslint-disable-next-line jsx-a11y/label-has-associated-control
```
mshaaban0, line 127: I think we should nest inputs inside `<label />`, I understand that it has `for` attribute but apparently this is not sufficient enough for all screen readers, here is a nice friendly [thread](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/718) 😅 
I'd say we should go for both, unless someone has a really good reason not to

mshaaban0 (commented):
Thanks @Chaoste 🍬

packages/components/controlled-input/src/ControlledInput.tsx
denkristoffer: The way I read that thread it's actually better to use ids and not nest?

mshaaban0: yeah, obviously there's a conflict people some say ID is better, and some say nesting. I suggest we go for both, unless there's a reason not to

packages/components/controlled-input/src/ControlledInput.styles.ts
Chaoste: Thanks for pointing it out. That will definitely save me a few lines in this file. I believe that mostly, this is better. But IMO, it might also decrease readability if it's easier to recognize the same selector block in various contexts.

e,g,
```
'& + label': {
        background: blue;
 },
'&:disabled' {
  '& + label': {
        background: red;
 },
},
'&:checked' {
  '& + label': {
        background: green;
 },
}
```
So in that case, I always have the same selector string '& + label' but easily grasp the different context it is used in (disabled, checked, normal).

If we always prefer merging those selectors, we should consider adding a linting rule. So this won't trigger further such feedback loops as this one :)

packages/components/controlled-input/src/ControlledInput.styles.ts
Chaoste: I tried replacing the percentage values (20% and 50%) with the px values from tokens (4px or 6px) but it's not the same. I believe that having those relative values is totally okay.

packages/forma-36-react-components/src/components/Checkbox/Checkbox.tsx
Chaoste: It was before but after I changed the ControlledInputProps to Omit<BoxProps<'div'>, 'ref'>`. It works :) Note that `ref` is still automatically typed by using `forwardRef`.

packages/components/controlled-input/src/ControlledInput.tsx
Chaoste: Thanks for the little journey Moe. I've read through the whole issue and the tweet incl. comments. I also had a look into all the linked references (W3, MDN, a11ysupport) and I also understand that it's better to stick to the standard (using the explicit form with `for`). Some even state that doing both will cause bugs so there is no evidence that we will gain anything from doing both. It's not explicitly stated at W3N/ MDN that you may do both. So I would even say that we break the standard by doing so.

Looking at [a11ysupport](https://a11ysupport.io/tech/html/label_element), the explicit form (`for`) has better support for screen readers as the implicit form (wrapping element)

mshaaban0: thanks for investigating, let's keep it as is then 👍🏼

denkristoffer: +1

> IMHO I prefer web standards over micro-optimizations for some not correctly functioning voice-over software that is not sticking to the standards.

That depends on the situation I'd say, let's be pragmatic. If the Dragon device is the most used screen reader then we'd be shooting ourselves in the foot by not adding a simply fix. However, to me, going with just ids seems to be fine here.

PR #947, opened 2021-05-03
feat: adds title attribute to typography components

djagya:
# Purpose of PR

This PR adds `title` attribute to all Typography components. 

`title` is a global HTML attribute and can be used on any HTML tag. There are multiple usages of it in our Contentful web app. To make Typescript happy we should have it listed in the props.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Typography/DisplayText/DisplayText.tsx
+++ b/packages/forma-36-react-components/src/components/Typography/DisplayText/DisplayText.tsx
@@ -11,6 +11,7 @@ export interface DisplayTextProps {
   children?: React.ReactNode;
   style?: React.CSSProperties;
   testId?: string;
+  title?: string;
 }
 
 export function DisplayText({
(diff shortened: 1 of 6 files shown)


packages/forma-36-react-components/src/components/Typography/DisplayText/DisplayText.tsx
```diff
@@ -11,6 +11,7 @@ export interface DisplayTextProps {
   children?: React.ReactNode;
   style?: React.CSSProperties;
   testId?: string;
+  title?: string;
```
mshaaban0, line 14: I think we should just do `interface DisplayTextProps extends HTMLProps<HTMLElement> { .. }`, that way we won't need to redefine html props

djagya: Yeah, it sounds better, thanks for the suggestion

djagya: `HTMLProps` pollutes the props type with unrelated props like `href`, `as`, so I used `HTMLAttributes`

mshaaban0 (approved):

PR #949, opened 2021-05-05
fix: allow target for IconButton

burakukula:
Hey!
A little issue that I face on the client side.

### DIFF
--- a/packages/forma-36-react-components/src/components/IconButton/IconButton.tsx
+++ b/packages/forma-36-react-components/src/components/IconButton/IconButton.tsx
@@ -5,36 +5,6 @@ import { Icon, IconProps } from '../Icon';
 import { TabFocusTrap } from '../TabFocusTrap';
 import styles from './IconButton.css';
 
-type IconButtonAnchorProps =
-  | {
-      /**
-       * Used to make the decision of either rendering as a <a> or as a <button> tag
-       */
-      href?: undefined;
-      /**
-       * Used with href to define a relationship between a linked resource and the current document
-       */
-      rel?: never;
-      /**
-       * Used with href to specify target attribute value
-       */
-      target?: never;
-    }
-  | {
-      /**
-       * Used to make the decision of either rendering as a <a> or as a <button> tag
-       */
-      href: string;
-      /**
-       * Used with href to define a relationship between a linked resource and the current document
-       */
-      rel?: React.AnchorHTMLAttributes<HTMLAnchorElement>['rel'];
-      /**
-       * Used with href to specify target attribute value
-       */
-      target?: React.AnchorHTMLAttributes<HTMLAnchorElement>['target'];
-    };
-
 export interface IconButtonProps
   extends React.HTMLAttributes<HTMLAnchorElement | HTMLButtonElement> {
   /**
--- a/packages/forma-36-react-components/src/components/IconButton/IconButton.tsx
+++ b/packages/forma-36-react-components/src/components/IconButton/IconButton.tsx
@@ -46,6 +46,10 @@ export interface IconButtonProps extends React.HTMLAttributes<HTMLElement> {
    * An ID used for testing purposes applied as a data attribute (data-test-id)
    */
   testId?: string;
+  /**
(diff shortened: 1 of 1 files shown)


denkristoffer (changes requested):
Can we destructure `target` and use it only on the `a` element?

packages/forma-36-react-components/src/components/IconButton/IconButton.tsx
```diff
@@ -59,7 +85,7 @@ export const IconButton = ({
   testId = 'cf-ui-icon-button',
   withDropdown = false,
   ...otherProps
-}: IconButtonProps) => {
+}: IconButtonProps & IconButtonAnchorProps) => {
```
burakukula, line 88: @denkristoffer thx for the feedback! 🙂 is this how I supposed to be doing that?

denkristoffer: I think the other change should work too, but if we're removing IconButton in v4 anyway it might not be worth messing with.

packages/forma-36-react-components/src/components/IconButton/IconButton.tsx
```diff
@@ -5,6 +5,36 @@ import { Icon, IconProps } from '../Icon';
...
+       * Used with href to define a relationship between a linked resource and the current document
+       */
+      rel?: React.AnchorHTMLAttributes<HTMLAnchorElement>['rel'];
+      /**
+       * Used with href to specify target attribute value
+       */
+      target?: React.AnchorHTMLAttributes<HTMLAnchorElement>['target'];
+    };
+
 export interface IconButtonProps extends React.HTMLAttributes<HTMLElement> {
```
mshaaban0, line 38: `export interface IconButtonProps extends React.HTMLAttributes<HTMLAnchorElement | HTMLButtonElement>`

I recommend extending the built-in types from typescript, so we don't have to redefine them

mshaaban0 (approved):

packages/forma-36-react-components/src/components/IconButton/IconButton.tsx
```diff
@@ -5,7 +5,38 @@ import { Icon, IconProps } from '../Icon';
...
+      /**
+       * Used with href to define a relationship between a linked resource and the current document
+       */
+      rel?: React.AnchorHTMLAttributes<HTMLAnchorElement>['rel'];
+      /**
+       * Used with href to specify target attribute value
+       */
+      target?: React.AnchorHTMLAttributes<HTMLAnchorElement>['target'];
+    };
+
```
denkristoffer, line 37:
```suggestion
```

packages/forma-36-react-components/src/components/IconButton/IconButton.tsx
burakukula: sure thing :) is this what you had in mind?

packages/forma-36-react-components/src/components/IconButton/IconButton.tsx
```diff
@@ -58,8 +85,9 @@ export const IconButton = ({
   onClick,
   testId = 'cf-ui-icon-button',
   withDropdown = false,
+  target,
   ...otherProps
-}: IconButtonProps) => {
+}: IconButtonProps & IconButtonAnchorProps) => {
```
denkristoffer, line 90:
```suggestion
}: IconButtonProps) => {
```

denkristoffer (approved):
🙏

packages/forma-36-react-components/src/components/IconButton/IconButton.tsx
burakukula: @denkristoffer sure, added. I would keep maybe both changes, so we are sure that target does not appear without href.

packages/forma-36-react-components/src/components/IconButton/IconButton.tsx
burakukula: @denkristoffer w8. I wanted to keep this change 😄

PR #960, opened 2021-05-10
fix: add CloudUpload icon

burakukula:


denkristoffer (changes requested):
The icon also has to be manually added to https://github.com/contentful/forma-36/blob/forma-v4/packages/components/icon/README.mdx

PR #967, opened 2021-05-12
docs: add README for Switch component

burakukula:


packages/forma-36-react-components/src/components/Switch/README.mdx
```diff
@@ -4,11 +4,42 @@ type: 'component'
 status: 'stable'
 slug: /components/switch/
 github: 'https://github.com/contentful/forma-36/tree/master/packages/forma-36-react-components/src/components/Switch'
-storybook: 'https://f36-storybook.contentful.com/?path=/story/components-switch--default'
+storybook: 'https://f36-storybook.contentful.com/?path=/story/components-switch--basic'
 ---
 
-Detailed documentation will be provided soon. For now, please have a look into the storybook for implementation details.
+Switch is a control that is used to quickly switch between two possible states. Switch represents a physical switch that allows users to turn things on or off, like a light switch.
```
denkristoffer, line 10: I find this a bit confusing, what about 

```suggestion
Switch is a control that is used to quickly switch between two possible states. Switch works like a physical switch that allows users to turn things on or off, like a light switch.
```
?

packages/forma-36-react-components/src/components/Switch/README.mdx
```diff
@@ -4,11 +4,42 @@ type: 'component'
...
+## Table of contents
+
+- [How to use Switch](#how-to-use-switch)
+- [Code examples](#code-examples)
+- [Content recommendations](#content-recommendations)
+- [Accessibility](#accessibility)
+
+## How to use Switch
+
+- use when an instant response of applied settings is required without an explicit action
```
denkristoffer, line 21: I don't really understand what "an instant response of applied settings" means. Do you all get it? I take it that it's the same information that is explained down below in "Switch vs Checkbox", I find that much clearer.

denkristoffer (changes requested):
Thank you! I have a question.

packages/forma-36-react-components/src/components/Switch/README.mdx
burakukula: ok, sure, maybe it's not clear. What I meant was, that within the switch action settings are instantly applied. I will try to rephrase it, since it's not clear. Thx for feedback

burakukula: @denkristoffer I actually see it as a repetition of a point 3 in the  "how to use section". I decided to remove it. Have a look now, please.

denkristoffer: Thanks, it's much clearer to me now 😁

denkristoffer (approved):
🙏

PR #969, opened 2021-05-17
Extract Note to separate package

nikazawila:
# Purpose of PR

Extract `Note` to separate package

## PR Checklist

- [x] Tests are passing


packages/components/note/stories/Note.stories.tsx
```diff
@@ -90,17 +87,20 @@ export const overview = () => (
       <SectionHeading as="h3">Note positive</SectionHeading>
     </Flex>
     <Flex marginBottom="spacingM">
-      <Note noteType={'positive'}>
+      <Note variant={'positive'}>
```
denkristoffer, line 90:
```suggestion
      <Note variant="positive">
```

packages/components/note/stories/Note.stories.tsx
```diff
@@ -90,17 +87,20 @@ export const overview = () => (
...
         A piece of information that is relevant to the context the user is
         currently in.
       </Note>
     </Flex>
-    <Note noteType={'positive'} title="Short, yet succint title" hasCloseButton>
-      <Flex marginBottom="spacingXs">
+    <Flex marginBottom="spacingM">
+      <Note
+        variant={'positive'}
```
denkristoffer, line 97:
```suggestion
        variant="positive"
```

packages/components/note/stories/Note.stories.tsx
```diff
@@ -70,17 +71,13 @@ export const overview = () => (
...
         currently in.
       </Note>
     </Flex>
     <Flex marginBottom="spacingM">
-      <Note
-        noteType={'warning'}
-        title="Short, yet succint title"
-        hasCloseButton
-      >
+      <Note variant={'warning'} title="Short, yet succint title" hasCloseButton>
```
denkristoffer, line 80:
```suggestion
      <Note variant="warning" title="Short, yet succint title" hasCloseButton>
```

packages/components/note/stories/Note.stories.tsx
```diff
@@ -70,17 +71,13 @@ export const overview = () => (
       <SectionHeading as="h3">Note warning</SectionHeading>
     </Flex>
     <Flex marginBottom="spacingM">
-      <Note noteType={'warning'}>
+      <Note variant={'warning'}>
```
denkristoffer, line 74:
```suggestion
      <Note variant="warning">
```

packages/components/note/stories/Note.stories.tsx
```diff
@@ -52,14 +53,14 @@ export const overview = () => (
...
-      <Note noteType={'negative'}>
+      <Note variant={'negative'}>
         A piece of information that is relevant to the context the user is
         currently in.
       </Note>
     </Flex>
     <Flex marginBottom="spacingM">
       <Note
-        noteType={'negative'}
+        variant={'negative'}
```
denkristoffer, line 63:
```suggestion
        variant="negative"
```

packages/components/note/stories/Note.stories.tsx
```diff
@@ -52,14 +53,14 @@ export const overview = () => (
       <SectionHeading as="h3">Note negative</SectionHeading>
     </Flex>
     <Flex marginBottom="spacingM">
-      <Note noteType={'negative'}>
+      <Note variant={'negative'}>
```
denkristoffer, line 56:
```suggestion
      <Note variant="negative">
```

packages/components/note/stories/Note.stories.tsx
```diff
@@ -14,18 +15,18 @@ export default {
...
   },
-};
+} as Meta;
 
 interface Args extends NoteProps {
   noteText?: string;
 }
 
-export const basic = ({ noteText, ...args }: Args) => (
+export const basic: Story<any> = ({ noteText, ...args }: Args) => (
```
denkristoffer, line 24:
```suggestion
export const basic: Story<any> = ({ noteText, ...args }) => (
```

packages/components/note/src/Note.tsx
```diff
@@ -0,0 +1,100 @@
...
+      {hasCloseButton && (
+        <IconButton
+          buttonType="secondary"
+          iconProps={{ as: Close }}
+          onClick={() => {
+            if (onClose) {
+              onClose();
+            }
+          }}
+          testId="cf-ui-note-close"
```
denkristoffer, line 89:
```suggestion
          testId={`${testId}-close`}
```

denkristoffer (changes requested):
Thanks, looks good! I just have a few nits.

packages/components/note/stories/Note.stories.tsx
nikazawila: @denkristoffer removing this removes the ability of editing the content of the Note in the story 🤔

denkristoffer (approved):
🙏

packages/components/note/stories/Note.stories.tsx
denkristoffer: I guess this should've been 

```suggestion
export const basic: Story<Args> = ({ noteText, ...args }) => (
```

?

nikazawila: went around it with using children, should be good too 🙏🏻

packages/components/note/src/Note.tsx
```diff
@@ -0,0 +1,100 @@
...
+      testId={testId}
+      ref={ref}
+      {...otherProps}
+    >
+      <div className={styles.icon}>
+        <Icon variant={variant} size={title ? 'medium' : 'small'} />
+      </div>
+      <div className={styles.info({ hasCloseButton })}>
+        {title && <div className={styles.title}>{title}</div>}
+        <div className={styles.content}>{children}</div>
```
mshaaban0, line 78: Should these elements use typography? title should be `h2` and probably content is `p`?

denkristoffer: I don't know if we can determine the heading level? Paragraph makes sense to me.

nikazawila: I probably would not use h2 for the title since we don't always know the context, would probably be okay with p but I don't really know what is the design/copy guideline for the content and usecases...
Happy to change whatever you feel like would be an improvement to the experience/accessibility in this case.

mshaaban0: yeah, you're right, I thought it's Notification 🙃 Should we expose the title element in `as` prop then, and require it to be set?

denkristoffer: If we use `article` as the root element we should be able to safely use `h1` for the title. How about that? I'd prefer some sane defaults to forcing people to set a header level.

nikazawila: @mshaaban0 how would you propose it looks like when using? 
`<Note title="this is title" as="h3" hasCloseButton>children go here</Note>` ?

denkristoffer: By the style guide it would probably be something like `<Note title="this is title" titleProps={{ as: "h3" }} hasCloseButton>children go here</Note>`

mshaaban0: The `article` tag is actually a good idea, as it will present it as self contained content, but I believe title should be `h2`, as `h1` should be preserved only to page title.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article
https://www.w3.org/WAI/tutorials/page-structure/headings/

nikazawila: so.. the Primitive would be the article and I use h2 for the title, does that sound okay for everyone?

denkristoffer: I believe `article` also signifies a "new" heading context but I think it's open for interpretation. `h2` works for me 👍

PR #976, opened 2021-05-21
chore: bump parcel to beta-3

mshaaban0:
# Purpose of PR

Looks like bumping parcel to latest gets rid of the memory leak issues we're seeing in the `forma-v4` branch, and it also reduces the build step by `15 seconds` 🌶️ 

I also added `--trace-warning` flag for the time being, so we can see what's going on in the branch

**Before**
Screenshot 2021-05-21 at 15 09 03

**After**
Screenshot 2021-05-21 at 14 58 12


denkristoffer (approved):
🙏

PR #954, opened 2021-05-06
feat: migrate f36-forms/controlled-input-field to v4 [HOMER-25] 

Chaoste:
# Purpose of PR

Extract the ControlledInputField component from react-components and move it to a new package f36-forms. Afterwards we can move RadioButtonField and CheckboxField.

- Create f36-forms package
- Move ControlledInputField from f36-react-components to f36-forms
- Migrate CSS to emotion

- [x] Evaluate !important css rules

## PR Checklist

- [x] I have read the relevant `readme.md` file(s)
- [x] All commits follow our [Git commit message convention](https://github.com/contentful/forma-36/tree/master/packages/forma-36-react-components#commits)
- [x] Tests are added/updated/not required
- [x] Tests are passing
- [x] Storybook stories are added/updated/not required
- [x] Usage notes are added/updated/not required
- [x] Has been tested based on [Contentful's browser support](https://www.contentful.com/faq/about-contentful/#which-browsers-does-contentful-support)
- [x] Doesn't contain any sensitive information
- [x] FormLabel is migrated to emotion

PR #977, opened 2021-05-21
feat: enable passing props down to label in ControlledInput [Homer 25]

Chaoste:
# Purpose of PR

For migrating ControlledInputField, we want to pass styles down to the label component. Thus, we add the prop `labelProps` in the same manner as `inputProps`.

[See related comment](https://github.com/contentful/forma-36/pull/954/files#r637008561)

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/components/inputs/src/controlled-input/ControlledInput.tsx
+++ b/packages/components/inputs/src/controlled-input/ControlledInput.tsx
@@ -10,6 +10,8 @@ import React, {
 import { cx } from 'emotion';
 import { Done, Minus } from '@contentful/f36-icons';
 import type { IconProps } from '@contentful/f36-icons';
+import { Label } from '@contentful/f36-forms';
+import type { LabelProps } from '@contentful/f36-forms';
 
 import { styles } from './ControlledInput.styles';
 import { Box, BoxProps } from '@contentful/f36-core';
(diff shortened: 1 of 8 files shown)


packages/components/inputs/src/controlled-input/ControlledInput.tsx
```diff
@@ -128,12 +138,9 @@ const _ControlledInput = (
       />
       {type === 'radio' ? (
         // eslint-disable-next-line jsx-a11y/label-has-associated-control
-        <label
-          className={cx(styles.ghost, styles.ghostRadioButton)}
-          htmlFor={id}
-        />
+        <label {...labelProps} className={labelClassnames} htmlFor={id} />
```
denkristoffer, line 141: Shouldn't this use the Label component?

Chaoste: Good point, I didn't know about Label when I was migrating the component :)

Chaoste: I tried but
```
export type LabelInternalProps = {
  /**
   * Label value to show
   */
  children: React.ReactText;
  ...
}
```

It doesn't allow me to pass actual components. I could only do it for the first of both cases which doesn't have children at all. But I need both to do the same so, `Label` doesn't work here :/

denkristoffer: Ah we should just change the type then, that was just me trying to narrow it down. Of course it should allow for icons.

denkristoffer (approved):

PR #971, opened 2021-05-18
fix(issue 894): add labelText prop support to TextInput component

anilkk:
# Purpose of PR
To fix the issue #894 

## PR Checklist

- [x] I have read the relevant `readme.md` file(s)
- [ ] All commits follow our [Git commit message convention](https://github.com/contentful/forma-36/tree/master/packages/forma-36-react-components#commits)
- [x] Tests are added/updated/not required
- [ ] Tests are passing
- [x] Storybook stories are added/updated/not required
- [ ] Usage notes are added/updated/not required
- [ ] Has been tested based on [Contentful's browser support](https://www.contentful.com/faq/about-contentful/#which-browsers-does-contentful-support)
- [x] Doesn't contain any sensitive information

PR #979, opened 2021-05-26
fix: add labelText prop support to TextInput component

### LINKED ISSUE #894: 🐛 Bug - Wrong output for Aria-Label on TextInput Component
# Forma 36 bug report

## Summary

During first steps (im fairly new to contentful) inside the app framework and using forma-36 I discovered a bug inside:
https://github.com/contentful/forma-36/blob/master/packages/forma-36-react-components/src/components/TextInput/TextInput.tsx

The TextInput-Component doesn't have a labelText-prop like other component (e.g. ControlledInput).
And instead of using this missing prop for arial-label, the name-prop is used instead.

```javascript
<input
        onKeyDown={handleKeyDown}
        aria-label={name}
        ...
      />
```

How this works now? Should I fix this and open a PullRequest?

## Environment

Environment independent

## Expected results

The component prop labelText on TextField Component is used as aria-label on rendered <input>-Tag

## Actual results

```javascript
  <TextField
    name="authToken"
    id="authToken"
    labelText="Authentication Token" // should be aria-label
    value={eurotext || ''}
    onChange={(e) =>
      this.setState({
        parameters: {
          ...this.state.parameters,
          eurotext: (e.target as HTMLTextAreaElement).value,
        },
      })
    }
    helpText="Authentication token for private rest api."
  />
```
But on the actual rendered element it uses the name-prop instead
image

mshaaban0:
# Purpose of PR

fixes #894

Addressed originally in #971 

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/TextInput/TextInput.tsx
+++ b/packages/forma-36-react-components/src/components/TextInput/TextInput.tsx
@@ -26,6 +26,7 @@ export type TextInputProps = {
     | 'date'
     | 'time';
   name?: string;
+  labelText?: string;
   id?: string;
   className?: string;
   withCopyButton?: boolean;
@@ -47,6 +48,7 @@ export const TextInput = ({
   isReadOnly = false,
   maxLength,
   name,
+  labelText,
   onBlur,
   onChange,
   onCopy,
@@ -113,7 +115,7 @@ export const TextInput = ({
     <div className={classNames}>
       <input
         onKeyDown={handleKeyDown}
-        aria-label={name}
+        aria-label={labelText}
         className={styles['TextInput__input']}
         id={id}
         name={name}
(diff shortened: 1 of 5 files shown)


packages/forma-36-react-components/src/components/TextInput/TextInput.tsx
```diff
@@ -26,6 +26,7 @@ export type TextInputProps = {
     | 'date'
     | 'time';
   name?: string;
+  labelText?: string;
```
nikazawila, line 29: massive nit: personally, I'd expect a label to show up along the input if I used that `labelText` prop and not for it to just be an aria-label.

nikazawila (approved):

packages/forma-36-react-components/src/components/TextInput/TextInput.tsx
mshaaban0: we went with `labelText` as this fix for v3, and this how it used on TextField, in v4 however, It's renamed to `label` which resolve to `aria-label` :) but I agree, it's a bit confusing

PR #990, opened 2021-06-14
fix: add keyboard support to dropdown elements [SPA-262]

mshaaban0:
# Purpose of PR

This PR fixes the issue of focus order on dropdown open and close
Previously the dropdown container won't be focused as the dropdown gets appended at the end of document body
Now dropdown container receives focus on open, and the toggle element receives the focus back on close
Same behaviour applied on submenus as well, it'll open when the `subToggleElement` is focused and clicked

https://user-images.githubusercontent.com/6163988/121927562-7d139080-cd3f-11eb-9ffd-b23020db2313.mov

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Dropdown/DropdownContainer/DropdownContainer.tsx
+++ b/packages/forma-36-react-components/src/components/Dropdown/DropdownContainer/DropdownContainer.tsx
@@ -58,9 +58,10 @@ export const DropdownContainer = forwardRef<
   });
 
   useEffect(() => {
+    dropdown.current?.focus();
+
     if (getRef && dropdown.current) {
       getRef(dropdown.current);
-      dropdown.current.focus();
     }
   }, [getRef]);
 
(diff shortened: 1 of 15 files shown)


packages/forma-36-react-components/src/components/Dropdown/DropdownContainer/DropdownContainer.tsx
```diff
@@ -60,6 +60,7 @@ export const DropdownContainer = forwardRef<
   useEffect(() => {
     if (getRef && dropdown.current) {
       getRef(dropdown.current);
+      dropdown.current.focus();
```
denkristoffer, line 63: Shouldn't this run outside of the `getRef` check?

denkristoffer:
I think we have to mock the `Date.now()` call with a static return value in the tests to get the snapshots to work.

denkristoffer (approved):
🙏

PR #994, opened 2021-06-22
fix(note): remove anchor styling

### LINKED ISSUE #866: 🐛 Bug - Visited links inside Note component are yellow
# Forma 36 bug report

Screenshot 2021-03-10 at 10 22 38

The links inside a Note should keep the same styles as links outside a Note
So when a link has been visited it will be dark blue

denkristoffer:
# Purpose of PR

This PR closes #866 by removing anchor styling completely from Note, so people have to use `<TextLink />` instead of `<a />` 😬 Is this how we want to tackle this, or should we just update the colours instead?

Edit: Actually, is this considered a breaking change? 🤔 

@gui-santos


mshaaban0 (approved):
I think this makes sense. 👍🏼 
Kudos to the documentation in examples 💯

burakukula (approved):
Looking good! Thx a lot for this fix!
I don't think this is a breaking change, it's a minor from my perspective. And yeah, I think we should keep the same style for links everywhere, but maybe someone from the design side would like to share perspective here.

suevalov (approved):

PR #996, opened 2021-06-22
fix(modal-controls): set full width with box-sizing border-box

### LINKED ISSUE #896: 🐛 Bug - Modal controls issue
The issue might be an implementation problem, not really sure right now. 

Screenshot 2021-03-29 at 12 31 27

I guess we actually don't need to calculate the width for the controls, or there is a reason to do it?

denkristoffer:
# Purpose of PR

Sets ModalControl styling to full width with `box-sizing: border-box` and plain old `padding` instead of trying to calculate width in some way that's clever-but-not-really.

Closes #896


mshaaban0 (approved):
Looks good 👍🏼

suevalov:
https://github.com/contentful/forma-36/discussions/1001

burakukula (approved):
🤝

Issue #1023, opened 2021-06-30
🐛 Bug - some people can't merge PRs


gui-santos:
# Forma 36 bug report

In some cases the `commit-check` does not run and blocks the PR

mshaaban0:
Yeah this is due to the CircleCI settings.

Enabling it shows warning about exposing ENV variables to the public which includes sensitive data npm tokens, etc.

We need to find a way to do that securely

denkristoffer:
@gui-santos so should this work now?

denkristoffer:
The original issue here should be gone by now. There's a new problem with building the website which should be fixed in #2164

PR #1031, opened 2021-07-02
fix: pill typescript workaround for rest props in react

janvandenberg:
# Purpose of PR
The Pill.tsx React component uses a rest props approach. This makes it unusable in TypeScript project where we want to provide a prop that is not defined in its PillProps interface. A way to fix this is adding a generic key value in the interface. I needed this to make drag-and-drop work for the Pill. Also there's no working drag-drop example in the docs of Forma-36, but that's another discussion. 

## PR Checklist

- [ x ] I have read the relevant `readme.md` file(s)
- [ x ] All commits follow our [Git commit message convention](https://github.com/contentful/forma-36/tree/master/packages/forma-36-react-components#commits)
- [ x ] Tests are passing
- [ ]  Storybook stories are added/updated/not required
- [ x ] Has been tested based on [Contentful's browser support](https://www.contentful.com/faq/about-contentful/#which-browsers-does-contentful-support)
- [ x ] Doesn't contain any sensitive information

### DIFF
--- a/packages/forma-36-react-components/src/components/Pill/Pill.tsx
+++ b/packages/forma-36-react-components/src/components/Pill/Pill.tsx
@@ -5,7 +5,7 @@ import { Icon } from '../Icon';
 import { TabFocusTrap } from '../TabFocusTrap';
 import styles from './Pill.css';
 
-export interface PillProps {
+export interface PillProps extends React.HTMLAttributes<HTMLElement> {
   label: string;
   onClose?: () => void;
   onDrag?: () => void;
@@ -14,8 +14,6 @@ export interface PillProps {
   style?: React.CSSProperties;
   dragHandleComponent?: React.ReactNode;
   variant?: 'idle' | 'active' | 'deleted';
-  // All other props
-  [key:string]: any;
 }
 
 export function Pill({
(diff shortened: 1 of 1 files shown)


packages/forma-36-react-components/src/components/Pill/Pill.tsx
```diff
@@ -14,6 +14,8 @@ export interface PillProps {
   style?: React.CSSProperties;
   dragHandleComponent?: React.ReactNode;
   variant?: 'idle' | 'active' | 'deleted';
+  // All other props
+  [key:string]: any;
```
mshaaban0, line 18: Thanks for spotting this, I think a better way for TS autocomplete would be to extend the interface to include html attributes, e.g.
```ts
export interface PillProps extends HTMLAttributes<HTMLElement> {}
```

janvandenberg: Agreed that would be better

janvandenberg:
Fixed it according to @mshaaban0 feedback

packages/forma-36-react-components/src/components/Pill/Pill.tsx
mshaaban0: Thanks

mshaaban0 (approved):

mshaaban0:
@all-contributors please add @janvandenberg for code

mshaaban0:
@janvandenberg I've cherry picked your commits in another PR #1032  to make our CI flow work, see #1023

Thanks again for the contribution

PR #1034, opened 2021-07-05
feat(pill): migrate to V4

massao:
# Purpose of PR

Migrate `Pill` component to v4

## PR Checklist

- [x] Tests are passing


packages/components/pill/src/Pill.tsx
```diff
@@ -0,0 +1,78 @@
...
+    onDrag,
+    className,
+    dragHandleComponent,
+    variant = 'idle',
+    ...otherProps
+  } = props;
+
+  const styles = getPillStyles(variant);
+
+  console.log(onDrag);
```
thebesson, line 37: rogue `console.log`

packages/components/pill/src/Pill.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { CommonProps, Box } from '@contentful/f36-core';
+import { TabFocusTrap } from '@contentful/f36-utils';
+import { Drag, Close } from '@contentful/f36-icons';
+import { PillVariants } from './types';
+import getPillStyles from './styles';
+
+export interface PillProps
+  extends React.HTMLAttributes<HTMLElement>,
+    CommonProps {
+  label: string;
```
suevalov, line 12: Ideally, we would have jsdoc comments for all 5 properties here, because there are picked up by tsdoc generator and end up on the documentation page.

packages/components/pill/src/Pill.tsx
```diff
@@ -0,0 +1,76 @@
...
+            <Close className={styles.icon} variant="muted" />
+          </TabFocusTrap>
+        </button>
+      )}
+    </Box>
+  );
+}
+
+/**
+ * TODO: Add description of component here.
```
suevalov, line 74: Don't forget to remove it

packages/components/pill/src/Pill.tsx
```diff
@@ -0,0 +1,76 @@
...
+      onDrag={onDrag}
+      ref={ref}
+      {...otherProps}
+    >
+      {onDrag &&
+        (dragHandleComponent ? (
+          dragHandleComponent
+        ) : (
+          <span className={styles.dragIcon}>
+            <Drag className={styles.icon} variant="muted" />
```
suevalov, line 51: I wonder if we need to apply any aria labels to the drag component here

packages/components/pill/src/Pill.tsx
```diff
@@ -0,0 +1,76 @@
...
+          </span>
+        ))}
+      <span aria-label={label} title={label} className={styles.label}>
```
suevalov, line 54: Not sure if we need `title` and `aria-label` at the same time

suevalov (commented):
Don't forget to add automatically generate props table to the documentation page.

Reviewing and modifying the documentation itself won't hurt either.

packages/components/pill/src/Pill.tsx
massao: This would be good, since we don't have any visible indication that this is a drag handle.

packages/components/pill/src/Pill.tsx
massao: Just checked and actually the `aria-label` should only be used when there is no visible text that acts as the label, so will keep only the title for it.

packages/components/pill/src/Pill.tsx
```diff
@@ -0,0 +1,85 @@
+import React from 'react';
+import { cx } from 'emotion';
+import { CommonProps, Box } from '@contentful/f36-core';
+import { TabFocusTrap } from '@contentful/f36-utils';
+import { Drag, Close } from '@contentful/f36-icons';
+import { PillVariants } from './types';
+import getPillStyles from './styles';
+
+export interface PillProps
+  extends React.HTMLAttributes<HTMLElement>,
```
suevalov, line 10: Do we actually want to allow all `React.HTMLAttributes<HTMLElement>` to be passed or just specific props + CommonProps will be enough?

suevalov: I started wondering because the props table generated automatically looks huge

suevalov (commented):
One last comment about generic HTMLDiv props. Otherwise, looks good.

packages/components/pill/README.mdx
```diff
@@ -60,4 +64,9 @@ Example of the Pill component with dragHandle, in that case Icon.
 
 ## Accessibility
 
-- Label value is passed in the component as a aria-label and title in the span element, like follows `aria-label={label} title={label}`.
+- Label value is passed in the component as a title in the span element, like follows `title={label}`.
+- The default drag component has a `aria-label="Drag handler"`, custom components should also include an `aria-label`
```
burakukula, line 68: It is not 100% clear to me what we mean by custom components. I think an example here could be nice.

packages/components/pill/src/Pill.tsx
```diff
@@ -0,0 +1,85 @@
...
+    CommonProps {
+  /**
+   * Text that will be shown on the pill
+   */
+  label: string;
+  /**
+   * Function that handles when the close icon is clicked. Close icon visibility also depends on if this is set
```
burakukula, line 17: Minor thing, I would rephrase here a bit: 
(... ) Close icon visibility depends on if this property is set.
Now when I think about it, it would be good to maybe extend to code styleguide with guidelines around comments. Or it's overkill?

burakukula:
@massao thanks for the contribution! Great work! 💯 
A small visual thing that I noticed is the alignment of the icon inside the Pill. Maybe there could be a way to vertically align it a bit more?

Screenshot 2021-07-08 at 09 25 06

packages/components/pill/stories/Pill.stories.tsx
```diff
@@ -51,7 +62,13 @@ export const CustomHandleComponent = (args: PillProps) => (
...
 );
 
-export const PillVariants = (args: PillProps) => (
+customHandleComponent.args = {
+  label: 'example.user@contentful.com',
+  onClose: action('clicked'),
+  onDrag: action('dragged'),
+};
+
+export const PillVariants: Story<PillProps> = (args: PillProps) => (
```
burakukula, line 71: A small thing that came to my mind is that maybe we could extend the existing overview story to contain the `deleted` variant with a close icon.

packages/components/pill/src/Pill.tsx
massao: yeah, that phrasing looks better, will update it for now, and we can talk about the guidelines for comments later, to have it consistent across all components.

packages/components/pill/src/Pill.tsx
massao: That's a good question, I kept it with the `React.HTMLAttributes<HTMLElement>` as it was like that on the v3, but we can remove it if it's not necessary.
Or maybe only use the `React.HTMLAttributes<HTMLDivElement>` it should remove some of the props from the generated table.
WDYT??

burakukula (approved):
👏 Awesome work! Thanks for the contribution!

suevalov (approved):

thebesson (approved):

PR #1041, opened 2021-07-05
feat: new colors tokens update

burakukula:
The first step of the implementation of this proposal https://github.com/contentful/forma-36/discussions/1005

Values are established in the new UIKit:

Screenshot 2021-07-05 at 14 32 33


packages/forma-36-tokens/src/tokens/colors/colors-contrast.js
```diff
@@ -1,7 +1,10 @@
+const colorsBlack = require('./colors-black');
+const colorsGray = require('./colors-gray');
+
 const colorsContrast = {
-  'color-contrast-dark': '#0C141C',
-  'color-contrast-mid': '#192532',
-  'color-contrast-light': '#263545',
+  'color-contrast-dark': colorsBlack['color-black'],
```
suevalov, line 5: are we going to deprecate these colors? or they'll be there but mapped to the scale colors?

packages/forma-36-tokens/src/tokens/colors/colors-elements.js
```diff
@@ -1,9 +1,11 @@
+const colorsGray = require('./colors-gray');
+
 const colorsElements = {
-  'color-element-darkest': '#b4c3ca',
-  'color-element-dark': '#c3cfd5',
-  'color-element-mid': '#d3dce0',
-  'color-element-light': '#e5ebed',
-  'color-element-lightest': '#f7f9fa',
+  'color-element-darkest': colorsGray['gray-500'],
```
suevalov, line 4: are we going to deprecate `color-element` colours?

packages/forma-36-tokens/src/tokens/colors/colors-contrast.js
burakukula: We would like to deprecate those colors and keep only `colors-gray`. 
I was wondering how to handle those cases. I guess we need to keep this file (most likely our clients use them?). So then I would need to do something like:
```
const colorsContrast = require('./colors-contrast-deprecated);

module.exports = colorsContrast;
```
And move this mapping into this file. Overkill or it makes sense?

burakukula: Ok, as discussed, I reexported colors in the main files and created additional `deprecated` files for them. 
Colors that we would keep are in `scale` files (then reexported in `colors-`) + `colors-semantic.js` all the rest will be deprecated.

suevalov (commented):
Looks good so far!

denkristoffer:
❤️ 

What's missing?

denkristoffer:
Since this is going into `master` we should add a CHANGELOG entry that the old tokens are deprecated. Are there any other ways we can let people know of the change?

burakukula:
@denkristoffer yeah, definitely a CHANGELOG entry. I think we would also announce on the community slack channel + maybe a note on the website?
We might have a blog post about it by @fabe 🙌

suevalov:
@denkristoffer @burakukula the changelog will be generated automatically and we'll just modify it afterward with a more detailed explanation

burakukula:
@denkristoffer @suevalov I think we could release this in alpha, so testing would be more relaxed? WDYT?

gui-santos (approved):
Wow! This is amazing

thebesson (approved):

denkristoffer (approved):
🙏

PR #1047, opened 2021-07-08
feat(tooltip-v4): refactor tooltip component for forma-v4

thebesson:
# Purpose of PR
- Migrate to emotion
- Update documentation
- Improve accessibility
- Update props to align with v4 style guide

##TODO

- [x] Solve the problem with inability to focus on interactive elements within the tooltip
- [x] Test accessibility 

## PR Checklist

- [x] Tests are passing


thebesson:
Accessibility problem: currently in forma-36 we allow people to use interactive elements inside our Tooltip component, and even give examples of such usage on the storybook. the interactive elements are unreachable with keyboard navigation. I browsed the internet, and it looks like this is not a recommended behaviour. I'm thinking for v4 we could replace content prop with label prop, and discouraging people from using complex interactive HTML inside the tooltip? https://forma-36-react-components.netlify.app/?path=/story/components-tooltip--content-hover

mshaaban0:
@thebesson Thanks for looking into this! I think it's necessary to allow users interact with the tooltip content whether it has interactive elements or not. e.g. a user could want to copy the content of the tooltip. 

However, I disagree with having interactive elements inside tooltips, I find it hard to spot, and as you mentioned keyboard navigation would be very tricky.

WCAG guidelines for [Content on Hover or Focus](https://www.w3.org/TR/WCAG21/#content-on-hover-or-focus)
> **Hoverable** If pointer hover can trigger the additional content, then the pointer can be moved over the additional content without the additional content disappearing;
**Persistent:** The additional content remains visible until the hover or focus trigger is removed, the user dismisses it, or its information is no longer valid.

domarku:
I agree with both. Tooltips in general should be used to provide only additional information, not information crucial to understanding the page or completing a task. Interactive elements in tooltips would pose one barrier more in obscuring interactions from users.

Thanks for the work @thebesson!

packages/components/tooltip/README.mdx
```diff
@@ -0,0 +1,54 @@
...
+
+---
+
+title: 'Tooltip'
+type: 'component'
+status: 'stable'
+slug: /components/tooltip/
+github: 'https://github.com/contentful/forma-36/tree/master/packages/components/tooltip'
+storybook: 'https://f36-storybook.contentful.com/?path=/story/components-tooltip--default'
+
```
suevalov, line 11: Please, add an automatically generated props table. We have an example of it in all other migrated components.

packages/components/tooltip/README.mdx
```diff
@@ -0,0 +1,54 @@
...
+</Tooltip>;
+```
+
+## Content recommendations
+
+- Use short and clear messages as the Tooltip’s content
+
+## Accessibility
+
+- Add `aria-describedby` property to the target element to make the contents of the tooltip more accessible
```
suevalov, line 47: Can we make it automatic? 

https://chakra-ui.com/docs/overlay/tooltip - generates id automatically, so consumers don't need to bother about adding `aria-describedby` manually.

packages/components/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,225 @@
...
+
+import { styles } from './Tooltip.styles';
+
+import tokens from '@contentful/f36-tokens';
+
+import { Portal } from '@contentful/f36-utils';
+
+export type TooltipPlace = Placement;
+
+export interface TooltipProps {
```
suevalov, line 22: extends CommonProps instead of retyping common props

packages/components/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,225 @@
...
+
+  const [isHoveringTarget, setIsHoveringTarget] = useState(false);
+  const [isHoveringContent, setIsHoveringContent] = useState(false);
+  useEffect(() => {
+    setShow(isHoveringContent || isHoveringTarget);
+  }, [isHoveringTarget, isHoveringContent]);
+
+  useEffect(() => {
+    if (isVisible) setShow(true);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
suevalov, line 153: why do we need this?

packages/components/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,225 @@
...
+  isVisible = false,
+  hideDelay = 0,
+  onBlur,
+  onFocus,
+  onMouseLeave,
+  onMouseOver,
+  targetWrapperClassName,
+  maxWidth = 360,
+  testId = 'cf-ui-tooltip',
+  place = 'auto',
```
suevalov, line 106: I'd probably rename it to `placement`. What do you think?

packages/components/tooltip/README.mdx
thebesson: sure, good example

packages/components/tooltip/src/Tooltip.tsx
thebesson: yeah, placement sounds much better

packages/components/tooltip/src/Tooltip.tsx
thebesson: this is something from the old component, I guess we can improve this part

packages/components/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,225 @@
...
+  }
+
+  const tooltip = (
+    <span
+      id={id}
+      ref={popperRef}
+      role="tooltip"
+      style={contentStyles}
+      className={cn(styles.tooltip, className)}
+      data-test-id={testId}
```
denkristoffer, line 176: Using our `Primitive` component here would set some of these props for us.

packages/components/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,225 @@
...
+  useEffect,
+  useState,
+  useRef,
+  MouseEvent,
+  FocusEvent,
+  CSSProperties,
+} from 'react';
+import { usePopper } from 'react-popper';
+import { Placement } from '@popperjs/core';
+import cn from 'classnames';
```
denkristoffer, line 11:
```suggestion
import { cx } from 'emotion';
```

packages/components/tooltip/src/Tooltip.tsx
```diff
@@ -0,0 +1,225 @@
...
+    return <HtmlTag className={targetWrapperClassName}>{children}</HtmlTag>;
+  }
+
+  const tooltip = (
+    <span
+      id={id}
+      ref={popperRef}
+      role="tooltip"
+      style={contentStyles}
+      className={cn(styles.tooltip, className)}
```
denkristoffer, line 175:
```suggestion
      className={cx(styles.tooltip, className)}
```

packages/components/tooltip/src/Tooltip.tsx
suevalov: +1 on this one

packages/components/tooltip/src/Tooltip.tsx
```diff
@@ -163,20 +158,17 @@ export const Tooltip = ({
...
-    <span
+    <Primitive
+      as="span"
       id={id}
       ref={popperRef}
       role="tooltip"
       style={contentStyles}
-      className={cn(styles.Tooltip, className)}
+      className={cx(styles.tooltip, className)}
       data-test-id={testId}
```
denkristoffer, line 172:
```suggestion
      testId={testId}
```

packages/components/tooltip/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "@contentful/f36-tooltip",
```
suevalov, line 2: this dependency has to be added to the main components package

packages/components/tooltip/src/Tooltip.tsx
```diff
@@ -91,23 +90,24 @@ export interface TooltipProps {
 export const Tooltip = ({
   children,
   className,
-  containerElement: ContainerElement = 'span',
+  as: HtmlTag = 'span',
   content,
-  id,
+  id = 'cf-ui-tooltip',
```
suevalov, line 95: this id should be optional and generate a unique number if the value is not passed

packages/components/tooltip/src/Tooltip.tsx
```diff
@@ -163,21 +158,18 @@ export const Tooltip = ({
   };
 
   if (!content) {
-    return (
-      <ContainerElement className={targetWrapperClassName}>
-        {children}
-      </ContainerElement>
-    );
+    return <HtmlTag className={targetWrapperClassName}>{children}</HtmlTag>;
```
suevalov, line 161: `return <Primitive as={HtmlTag} className={targetWrapperClassName}>{children}</Primitive>`

packages/components/tooltip/src/Tooltip.tsx
suevalov: use could use a convenient hook from Charka: https://github.com/chakra-ui/chakra-ui/blob/3205ad6d55f935f37db8f61d2dde7eaa2481578c/packages/tooltip/src/use-tooltip.ts#L108

packages/components/tooltip/package.json
```diff
@@ -0,0 +1,35 @@
...
+  "version": "4.0.0",
+  "private": true,
+  "description": "Forma 36: Tooltip component",
+  "scripts": {
+    "build": "parcel build index.ts"
+  },
+  "dependencies": {
+    "@babel/runtime": "^7.6.2",
```
suevalov, line 10: You need to specify all dependencies that are used in `src` folder here

packages/components/tooltip/README.mdx
```diff
@@ -40,4 +44,18 @@ Tooltips are very useful to communicate extra information related to an element
...
+<Tooltip placement="top" id="tooltip" content="Hi I am a Tooltip">
+  <TextLink>Hover me</TextLink>
+</Tooltip>;
+```
+
+## Props
+
+import { Props } from '@contentful/f36-docs-utils';
+
+<Props of="Tooltip" />
```
denkristoffer, line 61: This needs the `typescript` YAML frontmatter key to work. An example from Button: https://github.com/contentful/forma-36/blob/forma-v4/packages/components/button/README.mdx?plain=1#L8

packages/components/tooltip/src/Tooltip.tsx
```diff
@@ -8,30 +8,28 @@ import React, {
...
-import styles from './Tooltip.css';
+import { styles } from './Tooltip.styles';
 
 import tokens from '@contentful/f36-tokens';
 
 import { Portal } from '@contentful/f36-utils';
+import type { CommonProps } from '@contentful/f36-core';
+import { Box, useId } from '@contentful/f36-core';
 
 export type TooltipPlace = Placement;
```
denkristoffer, line 22: Maybe this should be `TooltipPlacement` to keep it similar to the prop?

suevalov: Agree with Kristoffer

packages/components/tooltip/README.mdx
```diff
@@ -1,10 +1,15 @@
...
+
 ---
+
 title: 'Tooltip'
 type: 'component'
 status: 'stable'
 slug: /components/tooltip/
-github: 'https://github.com/contentful/forma-36/tree/master/packages/forma-36-react-components/src/components/Tooltip'
+github: 'https://github.com/contentful/forma-36/tree/master/packages/components/tooltip'
 storybook: 'https://f36-storybook.contentful.com/?path=/story/components-tooltip--default'
```
massao, line 10:
```suggestion
storybook: 'https://f36-storybook.contentful.com/?path=/story/components-tooltip--basic'
```
Needs to point to a named export from the `stories` file, otherwise it ends in a blank storybook page

suevalov (approved):

massao (approved):

denkristoffer (approved):
Thank you @thebesson!

denkristoffer:
Let me just ping @0math and @gui-santos who I believe implemented hoverable content in the tooltip in the first place. We will have to rethink that pattern from now on 🙂

gui-santos:
Hi!

I actually agree with both @mshaaban0 and @domarku about interactive elements inside the tooltip
And I vaguely remember being against allowing this pattern, but I was still asked to support it

I will try to find my original PR to see what happened

I also remember seeing only one tooltip in user_interface using this pattern, but I don't remember which team was implementing it (maybe Moi?)

Anyway, I vote for not allowing interactive elements inside the tooltip

gui-santos:
I found the PR where the interactive pattern started 
https://github.com/contentful/forma-36/pull/636

denkristoffer:
Thanks Gui, let's cc @miretxu as well in case they want to chime in.

thebesson:
I removed an example with interactive elements inside the tooltip, but the type of `content` prop still allows React nodes as a prop, so it wouldn't break too much stuff, but I'm all for strictly typing it to `string`.

PR #1063, opened 2021-07-14
fix(dropdown): remove default value for getContainerRef

### LINKED ISSUE #1061: 🐛 Bug - devault value of getContainerRef should use useCallback to avoid focus on every render when focusContainerOnOpen is true
# Forma 36 bug report

## Summary

DropdownContainer component uses `useEffect` to focus the container on mount when `focusContainerOnOpen` is set to true. The issue is that that `useEffect` has `getRef` as a dependency, and because `getRef` is by default set to `getContainerRef` which in turn is by default set to `() => {}`, the `useEffect` executes on every re-render because `() => {}` is not considered a stable dependency.

Default value for `getContainerRef` should instead be something like `useCallback(() => {}, [])` (so it's stable across re-renders) or even `undefined` and then we check if it's a function before calling it?

## Environment

All browsers

## Steps to reproduce

https://codesandbox.io/s/interesting-swanson-o8zyz?file=/src/App.js

## Expected results

Dropdown is not re-focused on every re-render when I don't provide `getContainerRef`.

## Actual results

Dropdown is re-focused on every re-render when `getContainerRef` is not provided.

denkristoffer:
See #1061 for context:

The default value for `getContainerRef` in the `Dropdown` component triggers re-renders since it's not memoised. When `focusContainerOnOpen` is `true`, which it is by default, this means that the focus is moved to the dropdown element itself on every re-render. In reality there should be no need for this default value since `DropdownContainer` already checks if it's defined before calling it.

Closes #1061


mshaaban0 (approved):
Thanks for the quick fix 👍🏼

PR #1068, opened 2021-07-15
refactor: fix dependencies and improve types

suevalov:
# Purpose of PR

Fixes missing dependencies in packages and improve typings.

## Learning

We need to find an ESLint plugin that catches a situation when you use dependency but it's not specified in package.json of the corresponding package.

### DIFF
--- a/packages/components/button/package.json
+++ b/packages/components/button/package.json
@@ -8,7 +8,10 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.6.2",
+    "@contentful/f36-core": "^4.0.0",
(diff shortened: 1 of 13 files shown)


denkristoffer (approved):
Yes, we do 🙂

suevalov:
@denkristoffer I'll create a talk in my team backlog to improve ESLint setup.

denkristoffer:
@suevalov Parcel has actually fixed this in a recent nightly: https://github.com/parcel-bundler/parcel/pull/6564 🎉

PR #1069, opened 2021-07-15
feat: primitive optimizations

### LINKED ISSUE #948: 🐛 Bug - Primitive component performance impact
## Summary

We introduced `<Primitive />` component as a base component in V4 to ensure the correct usage of common props like `classNames`, `testId`, `style`, and `as`

We noticed the spread of the component actually consumes a lot of time to render the basic component in shadow DOM (See attached)

Let's use this issue to document our findings and find alternatives

## Results

Screenshot 2021-05-04 at 15 09 00

Screenshot 2021-05-04 at 15 17 43

## Alternatives
TBD

suevalov:
# Purpose of PR

Replace the Primitive component with `usePrimitive` and `useBox` hooks to avoid unnecessary components tree.

Closes #948 

Screenshot 2021-07-15 at 11 01 04


packages/components/empty-state/src/EmptyState.tsx
```diff
@@ -79,7 +79,7 @@ const _EmptyState = (props: EmptyStateProps, ref) => {
   } = props;
 
   return (
-    <Primitive
+    <Box
```
mshaaban0, line 82: Should this use `useBox` as well to avoid rendering `Box` in runtime?

packages/components/table/src/TableRow.tsx
```diff
@@ -44,7 +44,7 @@ export const TableRow = forwardRef<HTMLTableRowElement, TableRowProps>(
     forwardedRef,
   ) {
     return (
-      <Primitive
+      <Box
```
mshaaban0, line 47: Same here `useBox`?

mshaaban0 (approved):
Nice one! thanks for the fix 💯 
Can you please also do `useBox` as well for `button` component?

packages/components/empty-state/src/EmptyState.tsx
suevalov: I'm not sure if we need to optimize to this level here. I think components should be able to use <Box> when necessary without too much penalty.

PR #1074, opened 2021-07-19
docs: we are Forma 36

denkristoffer:
# Purpose of PR

👋

It happens quite often that people remove the space from the Forma 36 name and write it as "Forma36". This PR removes any last instances of such naming from our code. Hopefully this helps avoid confusion about how Forma 36 is written 🙂 

The only remaining question is if we should bother renaming the community Slack channel to `#forma-36` instead of `#forma36`.


burakukula (approved):
🙌

suevalov (approved):

PR #1092, opened 2021-07-20
feat: migration notes

burakukula:
Migration notes for tokens. WDYT?

### DIFF
--- a/packages/forma-36-tokens/MIGRATION.md
+++ b/packages/forma-36-tokens/MIGRATION.md
@@ -1,0 +1,78 @@
+# Forma 36 tokens - colors update in version 0.10.3
+
+In order to provide a better, more accessible experience for our customers, we decided to migrate our existing color palette into the new one.
+We are introducing a new color system that contains 6 base colors, along with gray and black/white. Each color consists of 9 shades, ranging from 100 to 900.
+
+The new approach allows us to meet all the a11y requirements. It also introduces a more consistent naming convention and the ability to scale the palette easily if needed.
+
+## What's new
+
+The main change in this release is the introduction of the new set of color tokens that will contain new namespaces and new values. The new palette is more vibrant, so you will see the change in Contentful products.
+
+Old tokens are still available, but they are deprecated from now on. You will see the deprecation note when using old namespaces with the recommendations on how to replace them.
+
+**What happened in this release?**
+
+Let's use blue, as an example.
+
+We are introducing a new scale that contains all the new namespaces, but we still keep the old tokens, that will be mapped to new values. As a result we get e.g:
+
+```js
+{
+  "color-blue-lightest":"#E8F5FF",
+  "color-blue-light":"#98CBFF",
+  "color-blue-mid":"#036FE3",
+  "color-blue-base":"#0059C8",
+  "color-blue-dark":"#0041AB",
+  "blue-100":"#E8F5FF",
+  "blue-200":"#CEECFF",
+  "blue-300":"#98CBFF",
+  "blue-400":"#40A0FF",
+  "blue-500":"#036FE3",
+  "blue-600":"#0059C8",
+  "blue-700":"#0041AB",
+  "blue-800":"#003298",
+  "blue-900":"#002A8E"
+}
+
+```
+
+The same approach we took with all our color tokens, so now old and new tokens are available to you, with the same, consistent values.
+
+**Note: We plan to remove the old, deprecated tokend in the next major version. But don't worry, we will help you to migrate your tokens to the new names right now.**
+
+## How to migrate your code to the new tokens
+
+In order to allow you to migrate your code to our new tokens, we created a codemod that will do all the work.
+
+Just run:
+
+`npx @contentful/f36-codemod <transform> <path> [...options]`
+
+- `transform` - name of transform, see available transforms below.
+- `path` - files or directory to transform
+- use the `--dry` option for a dry-run and use `--print` to print the output for comparison
+
+What will happen in your code, e.g:
+
+```tsx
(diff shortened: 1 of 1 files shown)


packages/forma-36-tokens/MIGRATION.md
```diff
@@ -0,0 +1,76 @@
+# Forma 36 tokens - colors update in version 0.10.3
+
+In order to provide a better, more accessible experience for our customers, we decided to migrate our existing color palette into the new one.
+We are introducing a new color system that contains 6 base colors, along with gray and black/white. Each color consists of 9 shades, ranging from 100 to 900.
+
+The new approach allows us to meet all the a11y requirements. It also introduces a more consistent naming convention and the ability to scale the palette easily if needed.
```
denkristoffer, line 6:
```suggestion
The new approach allows us to meet all accessibility requirements. It also introduces a more consistent naming convention and the ability to easily scale the palette in the future if needed.
```

Not sure if a11y is an established name or just a convention?

packages/forma-36-tokens/MIGRATION.md
```diff
@@ -0,0 +1,76 @@
...
+  "blue-100":"#E8F5FF",
+  "blue-200":"#CEECFF",
+  "blue-300":"#98CBFF",
+  "blue-400":"#40A0FF",
+  "blue-500":"#036FE3",
+  "blue-600":"#0059C8",
+  "blue-700":"#0041AB",
+  "blue-800":"#003298",
+  "blue-900":"#002A8E"
+}
```
denkristoffer, line 36: Maybe it makes more sense to show how the old tokens map to new ones here, rather than the values?

packages/forma-36-tokens/MIGRATION.md
```diff
@@ -0,0 +1,76 @@
...
+
+```
+
+The same approach we took with all our color tokens, so now old and new tokens are available to you, with the same, consistent values.
```
denkristoffer, line 40:
```suggestion
We took the same approach with all our color tokens, so now old and new tokens are available to you, with the same, consistent values.
```

denkristoffer: > so now old and new tokens are available to you, with the same, consistent values.

Not sure what is meant by "consistent" here.

packages/forma-36-tokens/MIGRATION.md
```diff
@@ -0,0 +1,76 @@
...
+
+**Note: We plan to remove the old, deprecated tokend in the next major version. But don't worry, we will help you to migrate your tokens to the new names right now.**
```
denkristoffer, line 42: > We plan to remove the old, deprecated tokend in the next major version

Do we?

packages/forma-36-tokens/MIGRATION.md
```diff
@@ -0,0 +1,76 @@
...
+
+## How to migrate your code to the new tokens
+
+In order to allow you to migrate your code to our new tokens, we created a codemod that will do all the work.
+
+Just run:
+
+`npx @contentful/f36-codemod <transform> <path> [...options]`
+
+- `transform` - name of transform, see available transforms below.
```
denkristoffer, line 52: > see available transforms below.

This information is missing. But I'd suggest we just show the full command for transforming colors and remove this line completely.

denkristoffer (changes requested):

packages/forma-36-tokens/MIGRATION.md
burakukula: yeah, maybe it's not clear. I will rephrase it.

packages/forma-36-tokens/MIGRATION.md
burakukula: I think we could plan for it, since we have a codemod? But dunno, let's check with the team.

packages/forma-36-tokens/MIGRATION.md
burakukula: How about something like that? WDYT?

denkristoffer: Works for me, thanks!

packages/forma-36-tokens/MIGRATION.md
denkristoffer: If we don't get a timely answer I think the document works without this line too.

denkristoffer:
Looks good, but I think we should decide on the line about removing tokens in v4 before we merge.

packages/forma-36-tokens/MIGRATION.md
thebesson: we can change it to "we can provide you with tools to migrate to the new names"?

packages/forma-36-tokens/MIGRATION.md
```diff
@@ -0,0 +1,103 @@
...
+  "blue-600":"#0059C8",
+  "blue-700":"#0041AB",
+  "blue-800":"#003298",
+  "blue-900":"#002A8E"
+}
+```
+
+The same approach we took with all our color tokens, so now old and new tokens are available to you.
+
+**Note: We plan to remove the old, deprecated tokens in the the version5 of Forma 36, but don't worry, we provide you with tools to migrate to the new names.**
```
denkristoffer, line 71:
```suggestion
**Note: We plan to remove the old, deprecated tokens in a future breaking change of Forma 36, but don't worry, we provide you with tools to migrate to the new names.**
```

denkristoffer (changes requested):

gui-santos (approved):
Loved it!

I +1 @denkristoffer suggestion on using "a future breaking change" in the text

denkristoffer (approved):

PR #1093, opened 2021-07-20
feat: extract DragHandle component

denkristoffer:
# Purpose of PR

Extracted `CardDragHandle` to a separate package since it's also being used in `EntityListItem`.

### DIFF
--- a/packages/components/drag-handle/src/DragHandle.tsx
+++ b/packages/components/drag-handle/src/DragHandle.tsx
@@ -138,4 +138,4 @@ const _DragHandle: PolymorphicComponentWithRef<
 export const DragHandle: PolymorphicComponent<
   DragHandleInternalProps,
   typeof DEFAULT_TAG
-> = React.forwardRef(_DragHandle);
+> = forwardRef(_DragHandle);
--- a/packages/components/drag-handle/src/DragHandle.tsx
+++ b/packages/components/drag-handle/src/DragHandle.tsx
@@ -1,0 +1,94 @@
+import React, { forwardRef, useCallback, useState } from 'react';
+import { css, cx } from 'emotion';
+import { usePrimitive } from '@contentful/f36-core';
+import type { CommonProps } from '@contentful/f36-core';
+import { Drag } from '@contentful/f36-icons';
+import tokens from '@contentful/f36-tokens';
+
+import { styles } from './DragHandle.styles';
+
+const generateStyles = ({ isActive, isFocused, isHovered }) => {
+  if (isActive || isFocused || isHovered) {
+    return css({
+      backgroundColor: tokens.colorElementLight,
+      cursor: isActive ? 'grabbing' : 'grab',
+    });
+  }
+};
+
+export type DragHandleInternalProps = {
+  /**
+   * Applies styling for when the component is actively being dragged by
+   * the user
+   */
+  isActive?: boolean;
+  /**
+   * Applies focus styling
+   */
+  isFocused?: boolean;
+  /**
+   * Applies hover styling
+   */
+  isHovered?: boolean;
+  /**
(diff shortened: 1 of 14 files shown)


denkristoffer:
The change to the new padded button component means there's more whitespace in the header here. I think this is good for actually making it easier to use the button, although the design suffers slightly. Thoughts @fabe @domarku?

domarku:
I don't mind it! As you said, it makes the click area around the button a bit larger.

packages/components/drag-handle/src/DragHandle.tsx
```diff
@@ -0,0 +1,96 @@
...
+      setisHovered(true);
+    }, []);
+
+    const handleMouseLeave = useCallback(() => {
+      setisHovered(false);
+    }, []);
+
+    return (
+      <Element
+        as="div"
```
mshaaban0, line 77: Should this be a `button` instead?

packages/components/drag-handle/src/DragHandle.tsx
```diff
@@ -0,0 +1,96 @@
...
+   */
+  isHovered?: boolean;
+  /**
+   * Label rendered in DragHandle - not visible on screen as its purpose
+   * is for screen readers only
+   */
+  label: string;
+};
+
+export type DragHandleProps = CommonProps & DragHandleInternalProps;
```
mshaaban0, line 40: Does this need to be an interface with `extend HTMLProps<ElementType> {...`?

denkristoffer: Do you think there's anything gained from being able to set common HTML props on the drag handle? The only thing I can think of is aria properties, but ideally we should be setting those ourselves.

packages/components/drag-handle/src/DragHandle.tsx
```diff
@@ -0,0 +1,96 @@
...
+    const { primitiveProps: commonProps, Element } = usePrimitive(otherProps);
+    const [isFocused, setisFocused] = useState(isFocusedProp);
+    const [isHovered, setisHovered] = useState(isHoveredProp);
+
+    const handleFocus = useCallback(() => {
+      setisFocused(true);
+    }, []);
+
+    const handleBlur = useCallback(() => {
+      setisFocused(false);
```
mshaaban0, line 64: I think we need to expose `onBlur` as a prop for consumers to be able to use it. same for other handlers as well. e.g.
```ts
const handleBlur = useCallback(() => {
  setisFocused(false);
  
  if (onBlur) onBlur();
}, []);
```

packages/components/drag-handle/src/DragHandle.tsx
mshaaban0: yeah typescript won't recognise all the compatible `{...otherProps}` that we didn't define, like `onClick` for example. 
I think this has to be the standard way we define interfaces for components, we had multiple issues opened because some props aren't supported like `title`, `onBlur` events, etc. I saw people adding `//ts-ignore` as well in consumer apps 😅

denkristoffer: > I saw people adding //ts-ignore as well in consumer apps 😅

Ouch, that's problematic. I agree with you, but I also don't think we should do it everywhere without thinking it through, there are likely cases where we don't want to support custom event handlers and such.

denkristoffer: This should work now with the polymorphic props.

mshaaban0 (approved):
Thanks for addressing the comments, Looks pretty good 💯

packages/components/drag-handle/src/DragHandle.tsx
mshaaban0: Yes, It does work with the polymorphic component. good to know 👍🏼

PR #1096, opened 2021-07-20
fix: fix issue with Tags

burakukula:
Fix background and text color for Tags.


fabe:
Not sure if it's the documentation, but the `featured` tag isn't purple. Otherwise looks good!

denkristoffer (approved):
🙏

PR #1108, opened 2021-07-22
chore(deps): fix website build by upgrading tokens

denkristoffer:
Fix master

### DIFF
--- a/packages/forma-36-website/package.json
+++ b/packages/forma-36-website/package.json
@@ -6,7 +6,7 @@
   "dependencies": {
     "@contentful/forma-36-fcss": "^0.3.0",
     "@contentful/forma-36-react-components": "^3.75.0",
-    "@contentful/forma-36-tokens": "^0.10.0",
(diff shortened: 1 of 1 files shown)


burakukula (approved):

mshaaban0 (approved):

packages/forma-36-website/package.json
```diff
@@ -6,7 +6,7 @@
   "dependencies": {
     "@contentful/forma-36-fcss": "^0.3.0",
     "@contentful/forma-36-react-components": "^3.75.0",
```
burakukula, line 8: We don't have to bump the rest of deps?

denkristoffer: I don't think we have to since they're still the same major version. The problem is that `^0.10.0` and `^0.11.0` are seen as different major versions. Which is why 0.X versions shouldn't be used for stable packages...

burakukula: I can bump it on my branch, no worries.

PR #1109, opened 2021-07-22
feat: create ButtonGroup component

massao:
# Purpose of PR

This PR create the ButtonGroup component as per the discussion on #1067.  
This is a initial version to be used on the migration of the ToggleButton component.

image

## PR Checklist

- [x] Tests are passing


packages/components/buttongroup/src/ButtonGroup.styles.ts
```diff
@@ -0,0 +1,71 @@
...
+        marginLeft: tokens.spacingS,
+        marginRight: tokens.spacingS,
+        '&:first-child': {
+          marginLeft: 0,
+        },
+        '&:last-child': {
+          marginRight: 0,
+        },
+      };
+  }
```
gui-santos, line 18: I see that besides the default "variant" we only have another case (separate) which makes this switch works like a simple `if (variant === 'separate')`
Shouldn't we replace it with an if statement then to keep things simple? (and if we increase the number of variants we return with the switch)
Or am I missing something here?

Maybe this is just my personal preference

packages/components/buttongroup/src/ButtonGroup.styles.ts
```diff
@@ -0,0 +1,71 @@
...
+    borderRadius: '0 !important',
+    marginRight: '-1px',
+    zIndex: tokens.zIndexDefault,
+    '&:first-child': {
+      borderBottomLeftRadius: `${tokens.borderRadiusMedium} !important`,
+      borderTopLeftRadius: `${tokens.borderRadiusMedium} !important`,
+    },
+    '&:last-child': {
+      borderBottomRightRadius: `${tokens.borderRadiusMedium} !important`,
+      borderTopRightRadius: `${tokens.borderRadiusMedium} !important`,
```
gui-santos, line 30: Uhmmm do we need to use `!important` here because the way emotion handle CSS classes?

Is there something we could do to avoid them?

gui-santos (changes requested):
Wow! Amazing job 🍕 

I made only one suggestion that maybe it's only my personal preference and, in that case, it can be ignored hahaha
And a question about the `!important` things

Besides that I'm ready to approve it
Thank you

packages/components/buttongroup/src/ButtonGroup.styles.ts
massao: This is because for some reason on gatsby the classes aren't being merged so the default border-radius of the button was overwriting this one.
I'm taking a deeper look to see if it's due to a plugin version or something, but on storybook they render correctly even without the `!important` flag

massao: Actually it's my bad, I was calling the `css()` method in the same value twice, fixed it and seems to be working fine on gatsby locally.

gui-santos (approved):
Great stuff!

packages/components/buttongroup/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "@contentful/f36-buttongroup",
```
suevalov, line 2: I'd probably make it a part of `button` package for the simplicity

packages/components/buttongroup/src/ButtonGroup.styles.ts
```diff
@@ -0,0 +1,76 @@
...
+import type { ButtonGroupVariants } from './types';
+import { CSSObject } from '@storybook/theming';
+
+const getGroupContentStyle = (
+  variant: ButtonGroupVariants,
+  withDivider: boolean,
+) => {
+  if (variant === 'separate') {
+    return {
+      marginLeft: tokens.spacingS,
```
suevalov, line 12: is it always `spacingS` according to design guidelines?

suevalov:
Could you please also add `separate` version to the Storybook, so it's captured by Chromatic?

packages/components/buttongroup/src/ButtonGroup.styles.ts
massao: There's no example of a separates button group on figma, I added `spacingS` to have the variant, but would need to double check with the designers.

packages/components/buttongroup/package.json
massao: Should we add the `ToggleButton` and other types also on the button package?

suevalov: others are debatable, but `ButtonGroup` can't be used without a `Button` so for me it makes sense to colocate them within one package.

packages/components/buttongroup/src/ButtonGroup.styles.ts
suevalov: yeah, let's double-check that

packages/components/buttongroup/package.json
massao: I think `ToggleButton` would also be the case, as it will use the Button as base and only compose on top of it, and would also be usable on the `ButtonGroup`, but we can check the others later

suevalov: if so, we can probably collocate it too

packages/components/buttongroup/src/ButtonGroup.styles.ts
massao: I may remove the option of having them as separate for now and we add it back later on if we actually need it, wdyt?

suevalov: i'd rather leave it as `spacingS` and then add a new property to control it

massao: Makes sense, I'll add another prop to control the spacing, and keep the spacingS as default and move it to the Button component then. 😄

packages/components/button/src/ButtonGroup/ButtonGroup.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { CommonProps, Box } from '@contentful/f36-core';
+import getStyles from './ButtonGroup.styles';
+import type { ButtonGroupVariants, ButtonGroupSpacing } from './types';
+
+export interface ButtonGroupProps extends CommonProps {
+  /**
+   * Determines how the Button Group will display the buttons
+   * @default collapse
+   */
+  variant?: ButtonGroupVariants;
```
suevalov, line 12: I wonder if it's possible to type that `withDivider` exists only when variant is `collapse`, and `spacing` only when variant is `separate`? 

something like:

```
{ variant: 'collapse', withDivider?: boolean } | { variant: 'separate', spacing: Spacing }
```

packages/components/button/src/ButtonGroup/types.ts
```diff
@@ -0,0 +1,11 @@
+import type { SpacingTokens } from '@contentful/f36-tokens';
+
+export type ButtonGroupVariants = 'separate' | 'collapse';
```
suevalov, line 3: a small thing, but I'd rather use adjectives instead of verbs here:

variant="collapsed" sounds more logical than variant="collapse"
variant="separated" (or even "spaced" because we have "spacing" property)

suevalov (approved):

PR #1114, opened 2021-07-25
feat: @contentful/f36-card

denkristoffer:
# Purpose of PR

This PR creates `@contentful/f36-card` with the components `Card`, `AssetCard`, `EntryCard`, and `InlineEntryCard` with new design. After considering different approaches I built this with a `BaseCard` component that can show a custom header component, as this seemed the best way to base all cards on the same foundational component to share props and styling as much as possible. 

Follow-up PRs:

- Loading state with skeletons for all cards.


packages/components/card/src/base-card/CardActions.tsx
```diff
@@ -0,0 +1,72 @@
+import React, { useCallback, useState } from 'react';
+import type { MouseEventHandler, ReactElement } from 'react';
+import { Button } from '@contentful/f36-button';
+import type { ButtonProps } from '@contentful/f36-button';
+import { MoreHorizontal } from '@contentful/f36-icons';
+import { Dropdown } from '@contentful/f36-components';
```
suevalov, line 6: shouldn't we wait till the Dropdown is migrated to be able to extract card to a separate package?

packages/components/card/src/card/README.mdx
```diff
@@ -19,12 +20,15 @@ A Card is a component to group information or content of a particular subject in
...
 - [Content recommendations](#content-recommendations)
 - [Accessibility](#accessibility)
+- [Props](#props)
 
 ## Component variations
 
 ```jsx
 <Card title="Title">
-  <Text as="p">This is the Card’s content</Text>
+  <Typography>
```
suevalov, line 29: `Typograpy` wrapper can be deleted

packages/components/card/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "@contentful/f36-drag-handle": "^4.0.0",
+    "@contentful/f36-core": "^4.0.0",
+    "@contentful/f36-icon": "^4.0.0",
+    "@contentful/f36-icons": "^4.0.0",
+    "@contentful/f36-tokens": "^4.0.0",
+    "@contentful/f36-typography": "^4.0.0",
+    "emotion": "^10.0.17"
+  },
+  "peerDependencies": {
+    "@contentful/f36-components": "^4.0.0",
```
suevalov, line 22: we need to remove this line before the official release.

suevalov:
In general, I like the idea of the BaseCard component that handles commonalities.

I guess we need to wait for the Dropdown component before merging this PR and also work on types a bit (so `href` becomes possible only when you specify <Card as="a" /> for example)

packages/components/card/src/base-card/CardActions.tsx
denkristoffer: I don't think we have to, but we can if you think that's better.

denkristoffer:
> and also work on types a bit (so href becomes possible only when you specify for example)

That's should be how it works now with https://github.com/contentful/forma-36/pull/1114/files#diff-73ea3ba88ccb975ddf6b2bc2987806c1cafa584382271a11cf23a392b192e149R69-R91 🙂

suevalov:
@denkristoffer `Menu` component is coming, so we hope to unblock this PR this Sprint.

suevalov:
@denkristoffer I've pushed the changes that use the Menu component instead of the Dropdown. We can finally focus on making PR ready for a review.

packages/components/card/src/asset-card/AssetCard.styles.ts
```diff
@@ -0,0 +1,42 @@
...
+      borderTopLeftRadius: tokens.borderRadiusMedium,
+      borderTopRightRadius: tokens.borderRadiusMedium,
+      boxSizing: 'border-box',
+      color: tokens.gray600,
+      fontSize: tokens.fontSizeM,
+      fontWeight: tokens.fontWeightNormal,
+      gridColumn: 'content',
+      gridRow: 'header',
+      lineHeight: tokens.lineHeightM,
+      paddingBottom: `calc(1rem * (10 / ${tokens.fontBaseDefault}))`,
```
suevalov, line 23: It would be nice to repalce all properties that use calculated spacing value (using fontBaseDefault) with the closest spacing token.

packages/components/card/src/asset-card/AssetCard.tsx
```diff
@@ -0,0 +1,113 @@
...
+const generateSizeStyles = ({
+  size,
+}: {
+  size: AssetCardInternalProps['size'];
+}): string => {
+  return css({
+    padding:
+      size === 'small'
+        ? tokens.spacingL
+        : `calc(1rem * (56 / ${tokens.fontBaseDefault}))`,
```
suevalov, line 24: Can we use spacing value here as well?

packages/components/card/src/asset-card/AssetCard.tsx
```diff
@@ -0,0 +1,113 @@
...
+   * Type of the entity represented by the card. Shown in the header of the card
+   */
+  type?: AssetType;
+}
+
+export type AssetCardProps = AssetCardInternalProps;
+
+export const AssetCard = ({
+  actions,
+  actionsButtonProps,
```
suevalov, line 57: do we really need to allow overriding `actionButtonProps`?

packages/components/card/src/base-card/BaseCard.tsx
```diff
@@ -0,0 +1,354 @@
...
+    styles.borderColor = tokens.colorPrimary;
+    styles.boxShadow = tokens.glowPrimary;
+  }
+
+  return css(styles);
+};
+
+export type CardElement = 'a' | 'article' | 'button' | 'div';
+
+type AnchorProps =
```
suevalov, line 69: I think these properties can be automatically recognized when you use `as="a"`. I'm not sure if you need to specify this type here.

packages/components/card/src/base-card/CardActions.tsx
```diff
@@ -0,0 +1,52 @@
...
+export type CardActionsProps = {
+  buttonProps?: Partial<Omit<ButtonProps<'button'>, 'ref'>>;
+  /**
+   * Child elements to be rendered in the component
+   */
+  children: React.ReactNodeArray;
+};
+
+export const CardActions = ({
+  buttonProps,
```
suevalov, line 17: do we use this property anywhere?

packages/components/card/src/card/__snapshots__/Card.test.tsx.snap
```diff
@@ -0,0 +1,281 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`calls an onClick function 1`] = `
```
suevalov, line 3: It would be nice to delete this snapshot test.

packages/components/card/src/asset-card/AssetCard.styles.ts
denkristoffer: @fabe @domarku can we do that? 🙂

packages/components/card/src/base-card/CardActions.tsx
denkristoffer: It's used by the `InlineEntryCard`

packages/components/card/src/asset-card/AssetCard.tsx
domarku: This 56px is just for the Storybook preview, correct?

packages/components/card/src/asset-card/AssetCard.styles.ts
domarku: Yes! Let's use `spacing-xs` here. It will reduce the height of the header section to `36px`, which is a good thing because it will contribute to items taking less vertical real-estate.

The only downside is that the "more" button's height will go down `4px`. I don't think that's terrible as it will still be close to `40px` in the size of its click-area.

packages/components/card/src/asset-card/AssetCard.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+const generateSizeStyles = ({
+  size,
+}: {
+  size: AssetCardInternalProps['size'];
+}): string => {
+  return css({
+    height:
+      size === 'small'
+        ? `calc(9px * ${tokens.fontBaseDefault})`
```
suevalov, line 23: can we use spacing tokens here?

packages/components/card/src/card/__snapshots__/Card.test.tsx.snap
```diff
@@ -0,0 +1,232 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`can be selected 1`] = `
```
suevalov, line 3: We decided to delete snapshot tests, so this file should be gone, together with a corresponding test.

suevalov (approved):
👏🏼 

Only two small comments

packages/core/src/utils/getEntityStatusStyles.ts
```diff
@@ -0,0 +1,38 @@
...
+        backgroundColor:
+          '#ffefd5' /* temporary hardcoded value until palette improvements */,
+      };
+    case 'new':
+      return {
+        color: tokens.colorWhite,
+        backgroundColor: tokens.colorPrimary,
+      };
+    default:
+      console.log(`Unknown EntityStatus: ${status}`);
```
gui-santos, line 35:
```suggestion
      console.warn(`Unknown EntityStatus: ${status}`);
```

Maybe warn? So it show very nice yellow message in the console 👌

packages/components/card/src/inline-entry-card/InlineEntryCard.tsx
```diff
@@ -0,0 +1,65 @@
...
+      bottom: 0,
+      content: '""',
+      display: 'block',
+      left: 0,
+      position: 'absolute',
+      top: 0,
+      width: tokens.spacing2Xs,
+    },
+  });
+};
```
gui-santos, line 31: Shouldn't we move all these `getStyles` functions to the `.styles.ts` files?

gui-santos (approved):
💎 

I have only two small suggestions that are not blocking this to be merged tbh

thebesson (approved):

packages/components/card/src/inline-entry-card/InlineEntryCard.tsx
denkristoffer: Yeah. Changed to a nice pattern I picked up from @bgutsol 👍

packages/components/card/src/asset-card/AssetCard.tsx
denkristoffer: I'm discussing this with @domarku

denkristoffer:
Merging, anything else can be done as a follow-up.

PR #1125, opened 2021-07-28
feat(TextLink) v4

bgutsol:
# Purpose of PR

Migrate `TextLink` component to v4

## PR Checklist

- [x] Tests are passing


packages/forma-36-react-components/src/components/TextLink/TextLink.tsx
```diff
@@ -40,6 +40,10 @@ export interface TextLinkProps {
   iconPosition?: IconPositionType;
 }
 
+/**
+ *
+ * @deprecated This component is deprecated, please use TextLink from '@contentful/f36-text-link'
+ */
 export function TextLink({
```
suevalov, line 47: Shouldn't we simply delete it?

packages/forma-36-react-components/src/components/TextLink/TextLink.stories.tsx
```diff
@@ -8,7 +8,7 @@ import '@contentful/forma-36-fcss/dist/styles.css';
 import { TextLink, textLinkColor } from './TextLink';
 
 export default {
-  title: 'Components/TextLink',
+  title: 'Components/TextLink(Deprecated)',
```
suevalov, line 11: let's remove it

packages/components/text-link/src/TextLink.styles.ts
```diff
@@ -0,0 +1,112 @@
+import { css } from 'emotion';
+import type { CSSObject } from '@emotion/serialize';
```
suevalov, line 2: Looks like this dependency is not specified in package.json, so it can't really be used here

packages/components/text-link/src/TextLink.styles.ts
```diff
@@ -0,0 +1,112 @@
...
+
+const textLinkIcon = css({
+  fill: 'currentColor',
+  transition: `fill ${tokens.transitionDurationShort} ${tokens.transitionEasingDefault}`,
+});
+
+const textLinkText = ({ alignIcon }: Pick<TextLinkProps, 'alignIcon'>) => {
+  if (alignIcon === 'start') {
+    return css({
+      marginLeft: '0.125rem',
```
suevalov, line 97: is there a suitable spacing value instead of a hardcoded number?

packages/components/text-link/src/TextLink.tsx
```diff
@@ -0,0 +1,103 @@
+import React, { HTMLProps } from 'react';
+import { cx } from 'emotion';
+import { Flex, CommonProps } from '@contentful/f36-core';
+import { styles } from './TextLink.styles';
+import { TextLinkVariant } from './types';
+import { Icon, IconComponent } from '@contentful/f36-icon';
+
+export interface TextLinkProps
+  extends Omit<
```
suevalov, line 9: I think we should use Polymorphic here to create proper typings based on `as` attribute passed here.

packages/components/text-link/src/TextLink.tsx
```diff
@@ -0,0 +1,103 @@
+import React, { HTMLProps } from 'react';
+import { cx } from 'emotion';
+import { Flex, CommonProps } from '@contentful/f36-core';
+import { styles } from './TextLink.styles';
+import { TextLinkVariant } from './types';
+import { Icon, IconComponent } from '@contentful/f36-icon';
```
suevalov, line 6: I guess `f36-icon` should be a dependency of this package, but I would argue that we need to change implementation here and require to pass the Icon component as an argument, but an icon name. 

The fact that we need to download all SVG to render a link is disturbing and we need to change that.

suevalov (changes requested):

packages/components/text-link/src/TextLink.styles.ts
bgutsol: no, I couldn't find any suitable common var.

packages/components/text-link/src/TextLink.tsx
bgutsol: What use cases do you see for `as` prop, that couldn't be covered with the current implementation? By current implementation I mean:
- pass href -> get `a`
- don't pass href -> get `button`
BTW the same approach used in the new Button component. For me there is no difference between the `Button` and `TextLink` components, only styling.
What do you think?

suevalov: I think it's a matter of consistency, all other basic components are using `as` attributes. A very cool feature of it that once you declare `as="a"` it automatically understands that `href`, `rel` and other specific attributes are possible, if it's button then we have attribute types for button.

If the new Button uses the same approach then we need to change that.

packages/components/text-link/src/types.ts
```diff
@@ -0,0 +1,6 @@
+import { ComponentVariant } from '@contentful/f36-core';
+
+export type TextLinkVariant =
+  | Exclude<ComponentVariant, 'warning'>
```
suevalov, line 4: I personally would always state all variants inline without using `ComponentVariant` but I don't have a strong objection.

suevalov (approved):
👍🏼  Good work!

P.S. I think we need to discuss snapshot tests. I see no point in using them anymore in this repository.

bgutsol:
> 👍🏼 Good work!
> 
> P.S. I think we need to discuss snapshot tests. I see no point in using them anymore in this repository.

Actually, I had the same concern. I will add it to our Monday PH topics.

packages/components/text-link/src/types.ts
bgutsol: Me neither 🙂 I just tried to keep consistency with `Button` https://github.com/contentful/forma-36/blob/37e988d17debe3499588e3b13975a347f212d824/packages/components/button/src/types.ts#L5-L7
We can talk about this as well

suevalov: yeah, let's talk

massao (approved):

packages/components/text-link/src/TextLink.tsx
denkristoffer: Sorry I'm late here, but I'm trying to understand why these changes were implemented. 

I can't see the code as it looked during this comment so maybe I'm overlooking something here, but the type `IconComponent` is just a wrapper for React's `ExoticComponent<any> | ComponentType<any>` – why would that or the `Icon` component itself require you to download all SVGs?

bgutsol: Maybe in the case if tree shaking is not working right, everything from `@contentful/f36-icon` ended up in the project bundle. Even if we import only one type and Icon component. 
But It just my guess, I didn't challenge enough that suggestion. 🤷‍♂️ 
Probably @suevalov have a better understanding.

PR #1127, opened 2021-07-28
feat(ToggleButton) V4

massao:
# Purpose of PR

Migrate `ToggleButton` to v4

## PR Checklist

- [x] Tests are passing


packages/components/button/src/ToggleButton/ToggleButton.tsx
```diff
@@ -0,0 +1,70 @@
+import React, { useState } from 'react';
+import { cx } from 'emotion';
+import { CommonProps } from '@contentful/f36-core';
+import { IconComponent } from '@contentful/f36-icon';
+import { Button } from '../Button';
+import getStyles from './ToggleButton.styles';
+
+export interface ToggleButtonProps extends CommonProps {
```
suevalov, line 8: shouldn't it just extend ButtonProps?

packages/components/button/src/ToggleButton/ToggleButton.tsx
```diff
@@ -0,0 +1,70 @@
...
+) {
+  const {
+    testId = 'cf-ui-toggle-button',
+    children,
+    className,
+    isDisabled,
+    isActive,
+    icon,
+    onToggle,
+  } = props;
```
suevalov, line 42: I think we need to destruct `{...rest}` and then apply `{...rest}` to the Button component.

suevalov (commented):
Please take a look at https://www.radix-ui.com/docs/primitives/components/toggle and apply the same a11y attributes.

Right now, it's just a button without any indication of the toggle state semantically.

packages/components/button/test/ToggleButton.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+      <ToggleButton className="my-extra-class">Toggle</ToggleButton>,
+    );
+
+    expect(container.firstChild).toMatchSnapshot();
+  });
+
+  it('renders the component active', () => {
+    const { container } = render(<ToggleButton isActive>Toggle</ToggleButton>);
+
+    expect(container.firstChild).toMatchSnapshot();
```
suevalov, line 26: a hot take, but don't we want to replace snapshots with queries verification? 

we could verify here that a particular aria attribute is present, for example

massao: uhmm thats true, I only kept like this as other components also used the snapshots tests too.

packages/components/button/src/ToggleButton/ToggleButton.tsx
massao: Actually was doing some testing, and I think it's better to have it extending only the `CommonProps`, since the toggle button should use the variants, or have different sizes. Extending the `ButtonProps` we would make it so those other props also be typed onto the `ToggleButton`

suevalov: Ok, if it makes sense more

suevalov (approved):
👍🏼

PR #1137, opened 2021-08-02
feat: improve Workbench a11y support

mshaaban0:
# Purpose of PR

Currently workbench is structured with divs only, this should be changed to use semantic html main aside  and header

We need to expose aria-label for these tags as well so screen reader users can determine the purpose of these tags

For more info see landmarks and page regions docs  https://www.w3.org/TR/WCAG20-TECHS/ARIA11.html

## PR Checklist

- [x] Tests are passing


burakukula (approved):
🎉

PR #1138, opened 2021-08-03
refactor: V4 button

massao:
# Purpose of PR

Update button v4 component:
* Better handle Polymorph.
* Use Icon as ReactElement.
Updated tests and removed snapshots from `button` package

## PR Checklist

- [x] Tests are passing


packages/components/button/src/types.ts
```diff
@@ -14,3 +16,69 @@ export type ButtonStylesProps = {
...
+  isActive?: boolean;
+  /**
+   * Disabled interaction and applies disabled styles
+   * @default false
+   */
+  isDisabled?: boolean;
+  /**
+   * Adds dropdown indicator icon
+   */
+  isDropdown?: boolean;
```
suevalov, line 45: I'm not sure about this property name... I'd rather make it `withChevron` or something like this. Or do we really need this property at all? Can't we just make a new component - `DropdownButton` that would pass ChevronDown as icon and inherit all other props?

packages/components/button/src/types.ts
```diff
@@ -14,3 +16,69 @@ export type ButtonStylesProps = {
...
+  /**
+   * The element used for the root node.
+   * @default button
+   */
+  as?: 'a' | 'button';
+  /**
+   * Button html type attribute
+   * @default button
+   */
+  type?: 'submit' | 'button' | 'reset';
```
suevalov, line 67: I guess that both `type` and `href` should be automatically interited from `as` property:

- `type` if `as="button"`
- `href`, `rel`, etc if as="a"

suevalov: Not sure if you need to specify it here explicitly

packages/components/button/src/types.ts
```diff
@@ -14,3 +16,69 @@ export type ButtonStylesProps = {
...
+   */
+  href?: string;
+}
+
+interface ButtonElementProps extends BaseButtonProps {
+  as?: 'button';
+  href?: never;
+}
+
+interface AnchorElementProps extends BaseButtonProps {
```
suevalov, line 79: If my assumption above is true, then this is not needed.

suevalov (commented):
I think after this PR, we need to create a new one that deletes a previous Button implementation from the main package.

suevalov:
Also, it seems that there were some changes on `master` branch that where not copied to the new implementation (css animation)... please update the branch with the latest changes in `forma-v4` and take a look if something needs to be ported.

burakukula:
@massao @suevalov Folks, a random question: `<Button icon={ChevronDown} />` will pass a11y requirements in the current implementation? We might need to force aria-label maybe for those buttons that are only icons? Or do we have that covered?

packages/components/button/src/types.ts
massao: Thats a good point, currently this prop only makes it so the right side icon shows up, if we go with the `DropdownButton` we would need to make it so we can align the icon to the right side, since currently it only renders the icon passed on the left side.
But the `withChevron` seems like a good solution too, but I don't have any strong preference over neither of them.

suevalov: We have `alignIcon` property in TextLink that we can bring here.

I prefer the DropdownButton solution more, because by doing this we can make Button package independent from icons package.

massao: Would the DropdownButton be a separate package?? Or would we need to make it so Button can accept two different icons to show on the left and right?

suevalov: I think it will be in the dropdown package.

suevalov (approved):

denkristoffer:
Hey @massao, sorry I'm late here. Can you explain the thoughts behind changing from the `IconComponent` type to generic `ReactElement`?

massao:
> Hey @massao, sorry I'm late here. Can you explain the thoughts behind changing from the `IconComponent` type to generic `ReactElement`?

Hey @denkristoffer this is to make it so we don't need to import all icons as dependencies for rendering buttons, and keep the same approach as on text-link: https://github.com/contentful/forma-36/pull/1125#discussion_r678929278

PR #1149, opened 2021-08-04
refactor: css-in-js enforcement

suevalov:
# Purpose of PR

Create styles only when a component is rendered.
Added ESLint rule to enforce this.
Removes old Button implementation and makes all other components to use new Button component.

**Open questions**

* Should we add `aria-hidden="true"` to all icons by default?
* Do we need to allow providing a custom icon size? 
* Should HelpText and ValidationMessage be polymorphic components?
* Do we want to inherit MarginProps or PaddingProps to components or do we want to leave it only for Box and Flex components?


packages/components/drag-handle/src/DragHandle.styles.ts
```diff
@@ -1,7 +1,7 @@
-import { css } from 'emotion';
+import { cx, css } from 'emotion';
 import tokens from '@contentful/f36-tokens';
 
-export const styles = {
+export const getStyles = () => ({
```
mshaaban0, line 4: doesn't this mean on each render emotion will generate the same styles with different classnames?

suevalov: no, it won't because class name value is a hash of properties and their values, new class name and styles will be generated only if values will be different.

massao (approved):

mshaaban0 (approved):

PR #1153, opened 2021-08-04
feat(tabs): migrate tabs to emotion

thebesson:
# Purpose of PR

This PR migrates Tabs to forma v4, using emotion for styling.

## PR Checklist

- [x] Tests are passing


thebesson:
Here are some questions regarding this component.

`Tabs` component is used both as Tabs: [https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel](tabs design pattern) and navigation: [https://www.w3.org/TR/wai-aria-practices-1.1/#aria_lh_navigation](navigation design pattern). 
 1. It may create some confusion about the usage; 
 2. it complicates implementation a bit, in my opinion; 
 3. navigation and tabs should have slightly different keyboard navigation. Currently, we use a standard Navigation pattern (tabbing through navigation items) keyboard navigation for tabs.
   
What I propose:
1. Create a separate `Navigation` component.
2. Remove all "navigation" logic from Tabs.
3. Implement at least [https://www.w3.org/TR/wai-aria-practices-1.1/examples/tabs/tabs-1/tabs.html](Automatic Activation Keyboard Navigation pattern), possibly in a separate PR.

packages/components/tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import type { CommonProps } from '@contentful/f36-core';
+
+import { getTabsStyles } from './Tabs.styles';
+export interface TabsProps extends CommonProps {
+  variant?: 'default' | 'horizontal' | 'vertical';
+  role?: 'navigation' | 'tablist';
```
thebesson, line 7: This prop defines if the component is navigation or tab list

packages/components/tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,42 @@
...
+  console.log('variant', variant);
+  console.log('styles', styles);
+
+  const elementProps = {
+    'data-test-id': testId,
+    className: styles.tabs,
+    style,
+  };
+
+  if (role === 'navigation') {
```
thebesson, line 29: If we are to implement some kind of hook to allow automatic selection of tabs, opposed to having a consumer handle it, it would add even more complexity.

thebesson:
I'm preparing a proposal for separating the navigation component from `tabs` component, to address the concerns from my comments. I want to do it in a separate PR so as not to increase the scope of the current PR.

packages/components/tabs/README.mdx
```diff
@@ -86,11 +89,25 @@ Be aware that the users will not see all the content at the same time. Make sure
...
 
 - every Tab should have concise copy
 - all content within a Tab should be related to it's label
 - to ensure an optimal cognitive load for users, do not use more than six (6) Tabs at any given time
+
+## Props
+
+### Tabs
+
+<Props of="Tabs" />
```
burakukula, line 105: Minor: for some reason props are not available in the deployed version. 🤔

packages/components/tabs/README.mdx
```diff
@@ -86,11 +89,25 @@ Be aware that the users will not see all the content at the same time. Make sure
...
+
+### Tab
+
+<Props of="Tab" />
+
+### TabPanel
+
+<Props of="Tab" />
```
burakukula, line 113: I think this should be most likely `<Props of="TabPanel" />` ? 🙂

packages/components/tabs/README.mdx
```diff
@@ -3,10 +3,13 @@ title: 'Tabs'
 type: 'component'
 status: 'stable'
 slug: /components/tabs/
-github: 'https://github.com/contentful/forma-36/tree/master/packages/forma-36-react-components/src/components/Tabs'
+github: 'https://github.com/contentful/forma-36/tree/master/packages/components/tabs'
 storybook: 'https://f36-storybook.contentful.com/?path=/story/components-tabs--default'
+typescript: ./src/TextLink.tsx
```
suevalov, line 8: the link is not correct

packages/components/tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import type { CommonProps } from '@contentful/f36-core';
+
+import { getTabsStyles } from './Tabs.styles';
+export interface TabsProps extends CommonProps {
+  variant?: 'default' | 'horizontal' | 'vertical';
```
suevalov, line 6: I don't think that the naming of the options is correct. If `variant` is `vertical`, it doesn't make tabs vertical, it just adds a vertical divider.

burakukula, line 6: I am a bit confused by these property values. Maybe it's just that it's about the divider and not really about the component itself and that's why I am confused. I think variant horizontal and vertical somehow suggest to me that it's about the whole component being vertical or horizontal. Dunno, maybe it's just me 🙃

thebesson: would it be better if property values would be called "verticalDivider" for example?

burakukula (approved):
@thebesson great work! 🤝
I am in favor of separating those 2 components if there is a different logic to cover a11y requirements and you this would be the best way to handle it. It would be also probably a bit easier to explain different use cases for those 2 components in the docs and create guidelines? Let's discuss this in the proposal. 🙂
I guess we could merge this PR and create a separate one to split those 2 components if decide to do so?

packages/components/tabs/src/Tabs.tsx
suevalov: I'm pretty sure that we need to implement this in v4, but we can do this in the follow up PR.

packages/components/tabs/src/Tabs.tsx
burakukula: It gives me clarity about what this variant actually does for the component, for sure. WDYT? Maybe other folks have a different perspective here. 🙂

packages/components/tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,40 @@
...
+  role?: 'navigation' | 'tablist';
+  children?: React.ReactNode;
+}
+
+export function Tabs({
```
suevalov, line 11: All components should be wrapped in `forwardRef`

packages/components/tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,40 @@
...
+  children,
+  variant = 'default',
+  testId = 'cf-ui-tabs',
+  role = 'tablist',
+  style,
+}: TabsProps): React.ReactElement {
+  const styles = getTabsStyles({ className, variant });
+
+  const elementProps = {
+    'data-test-id': testId,
```
suevalov, line 22: Ideally, we would use `Box` component inside, and just pass `testId={testId}` without providing a specific data attribute here.

suevalov (changes requested):

packages/components/tabs/src/Tabs.tsx
suevalov: yeah, I'd probably go with `vertical-divider`, `horizontal-divider` and `default`

packages/components/tabs/README.mdx
```diff
@@ -71,7 +74,7 @@ Be aware that the users will not see all the content at the same time. Make sure
 ### used with TabPanel
 
 ```jsx
-<>
+<div>
   <Tabs role="tablist">
     <Tab id="first" selected={selected === 'first'}>
```
suevalov, line 79: just realized that this example doesn't work because `selected` is not defined here. this PR would be a nice opportunity to make it work.

packages/components/tabs/src/Tab.tsx
```diff
@@ -1,35 +1,37 @@
...
-  selected = false,
-  style,
-  tabIndex = 0,
-  testId = 'cf-ui-tab',
-}: TabProps): React.ReactElement {
+function _Tab(
+  {
+    children,
+    className,
+    disabled = false,
```
suevalov, line 23: isDisabled?

packages/components/tabs/src/Tab.tsx
```diff
@@ -1,35 +1,37 @@
...
+    href,
+    id,
+    onSelect,
+    selected = false,
```
suevalov, line 27: isSelected?

packages/components/tabs/src/Tab.tsx
```diff
@@ -1,35 +1,37 @@
...
+    style,
+    tabIndex = 0,
+    testId = 'cf-ui-tab',
```
suevalov, line 30: it would be great to destuct `...otherProps` and apply them to the element as we do in all other places.

packages/components/tabs/README.mdx
thebesson: yes, just noticed it too

gui-santos:
I think @suevalov and @burakukula are already commenting everything that is really relevant
but while reviewing I had an idea and maybe it's something to be discussed with everyone
What if we use make `Tabs` to work with that pattern where the user would do the following?

```jsx
<Tabs>
  <Tabs.Tab>Tab I</Tabs.Tab>
  <Tabs.Panel>Panel of tabs I</Tabs.Panel>
</Tabs>
```

We already do that in `Workbench`, what if we do it for every one of these cases where we have multiple components to build a relevant UI?

suevalov:
@gui-santos In general, I like your proposal. We do the same with Modal as well. 

If we do it here, then we would need to do the same with the Table component.

gui-santos:
> If we do it here, then we would need to do the same with the Table component.

@suevalov Yes, I'm proposing to do this to every component where this pattern makes sense

suevalov (approved):

PR #1155, opened 2021-08-05
refactoring: more consistency and alignment about spacing props

suevalov:
# Purpose of PR

* Removes polymorphic possibilities from HelpText and ValidationMessage.
* Removes margin properties from all components (with an exception from Box, Flex, Grid, and Typography)
* Consistent way to inherit HTML properties via `PolymorphicComponentProps`


burakukula (approved):
👏

PR #1158, opened 2021-08-06
[BAU-49] fix: use dotenv only in development

gui-santos:
# Purpose of PR

In production, the website is ignoring the environment variables
And my suspicion is that the website is trying to use dotenv with a `.env.production` file that does not exist instead of getting the variables from Netlify
This PR will change that.... maybe

Also, since Osano’s script loads async we cannot hide the Cookie Preferences link by checking if Osano is there (It won't be there during load)
For now, I just made it show the link and I will think about a better way to handle this in another PR

## PR Checklist

- [ ] Tests are passing


suevalov:
@gui-santos 
https://www.gatsbyjs.com/docs/how-to/local-development/environment-variables/

suevalov:
> To expose a variable in the browser, you must preface its name with GATSBY_. So GATSBY_API_URL will be available in browser code but API_KEY will not.

gui-santos:
> > To expose a variable in the browser, you must preface its name with GATSBY_. So GATSBY_API_URL will be available in browser code but API_KEY will not.

@suevalov 🤦 

I totally missunderstood this yesterday

gui-santos:
@suevalov I updated the PR with `GATSBY_`, locally it does not work so I had to keep both versions of the env variables
let's see how this works in prod 🤷

suevalov:
@gui-santos why it doesn't work locally?

gui-santos:
> @gui-santos why it doesn't work locally?

@suevalov 
uhmmmm... tbh now I realize I will have to prefix then in the env files as well 🤦 
just forgot about how gatsby handles these things
now it should be fine hahahah

suevalov (approved):

PR #1059, opened 2021-07-13
feat: rename Paragraph to Text

denkristoffer:
# Purpose of PR

Renames `<Paragraph />` to `<Text />`. Closes #1030

PR #1159, opened 2021-08-06
feat: bring back paragraph component

suevalov:
# Purpose of PR

I think that our decision to rename Paragraph to Text was done prematurely. https://github.com/contentful/forma-36/pull/1059

We definitely need the Text component as our basic component for all typography components, but it shouldn't replace Paragraph by default.

1. Just renaming the component without changing the semantic is an unnecessary breaking change and an additional burden for the customers.

2. In many cases we need to have a component that comes with our default font-size, color, and letter spacing but without a default margin applied.

This PR rolls back the change done in #1059 

## Proposal for the follow-up PR

1. Create a new `<Text />` component that comes with default text styles, but you can change default text styles via props: `color`, `fontWeight`, `fontSize`, `letterSpacing`. By default, Text component has no margin and no padding and renders itself as `span`.

2. Rewrite all other typography components: `Paragraph`, `Heading`, etc to utilize `<Text />` under the hood.

3. Deprecate `<Typography>` component wrapper and apply default margin bottom to all typography components (except Text) no matter if they are wrapped in `Typography` or not.

4. **already done in v4** Allow all typography components to override the default margin.

```jsx
<Heading marginBottom="none" />
```


gui-santos (approved):
🆒

massao (approved):

burakukula (approved):
🙌

thebesson (approved):

PR #1161, opened 2021-08-06
chore: enforce callback codestyle

suevalov:
# Purpose of PR

According to our code style, all event handlers should start from 'handle' and properties should start from 'on'.

This PR introduces an ESLint rule that verifies that.

### DIFF
--- a/packages/forma-36-react-components/src/components/Modal/Modal.tsx
+++ b/packages/forma-36-react-components/src/components/Modal/Modal.tsx
@@ -94,8 +94,8 @@ export function Modal({
   testId,
   topOffset,
   ...otherProps
-}: ModalProps): React.ReactElement {
-  const allProps = {
+}: ModalProps) {
+  const props = {
     ...otherProps,
     allowHeightOverflow,
     position,
@@ -112,7 +112,7 @@ export function Modal({
         {otherProps.title && (
           <ModalHeader
             title={otherProps.title}
-            onClose={otherProps.onClose}
+            onClose={props.onClose}
             {...otherProps.modalHeaderProps}
           />
         )}
@@ -126,9 +126,9 @@ export function Modal({
   return (
     <ReactModal
       ariaHideApp={false}
-      onRequestClose={otherProps.onClose}
+      onRequestClose={props.onClose}
       isOpen={otherProps.isShown}
-      onAfterOpen={otherProps.onAfterOpen}
+      onAfterOpen={props.onAfterOpen}
       shouldCloseOnEsc={shouldCloseOnEscapePress}
       shouldCloseOnOverlayClick={shouldCloseOnOverlayClick}
       portalClassName={styles.Modal__portal}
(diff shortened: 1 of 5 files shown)


packages/forma-36-react-components/src/components/Modal/Modal.tsx
```diff
@@ -94,8 +94,8 @@ export function Modal({
   testId,
   topOffset,
   ...otherProps
-}: ModalProps): React.ReactElement {
-  const allProps = {
+}: ModalProps) {
+  const props = {
```
gui-santos, line 98: the props are being destructured in the argument of the function and then later we putting then together in a `props` variable again

I know this was already there before your PR, but could you fix this here as well?

packages/forma-36-react-components/src/components/Notification/NotificationsManager.tsx
```diff
@@ -177,6 +177,7 @@ export function NotificationsManager({
             duration={item.duration}
             key={item.id}
             hasCloseButton={item.canClose}
+            // eslint-disable-next-line react/jsx-handler-names
             onClose={item.close}
```
gui-santos, line 181: Why not renaming this one?

gui-santos (changes requested):

packages/forma-36-react-components/src/components/Notification/NotificationsManager.tsx
suevalov: because an item has a property called `.close` and we can not change it

packages/forma-36-react-components/src/components/Modal/Modal.tsx
suevalov: I'd prefer to fix it when we'll be migrating it to v4

packages/forma-36-react-components/src/components/Notification/NotificationsManager.tsx
gui-santos: Why can't we change it? I thought that was the reason for having the `forma-v4` branch

packages/forma-36-react-components/src/components/Modal/Modal.tsx
gui-santos: I'm confused, isn't this PR merging to `forma-v4` branch?

what do you mean by "migrating it to v4" if it's not merging to v4 branch?

suevalov: to correct myself: while migrating the component to Emotion and to a separate package.

gui-santos: got it! ok then 😃

packages/forma-36-react-components/src/components/Notification/NotificationsManager.tsx
suevalov: because this is a part of the public API for Notification component

```
const notification = Notification.success();
notification.close();
```

gui-santos: I still don't get it 100%, but I guess it's only me so no need for any further explanations ✌️

thebesson (approved):

massao (approved):

gui-santos (approved):

PR #1163, opened 2021-08-06
feat: Typography changes

suevalov:
# Purpose of PR

As discussed in https://github.com/contentful/forma-36/pull/1159, this PR introduces the following changes in the Typography system:

1. Create a new Text component that comes with default text styles, but you can change default text styles via props: fontColor, fontWeight, fontSize, letterSpacing. By default, Text component has no margin and no padding and renders itself as span. 
2. Rewrite all other typography components: Paragraph, Heading, etc to utilize Text under the hood. 
3. Deprecate Typography component wrapper and apply default margin bottom to all typography components (except Text) no matter if they are wrapped in Typography or not.


thebesson (approved):
Looks much neater now! 👏

massao (approved):

gui-santos (approved):
🎨

PR #1175, opened 2021-08-11
feat(copybutton): migrate copybutton to v4

massao:
# Purpose of PR

Migrate copybutton to v4

## PR Checklist

- [x] Tests are passing


packages/components/copybutton/src/CopyButton.tsx
```diff
@@ -0,0 +1,109 @@
...
+  tooltipText?: string;
+  /**
+   * Props that are passed to the tooltip component
+   */
+  tooltipProps?: Omit<TooltipProps, 'content' | 'children'>;
+  /**
+   * HTML element used to wrap the CopyButton
+   * @default div
+   */
+  as?: React.ElementType;
```
suevalov, line 36: Do we need `as` functionality here? I'd say no. As for me, it should always be a `button`

packages/components/copybutton/src/CopyButton.tsx
```diff
@@ -0,0 +1,109 @@
...
+      data-test-id={testId}
+      className={cx(styles.wrapper, className)}
+      {...otherProps}
+    >
+      <CopyToClipboard text={value} onCopy={handleOnCopy}>
+        <Tooltip
+          content={copied ? tooltipCopiedText : tooltipText}
+          {...tooltipProps}
+        >
+          <button
```
suevalov, line 95: Can't we use our `<Button>` component here?

packages/forma-36-react-components/src/components/TextField/__snapshots__/TextField.test.tsx.snap
```diff
@@ -51,13 +51,52 @@ exports[`renders the component with a copybutton 1`] = `
   margin-bottom: calc(1rem * (8 / 16));
 }
 
-.emotion-3 > button[disabled] {
-  pointer-events: none;
+.emotion-4 {
```
suevalov, line 54: Snapshots tests are back?

massao: This is from the `react-components` package, I only removed from the already migrated ones, but there are some components from there that are using the CopyButton

suevalov: Ah, I see. Makes sense now.

packages/components/copybutton/src/CopyButton.tsx
massao: We can, but I wasn't sure if it would make sense to add `Button` as a dependency to this package, if we are only going to use one variant, and size. And adding it as a dependency would also include any other dependency that the `@contentful/f36-button` may have.
But I don't have any strong preferences tho.

suevalov: got it, up to you. I believe your ability to make a right call :)

packages/components/copybutton/src/CopyButton.tsx
```diff
@@ -0,0 +1,109 @@
...
+        }
+      }, 1000);
+    },
+    [onCopy],
+  );
+
+  return (
+    <Box
+      as={as}
+      ref={ref}
```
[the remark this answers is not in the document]
denkristoffer: As much as I want to agree with that I don't like the idea of starting to decide where refs go or don't go. I think we should stick to the system of sending refs to the root element and then take a ref in `buttonProps` if necessary.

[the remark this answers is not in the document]
massao: Also just noticed that if we want to keep the functionality of unfocusing the button after the value is copied, we would need to access the button ref inside the component, as it's at the moment.
Or do we want to not call blur on the button??

packages/components/copybutton/src/CopyButton.tsx
massao: I'll keep as it is for now, but if we see the need to have the button variants here, we can update it to ur our `<Button>`

packages/components/copybutton/src/CopyButton.tsx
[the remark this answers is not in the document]
denkristoffer: My initial thought is that we shouldn't do anything uncommon just for the sake of it. @mshaaban0 do we want to blur?

suevalov (approved):

PR #1176, opened 2021-08-12
feat(modallauncher): add closeAll method to ModalLauncher

massao:
# Purpose of PR

Add `closeAll` method to the `ModalLauncher`.
It works by storing all the root elements ids, on a Map, when the `open` method is called, and iterating over the map when the `closeAll` method is called to render with `{ isShown: false }`, and waiting for the set delay before removing the element from the tree, so it keeps the closing animation.
Also removes the value from the map in case the modal is closed through the onClosed.

## PR Checklist

- [x] Tests are passing


suevalov (approved):
Looks good, thanks for a quick implementation.

PR #1188, opened 2021-08-16
feat: text link codemod 

suevalov:
# Purpose of PR

Adds a codemod to migrate TextLink component to v4.
Adds reusable functions that can be used in other codemods.

### DIFF
--- a/packages/forma-36-codemod/transforms/v4-text-link.js
+++ b/packages/forma-36-codemod/transforms/v4-text-link.js
@@ -46,9 +46,10 @@ module.exports = function (file, api) {
           if (value.value === 'right') {
             return j.literal('end');
           }
-          console.warn("TextLink: Couldn't automatically update alignIcon value");
+          console.warn(
+            "TextLink: Couldn't automatically update alignIcon value",
+          );
           return value;
-          }
         },
       });
 
--- a/packages/forma-36-codemod/transforms/v4-text-link.js
+++ b/packages/forma-36-codemod/transforms/v4-text-link.js
@@ -1,0 +1,14 @@
+const { getComponentLocalName } = require('../utils/getComponentLocalName');
+
+module.exports = function (file, api) {
+  const j = api.jscodeshift;
+
+  const localName = getComponentLocalName(j, file.source, {
+    componentName: 'TextLink',
+    from: '@contentful/forma-36-react-components',
+  });
+
+  console.log(localName);
+
+  return j(file.source).toSource();
+};
--- a/packages/forma-36-codemod/transforms/v4-text-link.js
+++ b/packages/forma-36-codemod/transforms/v4-text-link.js
@@ -1,5 +1,10 @@
-const { getComponentLocalName } = require('../utils/getComponentLocalName');
-const { renameProperties } = require('../utils/renameProperties');
+const {
+  getComponentLocalName,
+  renameProperties,
+  hasProperty,
+  addProperty,
+  changeProperties,
+} = require('../utils');
 
 module.exports = function (file, api) {
   const j = api.jscodeshift;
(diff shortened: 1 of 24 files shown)


packages/forma-36-codemod/transforms/v4-text-link.js
```diff
@@ -0,0 +1,84 @@
...
+          j,
+          icons: usedIcons,
+        });
+      }
+
+      return modifiedAttributes;
+    },
+  });
+
+  source = changeImport(j, source, {
```
suevalov, line 75: Probably sometimes we don't want a change of the import to happen... For example, in Compose codebase. It looks like we need to make it controllable.

packages/forma-36-codemod/transforms/v4-text-link.js
```diff
@@ -0,0 +1,84 @@
...
+    componentName,
+    from: '@contentful/forma-36-react-components',
+    to: '@contentful/f36-components',
+  });
+
+  source = addIconImports({ j, source, icons: usedIcons });
```
suevalov, line 81: Looks like some icon components names can clash with Forma component names:

* Heading
* Asset

Should we add postfix all icon components with with `Icon`?

`Heading` => `HeadingIcon`
`Asset` => `AssetIcon`

packages/forma-36-codemod/utils/index.js
```diff
@@ -0,0 +1,28 @@
...
+const { addProperty } = require('./addProperty');
+const { changeProperties } = require('./changeProperties');
+const { updatePropertyValue } = require('./updatePropertyValue');
+const { getProperty } = require('./getProperty');
+const { addImport } = require('./addImport');
+const { updateIcons, addIconImports } = require('./updateIcons');
+const { removeComponentImport } = require('./removeComponentImport');
+const { changeImport } = require('./changeImport');
+
+module.exports = {
```
burakukula, line 14: Awesome! 👏

burakukula (approved):
🎉 
Only one thing: we might want to extend the README a bit to cover your changes.

packages/forma-36-codemod/transforms/v4-text-link.js
burakukula: yeah, I think it would cover any potential cases like that in the future.

massao (approved):
👏

packages/forma-36-codemod/utils/getProperty.js
```diff
@@ -0,0 +1,16 @@
...
+
+  attributes.forEach((attribute) => {
+    if (found === null) {
+      if (attribute.name.name === propertyName) {
+        found = attribute;
+      }
+    }
+  });
+
+  return found;
```
bgutsol, line 15: Why not just:
```suggestion
  const found = attributes.find((attribute) => attribute.name.name === propertyName)

  return found || null;
```

packages/forma-36-codemod/transforms/v4-text-link.js
bgutsol: +1 to add postfix

bgutsol (commented):
Great work with utils functions 👏
One important comment for the `TextLink` codemod is to add `as="button"` to the places where we use `TextLink` without `href` prop. 
[Before the default was `button` element when you not provide `href`](https://github.com/contentful/forma-36/pull/1125/files#diff-2ecb4e98406d62467375ddf356870f24199fcc08728d2eee4d83c9fd7b82c442L92)
[In v4 the default is `a` element even if the `href` is not provided](https://github.com/contentful/forma-36/pull/1125/files#diff-3bbbaafb6051437ff8b342074584b898af8f71a6401a8c28f34cc34bcff9ce4bR38)

I did it, because we added `as` prop and IMO it should be the only one prop that defines component element. But we can also revisit component implementation if it's not convenient for us.

packages/forma-36-codemod/utils/getProperty.js
suevalov: cause Yoda I like :)

packages/forma-36-codemod/transforms/v4-text-link.js
suevalov: going to changes this in a subsequent PR

suevalov:
> One important comment for the TextLink codemod is to add as="button" to the places where we use TextLink without href prop.

Good remark! Changed the implementation, but we will need to learn if it's a good default while we're adopting it.

bgutsol (approved):

PR #1191, opened 2021-08-17
feat: Add postfix to all icon components

suevalov:
# Purpose of PR

As we found out while writing codemod for TextLink, current icons naming may collapse with Forma components. 

For example, `Asset` or `Heading`. This PR added `Icon` postfix to all icon components to avoid clashing names.


bgutsol (approved):
🥇

PR #1194, opened 2021-08-17
docs: provide README for NPM purposes

burakukula:
Since we started the internal testing of the v4 packages, we would like to provide small info for users that those packages are part of pre-release to prevent from using them. They can be found in NPM right now. 
We would rather avoid too extended communication, but still provide some info in the NPM.

NPM is picking up `README.md` files. I moved existing docs to `Component.mdx` which will be picked up by the website and added `README.md` which will be picked up by NPM. WDYT?


packages/forma-36-website/gatsby-node.js
```diff
@@ -36,7 +36,7 @@ exports.onPostBuild = () => {
 
 exports.createPages = ({ graphql, actions }) => {
   const { createPage } = actions;
-
+  console.log('graphql', graphql);
```
suevalov, line 39: probably should be deleted, right?

burakukula: ups

suevalov (approved):

denkristoffer:
Is the plan to change this back after release or what?

gui-santos (approved):
Looks good!

I suggest updating the plop-template as well
so when we create a new v4 package this message will be there from the start

burakukula:
> Changes look good 👍 Is the plan to change this back after release or what?

@denkristoffer i guess we would roll back, not really sure how we would approach it. We can make a decision later I guess.

PR #1196, opened 2021-08-17
feat: allow to migrate Tag to Badge by using codmod

burakukula:
Allow to migrate Tag to Badge. 

TODO:

`<Tag entityStatusType="draft">Draft</Tag>;` to `<EntityStatusBadge entityStatus="draft" />`


packages/forma-36-codemod/bin/inquirer-choices.js
```diff
@@ -19,6 +19,11 @@ const TRANSFORMS_CHOICES = [
     name: 'v4-text-link: Converts TextLink component from Forma v3 to v4',
     value: 'v4-text-link',
   },
+  {
+    name:
+      'v4-badge: Converts Tag component from Forma v3 to Badge component from v4',
+    value: 'v4-text-link',
```
suevalov, line 25:
```suggestion
    value: 'v4-badge',
```

packages/forma-36-codemod/transforms/v4-badge.js
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  source = changeProperties(j, source, {
+    componentName,
+    fn(attributes) {
+      let modifiedAttributes = attributes;
+
+      modifiedAttributes = renameProperties(modifiedAttributes, {
+        renameMap: {
+          tagType: 'variant',
```
suevalov, line 31: do we have the same variant values as they were in tagType? no renamed values?

packages/forma-36-codemod/utils/changeImportWithComponentName.js
```diff
@@ -0,0 +1,18 @@
+const { addImport } = require('./addImport');
+const { removeComponentImport } = require('./removeComponentImport');
+
+module.exports.changeImportWithComponentName = function changeImport(
```
suevalov, line 4: can we just modify `changeImport` function to accept optional outputComponentName property?

packages/forma-36-codemod/transforms/v4-badge.js
```diff
@@ -0,0 +1,56 @@
...
+      return modifiedAttributes;
+    },
+  });
+
+  source = changeComponentName(j, source, {
+    componentName,
+    outputComponentName: 'Badge',
+  });
+
+  // TODO: add and option to migrate <Tag entityStatusType="draft">Draft</Tag>; to <EntityStatusBadge entityStatus="draft" />;
```
suevalov, line 44: so, basically, we need to check if the component has `entityStatusType` property then we remove children and rename it to Badge?

packages/forma-36-codemod/transforms/v4-badge.js
```diff
@@ -0,0 +1,56 @@
...
+  getComponentLocalName,
+  renameProperties,
+  changeProperties,
+  changeImportWithComponentName,
+  changeComponentName,
+} = require('../utils');
+const { getFormaImport, shouldSkipUpdateImport } = require('../utils/config');
+
+module.exports = function (file, api) {
+  const j = api.jscodeshift;
```
gui-santos, line 11: What do you think about renaming this from `j` to `codeshift`?

I'm a pro descriptive names person hahaha

packages/forma-36-codemod/utils/changeComponentName.js
```diff
@@ -0,0 +1,15 @@
+module.exports.changeComponentName = function changeProperties(
+  j,
```
gui-santos, line 2: Same suggestion! It's hard to know what `j` is
I think a descriptive name would help us to maintain this code

packages/forma-36-codemod/transforms/v4-badge.js
burakukula: yeah, they are the same. 
But now when I think about it, we should actually remove `muted` since we deprecate it in v3. I will confirm that with the designers. but this change would need to be done in the API of the v4 Badge.
```
export type BadgeVariant =
  | 'negative'
  | 'positive'
  | 'primary'
  | 'secondary'
  | 'warning'
  | 'muted'
  | 'primary-filled'
  | 'featured';
```

packages/forma-36-codemod/transforms/v4-badge.js
burakukula: yeah, I guess that's it.

burakukula: I will give it a try.

packages/forma-36-codemod/utils/changeImportWithComponentName.js
```diff
@@ -0,0 +1,18 @@
...
+) {
+  let result = removeComponentImport(j, source, {
+    componentName,
+    importName: from,
+  });
+
+  const importResult = addImport(j, result, [
+    j.template.statement([`import { ${outputComponentName} } from "${to}"`]),
+  ]);
+  return importResult.source;
```
gui-santos, line 17: I think this codemod thing is magical, I still need to do one

I have one question about this two functions, I understand that it will replace old imports
but what happens in a scenario like this?

```jsx
import { Button, ComponentToReplace } from '@contentful/f36-components'
```

Will it do this?

```jsx
import { Button, ReplacedComponent } from '@contentful/f36-components'
```

Or will it create a new duplicated import?

```jsx
import { Button } from '@contentful/f36-components'
import { ReplacedComponent } from '@contentful/f36-components'
```

packages/forma-36-codemod/utils/changeImportWithComponentName.js
burakukula: Sure. Dunno, thought it would be useful to have a separate function for some reason 🙂. I'm gonna extend the `changeImport`.

gui-santos (approved):
LGTM

But I never used codemod, so I would wait for other people's reviews

I made a few questions because I'm not familiar with the tool, so maybe they will be irrelevant hahahah

packages/forma-36-codemod/transforms/v4-badge.js
burakukula: sure, of course. I would do it globally than in the whole package. Maybe separate PR?

packages/forma-36-codemod/utils/changeImportWithComponentName.js
burakukula: @gui-santos I don't understand the scenario tbh. What we do here is to replace:
```
import { Tag } from "@contentful/forma-36-react-components";
```

with 

```
import { Badge } from "@contentful/f36-components";
```

packages/forma-36-codemod/transforms/v4-badge.js
suevalov: it could a generic `removeChildren` function or something like this. if you'll struggle implementing that, we can take a look together.

packages/forma-36-codemod/transforms/v4-badge.js
suevalov: `j` is a common name in all codeshift examples you can find on Github

packages/forma-36-codemod/utils/changeImportWithComponentName.js
burakukula: ✅

packages/forma-36-codemod/utils/changeImportWithComponentName.js
suevalov: it should do `import { Button, ReplacedComponent } from '@contentful/f36-components'`

gui-santos: Ok, let's use `Tag` in the example, so what happens when I have something like this?

```jsx
import { Button, Heading, Table, Tag } from '@contentful/f36-components'
```

We only want to replace `Tag`, but it's being imported with other components
And by reading the code I understood (or misunderstood) that we are looking specifically for places in the code where `Tag` is being imported alone

packages/forma-36-codemod/utils/changeImport.js
```diff
@@ -4,15 +4,20 @@ const { removeComponentImport } = require('./removeComponentImport');
...
     componentName,
     importName: from,
   });
 
   const importResult = addImport(j, result, [
-    j.template.statement([`import { ${componentName} } from "${to}"`]),
+    j.template.statement([
+      `import { ${
+        outputComponentName ? outputComponentName : componentName
+      } } from "${to}"`,
```
massao, line 18: Wouldn't this remove other imports, in case more than one component is being imported from that package?

packages/forma-36-codemod/utils/changeImportWithComponentName.js
massao: This could be covered on the tests, with the fixtures, add another component being imported and checking what the output would be like.

packages/forma-36-codemod/transforms/v4-badge.js
gui-santos: got it!
Nevermind then

packages/forma-36-codemod/utils/changeImport.js
gui-santos: ☝️  That's what wanted to ask in my other comment hahaha
Massao just did it better 👏

burakukula: Got it now. Sorry, didn't get it what you mean at the beginning. Yup, good point, I will check it out.

massao (approved):
LGTM

PR #1197, opened 2021-08-17
fix: small text link adjustments

suevalov:
# Purpose of PR

Fixes small regressions that were found while integrating the TextLink component.

### DIFF
--- a/packages/components/icons/src/Gift.tsx
+++ b/packages/components/icons/src/Gift.tsx
@@ -3,6 +3,7 @@ import { generateIcon } from '@contentful/f36-icon';
 
 export const Gift = generateIcon({
   name: 'Gift',
+  viewBox: '0 0 18 18',
   path: (
     <Fragment>
       <g clipPath="url(#prefix__clip0)" fillRule="evenodd" clipRule="evenodd">
--- a/packages/components/text-link/src/TextLink.tsx
+++ b/packages/components/text-link/src/TextLink.tsx
@@ -109,7 +109,18 @@ const TextLink: PolymorphicComponentWithRef<
   }
 
   return (
-    <a {...commonProps} href={href}>
+    <a
+      {...commonProps}
+      onClick={
+        isDisabled
+          ? (e) => {
+              e.preventDefault();
+            }
+          : commonProps.onClick
+      }
+      href={href}
+      {...(isDisabled ? { tabIndex: -1 } : {})}
+    >
       {commonContent}
     </a>
   );
--- a/packages/components/text-link/src/TextLink.tsx
+++ b/packages/components/text-link/src/TextLink.tsx
@@ -79,7 +79,7 @@ const TextLink: PolymorphicComponentWithRef<
     <Flex as="span">
       {React.cloneElement(icon, {
         className: cx(icon.props.className, styles.textLinkIcon()),
-        size: 'tiny',
+        size: 'small',
       })}
     </Flex>
   ) : null;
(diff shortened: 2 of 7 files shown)


packages/components/text-link/src/TextLink.tsx
```diff
@@ -109,7 +109,18 @@ const TextLink: PolymorphicComponentWithRef<
...
+      {...commonProps}
+      onClick={
+        isDisabled
+          ? (e) => {
+              e.preventDefault();
+            }
+          : commonProps.onClick
+      }
+      href={href}
+      {...(isDisabled ? { tabIndex: -1 } : {})}
```
suevalov, line 122: so, disabled anchor cannot be focused

packages/components/text-link/src/TextLink.tsx
```diff
@@ -109,7 +109,18 @@ const TextLink: PolymorphicComponentWithRef<
   }
 
   return (
-    <a {...commonProps} href={href}>
+    <a
+      {...commonProps}
+      onClick={
```
suevalov, line 114: so disabled anchor cannot be clicked

denkristoffer: Didn't we render disabled TextLink as `span` before?

suevalov: no, we didn't

packages/components/icons/src/Gift.tsx
```diff
@@ -3,6 +3,7 @@ import { generateIcon } from '@contentful/f36-icon';
 
 export const Gift = generateIcon({
   name: 'Gift',
+  viewBox: '0 0 18 18',
```
burakukula, line 6: 🤔 why do we need this viewBox?

suevalov: because it was like that in v3 and it wasn't converted properly

burakukula (approved):
LGTM

packages/components/text-link/src/TextLink.tsx
denkristoffer: Okay, well that would be my suggestion here to get these behaviours automatically.

gui-santos (approved):

packages/components/text-link/src/TextLink.tsx
suevalov: I guess we'll have some typings problems in this case and we'd need to filter out some properties that are related only to `a` tag. I'd rather keep it this way.

gui-santos: Oh! I already approved the PR, but this ☝️  seems like a good idea

massao (approved):

packages/components/text-link/src/TextLink.tsx
suevalov: one more argument against it: we would still need to remove `onClick` handler on `span` element.

bgutsol (approved):
All good!
My only comment is to notify @domarku to approve this design changes (icon size and margin) and update Figma files, so we keep design and components in consistency

domarku:
Could we wait for the merge until later in the day? I want to understand the initial design decision from @fabe before we make the change. Thanks!

suevalov:
@domarku yes, we can

packages/components/text-link/src/TextLink.tsx
denkristoffer: I agree it's pretty bothersome for types 🙁

domarku:
Change approved!

PR #1200, opened 2021-08-18
feat: migrate datetime to v4

gui-santos:
# Purpose of PR

It migrates `DateTime` and `RelativeDate` to v4

- created `f36-datetime` package
- removed snapshots 
- updated README and stories
- DateTime:
  - changed `format` prop values to **not** use upper case values (I will update this in the code style guide) 
  - forwardRef
- RelateDate
  - renamed it to `RelativeDateTime`
  - forwardRef 
  - created `isRelativeToCurrentWeek` to support use case in Contentful internal products
  
## `isRelativeToCurrentWeek` in action

date_in_current_week

the date is within the current week

date_NOT_in_current_week

the date is NOT within the current week

## Where is dateUtils?

I will migrate those in a separated PR 🙏 

## Open Questions

Should we rename `DateTime` to `FormattedDateTime`?
I'm asking that because in the other repos that are not using f36 in contentful, they have their own components and that's how they are called

## PR Checklist

- [ ] Tests are passing


packages/components/datetime/DateTime.mdx
```diff
@@ -0,0 +1,80 @@
+---
+title: 'DateTime'
+type: 'component'
+status: 'stable'
+slug: /components/datetime/
+github: 'https://github.com/contentful/forma-36/tree/master/packages/components/components/datetime'
```
suevalov, line 6:
```suggestion
github: 'https://github.com/contentful/forma-36/tree/master/packages/components/datetime'
```

packages/components/datetime/DateTime.mdx
```diff
@@ -0,0 +1,80 @@
...
+storybook: 'https://f36-storybook.contentful.com/?path=/story/components-datetime--basic'
+---
```
suevalov, line 8: please add `typescript:` metadata and use PropsTable to visualize props

packages/components/datetime/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "@contentful/f36-datetime",
+  "version": "4.0.0",
+  "private": true,
```
suevalov, line 4: please, remove `private: true`, the plopfile got update already.

packages/components/datetime/package.json
```diff
@@ -0,0 +1,40 @@
...
+    "build": "parcel build index.ts"
+  },
+  "dependencies": {
+    "@babel/runtime": "^7.6.2",
+    "@contentful/f36-core": "^4.0.0",
+    "@contentful/f36-tokens": "^4.0.0",
+    "emotion": "^10.0.17"
+  },
+  "peerDependencies": {
+    "dayjs": "^1.10.6",
```
suevalov, line 16: I think it should be in `dependencies`

packages/components/datetime/src/DateTime.tsx
```diff
@@ -0,0 +1,60 @@
...
+   * The date that will be displayed. It accepts a JS Date, an ISO8601 Timestamp string, or Unix Epoch Milliseconds number
+   */
+  date: Date | string | number;
+  /**
+   * The format in which the date will be presented
+   *
+   * @default 'full'
+   **/
+  format?: 'full' | 'time' | 'weekday' | 'day';
+  testId?: string;
```
suevalov, line 19: both testId and className are available in CommonProps already.

packages/components/datetime/src/DateTime.tsx
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+import { CommonProps } from '@contentful/f36-core';
+
+import dayjs, { extend } from 'dayjs';
+import utcPlugin from 'dayjs/plugin/utc';
+extend(utcPlugin);
+
+export interface DateTimeProps extends CommonProps {
```
suevalov, line 8: do we want to extend it with Polymorphic<'time'> maybe to obtain all attributes that we would normally have for `<time>` tag?

packages/components/datetime/src/DateTime.tsx
```diff
@@ -0,0 +1,60 @@
...
+      break;
+    default:
+      template = 'ddd, DD MMM YYYY [at] h:mm A'; // Tue, 17 Aug 2021 at 3:45 PM
+  }
+
+  return dayjs(date).format(template);
+}
+
+function _DateTime(
+  { date, format = 'full', className, testId = 'f36-date-time' }: DateTimeProps,
```
suevalov, line 47: `testId` value doesn't follow `cf-ui` pattern

suevalov: we need to use {...otherProps} here

packages/components/datetime/src/DateTime.tsx
```diff
@@ -0,0 +1,60 @@
...
+  ref: React.Ref<HTMLTimeElement>,
+) {
+  return (
+    <time className={className} data-test-id={testId} ref={ref}>
```
suevalov, line 51: we could use `datatime` attribute here


https://www.w3schools.com/tags/tag_time.asp

> The datetime attribute of this element is used translate the time into a machine-readable format so that browsers can offer to add date reminders through the user's calendar, and search engines can produce smarter search results.

suevalov: `<Box as="time" testId={testId}> `, so you don't need to hardcode our `data-test-id` value

packages/components/datetime/src/RelativeDateTime.tsx
```diff
@@ -0,0 +1,77 @@
...
+  baseDate?: Date | string | number;
+  /**
+   * Sets the date to be relative only if it is in the current week
+   */
+  isRelativeToCurrentWeek?: boolean;
+  className?: string;
+  testId?: string;
+}
+
+const now = new Date();
```
suevalov, line 33: :important:

`now` should be calculated at the moment of rendering. move it to the component.

packages/components/datetime/src/RelativeDateTime.tsx
```diff
@@ -0,0 +1,77 @@
...
+import utcPlugin from 'dayjs/plugin/utc';
+import relativeTime from 'dayjs/plugin/relativeTime';
+import calendarPlugin from 'dayjs/plugin/calendar';
+import isTodayPlugin from 'dayjs/plugin/isToday';
+dayjs.extend(utcPlugin);
+dayjs.extend(relativeTime);
+dayjs.extend(calendarPlugin);
+dayjs.extend(isTodayPlugin);
+
+export interface RelativeDateTimeProps {
```
suevalov, line 13: All comments for `DataTime` are relevant to this file too.

suevalov (changes requested):

packages/components/datetime/DateTime.mdx
```diff
@@ -0,0 +1,80 @@
...
+
+## RelativeDateTime
+
+Provides a relative date string for resolutions ranging from seconds up through years. Should be used when absolute date and time is unnecessarily specific.
+
+### Examples
+
+If we consider today as August 18th of 2021 and we use pass the following values to the Component
+
+```tsx static=true
```
massao, line 44: Do we want to only show the code here? Or should we also show the rendered result?

packages/components/datetime/DateTime.mdx
gui-santos: got it!

packages/components/datetime/package.json
gui-santos: Oh really? Wouldn't that duplicate dayjs if the project using this already have dayjs? I'm always confused with this stuff

But I can fix it of course

packages/components/datetime/src/DateTime.tsx
gui-santos: ahhhh I guess another thing to update in the plop template

I will remove it

packages/components/datetime/src/DateTime.tsx
gui-santos: uhmmmm I think so, right? Good idea

packages/components/datetime/src/DateTime.tsx
gui-santos: yep yep!

packages/components/datetime/src/DateTime.tsx
gui-santos: uhmmm I thought we only needed `datetime` in the RelativeDateTime component

I will fix both things
I guess the second one will get fixed with the polymorphic solution 👌

packages/components/datetime/src/RelativeDateTime.tsx
gui-santos: 🤦

packages/components/datetime/src/RelativeDateTime.tsx
gui-santos: Do you mean polymorphic and common props?

packages/components/datetime/DateTime.mdx
gui-santos: 🤔 

I guess the rendered result as well

I was confused about the whole `static=true` thing
tbh, when do we want to show only the code?

massao: I believe that in most of the cases we will want to show the rendered component, we may only want to show the code when it are only snippets that don't necessary render a component.

packages/components/datetime/src/DateTime.tsx
gui-santos: Ok... I think I don't know how to use the types here

Because we want to allow the user to pass the attributes that the `time` tag has, right?
BUT we do not want to allow them to use `as` prop

Am I going to have to type this without using stuff from `f36-core/Primitive`?

packages/components/datetime/package.json
gui-santos: I changed it, but let me know about the duplication thing

packages/components/datetime/src/DateTime.tsx
suevalov: it's better to use Polymorphic props here

packages/components/datetime/package.json
suevalov: It won't duplicate. `^` version is rather relaxed range

packages/components/datetime/DateTime.mdx
```diff
@@ -0,0 +1,81 @@
+---
+title: 'DateTime'
+type: 'component'
+status: 'stable'
+slug: /components/datetime/
+github: 'https://github.com/contentful/forma-36/tree/master/packages/components/datetime'
+storybook: 'https://f36-storybook.contentful.com/?path=/story/components-datetime--basic'
+typescript: ./src/DateTime.tsx,.RelativeDateTime.tsx
```
suevalov, line 8:
```suggestion
typescript: ./src/DateTime.tsx,./src/RelativeDateTime.tsx
```

packages/components/datetime/DateTime.mdx
```diff
@@ -0,0 +1,81 @@
...
+
+```tsx
+import { RelativeDateTime } from '@contentful/f36-datetime';
+
+<RelativeDateTime date="2021-08-17T15:45" isRelativeToCurrentWeek />; // It will display "Yesterday at 3:45 PM"
+
+<RelativeDateTime date="2021-08-18T10:45" isRelativeToCurrentWeek />; // It will display "5 hours ago"
+
+<RelativeDateTime date="2021-08-18T15:45" isRelativeToCurrentWeek />; // It will display "Tomorrow at 3:45 PM"
+```
```
suevalov, line 81: PropTypes table is needed to be added explicitly.

gui-santos: Oh I forgot it again! 🤦

packages/forma-36-react-components/src/index.ts
```diff
@@ -108,7 +108,11 @@ export * from './components/Autocomplete';
 export * from './components/Card';
 export * from './components/Dropdown';
 export * from './components/EditorToolbar';
-export * from './components/DateTime';
+export { DateTime, RelativeDateTime } from '@contentful/f36-datetime';
```
suevalov, line 111: you also need to add `@contentful/f36-datetime` to the dependencies of `forma-36-react-components`

packages/components/datetime/DateTime.mdx
```diff
@@ -0,0 +1,91 @@
+---
+title: 'DateTime'
+type: 'component'
+status: 'stable'
+slug: /components/datetime/
```
massao, line 5: Needs to be the same slug as set on the gatsby config file, otherwise the link on the menu will point to the old version
```suggestion
slug: /components/date-time/
```

packages/components/datetime/DateTime.mdx
```diff
@@ -0,0 +1,91 @@
...
+| --------------- | ---------------------- |
+| `full`          | 12 Aug 2020 at 8:00 AM |
+| `day`           | 12 Aug 2020            |
+| `time`          | 8:00 AM                |
+| `weekday`       | Wed, 12 Aug            |
+
+### Examples
+
+```tsx
+import { DateTime } from '@contentful/f36-datetime';
```
massao, line 30: you can remove the `import` statement line the `codeblock`, otherwise it throws an error, and don't render the component.

packages/components/datetime/src/RelativeDateTime.tsx
```diff
@@ -0,0 +1,79 @@
...
+  /**
+   * If a value is passed to baseDate, then the component will compare both dates and return the time between them.
+   * If no value is passed then the date will be compared to "now"
+   *
+   * @default "Now"
+   */
+  baseDate?: Date | string | number;
+  /**
+   * Sets the date to be relative only if it is in the current week
+   */
```
bgutsol, line 30:
```suggestion
   * Sets the date to be relative only if it is in the current week
   * @default false
   */
```

packages/components/datetime/src/RelativeDateTime.tsx
```diff
@@ -0,0 +1,79 @@
...
+  isRelativeToCurrentWeek?: boolean;
+}
+
+const _RelativeDateTime = (
+  {
+    date,
+    baseDate,
+    isRelativeToCurrentWeek = false,
+    className,
```
bgutsol, line 39: You probably can leave `className` in `otherProps`, since you just pass it further

bgutsol (commented):
Great job @gui-santos .👏 I like how much cleaner it became thanks to neat props and examples 🧼

One question about the logic:
Is it intentionally that `isRelativeToCurrentWeek` ignores `baseDate`? So it working only with now (new Date()) even if you pass `baseDate` prop.

packages/components/datetime/DateTime.mdx
```diff
@@ -0,0 +1,91 @@
...
+### Examples
+
+If we consider today as August 18th of 2021 and we use pass the following values to the Component
+
+```tsx
+// import { RelativeDateTime } from '@contentful/f36-datetime';
+
+<RelativeDateTime date="2021-08-17T15:45" />; // passing a Timestamp as an ISOString
+<RelativeDateTime date={new Date()} />; // passing a JS Date
+<RelativeDateTime date={1629240300000} />; // passing a Unix Epoch in milliseconds
```
massao, line 52: This is still breaking on the documentation website, probably needs wrap the components with another component, and put the comments between `{}`, something like:
```tsx
<>
  <DateTime date="2021-08-17T15:45" /> { /* passing a Timestamp as an ISOString */ }
  <DateTime date={new Date()} /> { /* passing a JS Date */ }
  <DateTime date={1629240300000} /> { /* passing a Unix Epoch in milliseconds */ }
</>
```
but probably some other tag rather than `Fragment` to have spacing.

Also you can remove entirely the import statement, we removed the commented version from others components too.

gui-santos: Ahhh of course! How did I forget about this? 🤦

packages/components/datetime/src/RelativeDateTime.tsx
gui-santos: touché

packages/components/datetime/DateTime.mdx
```diff
@@ -0,0 +1,91 @@
...
+
+<RelativeDateTime baseDate="2021-08-17T15:45" date="2021-08-18T15:45" />
+```
+
+The component will display `in a day`
+
+### Using `isRelativeToCurrentWeek`
+
+Still using the last example. If you only need a relative date if the given date is within the current week,
+that is possible by using the boolean `isRelativeToCurrentWeek`
```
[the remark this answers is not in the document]
gui-santos: My writing is bad! that's why I need people reviewing my docs for real

So, I'm trying to say that if you want to show "Yesterday at 10:00 PM"
but if the day is more than seven days ago you want to show "12 Aug 2021"

Then you need to pass `isRelativeToCurrentWeek ={true}`

But it's so hard to explain it when I write it!!! do you have suggestions?

packages/components/datetime/DateTime.mdx
```diff
@@ -0,0 +1,91 @@
...
+<RelativeDateTime date={new Date()} />; // passing a JS Date
+<RelativeDateTime date={1629240300000} />; // passing a Unix Epoch in milliseconds
+```
+
+The component will display `a day ago`
+
+### Passing a `baseDate`
+
+Normally, the component compare your `date` with today’s date. But maybe you need to render a relative date
+that is compared with some other day that is not today. You can do that by using `baseDate`:
```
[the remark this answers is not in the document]
gui-santos: Amazing!
10x better

gui-santos:
> One question about the logic:
> Is it intentionally that `isRelativeToCurrentWeek` ignores `baseDate`? So it working only with now (new Date()) even if you pass `baseDate` prop.

@bgutsol no! You found a 🐛 
Thanks 🙌

packages/components/datetime/DateTime.mdx
gui-santos: I fixed all of these bugs and also decided to split this into two page

Screenshot 2021-08-20 at 17 56 36

packages/components/datetime/RelativeDateTime.mdx
```diff
@@ -0,0 +1,64 @@
...
+
+In `@contentful/f36-datetime`, there are two components (`DateTime` and `RelativeDateTime`) that will return a date following [Forma 36 guidelines](https://f36.contentful.com/guidelines/date-and-time/).
+
+## RelativeDateTime
+
+Provides a relative date string for resolutions ranging from seconds up through years. Should be used when absolute date and time is unnecessarily specific.
+
+### Examples
+
+If we consider today as August 18th of 2021 and we use pass the following values to the Component
```
burakukula, line 21:
```suggestion
If we consider today as of August 18th of 2021 and we pass the following values to the `<RelativeDateTime />` component
```

Is this what you meant?

packages/components/datetime/RelativeDateTime.mdx
```diff
@@ -0,0 +1,64 @@
...
+  {/* passing a Unix Epoch in milliseconds */}
+  <RelativeDateTime date={1629240300000} />
+</Grid>
+```
+
+The component will display `a day ago`
+
+### Passing a `baseDate`
+
+Normally, the component compare your `date` with today’s date. But if you want to compare it to some specific date provide `baseDate` prop.
```
burakukula, line 38:
```suggestion
Normally, the component compares your `date` with today’s date. If you want to compare it to some specific date you need to provide `baseDate` prop.
```

packages/components/datetime/RelativeDateTime.mdx
```diff
@@ -0,0 +1,64 @@
...
+
+```tsx
+<RelativeDateTime baseDate="2021-08-17T15:45" date="2021-08-18T15:45" />
+```
+
+The component will display `in a day`
+
+### Using `isRelativeToCurrentWeek`
+
+Still using the last example. If you only need a relative date if the given date is within the current week,
```
burakukula, line 48:
```suggestion
If you need a relative date if the given date is within the current week, that is possible by using the boolean `isRelativeToCurrentWeek`
```

packages/components/datetime/DateTime.mdx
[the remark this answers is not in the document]
bgutsol: I guess you rewrite somehow the changes back to your version here.

packages/components/datetime/DateTime.mdx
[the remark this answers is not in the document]
bgutsol: Yeah, it's hard to explain in a few words, maybe smth like:

> You can set `isRelativeToCurrentWeek`  to `true` if you want to indicate that the date is relative only within a week. And if the difference is more than one week, just show the date without relativeness. It will compare the date to today or `baseDate` if provided.

And it will be useful to add one more example with the date that has difference more than 7 days.

packages/components/datetime/RelativeDateTime.mdx
burakukula: Or maybe something like: 
If you need a relative date if the given date is within the current week, you can pass boolean `isRelativeToCurrentWeek` into the component

packages/components/datetime/stories/DateTime.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+    controls: {
+      matchers: {
+        date: /date$/,
+      },
+    },
+  },
+} as Meta;
+
+export const Basic: Story<DateTimeProps> = (args) => <DateTime {...args} />;
+Basic.args = {
```
burakukula, line 20: Maybe an overview of the possible variants would be helpful, both for DateTime and RelativeDateTime? Or it's overkill?

packages/components/datetime/DateTime.mdx
burakukula: @gui-santos great idea! but for some reason, this change is not there yet? 🤔

packages/components/datetime/DateTime.mdx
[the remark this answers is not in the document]
gui-santos: Good idea! Thanks

packages/components/datetime/DateTime.mdx
gui-santos: Really? It was my last commit

Maybe netlify is not re-deploying it

packages/components/datetime/RelativeDateTime.mdx
gui-santos: is `as of` correct? TIL

packages/components/datetime/stories/DateTime.stories.tsx
gui-santos: I think it's a good idea to always have some Overview! I will add it

packages/components/datetime/DateTime.mdx
gui-santos: Oh no! hahahaha
I forgot to commit my changes to gatsby-config! 🤦

gui-santos: It should be fixed now

suevalov (approved):

bgutsol (approved):

burakukula (approved):
Awesome work! Thank you!! 🏅

massao (approved):
🎉

PR #1201, opened 2021-08-18
feat(V4): Notification

massao:
# Purpose of PR

Migrate `Notification` to V4

## PR Checklist

- [x] Tests are passing


packages/components/notification/Notification.mdx
```diff
@@ -29,8 +32,11 @@ Notifications give immediate feedback about an action triggered or experienced b
 ### Basic example
 
 ```jsx
+// import { Button } from '@contentful/f36-button';
```
suevalov, line 35: I guess we can remove these comments. I remember removing them in other PRs.

packages/components/notification/src/Notification.tsx
```diff
@@ -7,20 +7,20 @@ import {
   ShowAction,
   CloseAction,
   CloseAllAction,
-  SetPositionAction,
+  SetPlacementAction,
   SetDurationAction,
   NotificationProps,
-  Position,
+  Placement,
```
suevalov, line 13: Open question: Do we really want to rename this property? 

In general, I'd avoid doing a breaking change in the component API if it doesn't contradict our guidelines. IMO, in this case, "Position" is an acceptable term.

packages/components/notification/src/Notification.tsx
```diff
@@ -54,20 +54,20 @@ function afterInit<PromiseValueType>(fn: Function) {
   };
 }
 
-const show = (intent: NotificationIntent) => (
+const show = (variant: NotificationVariant) => (
   text: string,
   settings?: {
     duration?: number;
-    canClose?: boolean;
+    withClose?: boolean;
```
suevalov, line 61: also, a questionable change for me. I'd rather not change it. It's not a component, but a part of an imperative API. IMO, it even better than `withClose`

packages/components/notification/src/NotificationItem/NotificationItem.tsx
```diff
@@ -0,0 +1,109 @@
...
+import { Heading, Paragraph } from '@contentful/f36-typography';
+
+import { NotificationVariant, NotificationCta } from '../types';
+import { getStyles } from './NotificationItem.styles';
+
+export interface NotificationItemProps extends CommonProps {
+  variant?: NotificationVariant;
+  withCloseButton?: boolean;
+  onClose?: Function;
+  testId?: string;
```
suevalov, line 22: `testId` already exists in CommonProps

packages/components/notification/src/NotificationItem/NotificationItemContainer.tsx
```diff
@@ -93,13 +95,13 @@ export class NotificationItemContainer extends Component<
           }
         }}
       >
-        <div
-          style={{ pointerEvents: 'all' }}
+        <Box
```
suevalov, line 98: I guess using `Box` here is an optional change. We don't pass any props outside, do it can stay as `div`

packages/components/notification/src/NotificationItem/NotificationItem.tsx
```diff
@@ -0,0 +1,109 @@
...
+      negative: <ErrorCircleIcon variant={variant} size={size} />,
+    };
+
+    return <Box className={cx(styles.icon)}>{iconVariants[variant]}</Box>;
+  }, [variant, styles.icon, title]);
+
+  const renderContent = () => (
+    <>
+      {title && (
+        <Heading as="h2" className={cx(styles.title)}>
```
suevalov, line 57: Do we need `cx` if only one style class is used?

suevalov: I think we need set marginBottom to "none" for both, Heading and Paragraph components here

packages/components/notification/src/Notification.tsx
massao: I changed this to keep it like we did for the `Tooltip`, and other components, but I'm fine with either too

packages/components/notification/src/NotificationItem/NotificationItem.tsx
massao: I had them updated on the style to use the correct spacing since it also needed the `font-size`, `line-height` and `color` properties, but we can move from there to here, if that would be better

packages/components/notification/src/Notification.tsx
massao: Yeah, also updated this one to follow what is set on the guidelines that when a prop should represent if a nested component should be rendered it should use the `with` prefix.
But considering that this is more and API than a component I'm fine with going back to `canClose`

packages/components/notification/src/index.ts
```diff
@@ -1,12 +1,10 @@
 export { Notification } from './Notification';
 export type { NotificationsAPI } from './Notification';
 export { NotificationItem } from './NotificationItem';
```
suevalov, line 3: do we actually need to export NotificationItem here? do we use it elsewhere?

packages/components/notification/src/NotificationItem/NotificationItem.tsx
suevalov: I just see the difference in a screenshot that shouldn't be there.

packages/components/notification/src/Notification.tsx
suevalov: let's discuss with the team

packages/components/notification/src/Notification.tsx
denkristoffer: I prefer we keep the name and prop values the same across them all. If not now, then when?

packages/components/notification/README.md
```diff
@@ -0,0 +1,4 @@
+# @contentful/f36-notification
```
bgutsol, line 1: Do we need to add this in every v4 components?

massao: Yes, because all the components are published as independent packages, so it needs the README on all of them to have this disclaimer

packages/components/notification/src/NotificationItem/NotificationItem.tsx
massao: Oh this may be because the new design has bigger spacing compared than the previous one

packages/components/notification/src/Notification.tsx
gui-santos: I'm with @denkristoffer on this one

We can make a codemod that rename props, no?

packages/components/notification/src/Notification.tsx
gui-santos: For me, is the same thing here that @denkristoffer said about the other prop

v4 is a chance to align things with breaking changes
if we don't break now, we will need v5 to do it and when we work in v5 we will probably have this discussion again hahahah
and in v5 we also won't break things so we will do in v6

you see where I'm going with this

packages/components/notification/Notification.mdx
```diff
@@ -30,7 +33,7 @@ Notifications give immediate feedback about an action triggered or experienced b
 
 ```jsx
```
gui-santos, line 34: What's the thing about adding `static=true` again?

Do we need it or not?

packages/components/notification/Notification.mdx
```diff
@@ -44,43 +47,43 @@ The `Notification` component can be configured in a number of different ways. He
...
   </Button>
   <Button
-    buttonType="warning"
+    variant="secondary"
     onClick={() => Notification.warning('This is a warning notification')}
   >
     Show warning notification
   </Button>
 </React.Fragment>
 ```
```
gui-santos, line 68: Kind of a nitpicky suggestion

To give some space between the buttons, you could do something like what we have in the Buttons README now
https://github.com/contentful/forma-36/blob/master/packages/forma-36-react-components/src/components/Button/README.mdx

basically wrap them with `<Grid rowGap="spacingXs" rows="repeat(4, auto)" justifyContent="start">`

packages/components/notification/README.md
gui-santos: I think we need to update the plop template

My branch only has one README with all the documentation, but I guess every package needs a `README.md` and `Component.mdx`, right?

packages/components/notification/src/NotificationItem/NotificationItem.tsx
```diff
@@ -0,0 +1,132 @@
...
+  const renderIcon = useCallback(() => {
+    const size = title ? 'medium' : 'small';
+    const iconVariants = {
+      positive: <CheckCircleIcon variant={variant} size={size} />,
+      warning: <WarningIcon variant={variant} size={size} />,
+      negative: <ErrorCircleIcon variant={variant} size={size} />,
+    };
+
+    return <Box className={cx(styles.icon)}>{iconVariants[variant]}</Box>;
+  }, [variant, styles.icon, title]);
```
gui-santos, line 71: Do we need to make a function and useCallback for this?

Would something like this work?

```jsx
const size = title ? 'medium' : 'small';
const iconVariants = {
  positive: <CheckCircleIcon variant={variant} size={size} />,
  warning: <WarningIcon variant={variant} size={size} />,
  negative: <ErrorCircleIcon variant={variant} size={size} />,
};

// ...

return (
    <Flex
      {...}
    >
      <Box className={cx(styles.icon)}>{iconVariants[variant]}</Box>
```

packages/components/notification/src/NotificationItem/NotificationItem.tsx
```diff
@@ -0,0 +1,132 @@
...
+      {cta?.label && (
+        <TextLink
+          {...cta?.textLinkProps}
+          as={cta?.textLinkProps?.as || 'button'}
+        >
+          {cta.label}
+        </TextLink>
+      )}
+    </>
+  );
```
gui-santos, line 94: Same here, it's a function returning JSX so it's a component

Maybe you could cut the middleman (the function) and just move the jsx to the `return`, you only replace the fragments here with the `Box` that is already there

wdyt?

packages/components/notification/stories/Notification.stories.tsx
```diff
@@ -0,0 +1,178 @@
...
+const getUniqueNumber = () => {
+  index += 1;
+  return index;
+};
+
+export default {
+  component: NotificationItem,
+  parameters: {
+    propTypes: NotificationItem['__docgenInfo'],
+  },
```
gui-santos, line 20: I think we don't need this anymore

These was a hack that we needed to do when we added storybook-docs, but our components didn't have README files I think

(I will update the plop file to not have this line someday)

gui-santos (changes requested):
Great job!

I only added some documentation and small refactors suggestions
Also, waiting for the decision on the props renaming idea

packages/components/notification/README.md
massao: Yeah, I think that it got updated to have the disclaimer, but not to have both files

packages/components/notification/Notification.mdx
massao: That is for not rendering the example and only showing the code snippet, I only used that for the codes about the API as they don't get rendered.
So it depends if we want to have the example working on the docs or only the code

packages/components/notification/src/Notification.tsx
suevalov: alright, let's rename it then

packages/components/notification/src/Notification.tsx
suevalov: I disagree with renaming this property. Why do we need to do a breaking change here? `canClose` is actually describing the intention better... Can the user close this notification?

packages/components/notification/stories/Notification.stories.tsx
massao: https://github.com/contentful/forma-36/pull/1207
Created a PR removing those lines and adding the `Component.mdx` file

packages/components/notification/src/NotificationItem/NotificationItem.tsx
bgutsol: TBH, I don't like this approach of using `useCallback` to create a node, they have zero impact on performance. And they only introduce confusion and bugs with dependency arrays. I saw this approach a couple of times already in our codebase.

packages/components/notification/Notification.mdx
gui-santos: Ahhhhhhhhhhhhh I got it
I will have to change my stuff hahahaha

Thanks 🙏

packages/components/notification/stories/Notification.stories.tsx
gui-santos: Legend! 🙌

packages/components/notification/src/Notification.tsx
massao: I agree that the `canClose` describes better the intention, and also this won't be used as a prop by the user itself, it will be passed through the settings when calling a show method from `Notification`.

bgutsol: I don't have a strong opinion on it. Either way seems fine for me. So I'd rather go forward with the approach that needs a least amount of work.

gui-santos (approved):
💎

packages/components/notification/src/Notification.tsx
suevalov: I'd root for leaving as it is now.

bgutsol (approved):
✨

Issue #1164, opened 2021-08-08
💡 Proposal - dropdown refactoring


bgutsol:
# Forma 36 contribution proposal

## The problem

At the moment we have variety of implementation with dropdown within different apps. They are not consistent, do not pass a11y rules, and hard to maintain because of custom complex logic.
The list of examples of current Dropdown usage:

1. Two places where we create custom `MultiSelect` component based on `Dropdown`: 
    - [first](https://github.com/contentful/user_interface/blob/f1874983b0a3e366836f0aaf3d72b61f8d5b39a7/src/javascripts/app/entry_editor/EntryReferences/MultiSelect/MultiSelect.tsx#L92)
    - [second](https://github.com/contentful/user_interface/blob/d390b23ee149c38b68802d608b151578e81f10f1/src/javascripts/app/OrganizationSettings/SpaceRoleEditor.js#L122)
 
![Screen Shot 2021-08-20 at 17 04 19](https://user-images.githubusercontent.com/7631029/130253767-5c274e3a-1101-423f-942b-a7bb8551b764.png)

2. Two places where we changing default `click` event as a trigger to `hover`: 
    - [first](https://github.com/contentful/user_interface/blob/d390b23ee149c38b68802d608b151578e81f10f1/src/javascripts/app/entity_editor/Components/FetchLinksToEntity/LinkedEntitiesBadge.js#L84), 
    - [second](https://github.com/contentful/creator-app-mvp/blob/54ce1619d3e097d81b3c2a2a7dd11cb8a2b12c88/src/features/page-editing-common/components/EntityReuseCounter/EntityReuseCounter.tsx#L82)

![Screen Shot 2021-08-20 at 17 05 40](https://user-images.githubusercontent.com/7631029/130253917-a31d36e7-bb70-418b-bb8b-aebde866936a.png)

3. Several places where we use `DropdownList` without `Dropdown`, and `CardActions` without `Card` components: 
    - [first](https://github.com/contentful/user_interface/blob/324c705288c750f3740ec4efc46175860770390b/src/javascripts/app/EntrySidebar/TasksWidget/View/Task.js#L169)
    - [second](https://github.com/contentful/user_interface/blob/043abbd930a993c99e78799a0539f43a3653e31f/src/javascripts/app/OrganizationSettings/Users/UserDetail/UserSpaceList.js#L91)
[...]

PR #1204, opened 2021-08-18
feat: @contentful/f36-popover

bgutsol:
# Purpose of PR
`Popover` component (replace for the current `Dropdown` component). Base for the next `Menu`, `Multiselect` and `Autocomplete` components.

[More details in the proposal](https://github.com/contentful/forma-36/issues/1164)

## PR Checklist

- [x] Tests are passing


packages/components/popover/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "@contentful/f36-popover",
+  "version": "4.0.0",
+  "private": true,
```
suevalov, line 4: don't forget to remove `private`, so the package gets published once merged to the main branch

packages/core/src/hooks/mergeRefs.ts
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+type ReactRef<T> = React.Ref<T> | React.MutableRefObject<T>;
+
+export function mergeRefs<T>(
```
massao, line 5: perhaps use the `use...` nomenclature, or move to utils instead of hooks folder?? 🤔

packages/components/popover/Popover.mdx
```diff
@@ -0,0 +1,59 @@
+---
+title: 'Popover'
+type: 'component'
+status: 'stable'
+slug: /components/popover/
+github: 'https://github.com/contentful/forma-36/tree/master/packages/components/popover'
+storybook: 'https://f36-storybook.contentful.com/?path=/story/components-popover--basic'
+typescript: ./src/Popover.tsx
+---
```
massao, line 9: Shouldn't we also render the props table?

massao (commented):
Also needs to add the popover to the gatsby config file for it to show up on the documentation website:
https://github.com/contentful/forma-36/blob/763439f1faad1ff40e7c4a1144ddec2485276b6b/packages/forma-36-website/gatsby-config.js

packages/components/popover/stories/Popover.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+        <Button onClick={() => setIsOpen(!isOpen)}>Toggle</Button>
+      </PopoverTrigger>
+      <PopoverContent>
+        <Box padding="spacingM">
+          <Text as="p" marginBottom="spacingM">
+            This is the content.
+          </Text>
+          <Button>Some action</Button>
+        </Box>
+      </PopoverContent>
```
gui-santos, line 33: IMO the padding should be handled by `PopoverContent`, we will always want some space between the content and the borders of the popover

We should be able to pass only what we want to show in the popover as children, like this:

```suggestion
      <PopoverContent>
          <Text as="p" marginBottom="spacingM">
            This is the content.
          </Text>
          <Button>Some action</Button>
      </PopoverContent>
```

packages/components/popover/stories/Popover.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const [isOpen, setIsOpen] = useState(false);
+
+  return (
+    <Popover {...args} isOpen={isOpen} onClose={() => setIsOpen(false)}>
+      <PopoverTrigger>
+        <Button onClick={() => setIsOpen(!isOpen)}>Toggle</Button>
+      </PopoverTrigger>
+      <PopoverContent>
+        <Box padding="spacingM">
+          <Text as="p" marginBottom="spacingM">
```
gui-santos, line 28: I think we have `Paragraph` again, no?

packages/components/popover/stories/Popover.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+  },
+} as Meta;
+
+export const Basic: Story<PopoverProps> = (args) => {
+  const [isOpen, setIsOpen] = useState(false);
+
+  return (
+    <Popover {...args} isOpen={isOpen} onClose={() => setIsOpen(false)}>
+      <PopoverTrigger>
+        <Button onClick={() => setIsOpen(!isOpen)}>Toggle</Button>
```
gui-santos, line 24: How is this button not blue in storybook?
Is `secondary` the default value of a button now?

packages/components/popover/stories/Popover.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+      </PopoverTrigger>
+      <PopoverContent>
+        <Box padding="spacingM">
+          <Text as="p" marginBottom="spacingM">
+            This is the content.
+          </Text>
+          <Button>Some action</Button>
```
gui-santos, line 31: Same here about button, why is it secondary even though you didn't pass `variant="secondary"`?

packages/components/popover/src/PopoverContent/PopoverContent.styles.ts
```diff
@@ -0,0 +1,12 @@
...
+import tokens from '@contentful/f36-tokens';
+
+export const getPopoverContentStyles = () => ({
+  container: css({
+    background: tokens.colorWhite,
+    border: 0,
+    borderRadius: tokens.borderRadiusMedium,
+    boxShadow: tokens.boxShadowDefault,
+    zIndex: tokens.zIndexDropdown,
+  }),
```
gui-santos, line 11: I think this is where we should add padding

packages/components/popover/src/PopoverContext.ts
```diff
@@ -0,0 +1,27 @@
...
+
+const PopoverContext = React.createContext<PopoverContextType | undefined>(
+  undefined,
+);
+
+export const usePopoverContext = () => {
+  const context = React.useContext(PopoverContext);
+
+  if (context === undefined) {
+    throw new Error('useCount must be used within a CountProvider');
```
gui-santos, line 21:
```suggestion
    throw new Error('usePopoverContext must be used within a PopoverContextProvider');
```

right?

packages/components/popover/src/PopoverContent/PopoverContent.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  if (!isOpen) {
+    return null;
+  }
+
+  return usePortal ? <Portal>{content}</Portal> : content;
+};
+
+const _PopoverContent = React.forwardRef(PopoverContent);
+export { _PopoverContent as PopoverContent };
```
gui-santos, line 47: Because we are doing this in our components, in some places the source code is showing `<_PopoverContent>` (especially in storybook)

in this PR https://github.com/contentful/forma-36/pull/1209 
I'm proposing inverting this names, so basically the one going inside forwardRef is the one with `_` and the one exported has the final name

```suggestion
export const PopoverContent = React.forwardRef(_PopoverContent);
```

Wdyt? this way we avoid the whole `as` thing

packages/components/popover/src/PopoverTrigger/PopoverTrigger.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { usePopoverContext } from '../PopoverContext';
+
+/**
+ * PopoverTrigger opens the popover. It must be an interactive element.
+ */
+export const PopoverTrigger: React.FC = (props) => {
```
gui-santos, line 7: I think we need to update the code style because it was kind agreed that we would prefer not to use `React.FC` and type only the props the component will use
even `children`

packages/components/popover/src/popperModifiers.ts
```diff
@@ -0,0 +1,18 @@
...
+  phase: 'beforeWrite',
+  requires: ['computeStyles'],
+  fn: ({ state }) => {
+    state.styles.popper.width = `${state.rects.reference.width}px`;
+  },
+  effect: ({ state }) => () => {
+    const reference = state.elements.reference as HTMLElement;
+    state.elements.popper.style.width = `${reference.offsetWidth}px`;
+  },
+};
```
gui-santos, line 18: Isn't this only being used by `Popover.tsx`?

I suggest moving it to that file, if it ever gets shared between to components then we can abstract

Just for maintainability purposes I think
It's also a personal preference of not jumping from file to file so... I don't know hahahah
not a blocker, more like a suggestion

gui-santos (changes requested):
Wow! Amazing job

I have only one some questions about the docs
and about the `placement` and `isAutoaligmentEnabled` props

Also a small CSS suggestion

packages/components/popover/stories/Popover.stories.tsx
massao: Yeah, the Button has the secondary as the default value for the variant

packages/components/popover/Popover.mdx
```diff
@@ -0,0 +1,59 @@
...
+
+Popover is used to display some content on top of another, and should be paired with a clickable trigger element.
+It is a base for other more specific component, like `Menu`, `Autocomplete` and `Multiselect`. Please, consider using these specific componenets to cover your needs.
```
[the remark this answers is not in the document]
bgutsol: We (with designers) decided to give the users more flexibility. Since we are removing `Dropdown` component which was something  in the middle between raw popover and specific components like `Menu` or `Multiselect`.

> I guess we have other components like this, right?
What do you mean?

packages/components/popover/Popover.mdx
```diff
@@ -0,0 +1,59 @@
...
+
+## Code examples
+
+- Pass trigger component as a child for `PopoverTrigger`.
+- Pass popover content as a child for `PopoverContent`
+
+```jsx
+<React.Fragment>
+  <TextLink variant="primary">Text link</TextLink>
+</React.Fragment>
```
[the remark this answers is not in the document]
bgutsol: Why the `TextLink`? ahah 😄 
I just forgot to update examples after copying it from `TextLink` 🤦‍♂️

packages/components/popover/stories/Popover.stories.tsx
bgutsol: You are right. TBH I was a bit confused what Typography component when to use

packages/components/popover/stories/Popover.stories.tsx
bgutsol: And it was aligned with our designers, right @massao ?

packages/components/popover/src/PopoverContent/PopoverContent.styles.ts
bgutsol: I don't think we should touch the inner content styles, since it can be customised in a different ways. 
For example for such divider you will have to remove somehow those paddings or set a negative margins: 
image

packages/components/popover/src/PopoverContext.ts
bgutsol: Totally!

packages/components/popover/stories/Popover.stories.tsx
bgutsol: Please check my answer in other comment: https://github.com/contentful/forma-36/pull/1204#discussion_r692996969
Basically, I think that adding padding is a lot easier the removing it after.

packages/components/popover/src/Popover.tsx
```diff
@@ -0,0 +1,224 @@
...
+          offset: [0, 4],
+        },
+      },
+      {
+        ...sameWidth,
+        enabled: isFullWidth,
+      },
+      {
+        name: 'preventOverflow',
+        enabled: isAutoalignmentEnabled,
```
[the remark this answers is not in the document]
denkristoffer: The `placement` from Popper.js is more of a hint than a requirement, so popper can actually ignore it. But I'm not sure we can or what to change that behaviour, so maybe this boolean flag is not worth keeping at all.

Although we should find and try to understand cases where `isAutoalignmentEnabled={false}` before removing.

packages/components/popover/Popover.mdx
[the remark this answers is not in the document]
gui-santos: Nevermind! hahahaha thanks for clarifying 👌

packages/components/popover/src/Popover.tsx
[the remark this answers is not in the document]
gui-santos: I was playing with it in Storybook and the only change I spot is this:

Screenshot 2021-08-20 at 16 52 07
(isAutoalignmentEnabled={false} and placement="auto")

Screenshot 2021-08-20 at 16 51 53
(isAutoalignmentEnabled={true} and placement="auto")

packages/components/popover/src/PopoverContent/PopoverContent.styles.ts
gui-santos: I guess with this image we have two things `PopoverContent` and `PopoverContentFooter` (?????)

So we could make an API like this: 
```
<PopoverContent>
  <MenuItem />
  <MenuItem />
</PopoverContent>
<PopoverContentFooter>
  <TextLink/>
</PopoverContentFooter>  
```

If ask the users to deal with the spacing themselves, they would have to do this:
```
<PopoverContent>
  <Box margin="spacingM">
     <MenuItem />
     <MenuItem />
  <Box/>
  <Box margin="spacingM">
     <TextLink/>
  <Box/>
</PopoverContent>
```

But we won't be able to prevent people from things like not using any spacing, using a bigger or smaller spacing than `spacingM`
I also can see some of the FE in Contentful complaining that they need to pass a `Box` all the time hahahaha

gui-santos: But maybe you agreed with the designers about making `Popover` flexible as you said in other comment
And this strict spacing that I'm talking about will come with `Menu` and the other components using `Popover`

then... just ignore me hahahaha

bgutsol: > And this more strict spacing that I'm talking about will come with Menu and the other components using Popover
💯 Let's see what we will have with designs and with specific components in the end. And then adjust this `Popover`. For now it's hard to peak the right decision, I even don't have new popover, menu, autocomplete etc. designs yet. 🤷‍♂️

packages/components/popover/src/Popover.tsx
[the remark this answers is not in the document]
bgutsol: I'll investigate it and get back to you 👍

packages/components/popover/src/PopoverContent/PopoverContent.styles.ts
denkristoffer: My two cents, Popover is a utility component that shouldn't do any styling. Keep that in `Menu` and `Modal` and whatever else.

packages/components/popover/src/Popover.tsx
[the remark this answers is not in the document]
bgutsol: Based on a small research I found out that:
1. `placement: auto` doesn't do the same as auto alignment. Auto alignment prevents from overflowing and overlapping. 
2. We have places in our projects where we intentionally set `isAutoalignmentEnabled=false`. So it's useful to give the end user control on auto alignment.
3. When `isAutoalignmentEnabled=false` Popper.js precisely follows the `placement` prop and does not adjust the position even if it's overlapped.

suevalov (approved):
Looks like solid work for me and a great foundation for the other components.

packages/components/popover/Popover.mdx
```diff
@@ -0,0 +1,77 @@
...
+## Table of contents
+
+- [How to use Popover](#how-to-use-Popover)
+- [Code examples](#code-examples)
+- [Content recommendations](#content-recommendations)
+- [Accessibility](#accessibility)
+
+## How to use Popover
+
+- If `Menu`, `Autocomplete` and `Multiselect` components are not covering your use cases, only then you can use `Popover`.
```
burakukula, line 23:
```suggestion
- Only If `Menu`, `Autocomplete` and `Multiselect` components are not covering your use cases, you should use `Popover`.
```

packages/components/popover/Popover.mdx
```diff
@@ -0,0 +1,77 @@
...
+- Double check with designers before using this component. Probably you can just use one of the more specific components, like `Menu`, `Autocomplete` and `Multiselect`
```
burakukula, line 24:
```suggestion
- Before using this component, double-check your design requirements. We are providing `Menu`, `Autocomplete`, and `Multiselect` for more specific use-cases and they can address your needs.
```

packages/components/popover/Popover.mdx
```diff
@@ -0,0 +1,77 @@
...
+- Keep in mind that everything related to accessibility for the popover content you have to develop on your own.
```
burakukula, line 25:
```suggestion
- Keep in mind that you will have to implement everything related to accessibility for the popover content.
```

packages/components/popover/src/Popover.tsx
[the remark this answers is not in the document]
gui-santos: Ahhhh I get it now! Could comment this in the prop and in the website docs? (number 3 at least, made me understand exactly what this prop does)

packages/components/popover/src/PopoverContent/PopoverContent.styles.ts
gui-santos: Yep... I changed my mind hahahah 
Let's do the padding thing in `Menu` and the other components 👌

burakukula (approved):
Awesome! 👏 
Just a couple of small suggestions around docs, but nothing really blocking.

packages/components/popover/src/Popover.tsx
```diff
@@ -0,0 +1,240 @@
...
+   */
+  placement?: Placement;
+
+  /**
+   * Boolean to disable automatic positioning of the element to fit the
+   * available viewport. Instead this forces the element to follow the given
+   * value of the `placement` prop.
+   *
+   * @default true
+   */
```
gui-santos, line 44:
```suggestion
  /**
   * Popper.js will try to prevent the Popover from overflowing and overlapping (the available viewport???).
   * If this prop is set to false, it disables that behavior and Instead it forces the element to follow the given
   * value of the `placement` prop, even if it's overlapped.
   *
   * @default true
   */
```

Something like this?????? don't trust me on writing text honestly

gui-santos (approved):
💎

thebesson (approved):

packages/components/popover/src/Popover.tsx
bgutsol: Is it better?
https://github.com/contentful/forma-36/pull/1204/commits/af1013eddb2e786463a43dad8ccb693b149be65f#diff-312f071358cb2435722193ae98706bfa3b73a5ea7b43eddf8f5e2ed66957146dR39-R46

denkristoffer:
@all-contributors add @bgutsol for code, maintenance

PR #1208, opened 2021-08-19
feat(codemod): pill codemod

massao:
# Purpose of PR

Adds codemod to be used to migrate `Pill` component, since it had no breaking change and only style changes, we only need to update the import


packages/forma-36-codemod/transforms/__tests__/v4-pill.test.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+const defineTest = require('jscodeshift/dist/testUtils').defineTest;
+
+const tests = ['v4-pill'];
+
+describe('v4-pill', () => {
```
suevalov, line 7: we should also add `v4-pill` option to CLI list

suevalov (approved):

PR #1228, opened 2021-08-25
refactor: adjustments to core utility types

bgutsol:
# Purpose of PR
It's unclear now when/how to use types for polymorphic components or for components that are not polymorphic but need to extend their own props with the default HTML element one.
That's why we ended up having slightly different approaches in our codebase:
- To extend own props we use [PolymorphicComponentProps with unchangeable as prop](https://github.com/contentful/forma-36/blob/c9991a83905b9ebac727946a69a941065ae20afb/packages/components/note/src/Note.tsx#L52)., It seems confusing, because we give `as` prop but fixing it to only one value.
- With the [DragHandle](https://github.com/contentful/forma-36/blob/673e7b39d637afe6d7afa8d1147aa2e24b838525/packages/components/drag-handle/src/DragHandle.tsx#L116) and [List](https://github.com/contentful/forma-36/blob/7e4dc05b740bfc47dbc52c40224dec424d6af841/packages/components/list/src/List.tsx#L31) components I didn't get the idea at all.
- We also for some reason include `CommonProps` initially [to Polymorphic component types](https://github.com/contentful/forma-36/blob/bf0fc7d5b9aa74009825588d822befe018c055ad/packages/core/src/Primitive/Primitive.tsx#L5) and then extending it in component one more time (ex. [Note](https://github.com/contentful/forma-36/blob/c9991a83905b9ebac727946a69a941065ae20afb/packages/components/note/src/Note.tsx#L29), [Button](https://github.com/contentful/forma-36/blob/ca7a3f7775bc2ee64ec80eb1fe09ddf7a9e91836/packages/components/button/src/types.ts#L20))
- The other question is why in some cases we use [usePrimitive](https://github.com/contentful/forma-36/blob/bf0fc7d5b9aa74009825588d822befe018c055ad/packages/core/src/Primitive/Primitive.tsx#L41) helper and in some we are not? What the purpose of this function? And do we need to use it everywhere where we have polymorphic component?

### The PR itself is just a proposal how everything described above can be done in a slightly different way. Just to simplify and bring a bit more clarity
New types working fine with storybook knobs generation and our website docs `<Props of="Component" />`

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/components/datetime/src/DateTime.tsx
+++ b/packages/components/datetime/src/DateTime.tsx
@@ -1,12 +1,10 @@
 import React from 'react';
-import { CommonProps } from '@contentful/f36-core';
+import { CommonProps, PropsWithHTMLElement } from '@contentful/f36-core';
 
 import type { DateType, DateFormat } from '../types';
 import { formatDateAndTime, formatMachineReadableDateTime } from './utils';
 
-export interface DateTimeProps
-  extends CommonProps,
-    React.AllHTMLAttributes<HTMLTimeElement> {
+interface DateTimeOwnProps extends CommonProps {
   /**
    * The date that will be displayed. It accepts a JS Date, an ISO8601 Timestamp string, or Unix Epoch Milliseconds number
    */
@@ -18,6 +16,8 @@ export interface DateTimeProps
    **/
   format?: DateFormat;
 }
+
+export type DateTimeProps = PropsWithHTMLElement<DateTimeOwnProps, 'time'>;
 
 const _DateTime = (
   {
--- a/packages/components/list/src/List.tsx
+++ b/packages/components/list/src/List.tsx
@@ -1,37 +1,57 @@
 import { cx, css } from 'emotion';
 import React from 'react';
 
-import { CommonProps, PropsWithHTMLElement } from '@contentful/f36-core';
+import {
+  CommonProps,
+  PolymorphicProps,
+  PolymorphicComponent,
+} from '@contentful/f36-core';
+
+const DEFAULT_TAG = 'ul';
+
 export interface ListInternalProps extends CommonProps {
   as?: 'ul' | 'ol';
   children?: React.ReactNode;
 }
 
-export type ListProps = PropsWithHTMLElement<ListInternalProps, 'ul'>;
+export type ListProps<
+  E extends React.ElementType = typeof DEFAULT_TAG
+> = PolymorphicProps<ListInternalProps, E>;
 
 /**
  * List is component that helps with vertical indexing of content.
  * Every list item begins with a bullet or a number.
  */
-export const List = React.forwardRef<HTMLUListElement, ListProps>(
-  ({ as, className, children, testId = 'cf-ui-list', ...otherProps }, ref) => {
-    return (
-      <ul
-        data-test-id={testId}
-        className={cx(
-          css({
-            margin: 0,
-            listStyleType: as === 'ul' ? 'disc' : 'decimal',
-          }),
-          className,
-        )}
-        ref={ref}
-        {...otherProps}
-      >
-        {children}
-      </ul>
-    );
-  },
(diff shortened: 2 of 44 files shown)


suevalov (commented):
I think it's the way better - more concise and easier to understand

packages/components/text-link/TextLink.mdx
```diff
@@ -87,3 +87,9 @@ Text links communicate actions and highlight linked resources to users.
 
 - Avoid using extra styling that that make link looks just like regular text.
 - If you use `TextLink` for linking to third party content and use `target="_blank"` prop, it is recommended to set `rel="noreferrer noopener"`. [More details about external links security](https://web.dev/external-anchors-use-rel-noopener/)
+
+## Props of DateTime
```
gui-santos, line 91:
```suggestion
## Props of TextLink
```

packages/components/datetime/src/DateTime.tsx
```diff
@@ -1,13 +1,11 @@
...
+import { CommonProps, PropsWithHTMLElement } from '@contentful/f36-core';
 
 import dayjs, { extend } from 'dayjs';
 import utcPlugin from 'dayjs/plugin/utc';
 extend(utcPlugin);
 
-export interface DateTimeProps
-  extends CommonProps,
-    React.AllHTMLAttributes<HTMLTimeElement> {
+interface DateTimeOwnProps extends CommonProps {
```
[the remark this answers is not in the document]
gui-santos: Yep, let's rename everything to `OwnProps` in another PR if we decide to do that

denkristoffer:
There's definitely some confusion here 😁

I think some of it is due to us changing our minds and refactoring things after the initial implementations of these ideas, which led to different understandings of how to implement these things. Speaking only for myself, `Primitive`, for example, was initially a component that I imagined as a base for all other components. It was changed to a hook for performance reasons, and I think I kept using it while others maybe didn't. Perhaps the polymorphic types have the same issue. I always imagined it as a global thing we would use everywhere to let consumers render components with the tags of their choice.  

It comes down to not having fully aligned on the strategy, I think. Overall we need a final cleanup and alignment before release 🙂

gui-santos (changes requested):
I liked it!

Hey, do you think using [type-fest](https://github.com/sindresorhus/type-fest) would help us?

bgutsol (commented):
So to sum up:
1. I will refactor old polymorphic types with the new one.
2. Instead of using polymorphic in the place where we just need to extend the props with the default html element attrs, I will use new type `PropsWithHTMLElement`
3. I will not rename `InternalProps` to `OwnProps` since we are using `InternalProps` everywhere and I don't see the value in such changes.
4. I will delete `usePrimitive` function and its usage.

suevalov:
> I will delete `usePrimitive` function and its usage.

I'd still leave it and use for `Box`, `Flex` and `Grid` components without exporting it out of the package.

packages/components/list/src/List.tsx
```diff
@@ -1,64 +1,37 @@
...
 
 /**
  * List is component that helps with vertical indexing of content.
  * Every list item begins with a bullet or a number.
  */
-export const List: PolymorphicComponent<
-  ListInternalProps,
-  typeof DEFAULT_TAG
-> = React.forwardRef(_List);
+export const List = React.forwardRef<HTMLUListElement, ListProps>(
```
suevalov, line 16: IMO, `List` should still be polymorphic.

suevalov (commented):
Just one comment about List

suevalov (approved):
🎖️ for 👨🏼‍💻

burakukula (approved):
Looks great! 🎉
Maybe we could come up later on with some form of guidelines for contributors if anyone would create a new component, how to use those. But we can tackle it later.

PR #1229, opened 2021-08-25
feat: v4 BaseInput

burakukula:
Attempt to create `BaseInput` component that would drive the `TextInput`, `Textarea` + potentially `Select`.
We could extend `BaseInput` while working on `Select`. 
I am not sure to which extend this component could be a base for `Checkbox` and `Radio`, we could give it a try, but from the first glance we might want to have something separate for those 2 cases - there is additional logic to allow custom styling of those components (previous ControlledInput and ControlledInputField was only used by `Checkbox` and `Radio`)

This would be an internal component. It would be used by, for example, by `TextInput`. In there we would handle label, validation, and help messages + there will be a property `is Standalone` that would allow hiding label (the `aria-label` is always in the input though)

Open questions:
 - @domarku @fabe do we want to allow icons in inputs at the beginning (search for example would be the case that could use that)

### DIFF
--- a/packages/components/inputs/src/base-input/BaseInput.tsx
+++ b/packages/components/inputs/src/base-input/BaseInput.tsx
@@ -1,0 +1,152 @@
+import React, {
+  useEffect,
+  useCallback,
+  FocusEvent,
+  KeyboardEvent,
+  ChangeEvent,
+  useState,
+  ElementType,
+} from 'react';
+import { cx } from 'emotion';
+
+import {
+  usePrimitive,
+  PolymorphicComponentWithRef,
+  PolymorphicComponentProps,
+  PolymorphicComponent,
+} from '@contentful/f36-core';
+import getInputStyles from './BaseInput.styles';
+import { BaseInputInternalProps } from './types';
+
+const DEFAULT_TAG: ElementType = 'input';
+
+export type BaseInputProps<
+  E extends React.ElementType
+> = PolymorphicComponentProps<E, BaseInputInternalProps, 'disabled'>;
+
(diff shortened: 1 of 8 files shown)


fabe:
> @domarku @fabe do we want to allow icons in inputs at the beginning (search for example would be the case that could use that)

Yep, I'm in favor of allowing icons as prefixes (passing in a Forma icon as a prop)

packages/components/inputs/src/base-input/BaseInput.tsx
```diff
@@ -0,0 +1,152 @@
...
+export type BaseInputProps<
+  E extends React.ElementType
+> = PolymorphicComponentProps<E, BaseInputInternalProps, 'disabled'>;
+
+const _BaseInput: PolymorphicComponentWithRef<
+  BaseInputInternalProps,
+  typeof DEFAULT_TAG
+> = (props, ref) => {
+  const {
+    as,
```
massao, line 32:
```suggestion
    as = DEFAULT_TAG,
```
Otherwise it renders as a `div`

massao (commented):
Also you will need to fix the export on the `f36-components`, you can remove this line:
https://github.com/contentful/forma-36/blob/589acecf3757615dff33eba428c44d99abdb94d9/packages/forma-36-react-components/src/index.ts#L2
and update with the `BaseInput` here:
https://github.com/contentful/forma-36/blob/589acecf3757615dff33eba428c44d99abdb94d9/packages/forma-36-react-components/src/index.ts#L64-L65

the export for inputs was duplicated for some reason

PR #1230, opened 2021-08-25
feat: v4 TextInput component

burakukula:
TextInput component that uses `BaseInput` under the hood. 

In the separate PR:
 - fix CopyButton, so that we would be able to propagate the `isDisbale` prop. It is needed for TextInput with CopyButton
 - when CopyButton is disabled, do not disable the Tooplit
 - after confirming with the design team, we would need to implement sizes for inputs to match with sizes of buttons.
 
 After all those changes ⬆️ I will create follow-up PR with additional improvements for input with CopyButton


packages/components/forms/stories/RadioButtonField.stories.tsx
```diff
@@ -22,7 +21,7 @@ export default {
 export const Basic = (args: RadioButtonFieldProps) => {
   const [activeOption, setActiveOption] = useState('yes');
   return (
-    <FieldGroup>
+    <Flex flexDirection="column">
```
burakukula, line 24: For some reason, FieldGroup and Form is not working. I don't want to lose too much time in fixing it, since we will remove it anyway.

packages/components/forms/src/text-input/TextInput.styles.ts
```diff
@@ -0,0 +1,31 @@
...
+      height: '100%',
+      borderRadius: `0 ${tokens.borderRadiusMedium} ${tokens.borderRadiusMedium} 0`,
+
+      '&:focus': {
+        zIndex: tokens.zIndexDefault,
+        borderColor: isInvalid
+          ? tokens.red600
+          : isDisabled
+          ? tokens.gray300
+          : tokens.blue600,
```
gui-santos, line 20: My suggestion here would be to replace the ternary with a function like this

```ts
const getBorderStyles = (isDisabled, isInvalid) => {
  if (isDisabled) {
    return tokens.gray300
  }

  if (isInvalid) {
    return tokens.red600
  }

  return tokens.blue600
}

// ...

borderColor: getBorderColor(isDisabled, isInvalid)
```

I think I saw this pattern in other components or something similar, btw
The reason for my suggestion is that it's easier to maintain and it returns early when the input is disabled

packages/components/forms/src/text-input/TextInput.styles.ts
```diff
@@ -0,0 +1,31 @@
...
+        boxShadow: isInvalid
+          ? tokens.glowNegative
+          : isDisabled
+          ? 'none'
+          : tokens.glowPrimary,
+      },
```
gui-santos, line 26: Same suggestion as the one above, in fact, I think you could merge both styles in the same function

packages/components/forms/src/text-input/TextInput.tsx
```diff
@@ -0,0 +1,141 @@
...
+  // Store a copy of the value in state.
+  // This is used by this component when the `countCharacters`
+  // option is on
+  const handleOnChange = useCallback(
+    (e: ChangeEvent<HTMLInputElement>) => {
+      setValueState(e.target.value);
+      if (onChange) onChange(e);
+    },
+    [onChange],
+  );
```
gui-santos, line 50: I think you don't need this `useCallback` here.... I might be wrong

I guess if the person using this component wants to optimise it, they can pass a `useCallback` as the value of their `onChange` props

But I'm basing my opinion on [this article](https://kentcdodds.com/blog/usememo-and-usecallback) which I may not have understood correctly hahahaha

packages/components/forms/src/text-input/TextInput.tsx
```diff
@@ -0,0 +1,141 @@
...
+import { ValidationMessage } from '@contentful/f36-validation-message';
+import { HelpText } from '@contentful/f36-helptext';
+import { CopyButton } from '@contentful/f36-copybutton';
+import getStyles from './TextInput.styles';
+import { TextInputProps } from './types';
+
+export const _TextInput = (
+  {
+    className,
+    testId = 'cf-ui-text-field',
```
gui-santos, line 14:
```suggestion
    testId = 'cf-ui-text-input',
```

packages/components/forms/src/text-input/TextInput.tsx
```diff
@@ -0,0 +1,141 @@
...
+          type="text"
+          onChange={handleOnChange}
+          placeholder={placeholder}
+          as="input"
+          name={name}
+          isInvalid={isInvalid}
+          isDisabled={isDisabled}
+          isReadOnly={isReadOnly}
+          icon={icon}
+          maxLength={maxLength}
```
gui-santos, line 105: I think the props that have the same name and will be passed down without any work on them should be included in `otherProps`, no?

do you know what I mean?

you don't need to do `placeholder={placeholder}`, just leave it included in `otherProps` and it will be passed down when you do `{...otherProps}`

packages/components/forms/src/text-input/TextInput.tsx
```diff
@@ -0,0 +1,141 @@
...
+  );
+  return (
+    <Flex flexDirection="column" fullWidth>
+      {!isStandalone && (
+        <Flex justifyContent="space-between" alignItems="center" fullWidth>
+          <Label htmlFor={id} required={isRequired}>
+            {label}
+          </Label>
+          {link && link}
+        </Flex>
```
gui-santos, line 86: I think this will render some extra spans even if we don't pass link, right?

```html
<span> <!-- This is generated by the first Flex -->
  <span> <!-- This is generated by the second Flex -->
    <label for="my-input">My label</label>
  </span>
  
  <input id="my-input" />
```

Maybe we could find a way to eliminate that second span, wdyt?

packages/components/forms/src/text-input/TextInput.tsx
```diff
@@ -0,0 +1,141 @@
...
+          isDisabled={isDisabled}
+          isReadOnly={isReadOnly}
+          icon={icon}
+          maxLength={maxLength}
+        />
+      )}
+
+      {!isStandalone && (
+        <>
+          {(helpText || countCharacters || validationMessage) && (
```
gui-santos, line 111:
```suggestion
      {!isStandalone && (helpText || countCharacters || validationMessage) && (
        <>
```

you can do this ☝️  and eliminate the second "if"

packages/components/forms/src/text-input/types.ts
```diff
@@ -0,0 +1,66 @@
...
+   */
+  onFocus?: EventHandler<FocusEvent<HTMLInputElement>>;
+  /**
+   * Allows to listen to an event when an element loses focus
+   */
+  onBlur?: EventHandler<FocusEvent<HTMLInputElement>>;
+  /**
+   * Allows to listen to an event when a key is pressed
+   */
+  onKeyDown?: KeyboardEventHandler<HTMLInputElement>;
```
gui-santos, line 65: Aren't these props coming from `BaseInputInternalProps` already?

packages/components/forms/stories/RadioButtonField.stories.tsx
gui-santos: I'm using `Flex` in `TextArea` as well

packages/components/forms/stories/TextInput.stories.tsx
```diff
@@ -0,0 +1,245 @@
...
+    />
+  );
+};
+
+Basic.args = {
+  label: 'Example label for input',
+  validationMessage: 'validation message',
+  helpText: 'help text',
+  placeholder: 'this is my placeholder',
+};
```
gui-santos, line 49: IMO the Basic story should be the most, well, basic thing you can do. In this case, I think we should pass only label and placeholder

Then maybe we can create a story showing how to handle errors and that would use `validationMessage`

One showing the helpText thing as well

packages/components/forms/stories/TextInput.stories.tsx
```diff
@@ -0,0 +1,245 @@
...
+import { SectionHeading } from '@contentful/f36-typography';
+import { Flex } from '@contentful/f36-core';
+import { SearchIcon } from '@contentful/f36-icons';
+
+export default {
+  title: 'Form Elements/TextInput',
+  component: TextInput,
+  parameters: {
+    propTypes: [TextInput['__docgenInfo']],
+  },
```
gui-santos, line 15: You can remove this, we don't need it anymore

gui-santos (changes requested):
Looking good!

I added a couple of suggestions, tell me what you think

Besides that, I noticed you only deleted the old stories files
is that it? Don't we need to delete more stuff?

burakukula:
> Looking good!
> 
> I added a couple of suggestions, tell me what you think
> 
> Besides that, I noticed you only deleted the old stories files
> is that it? Don't we need to delete more stuff?

Yeah, I'm gonna create a separate PR with that, I need to replace TextInput in components that are using it. Would it be ok?

packages/components/forms/src/text-input/TextInput.tsx
```diff
@@ -0,0 +1,141 @@
...
+import { Label } from '@contentful/f36-forms';
+import { ValidationMessage } from '@contentful/f36-validation-message';
+import { HelpText } from '@contentful/f36-helptext';
+import { CopyButton } from '@contentful/f36-copybutton';
+import getStyles from './TextInput.styles';
+import { TextInputProps } from './types';
+
+export const _TextInput = (
+  {
+    className,
```
gui-santos, line 13: Oh I forgot to talk about this

I don't see this `className` being used anywhere
I was actually going to ask "What should we do with className in these input/field components?"

packages/components/forms/src/text-input/TextInput.styles.ts
gui-santos: Actually, I have another suggestion... hahhahaah

This is how I'm doing in `TextArea`, I defined one class for `error` and `disabled` with the correct styles
then in the component I do this:

```jsx
className={cx(styles.textarea, {
  [styles.error]: Boolean(validationMessage),
  [styles.disabled]: Boolean(isDisabled)
})}
```

gui-santos: Actually, I have another suggestion... hahhahaah

This is how I'm doing in `TextArea`, I defined one class for `error` and `disabled` with the correct styles
then in the component I do this:

```jsx
className={cx(styles.textarea, {
  [styles.error]: Boolean(validationMessage),
  [styles.disabled]: Boolean(isDisabled)
})}
```

packages/components/forms/src/text-input/TextInput.mdx
```diff
@@ -0,0 +1,159 @@
...
+
+### Help text and validation message
+
+TextInput is able to display additional information for users like `helpText` or `validationMessage`
+
+```jsx
+<React.Fragment>
+  <TextInput
+    type="email"
+    labelText="E-mail address"
```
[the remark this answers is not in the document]
burakukula: Yeah, sorry, it should be label.  Fixign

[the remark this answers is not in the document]
burakukula: Yeah, label, sorry, mixed up

packages/components/forms/src/text-input/TextInput.tsx
burakukula: yeah, good point. Thanks!

packages/components/forms/src/text-input/TextInput.tsx
burakukula: But if we do just 
```
const handleOnChange =     (e: ChangeEvent<HTMLInputElement>) => {
      setValueState(e.target.value);
      if (onChange) onChange(e);
    }
```
(Or you had some different approach in mind?)
then we gonna get a warning `A component is changing an uncontrolled input of type text to be controlled`

packages/components/forms/src/text-input/TextInput.styles.ts
burakukula: yeah, good idea, did it the way you suggested.

packages/components/forms/src/text-input/TextInput.tsx
```diff
@@ -0,0 +1,135 @@
...
+        onCopy={onCopy}
+        className={copyButtonStyles}
+      />
+    </Flex>
+  );
+
+  return (
+    <Flex flexDirection="column" fullWidth className={className}>
+      {!isStandalone &&
+        (link ? (
```
burakukula, line 78: @gui-santos maybe something like this to avoid additional span?

packages/components/forms/src/base-input/index.ts
```diff
@@ -1,2 +1,3 @@
 export { BaseInput } from './BaseInput';
 export type { BaseInputProps } from './BaseInput';
+export type { BaseInputInternalProps } from './types';
```
burakukula, line 3: Maybe just temporary, we would rather avoid exposing this to users?

packages/components/forms/src/text-input/TextInput.styles.ts
massao: Also the `css` function accepts an array, so you could do something similar on the `.styles.ts` file, the `GhostCheckbox` has an example of that approach:
https://github.com/contentful/forma-36/blob/e2e1471ac8c59a9b91c155f075ddf57f3d8e34c6/packages/components/forms/src/base-checkbox/GhostCheckbox.styles.ts#L64

packages/components/forms/src/text-input/TextInput.tsx
massao: Since this handleOnChange is passed down to the `BaseInput`, isn't it better to have it inside the useCallback?? so the reference is always the same instead of creating a new anonymous function every render?? So React would use the same reference to the function??

packages/components/forms/src/text-input/TextInput.mdx
```diff
@@ -0,0 +1,98 @@
...
+- [Content recommendations](#content-recommendations)
+- [Accessibility](#accessibility)
+- [Props](#props)
+
+## How to use TextInput
+
+The TextInput component can be configured in a number of different ways. Use this guidance to determine the correct variation based on the context.
+
+TextInput is supporting different types of inputs, like `'text' | 'email' | 'file' | 'number' | 'password' | 'search' | 'url'`, where `text` is a default.
+Label property is always required and passed to the `<input />` as `aria-label`.
```
burakukula, line 26: I would rename it to aria-label, but I need to also do it for BaseInput I guess. Separate PR?

packages/components/forms/src/text-input/types.ts
```diff
@@ -0,0 +1,22 @@
+import { BaseInputInternalProps } from '../base-input/types';
```
burakukula, line 1: I should not probably do it, but we would need to refactor it anyway.

suevalov (approved):
👍🏼 as a first iteration.

I guess we can think more about how we can use the following pattern to extract CopyButton from the component itself: 


But I think we can merge it now in order to unblock FormControl development.

burakukula:
> 👍🏼 as a first iteration.
> 
> I guess we can think more about how we can use the following pattern to extract CopyButton from the component itself:
> https://chakra-ui.com/docs/form/input#left-and-right-addons
> 
> But I think we can merge it now in order to unblock FormControl development.

great! We can come back to this topic. I've already discussed it a bit with designers and they didn't see the need to have a lot of flexibility for the element that appears after input, but we can come back to this discussion together with Dominik.

packages/components/forms/src/text-input/TextInput.mdx
```diff
@@ -0,0 +1,78 @@
...
+- [Component variations](#component-variations)
+- [Content recommendations](#content-recommendations)
+- [Accessibility](#accessibility)
+- [Props](#props)
+
+## How to use TextInput
+
+The TextInput component can be configured in a number of different ways. Use this guidance to determine the correct variation based on the context.
+
+TextInput is supporting different types of inputs, like `'text' | 'email' | 'file' | 'number' | 'password' | 'search' | 'url'`, where `text` is a default.
```
gui-santos, line 25:
```suggestion
TextInput is supporting different types of inputs, like `'text' | 'email' | 'file' | 'number' | 'password' | 'search' | 'url'`, where `text` is the default.
```

packages/components/forms/src/text-input/TextInput.mdx
```diff
@@ -0,0 +1,78 @@
...
+Label property is always required and passed to the `<input />` as `aria-label`.
+
+## Code examples
+
+```jsx
+<React.Fragment>
+  <TextInput type="email" name="email" label="Email address" />
+</React.Fragment>
```
gui-santos, line 33: Why the Fragments?

packages/components/forms/src/text-input/TextInput.mdx
```diff
@@ -0,0 +1,78 @@
...
+## TextInput variations
+
+### Character count (`countCharacters`)
+
+The TextInput can be configured to display a character count. This is useful when an input field has a character limit.<br/>
+**Note**: The `maxLength` property must be also provided.
+
+```jsx
+<TextInput
+  type="text"
```
gui-santos, line 45: this prop is already the default right? 
I think we can remove defaults from examples so we provide only the minimum a dev needs to build the component
wdyt?

packages/components/forms/src/text-input/TextInput.mdx
```diff
@@ -0,0 +1,78 @@
...
+
+We provide a variant of the TextInput that is a composed with CopyButton. In order to use it add `withCopyButton` property to your implementation.
+
+```jsx
+<TextInput label="Token" name="token" id="input-2" withCopyButton />
+```
+
+## Content recommendations
+
+- Use direct, succinct copy that helps the user to successfully complete the form
```
gui-santos, line 65:
```suggestion
- Use direct and succinct copy that helps the user to successfully complete the form
```

Shouldn't we also mention that they should use the `HelpText` for this purpose?

packages/components/forms/src/text-input/TextInput.mdx
```diff
@@ -0,0 +1,78 @@
...
+- Do not preface the instructions with introductory text, such as "please"
+- To add additional context, link out to documentation
```
gui-santos, line 67: I didn't understand this, what does it mean?

packages/components/forms/src/text-input/TextInput.mdx
```diff
@@ -0,0 +1,78 @@
...
+- In cases where additional context is unnecessary, help text can be substituted for text in the label
```
gui-santos, line 68: Can it be replaced by label?

packages/components/forms/src/text-input/TextInput.mdx
```diff
@@ -0,0 +1,78 @@
...
+- Do not use punctuation for labels, help text and links
```
gui-santos, line 69:
```suggestion
- Do not use punctuation in labels, help text and links
```

Idk honestly... It sounds better for me with "in"
but I'm not a grammar teacher so feel free to ignore this

packages/components/forms/src/text-input/TextInput.styles.ts
```diff
@@ -0,0 +1,40 @@
...
+  }),
+  disabled: css({
+    button: css({
+      '&:focus': {
+        borderColor: tokens.gray300,
+        boxShadow: 'none',
+      },
+    }),
+  }),
+});
```
gui-santos: As I told you in slack, no need for two `css` calls probably

packages/components/forms/src/text-input/TextInput.tsx
```diff
@@ -0,0 +1,97 @@
...
+    isInvalid,
+    isDisabled,
+    withCopyButton,
+    onCopy,
+    ...otherProps
+  }: TextInputProps,
+  ref: React.Ref<HTMLInputElement>,
+) => {
+  const [valueState, setValueState] = useState<string | undefined>(value);
+  const styles = getStyles();
```
gui-santos, line 29: Since you are not passing any props to this function, I think it's better to do this call outside of the component

Otherwise when this component re-renders it will call `getStyles` and will build the same CSS classes every time

packages/components/forms/src/text-input/TextInput.tsx
```diff
@@ -0,0 +1,97 @@
...
+    maxLength,
+    isInvalid,
+    isDisabled,
+    withCopyButton,
+    onCopy,
+    ...otherProps
+  }: TextInputProps,
+  ref: React.Ref<HTMLInputElement>,
+) => {
+  const [valueState, setValueState] = useState<string | undefined>(value);
```
gui-santos, line 28:
```suggestion
  const [valueState, setValueState] = useState(value);
```

TS will infer the type of value here... I think

massao (approved):

packages/components/forms/src/text-input/TextInput.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+      {countCharacters && maxLength && (
+        <Flex justifyContent="flex-end">
+          <Box marginTop="spacingXs" marginLeft="spacingM">
+            <HelpText>
+              {valueState ? valueState.length : 0}/{maxLength}
+            </HelpText>
+          </Box>
+        </Flex>
+      )}
```
gui-santos, line 92: Uhmmmmmm 🤔 

I'm thinking now that we should do something else here

Because of this case:
Screenshot 2021-08-31 at 16 25 41

the counter needs to be aligned with ValidationMessage

Maybe we should build TextInput without the counter for now, and in another PR build an example with `FormControl` + `ValidationMessage` and then implement the counter somehow

packages/components/forms/src/text-input/TextInput.tsx
gui-santos: Btw, do we really need a state here?

couldn't we use ref to get the value in the input? I'm not sure either, just wondering

packages/components/forms/src/text-input/TextInput.tsx
```diff
@@ -0,0 +1,97 @@
...
+        {...otherProps}
+        testId={testId}
+        ref={ref}
+        label={label}
+        type="text"
+        onChange={handleOnChange}
+        as="input"
+        maxLength={maxLength}
+        className={styles.inputWithCopyButton}
+      />
```
gui-santos, line 56: If we do need `useState`, then I think here we should do `value={valueState}`

otherwise it's possible that `value !== valueState` 💥

packages/components/forms/src/text-input/types.ts
gui-santos: I think it's better if you export the type in `/base-input/index.ts`
and then here you keep importing from `../base-input`

wdyt?

packages/components/forms/src/text-input/types.ts
```diff
@@ -0,0 +1,22 @@
...
+export interface TextInputProps extends Omit<BaseInputInternalProps, 'as'> {
+  /**
+   * Controlls the character count for the input, works together with maxLength propery
+   * @default false
+   */
+  countCharacters?: boolean;
+  /**
+   * Pass max number of characters for the input, needed for countCharacters prop
+   */
+  maxLength?: number;
```
gui-santos, line 12: I think this is also a native `input` attribute, so maybe no need to put it in here 🤷 

that's why we also need `countCharacters` to tell if we should show the counter or not

packages/components/forms/stories/TextInput.stories.tsx
```diff
@@ -0,0 +1,105 @@
...
+export const Basic = (args: TextInputProps) => {
+  return <TextInput {...args} name="Example name" id="input-1" />;
+};
+
+Basic.args = {
+  label: 'Example label for input',
+  placeholder: 'this is my placeholder',
+};
+
+export const overview = () => (
```
gui-santos, line 32:
```suggestion
export const Overview = () => (
```

just for code consistency... not really important

gui-santos (changes requested):
It's looking good!

I got a few more tiny suggestions 🙏

packages/components/forms/src/text-input/TextInput.mdx
burakukula: Me neither, hahaha

packages/components/forms/src/text-input/TextInput.mdx
burakukula: This is actually not valid anymore since this component does not have labels, help texts nor links

packages/components/forms/src/text-input/TextInput.styles.ts
```diff
@@ -0,0 +1,36 @@
...
+  inputWithCopyButton: css({
+    borderRadius: `${tokens.borderRadiusMedium} 0 0 ${tokens.borderRadiusMedium}`,
+  }),
+  copyButton: css({
+    button: {
+      borderLeft: 'none',
+      height: '100%',
+      borderRadius: `0 ${tokens.borderRadiusMedium} ${tokens.borderRadiusMedium} 0`,
+
+      '&:focus': {
```
burakukula, line 14: most likely those styles will be moved to CopyButton (we would also adjust button secondary in that case) - it is discussed with designers

packages/components/forms/src/text-input/TextInput.mdx
massao: we could pass the label as `placeholder` and `aria-label` if the placeholder isn't defined, for cases where the TextInput won't be used with the `FormControl`, WDYT?

packages/components/forms/src/text-input/types.ts
burakukula: ok, I can try that. Maybe that's better, although it is temporary. we would approach it differently anyway right?

packages/components/forms/src/text-input/TextInput.mdx
burakukula: @massao I don't know if it would not be confusing. I would say, we enforce aria-label for this component and that's it. I think a placeholder is not required? Label is handled as a separate component. I would rename `label` to `aria-label` but I want to do it in separate PR, to not mix 2 different components (this change have to be done for BaseInput as well)

massao: placeholder can be required depending on the context, like in a search bar or something, it may not have the label, the github search bar itself as example, the `aria-label` and `placeholder` have the same value. `aria-label` to be accessible to screen readers and `placeholder` to be visually accessible.

packages/components/forms/src/text-input/TextInput.tsx
burakukula: @massao cool! thanks for the feedback! okidoks, `useCallback` it is then.

packages/components/forms/src/text-input/TextInput.mdx
burakukula: I would still keep them as separate properties - `aria-label` and `placeholder`. Placeholder is not required for input to pass a11y check, although it does not mean it’s usable. I will check with designers if they have any cases in mind then you might want to use an input without a placeholder.

packages/components/forms/src/text-input/types.ts
burakukula: hmmm, seems like when I import:

```
export type { BaseInputProps } from './BaseInput';
export type { BaseInputInternalProps } from './types';
```
in the `BaseInput` index. the build of the form package fail. I don't really get why.

burakukula: @gui-santos @massao I am not able to make it work in a different way than importing from `'../base-input/types'`. 🤷‍♀️

gui-santos (approved):

suevalov (approved):

PR #1235, opened 2021-08-25
refactor(checkbox): Create BaseCheckbox component

massao:
# Purpose of PR

Create the `BaseCheckbox` component that will handle the `Checkbox`, `RadioButton` and `Switch` components:

image

`BaseCheckbox` uses the same interface as `BaseInput` but since we have the custom checkmarks, we had to create a separate component to be able to handle it properly and keep it accessible.

## PR Checklist

- [x] Tests are passing


packages/components/inputs/src/base-checkbox/types.ts
```diff
@@ -0,0 +1,37 @@
+import type { ComponentProps } from 'react';
+import type { BaseInputInternalProps } from '../base-input/types';
+
+export type checkboxTypes = 'checkbox' | 'radio' | 'switch';
+
+export interface BaseCheckboxProps
+  extends Omit<
+    BaseInputInternalProps,
+    'type' | 'as' | 'placeholder' | 'isReadOnly' | 'icon' | 'isInvalid'
```
suevalov, line 9: I'm not sure if we need to exclude `isInvalid` here

packages/components/inputs/src/base-checkbox/types.ts
```diff
@@ -0,0 +1,37 @@
...
+  isChecked?: boolean;
+  /**
+   * Defines if the state is indeterminate
+   * @default false
+   */
+  isIndeterminate?: boolean;
+  /**
+   * Additional props that are passed to the input element
+   */
+  inputProps?: Partial<ComponentProps<'input'>>;
```
suevalov, line 29: Do we still need this? If yes, then in what cases?

suevalov (commented):
I looks really good in general, but I have one strong IMO that all refs should go directly to input not to label.

packages/components/inputs/src/base-checkbox/types.ts
burakukula: yeah, I think we actually gonna need it.

suevalov (approved):
All comments going to be addressed in the follow up PR

burakukula (approved):

PR #1244, opened 2021-08-26
fix(pill): small visual improvements

suevalov:
# Purpose of PR

Removes TabFocusTrap from `Pill` component.
Updates focused style to match `<Button variant="transparent" />`

### DIFF
--- a/packages/components/pill/src/styles.ts
+++ b/packages/components/pill/src/styles.ts
@@ -2,36 +2,21 @@ import { css } from 'emotion';
 import tokens from '@contentful/f36-tokens';
 import { PillVariants } from './types';
 
-const getCloseButtonStyle = (variant: PillVariants) => {
-  const variantStyle = (variant: PillVariants) => {
-    switch (variant) {
-      case 'active':
-        return {
-          borderLeft: `1px solid ${tokens.gray400}`,
-          '&:hover, &:focus, &:active': {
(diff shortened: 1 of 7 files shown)


packages/components/pill/src/styles.ts
```diff
@@ -1,38 +1,23 @@
...
-    borderBottomRightRadius: tokens.borderRadiusSmall,
+    borderLeft: `1px solid ${tokens.gray400}`,
+    borderTopLeftRadius: 0,
+    borderBottomLeftRadius: 0,
     padding: tokens.spacingXs,
-    background: 'transparent',
-    transition: `background ${tokens.transitionDurationShort} ${tokens.transitionEasingDefault}`,
-    ...variantStyle(variant),
+    minHeight: 'auto',
+    transition: `background ${tokens.transitionDurationShort} ${tokens.transitionEasingDefault},
```
bgutsol, line 12: Probably don't need it, since we have this transition style in `Button`

bgutsol (approved):
LGTM ✨
Only one minor suggestion, not a blocker

packages/components/pill/src/styles.ts
suevalov: we actually need this, because the button has also `border` transition and we don't need it here.

gui-santos (approved):

PR #1252, opened 2021-08-27
feat(forms): Create Checkbox component

massao:
# Purpose of PR

Creates a Checkbox component

## PR Checklist

- [ ] Tests are passing


packages/components/forms/src/checkbox/Checkbox.tsx
```diff
@@ -1,15 +1,37 @@
...
 export type CheckboxProps = _CheckboxProps;
 
-export const _Checkbox = (
-  { testId = 'cf-ui-checkbox', ...otherProps }: CheckboxProps,
-  ref: React.Ref<HTMLLabelElement>,
+const _Checkbox = (
+  { testId = 'cf-ui-checkbox', label, id, className, ...otherProps },
+  ref: React.Ref<HTMLDivElement>,
 ) => {
+  const styles = getStyles();
```
gui-santos, line 14: Since this will return the same thing every time
Wdyt about calling it outside of the component?

packages/components/forms/src/index.ts
```diff
@@ -1,10 +1,3 @@
...
-export * from './controlled-input-field';
-export * from './checkbox-field';
-export * from './radio-button-field';
-export * from './controlled-input';
+export * from './radio';
 export * from './checkbox';
-export * from './switch';
-export * from './radio-button';
-export * from './base-input';
-export * from './base-checkbox';
```
gui-santos, line 10: 👀 

All these components can be deleted?
Is that reason why parcel is failing in my branch then? hahahah

packages/components/forms/src/radio/Radio.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { Label } from '../Label';
+import { getStyles } from './Radio.styles';
+
+export type RadioProps = _RadioProps;
+
+const _Radio = (
+  { testId = 'cf-ui-radio-button', label, id, className, ...otherProps },
+  ref: React.Ref<HTMLDivElement>,
+) => {
+  const styles = getStyles();
```
gui-santos, line 14: same suggestion about the `getStyles()` here

gui-santos (approved):
Amazing job

And great 🧹  in all those files

Only got a tiny suggestion about `getStyles`
and I agree with @suevalov that maybe we should pass the ref to the inputs

packages/components/forms/src/checkbox/Checkbox.tsx
suevalov: we don't want to create styles before the component is rendered, that's why we can them here. this pattern is consistent across all components.

massao: Also as emotion works, it stores all the classNames, so it would return the memoized className

packages/components/forms/src/index.ts
massao: yeah, some were the old migrated version, and some where created on the `baseCheckbox` PR, but after the discussion we had, we don't need all of them anymore.

packages/components/forms/src/checkbox/Checkbox.tsx
gui-santos: so we are using emotion wrong in user_interface?

suevalov: yeah, we could use it better in user_interface

packages/components/forms/src/index.ts
burakukula: oh, the pure joy of solving conflicts 😄

packages/components/forms/src/radio/Radio.mdx
```diff
@@ -0,0 +1,78 @@
...
+title: 'Radio'
+type: 'component'
+status: 'stable'
+slug: /components/forms/src/radio/
+github: 'https://github.com/contentful/forma-36/tree/master/packages/components/forms/src/radio'
+storybook: 'https://f36-storybook.contentful.com/?path=/story/form-elements-radio--basic'
+typescript: ./Radio.tsx
+---
+
+`Radio` is a form component, that shows a checkbox with it's label, if you want to let the user select multiple options, use the `Checkbox` component.
```
burakukula, line 11:
```suggestion
`Radio` is a form component, that shows a radio input with its label. It is used to allow users to choose single option from the list. If you want to let the user select multiple options, use the `Checkbox` component.
```

burakukula (approved):
🎉

suevalov:
I think this PR has to be rebased and adapted to new types that @bgutsol did

PR #1254, opened 2021-08-30
refactor: compound components

suevalov:
# Purpose of PR

Makes `Grid`, `Tabs`, `List`, and `Accordion` to follow a compound component pattern.

### DIFF
--- a/packages/components/accordion/Accordion.mdx
+++ b/packages/components/accordion/Accordion.mdx
@@ -34,13 +34,13 @@ The accordion has two variations that define the alignment of the chevron icon: 
 
 ```jsx
 <Accordion>
-  <AccordionItem title="What payment methods do you accept?">
+  <Accordion.Item title="What payment methods do you accept?">
     <Text>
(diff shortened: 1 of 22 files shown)


denkristoffer (approved):
🙏

bgutsol (approved):
Awesome 👍

Only one comment: I remember @gui-santos was saying that we don't need to pass ['__docgenInfo'] in stories anymore. It's working automagically 🎩 Please correct me if I'm wrong @gui-santos .

gui-santos:
> Awesome 👍
> 
> Only one comment: I remember @gui-santos was saying that we don't need to pass ['__docgenInfo'] in stories anymore. It's working automagically 🎩 Please correct me if I'm wrong @gui-santos .

@bgutsol you are correct!

We had to do this when we started using storybook docs because at that time we didn't have a MDX file for each component, and that's how storybook recommended people to do the migration

suevalov:
Ah, good to know. We can delete all `['__docgenInfo']` as a follow up PR.

PR #1262, opened 2021-08-31
fix: allow isDisabled property for CopyButton

burakukula:
We would need to allow CopyButton to be disbaled on order to make it work nicely with input, when it's disbaled.

When CopyButton is disabled, we should not show the tooltip.

### DIFF
--- a/packages/components/copybutton/src/CopyButton.styles.ts
+++ b/packages/components/copybutton/src/CopyButton.styles.ts
@@ -31,4 +31,18 @@ export const getStyles = () => ({
       boxShadow: tokens.glowMuted,
     },
   }),
+  copyButtonDisbaled: css({
+    cursor: 'not-allowed',
+    backgroundColor: tokens.gray100,
+
+    '&:focus': {
+      borderColor: tokens.gray300,
+      boxShadow: 'none',
+    },
+
+    '&:active, &:active:hover': {
+      borderColor: tokens.gray300,
+      boxShadow: 'none',
+    },
+  }),
 });
--- a/packages/components/copybutton/src/CopyButton.styles.ts
+++ b/packages/components/copybutton/src/CopyButton.styles.ts
@@ -31,7 +31,7 @@ export const getStyles = () => ({
       boxShadow: tokens.glowMuted,
     },
   }),
-  copyButtonDisbaled: css({
+  copyButtonDisabled: css({
     cursor: 'not-allowed',
     backgroundColor: tokens.gray100,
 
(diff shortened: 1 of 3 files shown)


packages/components/copybutton/src/CopyButton.tsx
```diff
@@ -77,21 +83,33 @@ function _CopyButton(props: CopyButtonProps, ref: React.Ref<HTMLDivElement>) {
...
+              type="button"
+              ref={button}
+              className={cx(styles.copyButton)}
+              aria-label={`Copy ${value} to clipboard`}
+            >
+              <CopyIcon variant="muted" />
+            </button>
+          </Tooltip>
+        </CopyToClipboard>
+      )}
```
gui-santos, line 112: Instead of doing all this stuff

I think you can pass `disabled={isDisabled}` to the button and to the tooltip you pass `content=""` and it won't render it

Yes I know, this is not great DX
we should change this once we migrate the tooltip to v4
Or have we already done that?
If yes, we can just make it possible to force the tooltip not render

suevalov: Ideally, we should introduce `isDisabled` to a tooltip component and it it's true, we should just render `children`

burakukula: yeah, good idea. Thanks!

packages/components/copybutton/src/CopyButton.styles.ts
```diff
@@ -31,4 +31,18 @@ export const getStyles = () => ({
       boxShadow: tokens.glowMuted,
     },
   }),
+  copyButtonDisbaled: css({
```
massao, line 34:
```suggestion
  copyButtonDisabled: css({
```

packages/components/copybutton/src/CopyButton.tsx
```diff
@@ -35,6 +35,11 @@ export interface CopyButtonProps extends CommonProps {
    * @default Copy {value} to clipboard
    */
   label?: string;
+  /**
+   * Allows to disbale the copy button, when true, tooltip would not be shown
```
massao, line 39:
```suggestion
   * Allows to disable the copy button, when true the tooltip would not be shown
```

packages/components/tooltip/src/Tooltip.tsx
```diff
@@ -88,6 +88,11 @@ export interface TooltipProps extends CommonProps {
    * Defaults to `false`
    */
   usePortal?: boolean;
+  /**
+   * Prevents from showing the tooltip
```
massao, line 92:
```suggestion
   * Prevents showing the tooltip
```

massao (commented):
Just some small suggestions and typos

suevalov (approved):

PR #1265, opened 2021-09-01
 feat: migrate textarea to v4

gui-santos:
# Purpose of PR

- Migrate Textarea to v4
- removed the `size` prop from v3 because based on the design files and on what @burakukula  we don't want it anymore
- with the new approach to the form inputs, this component does NOT implement `FormLabel`, `ValidationMessage`, `HelpText`, and character counters. Those will be handled in the future once we have `FormControl `

## PR Checklist

- [ ] Tests are passing


packages/components/forms/src/textarea/Textarea.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import { cx } from 'emotion';
+
+import { BaseInput } from '../base-input';
+import type { BaseInputInternalProps } from '../base-input/types';
+import { getStyles } from './Textarea.styles';
+
+export type TextareaProps = Omit<BaseInputInternalProps, 'as'>;
```
massao, line 8: Can't we import the `BaseInputProps` directly? So it also has the types from the 'input' html tag?
Also maybe Omit the `type` property as it has no effect on the textarea component

packages/components/forms/src/textarea/README.mdx
```diff
@@ -0,0 +1,87 @@
...
+  ref={textareaRef}
+/>
+
+<Button onClick={() => textareaRef.current.focus()}>
+  Focus Textarea with ref
+</Button>
+```
+
+The `Textarea` is exposing the ref of the `textarea` tag as a prop,
+which means that you can pass a ref created with `React.createRef()` to it and it will work as expected.
```
massao, line 80:
```suggestion
which means that you can pass a ref created with `React.createRef()` or `useRef` hook to it and it will work as expected.
```

packages/components/forms/src/textarea/README.mdx
```diff
@@ -0,0 +1,87 @@
...
+
+## Accessibility
+
+- To ensure accessibility, provide a `name` prop
+- In case the input is required, use `isRequired` prop
+- Provide validation to the input and use `isInvalid` to control the error state of `Textarea`. Also show meaninful error messages with the `ValidationMessage` component.
```
massao, line 86: Maybe the part of the `ValidationMessage` would be better off on the `FormControl` component??
As it's not strictly necessary to have a validation for using the textarea by itself, but it should be good to have if it's being used on a Form, but not sure how to phrase that. 🤔

packages/components/forms/src/base-input/BaseInput.styles.ts
```diff
@@ -24,6 +23,11 @@ const getStyles = ({ isDisabled, isInvalid }) => ({
     cursor: isDisabled ? 'not-allowed' : 'auto',
     width: '100%',
 
+    // if the input is a textarea, the height is resizeble
```
massao, line 26: Can't this styling be passed from the `TextField` or `TextArea` component? Since emotion would merge the passed style with this, that should be the base style?

gui-santos: but do you think it's correct if someone does this `<BaseInput as="textarea" />` and this textarea has a max-height of 40 px?

massao: Ah good point, I forgot that we would still export the `BaseInput`

packages/components/forms/src/textarea/README.mdx
gui-santos: Yep! This is meant to be only recommendations for a11y forms
And I'm recommending them to use `ValidationMessage` because it is red and has the `!` icon
I mean it will look like the designs

But they should do that in FormControl of couse


Ah do you mean move this orientation to FormControl’s documentation? 
That's fine for me

packages/components/forms/src/textarea/Textarea.tsx
gui-santos: Yes we can

packages/components/forms/src/textarea/README.mdx
massao: yeah, exactly having this part on the FormControl documentation, but we can have on both actually.

packages/components/forms/src/textarea/Textarea.tsx
gui-santos: Maybe we can't?
Screenshot 2021-09-01 at 15 51 06

packages/components/forms/src/textarea/README.mdx
gui-santos: I will leave it here then and will try to write something about this being relevant when the textarea is in a form then 👌

packages/components/forms/src/textarea/Textarea.tsx
massao: But shouldn't the label be required anyway? Or `ariaLabel` as it's going to be renamed to: https://github.com/contentful/forma-36/pull/1266

gui-santos: I'm not doing anything in Textarea
The component is using the BaseInputProps directly (well... omitting two props but still)

Isn't the error here saying that to `BaseInput` `label` is required but to `Textarea` (and therefore `BaseInputProps`) `label` is optional???

which means, `BaseInputProps` and `BaseInputInternalProps` are treating label differently and `BaseInputProps` does not reflect the real API of `BaseInput`

suevalov (approved):

packages/components/forms/src/textarea/Textarea.tsx
massao: Ohh got it, the `BaseInputProps` isn't exported from `types.ts`
either:
```js
import type { BaseInputProps } from '../base-input';
```
or
```js
import { BaseInput, BaseInputProps } from '../base-input';
 ```

gui-santos: Screenshot 2021-09-01 at 17 08 33

There is still something to be done here

massao: How are you using the type?? it should be `Omit<BaseInputProps<'textarea'>, 'as' | 'type'>`

gui-santos: voilá! I was missing the `<'textarea'>` in BaseInputProps

massao (approved):

PR #1272, opened 2021-09-02
feat: border-radius-round token

denkristoffer:
# Purpose of PR

This PR introduces a border-radius token to make things completely round. There are a few places in our UI where we need round items, and I figured having a token for it might be nice. On the other hand, maybe this will lead people to overuse it.

Thoughts?

### DIFF
--- a/packages/forma-36-tokens/src/tokens/border-radius/border-radius.js
+++ b/packages/forma-36-tokens/src/tokens/border-radius/border-radius.js
@@ -1,6 +1,7 @@
 const borderRadius = {
   'border-radius-small': '4px',
   'border-radius-medium': '6px',
+  'border-radius-round': '999999999px',
 };
 
 module.exports = borderRadius;


packages/forma-36-tokens/src/tokens/border-radius/border-radius.js
```diff
@@ -1,6 +1,7 @@
 const borderRadius = {
   'border-radius-small': '4px',
   'border-radius-medium': '6px',
+  'border-radius-round': '999999999px',
```
gui-santos, line 4: What happens if you use `100%`?

denkristoffer: The `px` value is a force of habit for me since I've had trouble with percentages earlier on, where they didn't round an element as I wanted it to. So you give me a chance to braindump some CSS theory here and also try to explain this to myself as well 🤓 

When you use percentage values for `border-radius`, you are referring to border box of the element, so for instance, settings border radius `100%` on the top left corner would mean a curve from the bottom left corner of the box to the top right corner of the box, [like this](https://codesandbox.io/s/autumn-wind-s9kzj?file=/index.html):

Screenshot 2021-09-03 at 10 00 37

So it doesn't make sense to use `100%` on all sides of an element as the radii (radiuses?) would overlap. So browsers are actually required by the CSS spec to [scale it down to `50%` anyway](https://codesandbox.io/s/admiring-ardinghelli-g7ff3?file=/index.html). So actually, `50%` is the value you need 🙂 

`px` values work differently than percentages, they are not relative to the border box. For square elements this doesn't matter, the `999999999px` value [results in the same thing](https://codesandbox.io/s/late-meadow-zxlkc?file=/index.html). But when it comes to rectangular elements, [we see a difference](https://codesandbox.io/s/confident-agnesi-m5kmt?file=/index.html):
Screenshot 2021-09-03 at 10 18 03

Why? I'm actually not sure I understand this myself yet, I will see if I can dig deeper. I suppose the browser in this case scales the overlapping values differently 🤔 Personally I prefer the behaviour of the `px` rectangle, but that is up for discussion.

gui-santos: Ahhhh got it!

Wow thanks for sharing the research 🙌

gui-santos (approved):
💎

packages/forma-36-tokens/src/tokens/border-radius/border-radius.js
thebesson: I would challenge if we need this or not as a token. First of all, because of the different ways people can round borders in CSS, as you explained earlier. second of all, it makes sense to me to have specific border radius values as part of the library, as it is part of the design guidelines, we want to have the same border radiuses across the app. but fully round elements are not exactly specific to our library. anyway, that's just how I see it.

denkristoffer: Okay, I can understand that, but I also think that the point of the design system is to decide on a common approach and then make it easy to use that approach everywhere. In this case, to me, this is like providing a token for the colour black. I want to help avoid ending up with different hardcoded values for the same thing.

massao (approved):
🎉

packages/forma-36-tokens/src/tokens/border-radius/border-radius.js
denkristoffer: @massao I still don't understand why a `100%` value scales differently than `999999px` since they both create overlapping borders. Can you explain it? 😬

massao: @denkristoffer sure, basically it's because `percentage` 😄 
Like I said, the border-radius accepts two values for each corner, being  `X/Y`.
Considering a rectangle with `{ width: 200px; height: 100px; }`, when setting the `border-radius: 50%`(or higher for that matter, since it will replicate for all other corners it will round down to 50%), it is the same as `border-radius: 50%/50%`(X/Y), which in turn would be the same as `border-radius: 100px/50px` or 50% the size of our border-box.
While with `border-radius: 999999px`(same as `border-radius: 999999px/999999px `), it will reduce that amount of pixels until it fits in all corners, and since it's being passed only one value both radii should be the same, in the example, to fit all 4 curves within the rectangle, it gets as small as `border-radius: 50px/50px`,

suevalov (approved):

domarku:
I can't find any examples of it's use in our products so far. Could you point me to those few places in our UI, @denkristoffer?

denkristoffer:
@domarku When we render user avatars we round them. There are also a few other cases of `border-radius: 100%` in our legacy CSS but I can't tell you for certain where they are rendered.

domarku:
Ah, yes! Good find.

I'm not sure if that one use case is enough for us to introduce this as a token, because I don't see us using the "100%" border-radius any time in the future. Seems like it's totally fine that it's one line of CSS in that part of the code.

denkristoffer:
Fair enough, my only "fear" is that we have a bunch of different values – `100%`, `50%`, `9999999px`, etc. – for the same thing, and we won't catch them all with a search if we ever want to replace them.

PR #1273, opened 2021-09-02
fix: TextInput/BaseInput default values

burakukula:


packages/components/forms/src/text-input/TextInput.tsx
```diff
@@ -22,43 +23,42 @@ export const _TextInput = (
   }: TextInputProps,
   ref: React.Ref<HTMLInputElement>,
 ) => {
-  const [valueState, setValueState] = useState(value);
+  const textInputRef = useRef(null);
```
burakukula, line 26: @gui-santos you were right 🙂 (Although I don't know if I do it correctly. probably no)

packages/components/forms/src/text-input/TextInput.tsx
```diff
@@ -22,43 +23,42 @@ export const _TextInput = (
...
     [styles.invalid]: Boolean(isInvalid),
   });
 
+  const handleCopy = (e) => {
+    if (onCopy) {
+      onCopy(e);
+    }
+
+    textInputRef.current.select();
+    document.execCommand('copy');
```
suevalov, line 40: does it work in all browsers we support?

packages/components/forms/src/text-input/types.ts
```diff
@@ -11,4 +11,9 @@ export interface TextInputProps
    * Allows to listen to an copy button
    */
   onCopy?: (value: string) => void;
+  /**
+   * Set's default value for text input
+   * @default false
```
suevalov, line 16:
```suggestion
```

packages/components/forms/src/base-input/BaseInput.tsx
```diff
@@ -109,7 +102,8 @@ function _BaseInput<E extends React.ElementType = typeof DEFAULT_TAG>(
       style={style}
       placeholder={placeholder}
       className={cx(styles.input, iconClassName, className)}
-      value={valueState}
+      value={value}
+      defaultValue={defaultValue}
```
massao, line 106: We have to pass only one of those, because if we pass value even empty `value=""` it will consider the value, and set the input as controlled, and will not change it's value unless an onChange is also defined

packages/components/forms/src/base-input/BaseInput.tsx
```diff
@@ -86,16 +83,12 @@ function _BaseInput<E extends React.ElementType = typeof DEFAULT_TAG>(
...
-    setValueState(value);
-  }, [value]);
-
   const iconContent = icon && (
     <Box as="span" className={styles.iconPlaceholder}>
       {React.cloneElement(icon, {
         size: 'small',
         variant: 'muted',
-        ariaHiden: true,
+        'aria-hidden': true,
```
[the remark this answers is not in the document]
massao: There was a typo on the previous version: `ariaHiden` instead of `ariaHidden`, maybe that's the reason??

packages/components/forms/src/text-input/TextInput.tsx
```diff
@@ -22,43 +23,42 @@ export const _TextInput = (
...
+  };
+
   const inputWithCopyButton = (
     <Flex className={className}>
       <BaseInput
         {...otherProps}
         testId={testId}
-        ref={ref}
+        ref={textInputRef}
```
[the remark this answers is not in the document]
massao: Something like this: https://github.com/contentful/forma-36/pull/1252/commits/e75fd98d399a76135f75d7a417ecb62b6d6acb8d

packages/components/forms/src/text-input/TextInput.tsx
burakukula: yeah.

packages/components/forms/src/base-input/BaseInput.tsx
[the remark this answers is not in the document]
burakukula: 🤦‍♀️ on the typo thing 😁
Unfortunately, I am still getting the same warning.

packages/components/forms/src/base-input/BaseInput.tsx
burakukula: @massao WDYT?

suevalov: IMO, we need to pass both, but `value` should be undefined by default

suevalov (approved):

PR #1275, opened 2021-09-02
feat: Improve form control api and docs

suevalov:
# Purpose of PR

* Adds default marginBottom to FormLabel, ValidationMessage, and HelpText components. Allows to override it with margin props.
* Adds different examples of using FormControl in the storybook


bgutsol (approved):

packages/components/forms/stories/FormControl.stories.tsx
```diff
@@ -13,12 +21,74 @@ export default {
 
 export const Basic = (args: FormControlInternalProps) => {
   return (
-    <FormControl {...args}>
-      <FormControl.Label isRequired>Name</FormControl.Label>
-      <TextInput label="input" />
+    <>
+      <FormControl {...args}>
+        <FormControl.Label isRequired>Name</FormControl.Label>
+        <TextInput label="input" />
```
gui-santos, line 27: I thought the label, in this case, would be provided by `FormControl.Label`
why do we need to pass label here?

packages/components/forms/stories/FormControl.stories.tsx
```diff
@@ -13,12 +21,74 @@ export default {
...
+            <TextLink
+              as="button"
+              icon={<LockIcon />}
+              onClick={() => {
+                setIsDisabled(true);
+              }}
+            >
+              Lock
+            </TextLink>
+          )}
```
gui-santos, line 84:
```suggestion
            <TextLink
              as="button"
              icon={<LockIcon />}
              onClick={() => {
                setIsDisabled(prevState => !prevState);
              }}
            >
              {isDisabled ? 'Unlock to edit' : 'Lock'}
            </TextLink>
          )}
```

I think it's possible to do the same without repeating the code as much, wdyt?

P.S.: it also teaches people about prevState

gui-santos (commented):
It looks good

I only have suggestions on the stories

packages/components/forms/stories/FormControl.stories.tsx
suevalov: Right now it's enforced by TextInput API, but @burakukula is going to remove it.

packages/components/forms/stories/FormControl.stories.tsx
suevalov: yeah, it's a great suggesion

gui-santos (approved):

massao (approved):

PR #1276, opened 2021-09-02
refactor: Make BaseCheckbox possible to be controlled and uncontrolled.

massao:
# Purpose of PR

- Update BaseCheckbox to be able to either controlled or uncontrolled.
- Update it to use children instead of `label` property.

## PR Checklist

- [ ] Tests are passing


packages/components/forms/src/base-checkbox/BaseCheckbox.mdx
```diff
@@ -2,7 +2,7 @@
 title: 'BaseCheckbox'
 type: 'component'
 status: 'stable'
-slug: /components/base-checkbox/
+slug: /components/inputs/src/base-checkbox/
```
massao, line 5:
```suggestion
slug: /components/base-checkbox/
```

packages/components/forms/src/base-checkbox/GhostCheckbox.styles.ts
```diff
@@ -13,31 +13,31 @@ const getBaseGhostStyles = ({ isDisabled }): CSSObject => ({
...
       fill: !isDisabled ? tokens.colorWhite : tokens.gray300,
     },
-  };
-
-  const checkedStyle = {
-    backgroundColor: !isDisabled ? tokens.blue600 : tokens.gray300,
-    borderColor: !isDisabled ? tokens.blue600 : tokens.gray300,
-    '& svg': {
-      fill: !isDisabled ? tokens.colorWhite : tokens.gray600,
+    'input:indeterminate + &, input:checked + &': {
```
massao, line 28: This change was needed so it updates the GhostCheckbox correctly if the component is uncontrolloed.

suevalov (approved):
👏🏼 awesome! I like it the way more!

packages/components/forms/src/base-checkbox/BaseCheckbox.tsx
```diff
@@ -99,40 +84,43 @@ function _BaseCheckbox(
...
-        aria-label={label}
       />
       <GhostCheckbox
         type={type}
-        isChecked={checked}
         isDisabled={isDisabled}
         isIndeterminate={isIndeterminate}
       />
-    </Flex>
+      {children}
```
burakukula, line 122: I find it a bit less intuitive to pass a label to the component like the checkbox or radio as children. Maybe it's just me. 
I would rather expect to have something like we had before, so:
```
<Checkbox
  label="my label over here"
/>
```

But again, might be just me. 🤷‍♀️

massao: I think this may be just a familiarity bias, kind of how we think it's fine to have the value of the `textarea` inside the tags, and the for the `input` it has the value.

packages/components/forms/src/base-checkbox/BaseCheckbox.mdx
```diff
@@ -22,50 +22,94 @@ BaseCheckbox is used to show a controlled checkbox or radio button.
...
-        label="Take the blue pill"
         name="pill-decision"
         value="blue"
         isChecked={activeOption === 'blue'}
         onChange={(e) => {
           setActiveOption(e.target.value);
         }}
-      />
+      >
+        Take the blue pill
```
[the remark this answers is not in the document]
massao: that was exact my reaction when I was migrating the documentation 🤣

PR #1277, opened 2021-09-02
feat: Select component

suevalov:
# Purpose of PR

Migrates `<Select>` to v4


packages/components/forms/src/select/Select.styles.ts
```diff
@@ -0,0 +1,87 @@
+import tokens from '@contentful/f36-tokens';
+import { css, cx } from 'emotion';
+import { SelectWidth } from './types';
+
+function getStylesForSize(props: { width?: SelectWidth }) {
```
burakukula, line 5: I think we need to rethink the size prop. Selects should match inputs, and inputs need to match buttons in terms of height. 😁 But we can refactor later I guess.

packages/components/forms/src/select/Select.styles.ts
```diff
@@ -0,0 +1,87 @@
...
+export function getSelectStyles(props: {
+  isInvalid?: boolean;
+  isDisabled?: boolean;
+  width?: SelectWidth;
+}) {
+  const select = css({
+    width: '100%',
+    display: 'block',
+    appearance: 'none',
+    padding: '10px',
```
burakukula, line 30: I guess it should be the same like for inputs: `padding: `10px ${tokens.spacingS}`

packages/components/forms/src/select/Select.styles.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  const disabled = css({
+    backgroundColor: tokens.gray100,
+    cursor: 'not-allowed',
+    color: tokens.gray600,
+    '&:focus, &:active': { borderColor: tokens.gray300 },
+  });
+
+  const invalid = css({
+    borderColor: props.isDisabled ? tokens.red300 : tokens.colorNegative,
```
burakukula, line 59: Most likely would be `tokens.red600` not `tokens.colorNegative` - select and input should match styles.

packages/components/forms/src/select/Select.styles.ts
```diff
@@ -0,0 +1,87 @@
...
+}) {
+  const select = css({
+    width: '100%',
+    display: 'block',
+    appearance: 'none',
+    padding: '10px',
+    paddingRight: tokens.spacingL,
+    backgroundColor: tokens.colorWhite,
+    color: tokens.gray700,
+    fontSize: tokens.fontSizeM,
```
burakukula, line 34: we probably miss `lineHeight: tokens.lineHeightM,` over here

packages/components/forms/src/select/Select.styles.ts
```diff
@@ -0,0 +1,87 @@
...
+    fontFamily: tokens.fontStackPrimary,
+    borderRadius: tokens.borderRadiusMedium,
+    boxShadow: tokens.insetBoxShadowDefault,
+    height: '40px',
```
burakukula, line 38: placeholder color should also match input so `color: tokens.gray500,`

massao (approved):
🥇

burakukula (approved):
Looks good! 👏

PR #1278, opened 2021-09-03
feat: V4/menu

bgutsol:
# Purpose of PR

Menu component

**Tests and documentation will be introduced in the separate PR, in order to not block this component**

## PR Checklist

- [x] Tests are passing


packages/components/menu/package.json
```diff
@@ -0,0 +1,35 @@
...
+  "name": "@contentful/f36-menu",
+  "version": "4.0.0",
+  "description": "Forma 36: Menu component",
+  "scripts": {
+    "build": "parcel build index.ts"
+  },
+  "dependencies": {
+    "@babel/runtime": "^7.6.2",
+    "@contentful/f36-core": "^4.0.0",
+    "@contentful/f36-tokens": "^4.0.0",
```
suevalov, line 11: popover is missing as a dependency

packages/components/menu/src/Menu.tsx
```diff
@@ -0,0 +1,226 @@
...
+
+  return { isOpen: isOpenValue, isControlled, handleClose, handleOpen };
+};
+
+interface UseArrowKeyNavigationProps {
+  itemsContainerRef: React.MutableRefObject<HTMLElement>;
+  itemsSelector: string;
+}
+
+const useArrowKeyNavigation = ({
```
denkristoffer, line 178: Worth extracting? https://github.com/contentful/forma-36/discussions/1008

suevalov (approved):
👏🏼

packages/components/menu/src/MenuItem/MenuItem.styles.ts
```diff
@@ -0,0 +1,27 @@
...
+export const getMenuItemStyles = () =>
+  css({
+    display: 'block',
+    width: '100%',
+    background: 'none',
+    border: 0,
+    outline: 'none',
+    textDecoration: 'none',
+    fontSize: tokens.fontSizeM,
+    lineHeight: tokens.lineHeightCondensed,
```
gui-santos, line 13:
```suggestion
    lineHeight: tokens.lineHeightM,
```

But based on the figma file, it looks like the height of each item is `32px`

I'm not sure if you want to achieve it with line height or other css attribute

packages/components/menu/src/MenuItem/MenuItem.styles.ts
```diff
@@ -0,0 +1,27 @@
...
+    color: tokens.gray600,
+    textAlign: 'left',
+    padding: '6px 16px',
```
gui-santos, line 16: why not tokens here?

packages/components/menu/src/MenuList/MenuList.tsx
```diff
@@ -0,0 +1,58 @@
...
+    >
+      {headerItem &&
+        React.cloneElement(headerItem, {
+          className: cx(styles.headerItem, headerItem.props.className),
+        })}
+      {children}
+      {footerItem &&
+        React.cloneElement(footerItem, {
+          className: cx(styles.footerItem, footerItem.props.className),
+        })}
```
gui-santos, line 53: Shouldn't we transform headerItem and footerItem into Components? (e.g.: `Menu.ListHeader`)

packages/components/menu/src/MenuSubheading/MenuSubheading.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+  return (
+    <Subheading
+      testId={testId}
+      className={cx(styles, className)}
+      marginBottom="none"
+      {...otherProps}
+    >
+      {children}
+    </Subheading>
```
gui-santos, line 26: Based on the designs this should be a SectionHeading

Screenshot 2021-09-06 at 16 36 23

gui-santos (changes requested):
Great job!

The code looks good and the api is great, but I'm still not sure about the styles
Some parts are not looking like the designs

this is the pr
Screenshot 2021-09-06 at 16 34 30

This is the design on Figma
Screenshot 2021-09-06 at 16 36 23

The spacing between the popover and the first and last items is too small
it should be the same space that they have on their left side

The title should be using SectionHeading I think

And each MenuItem should have 32px of height

I suggest working with @domarku to fix those things

packages/components/menu/src/MenuItem/MenuItem.styles.ts
suevalov: updated

packages/components/menu/src/MenuList/MenuList.tsx
suevalov: @bgutsol said the reasoning basically was that by doing them components we cannot control a place where they will be put. `ListHeader` should always be the first and `ListFooter` should always be the last.

packages/components/menu/src/MenuSubheading/MenuSubheading.tsx
suevalov: Updated

packages/components/menu/src/Menu.tsx
suevalov: I guess, we'll extract this if we'll be using it in any other component.

packages/components/menu/src/Menu.tsx
```diff
@@ -0,0 +1,226 @@
...
+
+  /**
+   * If `true`, the Menu will be initially opened.
+   */
+  defaultIsOpen?: boolean;
+
+  /**
+   * Callback fired when the Menu opens
+   */
+  onOpen?: Function;
```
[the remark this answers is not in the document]
suevalov: a good suggestion, done for both functions

packages/components/menu/src/MenuList/MenuList.tsx
gui-santos: I think we could
I noticed the same thing in the workbench, but since we didn't want to change the Workbench I didn't do it

The strategy would be to check the children of MenuList, then you can pick `ListHeader` and `ListFooter` by checking the `displayName` for example and render them always on top and bottom

That way it would not matter if someone does 

```tsx
<MenutItem>
<MenutItem>
<MenutItem>
<MenutFooter>
<MenutHeader>
```

I wanted to do that for the Sidebars in workbench, because now you need to position them correctly and set `position` prop

but that's not necessary

suevalov: I'll check what you're proposing and see if I'll be able to implement that.

gui-santos: but, honestly, we don't need to do it now

We can do it later if we want... if we don't want it we just don't do it

gui-santos (approved):
🍗

packages/components/menu/src/MenuList/MenuList.tsx
gui-santos: If you want to check, the Workbench is doing that for the header

```jsx
const header: React.ReactNode[] = [];
  const content: React.ReactNode[] = [];

  React.Children.forEach(
    children,
    // eslint-disable-next-line
    (child: any) => {
      if (child.type?.name === 'WorkbenchHeader') {
        header.push(child);
      } else {
        content.push(child);
      }
    },
  );
```

In v3, we had that `any` as the type of the child and I tried to replace it with proper typing but I couldn't
I tried using ReactNode, Child, Element, etc... all of them had some issue

I'm still trying to figure out the best way in TS for this pattern

suevalov: @gui-santos pushed the change, as you suggested.

packages/components/menu/src/MenuList/MenuList.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { MenuListHeader } from './MenuListHeader';
+import { MenuListFooter } from './MenuListFooter';
+
+interface MenuListInternalProps extends CommonProps {
+  children?: React.ReactNodeArray;
+}
+
+function assertChild(child: any): child is { type: { displayName: string } } {
+  return Boolean(child?.type?.displayName);
+}
```
gui-santos, line 16: Nice! 👌

packages/components/menu/src/MenuList/MenuList.tsx
```diff
@@ -0,0 +1,69 @@
...
+        header = (child as unknown) as React.ReactElement;
+        appendChild = false;
+      } else if (child.type.displayName === MenuListFooter.displayName) {
+        footer = (child as unknown) as React.ReactElement;
+        appendChild = false;
+      }
+    }
+    if (appendChild) {
+      items.push((child as unknown) as React.ReactElement);
+    }
```
gui-santos, line 47: If I understood this right, we want to check if the child has `displayName`
then if the display name is `ListHeader`, we assign it to the `header` variable
if the display name is `ListFooter`, we assign it to the `footer` variable
otherwise, we just append it to the `items` array

is that correct?


What do you think about this 👇 ?
It would reduce the number of if statements and we would not need the appendChild logic
I'm not sure if I got it right though

```jsx
const hasDisplayName = assertChild(child)
    
if (hasDisplayName && child.type.displayName === MenuListHeader.displayName) {
  header = (child as unknown) as React.ReactElement;
} else if (hasDisplayName && child.type.displayName === MenuListFooter.displayName) {
  footer = (child as unknown) as React.ReactElement;
} else {
  items.push((child as unknown) as React.ReactElement);
}
```

This is not a blocker btw, I've already approved the PR
It's just an idea that came to my mind when reading the solution

gui-santos (approved):
Just made one suggestion that I don't think is a blocker

It's just an idea that I had to "simplify" a bit the logic of the children handling

packages/components/menu/src/MenuList/MenuList.tsx
suevalov: unfortunately, assert functions don't work this way in Typescript, so I had to come with a way I implemented it.

gui-santos: Ohhhhh of course!

the solution looks great anyway 🙌

massao (approved):

PR #1281, opened 2021-09-03
feat: add GlobalStyles

suevalov:
# Purpose of PR

Introduces a GlobalStyles component that allows to override all default styles in the browser and set a correct default font-size and font-family.

### DIFF
--- a/packages/components/typography/src/Text.tsx
+++ b/packages/components/typography/src/Text.tsx
@@ -45,7 +45,7 @@ function _Text<E extends React.ElementType = typeof DEFAULT_TAG>(
     fontStack = 'fontStackPrimary',
     fontWeight = 'fontWeightNormal',
     fontColor = 'gray700',
-    lineHeight,
+    lineHeight = 'lineHeightDefault',
     children,
     isTruncated,
     as,
--- a/packages/components/typography/src/Text.tsx
+++ b/packages/components/typography/src/Text.tsx
@@ -44,7 +44,7 @@ function _Text<E extends React.ElementType = typeof DEFAULT_TAG>(
     fontSize = 'fontSizeM',
     fontStack = 'fontStackPrimary',
     fontWeight = 'fontWeightNormal',
-    fontColor = 'gray700',
+    fontColor = 'gray800',
     lineHeight = 'lineHeightDefault',
     children,
     isTruncated,
--- a/packages/components/typography/src/Text.tsx
+++ b/packages/components/typography/src/Text.tsx
@@ -45,7 +45,7 @@ function _Text<E extends React.ElementType = typeof DEFAULT_TAG>(
     fontStack = 'fontStackPrimary',
     fontWeight = 'fontWeightNormal',
     fontColor = 'gray800',
-    lineHeight = 'lineHeightDefault',
+    lineHeight = 'lineHeightM',
     children,
     isTruncated,
     as,
--- a/packages/core/src/GlobalStyles/GlobalStyles.tsx
+++ b/packages/core/src/GlobalStyles/GlobalStyles.tsx
@@ -21,7 +21,6 @@ export const GlobalStyles = ({
           min-height: 100%;
           font-size: 16px;
           font-family: ${tokens.fontStackPrimary};
-          line-height: ${tokens.lineHeightDefault};
           vertical-align: baseline;
         }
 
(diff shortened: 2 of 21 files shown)


massao (approved):

packages/core/src/GlobalStyles/GlobalStyles.tsx
```diff
@@ -0,0 +1,45 @@
...
+          border: 0;
+          min-height: 100%;
+        }
+
+        body {
+          font-size: ${tokens.fontSizeM};
+          font-family: ${tokens.fontStackPrimary};
+          color: ${tokens.gray800};
+          line-height: ${tokens.lineHeightDefault};
+          vertical-align: baseline;
```
massao, line 29: should we set: `boxSizing: "border-box"`?
It would fix the issue on the search bar:
image

suevalov: Compose has the following code:

```
*,
html {
  box-sizing: border-box;
}
```

Since I'm CSS Junior Developer, I'd like to ask for opinions :)

massao: I think this would cover all elements and pseudo-elements and can be easily overwritten if needed:
```css
html {
  box-sizing: border-box;
}
*, *:before, *:after {
  box-sizing: inherit;
}
```

packages/components/typography/src/Text.tsx
```diff
@@ -44,8 +44,8 @@ function _Text<E extends React.ElementType = typeof DEFAULT_TAG>(
     fontSize = 'fontSizeM',
     fontStack = 'fontStackPrimary',
     fontWeight = 'fontWeightNormal',
-    fontColor = 'gray700',
-    lineHeight,
+    fontColor = 'gray800',
+    lineHeight = 'lineHeightDefault',
```
gui-santos, line 48: I think we decided to deprecate the tokens that have "functions" in their names like "default"
and we wanted to use the tokens named with their sizes, because it's more scalable

more in this [thread](https://contentful.slack.com/archives/GFB336BAA/p1612430272018100)

and based on this Figma file (Outdated, but the only one I could find... posted the link of the file in slack)
Screenshot 2021-09-06 at 15 28 31



the line height here is `lineHeightM`

packages/core/src/GlobalStyles/GlobalStyles.tsx
```diff
@@ -0,0 +1,47 @@
...
+          padding: 0;
+          border: 0;
+          min-height: 100%;
+        }
+
+        body {
+          font-size: ${tokens.fontSizeM};
+          font-family: ${tokens.fontStackPrimary};
+          color: ${tokens.gray800};
+          line-height: ${tokens.lineHeightDefault};
```
gui-santos, line 29:
```suggestion
          line-height: ${tokens.lineHeightM};
```

packages/core/src/GlobalStyles/GlobalStyles.tsx
gui-santos: +1 to ☝️

packages/forma-36-website/src/components/Layout.css
```diff
@@ -1,26 +1,17 @@
...
 pre {
   margin: 0;
   white-space: normal;
-  font-family: var(--font-stack-primary);
+  font-family: var(--font-stack-primary) !important;
 }
 
 pre.prism-code {
-  font-family: var(--font-stack-monospace);
+  font-family: var(--font-stack-monospace) !important;
```
gui-santos, line 14: Why `!important`?

gui-santos (changes requested):
Looks good!

I'm only questioning if the line height token is the correct one
And as I say in my comment, I'm basing my question on the last information I could find

packages/forma-36-website/src/components/Layout.css
massao: I believe that the prism library we use for highlighting the code already has a font-family set. And due to specificity it needs the !important flag, to correct override that.

gui-santos: got it

packages/components/typography/src/Text.tsx
suevalov: got it

suevalov:
@gui-santos good suggestions about line height, it makes sense for me 👍🏼

gui-santos (approved):
🍭

PR #1268, opened 2021-09-01
feat: add default id to workbench [SPA-237]

Inbal-Gordon:
# Purpose of PR
Added an ID to workbench 

## PR Checklist

- [x] I have read the relevant `readme.md` file(s)
- [x] All commits follow our [Git commit message convention](https://github.com/contentful/forma-36/tree/master/packages/forma-36-react-components#commits)
- [x] Tests are added/updated/not required
- [x] Tests are passing
- [x] Storybook stories are added/updated/not required
- [x] Usage notes are added/updated/not required
- [x] Has been tested based on [Contentful's browser support](https://www.contentful.com/faq/about-contentful/#which-browsers-does-contentful-support)
- [x] Doesn't contain any sensitive information

PR #1283, opened 2021-09-03
feat: migrate workbench to v4

gui-santos:
# Purpose of PR

- Created new `@contentful/f36-workbench` package
- the workbench components had a `labelText` prop that was used to set `aria-label`, I removed in favor of extending the html props of their HTML tags
- Workbench got removed from `@contentful/f36-components`

## PR Checklist

- [ ] Tests are passing


packages/components/workbench/Workbench.mdx
```diff
@@ -0,0 +1,11 @@
+---
```
suevalov, line 1: looks like a duplicate of an other file

packages/components/workbench/src/Workbench.mdx
```diff
@@ -0,0 +1,72 @@
...
+storybook: 'https://f36-storybook.contentful.com/?path=/docs/components-workbench--basic'
+typescript: ./Workbench.tsx,./WorkbenchHeader/WorkbenchHeader.tsx,./WorkbenchContent/WorkbenchContent.tsx,./WorkbenchSidebar/WorkbenchSidebar.tsx
+---
+
+import { Props } from '@contentful/f36-docs-utils';
+
+An app shell that structures regions of content and interaction.
+The workbench assembles the outer app shell which defines regions to structure content and interactions.
+
+## Table of contents
```
suevalov, line 16: we need to mention that this package is not included by default and the components can be imported from another package

packages/components/workbench/src/WorkbenchContent/WorkbenchContent.tsx
```diff
@@ -0,0 +1,35 @@
...
+  children,
+  className,
+  type = 'default',
+  testId = 'cf-ui-workbench-content',
+  ...otherProps
+}: WorkbenchContentProps) {
+  const styles = getWorkbenchContentStyles(type);
+
+  return (
+    <main
```
suevalov, line 27: could you please incorporate latest changes from master? https://github.com/contentful/forma-36/pull/1268

packages/components/workbench/src/WorkbenchHeader/WorkbenchHeader.styles.ts
```diff
@@ -0,0 +1,40 @@
...
+    alignItems: 'center',
+    flexShrink: 0,
+    width: '100%',
+    height: '70px',
+    padding: `0 ${tokens.spacingL}`,
+    borderBottom: `1px solid ${tokens.gray300}`,
+    backgroundColor: tokens.gray100,
+    zIndex: tokens.zIndexWorkbenchHeader,
+    boxSizing: 'border-box',
+    ...(hasBackButton && { paddingLeft: 0 }),
```
massao, line 16: IMO it would look better if you pass an array to the `css()` method here, instead of manually destructuring

packages/components/workbench/src/WorkbenchSidebar/WorkbenchSidebar.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import { cx } from 'emotion';
+import { CommonProps, PropsWithHTMLElement } from '@contentful/f36-core';
+
+import { getWorkbenchSidebarStyles } from './WorkbenchSidebar.styles';
+
+export interface WorkbenchSidebarOwnProps extends CommonProps {
+  children: React.ReactNode;
+  position?: 'left' | 'right';
```
massao, line 9: Shouldn't we add a description and default value? Also for the other components.

packages/components/workbench/src/Workbench.mdx
```diff
@@ -0,0 +1,72 @@
+---
```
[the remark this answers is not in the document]
gui-santos: I don't know tbh, I try to replicate the mdx in other components

I will move to the root 👌

packages/components/workbench/src/Workbench.tsx
```diff
@@ -0,0 +1,61 @@
...
+function _Workbench(
+  { children, className, testId = 'cf-ui-workbench' }: WorkbenchProps,
+  ref: React.Ref<HTMLDivElement>,
+) {
+  const styles = getWorkbenchStyles();
+
+  const header: React.ReactNode[] = [];
+  const content: React.ReactNode[] = [];
+
+  React.Children.map(
```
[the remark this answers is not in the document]
gui-santos: I will try it!

packages/components/workbench/src/Workbench.mdx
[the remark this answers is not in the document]
massao: Ohh yeah, when we have multiple components with their own mdx file, we separate it, for example the `forms` one, but on others we have a single mdx with the documentation and props table of all the components, the Tabs for example
This being a compound and expected to be used together I would argue that it should have all documentation on the same page

packages/components/workbench/src/WorkbenchHeader/WorkbenchHeader.styles.ts
gui-santos: do you mean?

```ts
css([{ ... }, hasBackButton && { paddingLeft: 0 }])
```

I usually prefer to do that in the components, but I thought the team would prefer me doing all this emotion stuff in the getStyles function

packages/components/workbench/src/WorkbenchSidebar/WorkbenchSidebar.tsx
gui-santos: the default should be 'left', I thought I added it in the arguments of the component... no?

I didn't add description because someone told me just to migrate and deprecate it
But you are right! I will add them

packages/components/workbench/Workbench.mdx
gui-santos: What do you mean?

packages/components/workbench/src/WorkbenchHeader/WorkbenchHeader.styles.ts
massao: yeah, I just think that approach looks better than the destructuring one, but isn't a big issue anyway 😄 
Would also be fine with having a `hasBackButton` style and merging it with `cx` on the component as you said

packages/components/workbench/src/WorkbenchContent/WorkbenchContent.tsx
gui-santos: Do we still need it?
I'm extending the `<main>` attributes, so some one can do `<Workbench.Content id="some-id">` and it will work

We usually implement testId's but not ids
I tried to understand why this change was made in the PR you sent me, but there is no explanation

I volunteer to work on the Workbench v4 adoption in the product where this id is necessary and then I pass as a prop

wdyt?

packages/components/workbench/src/WorkbenchSidebar/WorkbenchSidebar.tsx
massao: I think that just a quick description should be fine, even tho it's deprecated it's still used, so the description would help.
But we shouldn't spend too much effort on it.

packages/components/workbench/Workbench.mdx
gui-santos: Ohhhh now I saw it

packages/components/workbench/src/WorkbenchContent/WorkbenchContent.tsx
massao: That is to be used with the `Skip to Content`, it jumps to that id provided there.

packages/components/workbench/Workbench.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+<Props of="Workbench" />
+
+## Props of Workbench.Header
+
+<Props of="WorkbenchHeader" />
+
+## Props of Workbench.Content
+
+<Props of="WorkbenchContent" />
```
suevalov, line 70: I don't see props table showing up in the documentation website.

gui-santos: Yep, I'm also not seeing them
I thought someone would be able to tell me why

gui-santos: Oh I discovered why! hahahahah

I forgot to put `./src` in the path of the TS files in the .mdx

suevalov (approved):
👏🏼

PR #1284, opened 2021-09-03
feat: create CheckboxGroup and RadioGroup

massao:
# Purpose of PR

Create CheckboxGroup and RadioGroup to make it easier to handle multiple inputs.
* Let me know if you think they should cover something more.
* I started them as separate files, but decided to merge into `BaseCheckboxGroupContext`, and `BaseCheckboxGroup`  as most of the code was the same, but let me know if you prefer the separate files approach.
* Didn't use the Compound Solution as there may be cases where users may want to use the `Checkbox` or `Radio` without a Group

## PR Checklist

- [ ] Tests are passing


massao:
Should we do the same for the `Switch` component??

suevalov:
@massao do you think a case for a group of switches?

massao:
> @massao do you think a case for a group of switches?

@suevalov No I couldn't think of any, but if someone can think of one, we could create one, or we can merge this one and if a use case comes to mind later we create a new PR

suevalov:
@massao I don't see use cases either, so I guess we can skip it for now.

suevalov (approved):
Looks amazing! I love these new components.

PR #1285, opened 2021-09-04
feat: Introduce size to form elements

burakukula:
Introducing sizes to input and selects.

TODO in a separate PR: refactor switch component so it allows `default` and `small` size.


packages/components/forms/src/base-input/types.ts
```diff
@@ -72,4 +72,8 @@ export interface BaseInputInternalProps extends CommonProps {
    * Allows to listen to an event when an element get focused
    */
   onFocus?: FocusEventHandler<HTMLTextAreaElement | HTMLInputElement>;
+  /**
+   * Allows setting size small to the input
+   */
+  size?: 'small' | 'default';
```
massao, line 78: Shouldn't we have the sizes as `small`, `medium` and so on as we have in other components? And have the medium being the default?

gui-santos: Yes!
It's in the code style guide already
https://github.com/contentful/forma-36/blob/forma-v4/docs/code-style-guide.md#commonly-used-props

packages/components/copybutton/src/CopyButton.styles.ts
```diff
@@ -1,12 +1,27 @@
 import { css } from 'emotion';
 import tokens from '@contentful/f36-tokens';
+import type { CSSObject } from '@emotion/serialize';
 
-export const getStyles = () => ({
+const sizeToStyles = ({ size }): CSSObject => {
```
gui-santos, line 5: Just a nit pick suggestion
IMO naming the function with `getSizeStyles` would make it easier to understand what it is doing

what did you mean with sizeTOstyles? What is this "to" in the name exactly?

packages/components/copybutton/src/CopyButton.styles.ts
```diff
@@ -1,12 +1,27 @@
...
+      };
+    default:
+      return {
+        height: `40px`,
+        width: `40px`,
+      };
+  }
+};
+
+export const getStyles = ({ size }) => ({
```
gui-santos, line 20: Btw, since it's only two options

You could move that logic here and simplify it, like this:

```ts
export const getStyles = ({ size }) => {
const sizeValue =  size === 'small' ? '32px' : '40px';

return ({
  wrapper: {
    // ...
    width: sizeValue
    height: sizeValue
  },
  // all the styles you have already
})
```

packages/components/copybutton/src/CopyButton.styles.ts
```diff
@@ -1,12 +1,27 @@
...
+  switch (size) {
+    case 'small':
+      return {
+        height: `32px`,
+        width: `32px`,
+      };
+    default:
+      return {
+        height: `40px`,
+        width: `40px`,
```
gui-santos, line 15: Why the backtick here if you're not using any js inside the string?

packages/components/forms/src/base-input/BaseInput.styles.ts
```diff
@@ -1,7 +1,26 @@
...
 
-const getStyles = ({ as, isDisabled, isInvalid }) => ({
+const sizeToStyles = ({ size }): CSSObject => {
+  switch (size) {
+    case 'small':
+      return {
+        fontSize: tokens.fontSizeM,
+        lineHeight: tokens.lineHeightM,
+        padding: `${tokens.spacing2Xs} ${tokens.spacingXs}`,
+        height: `32px`,
```
gui-santos, line 12: This is what I meant in the other comment

In `padding`, the use of the backtick makes sense because you are passing the tokens to the string
but in `height` you don't, so it should be just a single quote

packages/components/forms/src/base-input/BaseInput.styles.ts
```diff
@@ -1,7 +1,26 @@
...
+    default:
+      return {
+        fontSize: tokens.fontSizeM,
+        lineHeight: tokens.lineHeightM,
+        height: `40px`,
+      };
+  }
+};
+
+const getStyles = ({ as, isDisabled, isInvalid, size }) => ({
```
gui-santos, line 23: Same suggestion i gave before, since you only have two size variants, the logic in `sizeToStyles` could be moved here and transformed into a if statement

Even if you keep in another function, it's only two cases.... does it need to be a switch if it's only two cases?

packages/components/forms/src/select/Select.styles.ts
```diff
@@ -1,10 +1,26 @@
...
+        height: `32px`,
+      };
+    default:
+      return {
+        fontSize: tokens.fontSizeM,
+        lineHeight: tokens.lineHeightM,
+        height: `40px`,
+      };
+  }
+};
```
gui-santos, line 21: same as commented in the other styles files

packages/components/forms/src/base-input/types.ts
massao: @gui-santos I'll do this change then

packages/components/forms/src/select/Select.styles.ts
```diff
@@ -60,8 +77,8 @@ export function getSelectStyles(props: {
     ),
     select: cx(
       select,
-      props.isDisabled ? disabled : null,
-      props.isInvalid ? invalid : null,
+      isDisabled ? disabled : null,
+      isInvalid ? invalid : null,
```
gui-santos, line 81:
```suggestion
      isDisabled && disabled,
      isInvalid && invalid,
```

I think this will work, maybe you will need to do: 

```
cx([
  select,
  isDisabled && disabled,
  isInvalid && invalid,
])
```

packages/components/forms/src/base-input/BaseInput.styles.ts
```diff
@@ -1,7 +1,26 @@
...
+        fontSize: tokens.fontSizeM,
+        lineHeight: tokens.lineHeightM,
+        padding: `${tokens.spacing2Xs} ${tokens.spacingXs}`,
+        height: `32px`,
+      };
+    default:
+      return {
+        fontSize: tokens.fontSizeM,
+        lineHeight: tokens.lineHeightM,
+        height: `40px`,
```
gui-santos, line 18: This is creating a weird scroll in the textarea component

Screenshot 2021-09-06 at 14 41 30

I suggest not applying this height if `as === 'textarea'` or pass this as min-height when base input is textarea

gui-santos (changes requested):
Looking good!

Besides my suggestions, I agree with @massao regarding the values of the `size` prop

massao:
Also noticed an issue on the Select component, will create a follow up PR to address that.
* We were passing the `isSelected` to the option, when the correct approach is passing a `value` prop to the select: https://reactjs.org/docs/forms.html#the-select-tag

gui-santos:
> Also noticed an issue on the Select component, will create a follow up PR to address that.
> 
> * We were passing the `isSelected` to the option, when the correct approach is passing a `value` or `defaultValue` prop to the select: https://reactjs.org/docs/forms.html#the-select-tag
>   And it would work as a `controlled/uncontrolled` input as the other inputs

@massao 

I also noticed this in chromatic
Screenshot 2021-09-06 at 15 46 17

select input border is not red when `isInvalid = true`

Will you fix that here or in another PR?

gui-santos (approved):
Looks good!

The only problem is that red border, but maybe you will fix that in the other PR you said you will open to fix Select

massao:
> Looks good!
> 
> The only problem is that red border, but maybe you will fix that in the other PR you said you will open to fix Select

Yes, I have fixed this on the other PR already. but nice catch

PR #1291, opened 2021-09-06
fix(Modal): story improvement & allow scrolling in ModalCofirm []

Chaoste:
# Purpose of PR

## Topic 1:
A customer is complaining that a confirmation modal with some content doesn't render properly on smaller devices. The content overflows and makes it impossible to submit the dialog.
image

This works for the normal modal while it doesn't work for the confirmation one because we render a `<div>` container breaking the intended scrolling behavior of the modal.

After the fix:
Screenshot 2021-09-06 at 17 35 02

## Topic 2:
While investigating this bug and clicking through modals in the storybook, I was getting annoyed by the ModalLauncher stories always creating a modal on mount. I guess it's part of the story but at least I expect it to close them again if I switch to another story. With the unmount function I'm doing this now.

## PR Checklist

- [o] Tests are added/updated/not required
- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Modal/ModalConfirm/ModalConfirm.tsx
+++ b/packages/forma-36-react-components/src/components/Modal/ModalConfirm/ModalConfirm.tsx
@@ -173,7 +173,7 @@ export function ModalConfirm({
     >
       {() => {
         return (
-          <div>
+          <React.Fragment>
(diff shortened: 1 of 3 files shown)


massao (approved):
🎉

gui-santos (approved):
👏 👏 👏

suevalov (approved):

denkristoffer (approved):

PR #1292, opened 2021-09-06
refactor:(select) use defaultValue and value props instead of isSelected

massao:
# Purpose of PR

* Refactor `Select` component to use `value` and `defaultValue` instead of isSelected on the option.
* Update documentation with examples.

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/packages/components/forms/src/checkbox/Checkbox.tsx
+++ b/packages/components/forms/src/checkbox/Checkbox.tsx
@@ -29,7 +29,7 @@ const _Checkbox = (props: CheckboxProps, ref: React.Ref<HTMLInputElement>) => {
     name,
   });
 
-  const formProps = useFormControl({
+  const { isReadOnly, ...formProps } = useFormControl({
     id,
     isDisabled,
     isInvalid,
--- a/packages/components/forms/src/checkbox/Checkbox.tsx
+++ b/packages/components/forms/src/checkbox/Checkbox.tsx
@@ -29,6 +29,7 @@ const _Checkbox = (props: CheckboxProps, ref: React.Ref<HTMLInputElement>) => {
     name,
   });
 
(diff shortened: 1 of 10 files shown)


packages/components/forms/src/checkbox/Checkbox.tsx
```diff
@@ -29,7 +29,7 @@ const _Checkbox = (props: CheckboxProps, ref: React.Ref<HTMLInputElement>) => {
     name,
   });
 
-  const formProps = useFormControl({
+  const { isReadOnly, ...formProps } = useFormControl({
```
massao, line 32: Had to do this change on `Checkbox`, `Radio` and `Switch` as they can't have the readonly state.
And this was giving warnings on formControl.

suevalov (approved):
Makes sense, the way better and consistent with other form components

packages/components/forms/src/checkbox/Checkbox.tsx
gui-santos: Where is `isReadOnly` being used then? 
eslint is not complaining that this is not being used? 👀

gui-santos (approved):
Looks good!

I just didn't understand how is `isReadOnly` not being used anywhere and eslint does not complain

packages/components/forms/src/checkbox/Checkbox.tsx
massao: It isn't used on the Checkbox at all, and since we use the rest operator here, if we use the `formProps` variable, eslint understand that we are only removing that key from the object.

gui-santos: #TIL 
Could you please comment on that line that you are just removing it from the object?

At least, I won't remember this in 2 weeks and then I will ask myself again "What is happening here?"

PR #1295, opened 2021-09-08
fix(Button): hide spinner when loading is false

Chaoste:
# Purpose of PR

We want to use the button with a loading state in our use case. We observe buggy behaviour when removing the loading animation (the animation comes back and never disappears). 

We wanted to check if the button actually supports removing the loading animation. Because the storybook is missing an example so far I added one and tested with it. It works in the storybook, so we will do further investigations. I think it's good to keep this story for others.

*Update:*
I found a fix in [this github issue](https://github.com/reactjs/react-transition-group/issues/338#issuecomment-442538453). By checking the transition status on our own we can make sure that it's really unmounted if the status is "unmounted".

### See the buggy behaviour
https://user-images.githubusercontent.com/9327071/132471918-6642c0fc-d701-41fb-ad32-49238f14a808.mov

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Button/Button.stories.tsx
+++ b/packages/forma-36-react-components/src/components/Button/Button.stories.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useCallback } from 'react';
 
 import { Button, ButtonProps } from './Button';
 import { Flex } from '../Flex';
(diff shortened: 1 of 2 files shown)


packages/forma-36-react-components/src/components/Button/Button.stories.tsx
```diff
@@ -318,3 +318,39 @@ export const Overview = (args: ButtonProps) => (
...
+              onClick={submitForm}
+            >
+              Submit me
+            </Button>
+          </Flex>
+        </Flex>
+      </Flex>
+    </>
+  );
+};
```
gui-santos, line 356: Could you move this story up?

We try to leave the "Overview" story as the last one always

Screenshot 2021-09-09 at 10 49 15

gui-santos (approved):
Amazing fix!

I only have a nit pick request

Thanks a lot for fixing this 🙌

PR #1297, opened 2021-09-08
refactor: Add default styling for BaseCheckboxGroup

massao:
# Purpose of PR

Add a default styling for the BaseCheckboxGroup.

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/packages/components/forms/src/base-checkbox/BaseCheckbox.tsx
+++ b/packages/components/forms/src/base-checkbox/BaseCheckbox.tsx
@@ -1,5 +1,4 @@
 import React, { useCallback, useEffect } from 'react';
-import { cx } from 'emotion';
 import { useForwardedRef, PropsWithHTMLElement } from '@contentful/f36-core';
 import type { BaseCheckboxInternalProps } from './types';
 import { GhostCheckbox } from './GhostCheckbox';
--- a/packages/components/forms/src/base-checkbox/BaseCheckbox.tsx
+++ b/packages/components/forms/src/base-checkbox/BaseCheckbox.tsx
@@ -121,7 +121,7 @@ function _BaseCheckbox(
         isIndeterminate={isIndeterminate}
         size={size}
       />
-      {children}
+      <span>{children}</span>
     </Text>
   );
 }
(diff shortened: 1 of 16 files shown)


packages/components/forms/src/base-checkbox/BaseCheckbox.tsx
```diff
@@ -90,44 +90,47 @@ function _BaseCheckbox(
...
-      />
-      <GhostCheckbox
-        type={type}
-        isDisabled={isDisabled}
-        isIndeterminate={isIndeterminate}
-        size={size}
-      />
-      {children}
-    </Text>
+    <Box className={className}>
```
suevalov, line 93: I guess we can use `React.Fragment` here. Can we?

massao: Not with the way it's at the moment, because we only add the default spacing if the Checkbox/Radio is used within a group, and the context pass down the className with the styling and apply it here so it gets the same spacing in case there are checkboxes with helpText and some without helpText.
But it doesn't need to be a box necessarily either, or we could have the Context set a different prop that would be used to generate the styles on this component, so it knows if the spacing should be on the `helpText` or the `label`, or should we always add a bottom spacing regardless if it's inside a group or not??

packages/components/forms/src/base-checkbox/BaseCheckbox.tsx
```diff
@@ -90,44 +86,48 @@ function _BaseCheckbox(
...
+        />
+        <GhostCheckbox
+          type={type}
+          isDisabled={isDisabled}
+          isIndeterminate={isIndeterminate}
+          size={size}
+        />
+        {children}
+      </Text>
+      {helpText && <HelpText className={styles.helpText}>{helpText}</HelpText>}
```
suevalov, line 129: a small aria improvement: I would be great if we could add `id={id + '-helptext'}` to HelpText element and add `aria-describedby={id+ '-helptext'}` to <input> when help text is defined

suevalov (approved):
Looks great, just one aria tip

PR #1298, opened 2021-09-08
feat: Stack component

suevalov:
# Purpose of PR

In many scenarios, we want to have a simple way to have a row or a column of elements evenly spaced between.

Right, now it's can be achieved with a rather verbose way using `Flex` or rather complicated using `Grid`

This PR proposes a new core component - Stack. Stack is a layout component that makes it easy to stack elements together and apply a space between them. It uses a modified version of the CSS lobotomized owl selector to add spacing between its children.

**Before** 

```jsx
<Flex flexDirection="row">
  <Box marginRight="spacingXs">
    <Button variant="primary" icon={icon && <Icon as={icons[icon]} />}>
      Primary
    </Button>
  </Box>
  <Box marginRight="spacingXs">
    <Button
      variant="secondary"
      icon={icon && <Icon as={icons[icon]} />}
    >
      Secondary
    </Button>
  </Box>
</Flex>
```

**After**
```jsx
<Stack flexDirection="row" spacing="spacingXs">
  <Button variant="primary" icon={icon && <Icon as={icons[icon]} />}>
    Primary
  </Button>

  <Button variant="secondary" icon={icon && <Icon as={icons[icon]} />}>
    Secondary
  </Button>
</Stack>
```

---

**Before**

```jsx
<Grid rowGap="spacingXs" rows="repeat(4, auto)" justifyContent="start">
  <Button variant="primary">Primary</Button>
  <Button variant="secondary">Secondary</Button>
  <Button variant="negative">Negative</Button>
  <Button variant="positive">Positive</Button>
  <Button variant="transparent">Transparent</Button>
</Grid>
```

**After**

```jsx
<Stack spacing="spacingXs">
  <Button variant="primary">Primary</Button>
  <Button variant="secondary">Secondary</Button>
  <Button variant="negative">Negative</Button>
  <Button variant="positive">Positive</Button>
  <Button variant="transparent">Transparent</Button>
</Stack>
```


massao (approved):
❤️

packages/core/src/Stack/Stack.styles.ts
```diff
@@ -0,0 +1,28 @@
...
+    return css({
+      '> :not(style) ~ :not(style)': {
+        marginTop: tokens[spacing],
+        marginInline: 0,
+        marginBottom: 0,
+      },
+    });
+  }
+  return css({
+    '> :not(style) ~ :not(style)': {
```
gui-santos, line 22: What is this selecting?

not style tags?

packages/core/src/Stack/Stack.tsx
```diff
@@ -0,0 +1,74 @@
...
+      FlexInternalProps,
+      'isInline' | 'alignItems' | 'justifyContent' | 'flexWrap'
+    > {
+  /**
+   * Defines how flexbox items are ordered within a flexbox container. */
+  flexDirection?: 'row' | 'column';
+  /**
+   * The space between each stack item
+   */
+  spacing?: Spacing;
```
gui-santos, line 23: is it a `spacing` or a `gap`?
Idk either, I'm just thinking that this works similarly with rowGap, columnGap, and gap in `display: grid`

Naming is the hardest part of coding

gui-santos (approved):
Pretty cool, huh!

🍬

packages/core/src/Stack/Stack.tsx
suevalov: We have `spacing` in button group already, so I decided to go this way.

denkristoffer:
This seems like using `Flex` with the `gap` property to me?

packages/core/src/Stack/Stack.tsx
gui-santos: Oh I think there it's "spacing" hahahahaha

I mean, in my head, gap is like blank space between elements (like the stack here and grid)
and spacing is more what padding does in general

But it was just a comment, I approved the PR
naming is a matter of convention 👍

packages/core/src/Stack/Stack.styles.ts
gui-santos: got it!

if this is a lobotomized owl `* + *`
how should we call this `* ~ *`? it looks like a sad pokemon

suevalov:
@denkristoffer, hmm, that's true, but our `Flex` gap attribute doesn't accept spacing values (is it a bug?), and also Stack component aligns items in center by default and applies spacingM by default.

packages/core/src/Stack/Stack.styles.ts
suevalov: "defeated lobotomized pokemon"

denkristoffer:
Having `Stack` would replace `Flex` for me, or the other way around. Do you see value in having both?

> is it a bug?

Yeah good point, it probably should.

Lelith (approved):
this is so simple and genius

suevalov:
@denkristoffer yeah, I see value in having both:

1. We don't want to have defaults in Flex for backward compatibility reasons.
2. Flex for me, is a bit more low-level component, that you would eventually use if you're not satisfied with Stack flexibility. 

P.S. I can fix `gap` in Flex to accept token spacing.

packages/core/src/Stack/Stack.styles.ts
Lelith: i think its a Wobbuffet.

denkristoffer:
Okay well if most people want both then I don't want to stand in the way 🙂

packages/core/src/Stack/Stack.styles.ts
gui-santos: you are right!!!!

Wobbuffet

PR #1299, opened 2021-09-09
fix: proper compound modal

suevalov:
# Purpose of PR

Previous compound implementation produced wrong types and typescript was complaining on the consumer side.


massao (approved):

massao:
Codewise it looks good, but seems that it's still complaining about the Compound components on the build 🤔

PR #1304, opened 2021-09-10
docs: add Formik examples in storybook 

massao:
# Purpose of PR

Integrates our Form Elements with Formik and creates tests for them
image

I will add documentation in the Website in another PR

### DIFF
--- a/packages/components/forms/stories/Formik.stories.tsx
+++ b/packages/components/forms/stories/Formik.stories.tsx
@@ -39,6 +39,9 @@ export const Basic = () => {
       initialValues={{
         name: '',
         description: '',
+        cities: 'Mumbai',
+        favoritFruit: 'apples',
+        fruits: ['apples'],
         checkbox: false,
       }}
       onSubmit={onSubmit}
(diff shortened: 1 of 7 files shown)


packages/components/forms/stories/Formik.stories.tsx
```diff
@@ -0,0 +1,337 @@
...
+          <Button variant="primary" type="submit">
+            Submit
+          </Button>
+        </Form>
+      )}
+    </Formik>
+  );
+};
+
+export const WithUseFormik = () => {
```
massao, line 228: I added this example so we can show that it also works with this approach, but they recommend using the `<Formik>` and `<Field>` components

denkristoffer:
My take would to combine these stories into one, do we need to show (and update) this many examples?

suevalov:
I'd probably double up on @denkristoffer suggestions to combine all stories into one with a complex form that represents all our form components.

Otherwise, the PR looks great.

denkristoffer (approved):
🙏

PR #1320, opened 2021-09-16
feat: V4/entity list

bgutsol:
# Purpose of PR

V4 Entity List component

## PR Checklist

- [x] Tests are passing


packages/components/entity-list/src/EntityList.tsx
```diff
@@ -0,0 +1,25 @@
...
+}
+
+function _EntityList(props: EntityListProps, ref: React.Ref<HTMLUListElement>) {
+  const styles = getEntityListStyles();
+
+  return (
+    <ul
+      data-test-id={props.testId || 'cf-ui-entity-list'}
+      ref={ref}
+      className={cx(styles.root, props.className)}
```
suevalov, line 18:
```suggestion
      className={cx(styles.root, props.className)}
      style={props.style}
```

packages/components/entity-list/src/EntityListItem/EntityListItem.styles.ts
```diff
@@ -0,0 +1,73 @@
...
+    flexGrow: 1,
+    minWidth: 0,
+  }),
+  media: css({
+    display: 'flex',
+    flexShrink: 0,
+    alignItems: 'center',
+    justifyContent: 'center',
+    backgroundColor: tokens.gray200,
+    width: '2.875rem',
```
domarku, line 44: Let's change the width and the height of the media here to `spacing-2xl`

That will result in the EntityList item with a computed height of `64px`, which is equal to our `spacing-3xl`.

domarku (approved):

packages/components/entity-list/src/EntityListItem/EntityListItem.styles.ts
bgutsol: Can I ask what the reason for such changes? I don't think it's so necessary to couple everything with our spacing tokens, especially if it's about sizes not the spacing itself. So I wouldn't change the old design just for the sake of reusing spacing tokens. WDYT @domarku?

packages/components/entity-list/src/EntityListItem/EntityListItem.styles.ts
```diff
@@ -0,0 +1,73 @@
...
+    display: 'block',
+    width: '100%',
+    height: '100%',
+    objectFit: 'cover',
+  }),
+  contentType: css({
+    marginLeft: tokens.spacingXs,
+  }),
+  description: css({
+    marginTop: '0.25rem',
```
burakukula, line 59: I guess you could use `spacing-2xs` over here :)

packages/components/entity-list/EntityList.mdx
```diff
@@ -16,25 +19,25 @@ The EntityList component is used to wrap EntityListItem components. The EntityLi
 
 ## How to use EntityList
 
-This component should be used in combination with the EntryListItem components, and should be used to entries or assets when in a list context, for example, a multiple entry reference field. It differs from existing EntryCard/AssetCard components as it's intended use is for lists of entities, not individual ones.
+This component should be used in combination with the EntityList.Item component, and should be used to entries or assets when in a list context, for example, a multiple entry reference field. It differs from existing EntryCard/AssetCard components as it's intended use is for lists of entities, not individual ones.
```
burakukula, line 22:
```suggestion
This component should be used in combination with the EntityList.Item component. Its main purpose is to be used to entries or assets when in a list context, for example, a multiple entry reference field. It differs from existing EntryCard/AssetCard components as its intended use is for lists of entities, not individual ones.
```

burakukula: WDYT? 🙂

burakukula (approved):
LGTM 👏

packages/components/entity-list/EntityList.mdx
bgutsol: Awesome, thanks ☺️

suevalov (approved):

packages/components/entity-list/src/EntityListItem/EntityListItem.styles.ts
domarku: As discussed in standup today, the reason to keep sizes coupled with our spacing tokens is to keep everything on a modular scale and have all of our elements better sit on a visual grid. I don't mind if the media height and width is a hardcoded value, but would still go for `3rem` (spacing-2xl).

burakukula (approved):

PR #1324, opened 2021-09-20
fix: remove migration note regarding colors migration

burakukula:
This might be a moment to remove the note regarding the colors update from the website? It's been there for 2 months. WDYT?


denkristoffer (approved):
👍

PR #1325, opened 2021-09-20
feat: add useKeyboard hook to f36-utils

mshaaban0:
# Purpose of PR

closes https://github.com/contentful/forma-36/discussions/1008

Example usage

```js
useKeyboard({
   event: 'keydown', // Optional, default is `keydown`
   ref: yourReactRef, // Optional, by default event is attached to document
   keys: {
     ArrowUp: (e) => handleArrowUp(e),
     Tab: (e) => handleTab(e)
   }
})
```

## Todo
- [x] Add tests

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/components/utils/src/useKeyboard/useKeyboard.spec.tsx
+++ b/packages/components/utils/src/useKeyboard/useKeyboard.spec.tsx
@@ -1,0 +1,66 @@
+import React, { useRef } from 'react';
+import { render, fireEvent, getByText } from '@testing-library/react';
+
+import { useKeyboard } from './useKeyboard';
+
+const handleEnter = jest.fn();
+const handleArrowUp = jest.fn();
+
+const Component = () => {
+  useKeyboard({
+    keys: {
+      ArrowUp: handleArrowUp,
+      Enter: handleEnter,
+    },
+  });
+
+  return <button type="button">Button</button>;
+};
+
+const ComponentWithRef = () => {
+  const divRef = useRef(null);
+
+  useKeyboard({
+    ref: divRef,
+    keys: {
+      ArrowUp: handleArrowUp,
+      Enter: handleEnter,
+    },
+  });
+
+  return (
+    <button type="button" ref={divRef}>
+      Button
+    </button>
+  );
+};
+
+describe('useKeyboard', () => {
+  it('should attach handler to document by default', () => {
+    const events = {};
+    document.addEventListener = jest.fn((event, callback) => {
+      events[event] = callback;
+    });
+
+    render(<Component />);
+
+    expect(document.addEventListener).toHaveBeenCalledWith(
+      'keydown',
+      expect.any(Function),
+    );
+  });
+
+  it('should trigger handlers correctly', () => {
+    const { container } = render(<ComponentWithRef />);
+
+    fireEvent.keyDown(getByText(container as HTMLElement, 'Button'), {
+      key: 'Enter',
+    });
+    expect(handleEnter).toHaveBeenCalled();
+
+    fireEvent.keyDown(getByText(container as HTMLElement, 'Button'), {
+      key: 'ArrowUp',
+    });
+    expect(handleArrowUp).toHaveBeenCalled();
+  });
+});
--- a/packages/components/utils/src/useKeyboard/useKeyboard.spec.tsx
+++ b/packages/components/utils/src/useKeyboard/useKeyboard.spec.tsx
@@ -1,5 +1,5 @@
 import React, { useRef } from 'react';
-import { render, fireEvent, getByText } from '@testing-library/react';
+import { render, fireEvent, screen } from '@testing-library/react';
 
 import { useKeyboard } from './useKeyboard';
 
@@ -51,14 +51,14 @@ describe('useKeyboard', () => {
   });
 
   it('should trigger handlers correctly', () => {
-    const { container } = render(<ComponentWithRef />);
+    render(<ComponentWithRef />);
 
-    fireEvent.keyDown(getByText(container as HTMLElement, 'Button'), {
+    fireEvent.keyDown(screen.getByRole('button'), {
       key: 'Enter',
     });
     expect(handleEnter).toHaveBeenCalled();
 
-    fireEvent.keyDown(getByText(container as HTMLElement, 'Button'), {
+    fireEvent.keyDown(screen.getByRole('button'), {
       key: 'ArrowUp',
     });
     expect(handleArrowUp).toHaveBeenCalled();
--- a/packages/components/utils/src/useKeyboard/useKeyboard.ts
+++ b/packages/components/utils/src/useKeyboard/useKeyboard.ts
@@ -6,7 +6,7 @@ export interface UseKeyboardProps {
    * value is a callback function to be called when key matches
    */
   keys: {
-    [key: KeyboardEvent['key']]: (e: KeyboardEvent) => void;
+    [key: string]: (e: KeyboardEvent) => void;
   };
   /**
    * @description defines the attached event type
--- a/packages/components/utils/src/useKeyboard/useKeyboard.ts
+++ b/packages/components/utils/src/useKeyboard/useKeyboard.ts
@@ -48,7 +48,7 @@ export const useKeyboard = (props: UseKeyboardProps) => {
   );
 
   useEffect(() => {
-    if (ref) {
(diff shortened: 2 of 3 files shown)


packages/components/utils/src/useKeyboard/useKeyboard.ts
```diff
@@ -0,0 +1,55 @@
...
+export interface UseKeyboardProps {
+  /**
+   * @description defines which element is the event attached to
+   * @default document
+   */
+  ref?: Document | Window | HTMLElement;
+  /**
+   * @description defines which key to look for i.e. `ArrowUp`, `Escape`, `Shift`
+   */
+  key: KeyboardEvent['key'];
```
mshaaban0, line 12: This resolves to a `string`, I couldn't find the proper types with the full list of keys, any ideas?

packages/components/utils/src/useKeyboard/useKeyboard.ts
```diff
@@ -0,0 +1,55 @@
+import { useState, useEffect, useCallback } from 'react';
+
+export interface UseKeyboardProps {
+  /**
+   * @description defines which element is the event attached to
+   * @default document
+   */
+  ref?: Document | Window | HTMLElement;
```
bgutsol, line 8: I wouldn't call it `ref`, since it can be confused with `ReactRef`. I'd rather call it just an element.

bgutsol: What if I want to pass `ref.current` which can be `null`? 
```tsx
const myRef = React.createRef(null)
```
it is the most common pattern.
In this case I won't be able to use this hook, cause I can't pass `null` as an argument. And I can't wrap hooks into conditionals.

packages/components/utils/src/useKeyboard/useKeyboard.ts
```diff
@@ -0,0 +1,55 @@
...
+  /**
+   * @description defines which key to look for i.e. `ArrowUp`, `Escape`, `Shift`
+   */
+  key: KeyboardEvent['key'];
+  /**
+   * @description defines the attached event type
+   * @default 'keyup'
```
bgutsol, line 15: Why `keyup` is a default? 
IMO `keydown` would be/should be more common.

bgutsol (commented):
Good 👍
Some small comments and a general thought to consider:
What if I want/need to pass a handler as a property on an element?
Wouldn't it be more flexible if we will not ask for ref (element) to attach handlers, but rather return an object with all the handlers. So the user can just spread it as a props or pick some specific ones if they need it. WDYT?
For example:
```tsx
const keyboardHandlers = useKeyboard({
ArrowUp: () => console.log('handle ArrowUp')
Tab: () => console.log('handle Tab')

return <List {...keyboardHandlers}></List>
})

```

packages/components/utils/src/useKeyboard/useKeyboard.ts
mshaaban0: I recall reading something about a11y and that `keyup` allow users to cancel events with `esc` but I can't find any resources. I can validate that and change if my assumption is incorrect 👍🏼

packages/components/utils/src/useKeyboard/useKeyboard.ts
denkristoffer: I don't think such a type exists so this is probably fine.

packages/components/utils/src/useKeyboard/useKeyboard.ts
mshaaban0: I went with passing the full `ref` object as the hook needs to monitor current when it changes from `null` to defined. 

That way the consumer wouldn't need to fiddle much with it, and by default the event is attached to `document`

packages/components/utils/src/useKeyboard/useKeyboard.spec.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  it('should trigger handlers correctly', () => {
+    const { container } = render(<ComponentWithRef />);
+
+    fireEvent.keyDown(getByText(container as HTMLElement, 'Button'), {
+      key: 'Enter',
+    });
+    expect(handleEnter).toHaveBeenCalled();
+
+    fireEvent.keyDown(getByText(container as HTMLElement, 'Button'), {
```
denkristoffer, line 61: you can use `screen` from `@testing-library/react` here instead of having to cast types here.

mshaaban0: Nice, thanks

packages/components/utils/src/useKeyboard/useKeyboard.ts
```diff
@@ -0,0 +1,60 @@
...
+  useEffect(() => {
+    if (ref) {
+      element = ref.current;
+    }
+
+    element.addEventListener(event, handleKeyEvent);
+    return () => {
+      element.removeEventListener(event, handleKeyEvent);
+    };
+  }, [element, ref, event, handleKeyEvent]);
```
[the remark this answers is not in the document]
mshaaban0: Hey you're right, element is not necessary as it doesn't change, left over from previous implementation.
`ref` is there because as it should be watched for changes and re-trigger the events addition/removal.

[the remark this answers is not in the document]
bgutsol: But `ref` is a mutable object, it will not change. So when ref.current changes its value from `null` to some `element` it won't trigger this effect.

packages/components/utils/src/useKeyboard/useKeyboard.ts
```diff
@@ -0,0 +1,60 @@
...
+      const isKey = Object.prototype.hasOwnProperty.call(keys, e.key);
+      if (isKey) {
+        keys[e.key](e);
+      }
+    },
+    [keys],
+  );
+
+  useEffect(() => {
+    if (ref) {
```
bgutsol, line 51: Since ref is a mutable object it will always exist, so always `true`. You might rather check `ref.current`

packages/components/utils/src/useKeyboard/useKeyboard.ts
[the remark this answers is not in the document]
mshaaban0: The goal here is not to monitor `ref.current`, it's to monitor `ref` itself as it's `undefined` by default if the consumer doesn't pass it we fallback to `document`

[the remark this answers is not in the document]
mshaaban0: I removed it and the events gets attached properly. also I don't think users will dynamically change `ref` so no need to watch it as well. Thanks for the feedback 👍🏼

denkristoffer (approved):
Looks good to me 👍 How about support for key combinations?

packages/components/utils/src/useKeyboard/useKeyboard.ts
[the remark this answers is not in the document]
bgutsol: I don't agree that users will not dynamically change ref, here is the example: https://github.com/contentful/forma-36/pull/1343 
In this case I can't attach events on the button with ref. It attaches to document, since the ref.current is `null` on the first render.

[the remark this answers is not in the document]
bgutsol: and what if I don't want to attach the events to a document when my `ref.current` is `null`? Also something to consider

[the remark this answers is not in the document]
mshaaban0: that shouldn't be the case. If you provide a ref it must be attached to the current. I'll have a look, thanks

mshaaban0:
@denkristoffer Good point :) I completely forgot about it, how to you foresee the api?
Would something like 👇🏼 be alright? I can basically parse the string and look for `+` with some magic

```js
useKeyboard({
   event: 'keydown', // Optional, default is `keydown`
   ref: yourReactRef, // Optional, by default event is attached to document
   keys: {
     'metaKey + ArrowUp': (e) => handleArrowUp(e),
     Tab: (e) => handleTab(e)
   }
})
```

denkristoffer:
I'm just thinking out loud here but it could also be an array of keys. We don't have to land it in this PR as long as we think about making it possible without breaking the API.

anwaar931 (approved):

gui-santos (approved):
🏆

PR #1336, opened 2021-09-21
feat: migrate autocomplete to v4

gui-santos:
# Purpose of PR

- Migrates Autocomplete to v4
- Deletes files in f36-react-components
- uses [downshift](https://github.com/downshift-js/downshift/tree/master/src/hooks/useCombobox) to provide keyboard navigation
- changes Autocomplete API, but that was an "alpha" component
- updates documentation in the website and storybook

I will implement tests in another PR

## PR Checklist

- [x] Tests are passing


packages/components/autocomplete/src/Autocomplete.tsx
```diff
@@ -0,0 +1,246 @@
...
+} from '@contentful/f36-forms';
+import { IconButton } from '@contentful/f36-button';
+import { CloseIcon, ChevronDownIcon } from '@contentful/f36-icons';
+import { Popover } from '@contentful/f36-popover';
+
+import { getAutocompleteStyles } from './Autocomplete.styles';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export interface AutocompleteProps<ItemType = any>
+  extends CommonProps,
```
Lelith, line 19: do we maybe also need props for the size of the dropdown list ?

Lelith (approved):
thank you for putting all the work in this!
i have some minor comments and some improvements that can go into the next iteration
- adding more tests for e.g. calling the callbacks with the expected elements
- the pop-over list is way smaller than the input field, would be good to either make it by default the same width as the input element or add a prop which is able to adjust the width
- enable optgroup as part of the items object, so maybe if i hand over a nested object that it takes the first level as the optgroup label and the childs as the optgroup entries?
- open the list of elements when user focuses, not only when they start typing
- as a keyboard  only user i currently have no possibility of observing the whole list without starting to type. which might be frustrating.

packages/components/autocomplete/Autocomplete.mdx
```diff
@@ -0,0 +1,241 @@
...
+- [How to use Autocomplete](#how-to-use-autocomplete)
+- [Code examples](#code-examples)
+- [Content recommendations](#content-recommendations)
+- [Accessibility](#accessibility)
+
+## How to use Autocomplete
+
+The Autocomplete requires 3 props to work:
+
+- `items`: It’s an array containing the items that will be shown as selectable options when the user types something in the `TextInput`.
```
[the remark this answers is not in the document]
gui-santos: It can be an array of anything actually

I just added an example with objects because I thought it would be a good example of how to use complex data with the Autocomplete

packages/components/autocomplete/Autocomplete.mdx
```diff
@@ -0,0 +1,241 @@
...
+  };
+
+  return (
+    <Stack flexDirection="column" alignItems="start">
+      <Autocomplete
+        items={fruits}
+        onFilter={handleFilter}
+        onSelectItem={handleSelectItem}
+        // This prop is the function that will tell the component how to extract a string that will be used as inputValue
+        itemToString={(item) => item.name}
```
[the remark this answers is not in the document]
gui-santos: Oh really? My idea was to extend the previous examples, I was trying to teach by building on top of previous knowledge

Could you give an example of what can I do here?

packages/components/autocomplete/Autocomplete.mdx
```diff
@@ -0,0 +1,241 @@
...
+```
+
+## Content recommendations
+
+- Autocomplete label should be short, contain 1 to 3 words
+- Label should be written in a sentence case (the first word capitalized, the rest lowercase)
+
+## Accessibility
+
+- dismisses the dropdown when selecting with the enter key
```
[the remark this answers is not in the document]
gui-santos: Could you elaborate? Maybe write these things and I will add them

I'm not really good at writing or at a11y notes

packages/components/autocomplete/src/Autocomplete.styles.ts
```diff
@@ -0,0 +1,47 @@
...
+    top: '4px',
+    right: '4px',
+    zIndex: 1,
+  }),
+  list: css({
+    overflowY: 'auto',
+    listStyle: 'none',
+    padding: `${tokens.spacingXs} 0`,
+    margin: 0,
+    maxHeight: '360px',
```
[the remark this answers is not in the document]
gui-santos: I'm not sure if we do

I came with this value magically, but I considered an N number of items and used it as maxHeight
I forgot how many items.... maybe 5?

packages/components/autocomplete/src/Autocomplete.tsx
gui-santos: what do you mean by size of the dropdown  list?

gui-santos:
> thank you for putting all the work in this!
> i have some minor comments and here are some additional improvements that can go into the next iteration:

@Lelith thank you for the review

> * adding more tests for e.g. calling the callbacks with the expected elements

As I said in the PR description, I'm working on the tests and I will add them in following PR

> * the pop-over list is way smaller than the input field, would be good to either make it by default the same width as the input element or add a prop which is able to adjust the width

Since we didn't have any designs for this component, I kept the design as close as it was in v3
Maybe it's time for @domarku and @fabe to take a look on it and improve it

> * enable optgroup as part of the items object, so maybe if i hand over a nested object that it takes the first level as the optgroup label and the childs as the optgroup entries?

Same as the above, it's not a feature in v3. So I figured it's still not a feature we want to add to the component
Maybe in the future someone get's this demand and then we implement it

> * open the list of elements when user focuses, not only when they start typing

I think it's possible to make the list to open when the user focuses the input, should I do that now?

> * as a keyboard  only user i currently have no possibility of observing the whole list without starting to type. which might be frustrating.

Isn't this fixed by the other suggestion, the one above it?

suevalov (changes requested):
I was playing around with the component on the documentation website and stuck in a state when "No Matches" is always showing.

suevalov:
Few questions:


1. I wonder if we should cover the case when options are async (based on the input value). It's relevant for big amount of options, we don't want to prefetch them all.

2. Didn't we decide to rename component to something else? @domarku

gui-santos:
> I was playing around with the component on the documentation website and stuck in a state when "No Matches" is always showing.

@suevalov yep, for some reason, our implementation of `react-live` does is not passing the array to the component

I think it's a bug in the website, I already debugged a bit and I didn't find the solution yet
but please check my markdown files
maybe you find something in them that I'm missing

packages/components/autocomplete/Autocomplete.mdx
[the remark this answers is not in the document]
thebesson: maybe 'fruitName"? it's "data specific", doesn't look like standard component API and may convey that the user can use any data structure with this component and that they control what is shown as a label.

packages/components/autocomplete/src/Autocomplete.tsx
thebesson: maxHeight?

packages/components/autocomplete/src/Autocomplete.tsx
```diff
@@ -0,0 +1,246 @@
...
+    testId = 'cf-autocomplete',
+  } = props;
+
+  const formProps = useFormControl({
+    id,
+    isInvalid,
+    isDisabled,
+    isRequired,
+    isReadOnly,
+  });
```
bgutsol, line 108: You probably don't need `useFormControl` here. It will be too lengthy to explain it here, so I've created a PR
https://github.com/contentful/forma-36/pull/1342

bgutsol (changes requested):
Great work @gui-santos 🏅 👏

### I have some comments/suggestion:
I think we should also expose some of the popover props, for example:
`isFullWidth` - will match the inputText width.
`placement` - it can be handy in some cases
`isAutoalignmentEnabled`

Also how can I change the maxHeight for the popover list? Now it's `360px` and you are not able to change it. We can provide some extra prop, like `listMaxHeight` or expose the `listClassName`. Second approach will also make it possible to provide custom `width` for the list.

gui-santos:
@bgutsol 

> #### I have some comments/suggestions:
> 1. I think we should also expose some of the popover props, for example:
>    `isFullWidth` - will match the inputText width.
>    `placement` - it can be handy in some cases
>    `isAutoalignmentEnabled`

I will do that! It makes sense

> 2. How can I change the maxHeight for the popover list? Now it's `360px` and you are not able to change it. We can provide some extra prop, like `listMaxHeight` or expose the `listClassName`. Second approach will also make it possible to provide custom `width` for the list.

Do we want to allow the user to set their own maxHeight? 
I'm against exposing the listClassName because IMO people should avoid overwriting the styles, if they reaaaally need it, the div wrapping everything has className and they could style anything by using CSS selectors 👌 

I think I will implement the `width` prop since it’s a feature that seems to be expected by more than one person hahahah
but since we don't have any designs, I will limit it to 'auto' or 'full' (same width as the input)

packages/components/autocomplete/src/Autocomplete.tsx
gui-santos: made a comment in the PR

I guess we need some documentation around this hook, I still don't get it

packages/components/autocomplete/src/Autocomplete.tsx
```diff
@@ -0,0 +1,246 @@
...
+  /**
+   * This is the function that will be called when the user selects one of the "options" in the list.
+   * It receives the selected item as an argument and it needs to return a string that will be set as the value of `TextInput`.
+   */
+  onSelectItem: (item: ItemType) => void;
+  /**
+   * This is the function that will be called for each "item" passed in the `items` prop.
+   * It receives the "item" as an argument and returns a ReactNode.
+   */
+  renderItem?: (item: ItemType) => React.ReactElement;
```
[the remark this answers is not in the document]
gui-santos: Oh! that's why I was not being able to pass `(item) => item.name` in my example

Totally missed changing this type
good catch, thanks

packages/components/autocomplete/src/Autocomplete.tsx
Lelith: `maxHeight` and `width`

burakukula (approved):
@gui-santos amazing work! 👏 I love how straightforward code examples are, with comments and hints. I think it's very useful to understand how this component works and how to use it.
+1 for comments from other folks and small suggestions for the README.

packages/components/autocomplete/Autocomplete.mdx
```diff
@@ -0,0 +1,239 @@
...
+        <FormControl.Label>Favorite fruit:</FormControl.Label>
+
+        <Autocomplete
+          items={fruits}
+          onFilter={handleFilter}
+          onSelectItem={handleSelectItem}
+        />
+
+        {isInvalid && (
+          <FormControl.ValidationMessage>Error</FormControl.ValidationMessage>
```
mshaaban0, line 214: Unrelated to this PR, but this component renders paragraph with `aria-live="polite"` as it requires the user immediate attention, i think it should be `aria-live="assertive"`

mshaaban0 (approved):
Nice one! tested with screen reader and works like a charm 💯

packages/components/autocomplete/Autocomplete.mdx
suevalov: @mshaaban0 do you mind opening a PR for validation message component?

packages/components/autocomplete/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "@contentful/f36-autocomplete",
+  "version": "4.0.0",
+  "description": "Forma 36: Autocomplete component",
+  "scripts": {
+    "build": "parcel build index.ts"
+  },
+  "dependencies": {
+    "@babel/runtime": "^7.6.2",
+    "@contentful/f36-button": "^4.0.0",
```
bgutsol, line 10: forgot to add a popover as a dependency

packages/components/autocomplete/src/Autocomplete.tsx
gui-santos: I will add those, but I will give maxHeight a default value because I don't think we should allow huge lists

what I mean is that a list with 20 items and without maxHeight would be 1200px tall 👀

packages/components/autocomplete/Autocomplete.mdx
burakukula: created quick fix https://github.com/contentful/forma-36/pull/1350

gui-santos:
> Nice one! tested with screen reader and works like a charm 💯

@mshaaban0 that’s more on downshift than on me tbh hahahah
amazing library!

packages/components/autocomplete/Autocomplete.mdx
gui-santos: Ohhhh good catch! I will do it in another PR

packages/components/autocomplete/package.json
gui-santos: nice catch! I will fix it

packages/components/autocomplete/src/Autocomplete.styles.ts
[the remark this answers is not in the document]
gui-santos: this was moved to a prop and the default value is the height of 5 single line items, cool?

packages/components/autocomplete/Autocomplete.mdx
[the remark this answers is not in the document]
gui-santos: I ended up changing the data structure to this:

```
const fruits = [
    { id: 1, name: 'Apple', emoji: 🍎 },
    { id: 2, name: 'Ananas', emoji: 🍍 },
    { id: 3, name: 'Avocado', emoji: 🥑 },
    { id: 4, name: 'Banana', emoji: 🍌 },
    { id: 5, name: 'Coconut', emoji: 🥥 },
  ];
```

and then I put this in line 113 to 115

```
// This prop is the function that will tell the component how to render each item in the options list
// In this example the first item will render "Apple 🍎"
renderItem={(item) => `${item.name} ${item.emoji}`}
```

wdyt?

bgutsol (approved):
👏

suevalov (approved):

packages/components/autocomplete/Autocomplete.mdx
gui-santos: I guess I replied to moe before refreshing the page hihihi

PR #1338, opened 2021-09-21
feat: product icons package for v4

burakukula:


burakukula:
now when I think about it, maybe we should actually add this component to experience components. Apps are using more custom icons (I think they use some additional icons in the navigation 🤔), that could be part of this package WDYT? Maybe it's a better idea since the product icons are really Contentful specific?

suevalov:
@burakukula yeah, I think it's a good idea to move it to experience-packages

burakukula:
Closing here, to open in experience components.

PR #1339, opened 2021-09-21
feat(tooltip-v4): update tooltip-v4 codemod to add console error for wrong prop type

thebesson:
# Purpose of PR

To make this change visible, I added a "console.error". If consumer won't update the property, it will lead to type errors.

### DIFF
--- a/packages/forma-36-codemod/transforms/v4-tooltip.js
+++ b/packages/forma-36-codemod/transforms/v4-tooltip.js
@@ -1,34 +1,21 @@
 const { modifyPropsCodemod } = require('./common/modify-props-codemod');
-const { pipe } = require('./common/pipe');
-const { getComponentLocalName } = require('../utils');
-const { getFormaImport } = require('../utils/config');
+const { hasProperty, getProperty } = require('../utils');
 
-module.exports = pipe([
-  (file, api) => {
-    const j = api.jscodeshift;
-
-    let source = file.source;
-
-    const componentName = getComponentLocalName(j, source, {
-      componentName: 'Tooltip',
-      importName: getFormaImport(),
-    });
-
-    if (!componentName) {
-      return source;
+module.exports = modifyPropsCodemod({
+  componentName: 'Tooltip',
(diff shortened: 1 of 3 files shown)


suevalov (approved):

packages/forma-36-codemod/transforms/v4-tooltip.js
```diff
@@ -1,34 +1,21 @@
...
-    });
-
-    if (!componentName) {
-      return source;
+module.exports = modifyPropsCodemod({
+  componentName: 'Tooltip',
+  beforeRename: (attributes) => {
+    if (
+      hasProperty(attributes, { propertyName: 'content' }) &&
+      typeof getProperty(attributes, { propertyName: 'content' } !== 'string')
```
gui-santos, line 9: I know about cases in `user_interface` that we are passing `undefined` to the content prop

why? because in v3 that's how we were conditionally showing/hiding a tooltip
(Really bad pattern I know, but it is what it is)

So my question is, can the tooltip be shown/hidden conditionally with in a proper way in v4?

If not, then I think we need to update the tooltip component

gui-santos (approved):
I think it looks great

but it reminded me of something in the Tooltip API that we wanted to fix in v4
and I don't know if we did so I posted a question in the PR :)

packages/forma-36-codemod/transforms/v4-tooltip.js
suevalov: Good call out. We've added a new property `isDisabled` to control that

suevalov: Looks like we need to add it when content is undefined or ""

thebesson: in the code for Tooltip component we check for "isDisabled" and for the content. there are cases where tooltip would be empty, till some event in front end triggers state changes and populates this prop value with a string. for example, in field editors, after fetching data for scheduled actions.

massao (approved):

PR #1342, opened 2021-09-22
v4/autocomplete PR suggestions

bgutsol:
refactoring suggestions for Autocomplete component

### DIFF
--- a/packages/components/autocomplete/src/Autocomplete.tsx
+++ b/packages/components/autocomplete/src/Autocomplete.tsx
@@ -3,11 +3,7 @@ import { cx } from 'emotion';
 import { useCombobox } from 'downshift';
 
 import { CommonProps, mergeRefs } from '@contentful/f36-core';
-import {
-  TextInput,
-  TextInputProps,
-  useFormControl,
-} from '@contentful/f36-forms';
+import { TextInput, TextInputProps } from '@contentful/f36-forms';
 import { IconButton } from '@contentful/f36-button';
 import { CloseIcon, ChevronDownIcon } from '@contentful/f36-icons';
 import { Popover } from '@contentful/f36-popover';
@@ -99,14 +95,6 @@ function _Autocomplete<ItemType>(
     testId = 'cf-autocomplete',
   } = props;
 
-  const formProps = useFormControl({
-    id,
-    isInvalid,
-    isDisabled,
-    isRequired,
-    isReadOnly,
-  });
-
   const styles = getAutocompleteStyles();
 
   const [filteredItems, setFilteredItems] = useState(items);
@@ -152,7 +140,11 @@ function _Autocomplete<ItemType>(
   });
 
   const comboboxProps = getComboboxProps();
-  const inputProps = getInputProps();
+  const {
+    'aria-labelledby': _labelledby,
+    id: _inputId,
+    ...inputProps
+  } = getInputProps();
   const toggleProps = getToggleButtonProps();
   const menuProps = getMenuProps();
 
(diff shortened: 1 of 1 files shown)


packages/components/autocomplete/src/Autocomplete.tsx
```diff
@@ -152,7 +140,11 @@ function _Autocomplete<ItemType>(
   });
 
   const comboboxProps = getComboboxProps();
-  const inputProps = getInputProps();
+  const {
+    'aria-labelledby': _labelledby,
+    id: _inputId,
```
gui-santos, line 145: I thought downshift was needing this
did you test it and it still works? keyboard and everything?

bgutsol: Yes, everything is working as before.
But TBH before you were also replacing `id` with `{...formProps}`, so nothing really changed.
Regarding `aria-labelledby` I doubt it has some other purpose that a11y.

gui-santos: but if I was not passing an id, downshift was passing its own id

I wonder how it's working now with undefined id

bgutsol: If you wrap with FormControl it gets id from the context.

gui-santos: but we are not wrapping it every time

Autocomplete must work outside of FormControl and without an id prop

Did you test without FormControl?

bgutsol: Yes, it's working, but it's not my point.
The point is, that in your implementation you have 3 cases:
1. When user not wrapping `Autocomplete` in `FormControl` and not passing the `id` prop - the id will be from downshift.
2. When user wrapping `Autocomplete` in `FormControl` and not passing the `id` prop - the id will be from FormControl context.
3. When user passing the `id` prop - the id will be from props. 

So in 2nd and 3rd cases you setting your custom `id` on `TextInput` not the one that downshift provides. If it breaks because of it it should already be broken in your branch. Do you see?

So my point is: we should either always set custom id or always the `id` that downshift provides.

gui-santos: > So my point is: we should either always set custom id or always the id that downshift provides.

Why one or the other?
I don't think we should ask the user to add id always

anyway, I think I got it how not to use the hook and I will make the changes to make it work
thanks!

bgutsol: > Why one or the other?

Because if downshift `id` do nothing then we can drop it, and always use our custom id, if it's provided. I didn't mean to make it mandatory to force user to provide it always.

gui-santos: got it now!

gui-santos (approved):

packages/components/autocomplete/src/Autocomplete.tsx
```diff
@@ -99,14 +95,6 @@ function _Autocomplete<ItemType>(
     testId = 'cf-autocomplete',
   } = props;
 
-  const formProps = useFormControl({
```
suevalov, line 102: Don't we want to integrate form control with autocomplete?

bgutsol: It has support out of the box, since we are using `TextInput` that's watching context value from `FormControl`.

suevalov: ah, that's nice!

suevalov (approved):

PR #1344, opened 2021-09-22
fix(Menu): Allow single menu item

suevalov:
# Purpose of PR

More relaxed types to allow one Menu.Item as a child of Menu.List


packages/components/menu/src/MenuList/MenuList.tsx
```diff
@@ -8,7 +8,7 @@ import { MenuListHeader } from './MenuListHeader';
 import { MenuListFooter } from './MenuListFooter';
 
 interface MenuListInternalProps extends CommonProps {
-  children?: React.ReactNodeArray;
+  children?: React.ReactNodeArray | React.ReactNode;
```
bgutsol, line 11:
```suggestion
  children?: React.ReactNode;
```
In such case we probably don't need `React.ReactNodeArray` at all

bgutsol (approved):

PR #1345, opened 2021-09-22
fix: remove select on disabled and readonly

suevalov:
# Purpose of PR

Completely removes `(e.target as HTMLInputElement).select();`. Fixes https://github.com/contentful/field-editors/pull/867

### DIFF
--- a/packages/components/forms/src/base-input/BaseInput.tsx
+++ b/packages/components/forms/src/base-input/BaseInput.tsx
@@ -57,14 +57,11 @@ function _BaseInput<E extends React.ElementType = typeof DEFAULT_TAG>(
   const handleFocus = useCallback(
     (e: FocusEvent<HTMLInputElement | HTMLTextAreaElement>) => {
       e.persist();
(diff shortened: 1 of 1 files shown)


packages/components/forms/src/base-input/BaseInput.tsx
```diff
@@ -57,14 +57,11 @@ function _BaseInput<E extends React.ElementType = typeof DEFAULT_TAG>(
   const handleFocus = useCallback(
     (e: FocusEvent<HTMLInputElement | HTMLTextAreaElement>) => {
       e.persist();
```
denkristoffer, line 59:
```suggestion
```

suevalov: we don't need it as well? then we don't need custom `handleFocus` at all too

denkristoffer: You're right, I would remove it completely.

bgutsol (approved):

PR #1355, opened 2021-09-24
test: add tests to v4 autocomplete

gui-santos:
# Purpose of PR

Besides adding tests I made the Popover.Content to always be in the DOM
That's necessary for a better a11y of the Autocomplete and it makes downshift stop throwing that error about the menuRef that you probably saw already

I made this the default behaviour, do you think it should be optional? Something that we control with a prop?

But I really don't know if I'm breaking something with my solution for that hahahaha
(I'm just using `display: 'none' when !isOpen`)
so please test it!

## PR Checklist

- [ ] Tests are passing


Lelith (commented):
Looks good to me!

Lelith (approved):

bgutsol (approved):
💎

PR #1371, opened 2021-10-01
fix: fix styling in checkbox and remove exports of base components

burakukula:
Aligning checkbox to label using flex and removing margins, it should solve the issue with positioning custom element to the checkbox on the client side. 

Even though we don't want to encourage folks to create their own custom labels, we already have cases that need's them. That's why I added a story for that. WDYT?


suevalov (commented):
Looks like we also have `BaseCheckbox.mdx` and `README.md` files (and a link from documentation site) that needs to be removed.

burakukula:
@suevalov yeah, make sense. I removed both for BaseCheckbox and for BaseInput then

suevalov (approved):

gui-santos (approved):

PR #1372, opened 2021-10-01
feat(codemod): add Radio codemod

massao:
# Purpose of PR

* Add Radio codemod
* Fix issue when some props were expressions and we had to pass as children

### DIFF
--- a/packages/forma-36-codemod/transforms/__testfixtures__/v4-radio.input.js
+++ b/packages/forma-36-codemod/transforms/__testfixtures__/v4-radio.input.js
@@ -1,0 +1,51 @@
+import { RadioButton, RadioButtonField } from '@contentful/forma-36-react-components';
+
+<RadioButton id="Radio" labelText="some label text" name="option1" />;
+
+const isDisabled = false;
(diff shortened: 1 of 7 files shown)


packages/forma-36-codemod/transforms/__testfixtures__/v4-radio.input.js
```diff
@@ -0,0 +1,51 @@
...
+<RadioButtonField
+  className="radio-button"
+  id="unit-option-percent"
+  checked={selectedUnit === 'percent'}
+  labelText="percent"
+  value="percent"
+  onChange={() => {}}
+  labelIsLight
+/>;
+<RadioButtonField
```
thebesson, line 33: `RadioButtonField` is not transformed to `Radio` with `FormControl`?

massao: Only if needed, as you check on the `output` file, if it needs the validation message it get wrapper on the `FormControl`

thebesson (approved):

PR #1411, opened 2021-10-15
fix: export get match utility of the Autocomplete package

gui-santos:
# Purpose of PR

Oops! I forgot to do this last time
and because of that it's impossible to use `getMatch`

## PR Checklist

- [ ] Tests are passing


gui-santos:
Hey @suevalov and @mshaaban0 I request a re-review from you because moving the function to f36-utils changed a lot of stuff, so I think it's better if you give one last look into it

Thanks for the suggestion

PR #1459, opened 2021-10-29
fix: autocomplete style & empty groups

Chaoste:
# Purpose of PR

- I don't see the need in rendering an empty group. It should be our default behaviour to not render this until needed otherwise.
- Currently, we need to define the font explicitly for the AutoCompleteItems (only for those). Same as for all components, this should be defined in the library.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/components/autocomplete/src/Autocomplete.tsx
+++ b/packages/components/autocomplete/src/Autocomplete.tsx
@@ -279,7 +279,7 @@ function _Autocomplete<ItemType>(
 
           {!isLoading && isShowingNoMatches && (
             <div className={cx(styles.item, styles.disabled)}>
-              <Subheading margin="none" className={styles.noMatchesTitle}>
+              <Subheading className={styles.noMatchesTitle}>
                 {noMatchesMessage}
               </Subheading>
             </div>
--- a/packages/components/autocomplete/src/Autocomplete.tsx
+++ b/packages/components/autocomplete/src/Autocomplete.tsx
@@ -275,7 +275,7 @@ function _Autocomplete<ItemType>(
 
           {!isLoading && items.length === 0 && (
             <div className={cx(styles.item, styles.disabled)}>
-              <Subheading margin="spacingXs" className="styles.NoMatchesTitle">
+              <Subheading margin="none" className={styles.noMatchesTitle}>
                 {noMatchesMessage}
               </Subheading>
             </div>
--- a/packages/components/autocomplete/src/Autocomplete.tsx
+++ b/packages/components/autocomplete/src/Autocomplete.tsx
@@ -278,7 +278,7 @@ function _Autocomplete<ItemType>(
             ))}
 
           {!isLoading && isShowingNoMatches && (
-            <div className={cx(styles.item, styles.disabled)}>
+            <div className={styles.item}>
               <Subheading className={styles.noMatchesTitle}>
                 {noMatchesMessage}
               </Subheading>
--- a/packages/components/autocomplete/src/Autocomplete.tsx
+++ b/packages/components/autocomplete/src/Autocomplete.tsx
@@ -8,7 +8,7 @@ import { TextInput, TextInputProps } from '@contentful/f36-forms';
 import { CloseIcon, ChevronDownIcon } from '@contentful/f36-icons';
 import { SkeletonContainer, SkeletonBodyText } from '@contentful/f36-skeleton';
 import { Popover } from '@contentful/f36-popover';
-import { SectionHeading } from '@contentful/f36-typography';
+import { Subheading, SectionHeading } from '@contentful/f36-typography';
 
 import { AutocompleteItems } from './AutocompleteItems';
 import { getAutocompleteStyles } from './Autocomplete.styles';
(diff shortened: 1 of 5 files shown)


Lelith (approved):
Looks good to me!

gui-santos:
I think something weird is happening

Screenshot 2021-11-01 at 14 49 56

When using grouped items and there are no matches to what I typed, the autocomplete is not showing the "No matches" message

This is how it is expected to look like:
Screenshot 2021-11-01 at 14 50 04

gui-santos (changes requested):
I think something weird is happening

Screenshot 2021-11-01 at 14 49 56

When using grouped items and there are no matches to what I typed, the autocomplete is not showing the "No matches" message

This is how it is expected to look like:
Screenshot 2021-11-01 at 14 50 04

domarku:
Thanks for the PR @Chaoste 

While we're at it, could you improve the styling of _No matches_?

I suggest we use the copy "No matches found" using [Subheading](https://5fd1dda724cc620021ace8c5-ypfehswhhc.chromatic.com/?path=/story/typography-subheading--basic) in `--gray-500` with the same amount of 1rem spacing above as below.

Preview:
Screen Shot 2021-11-01 at 15 04 34

The hack I did in the browser to get it:
Screen Shot 2021-11-01 at 15 04 43
Screen Shot 2021-11-01 at 15 04 48

Lelith:
@gui-santos 
@Chaoste  
i think the issue of not rendering empty group titles and the "no matches found" is better done in the filter function. Which then gets a little more complex. but the logic of the component then applies just like with non-grouped items. 

In the context of the filter function, we know exactly how the object looks like and how filtering should behave. So i think in this case, it is okay to let the builder who uses the component decide what kind of display they want is best for them.

packages/components/autocomplete/stories/Autocomplete.stories.tsx
```diff
@@ -143,14 +148,22 @@ export const UsingGroupedItems = (args: AutocompleteProps<GroceryList>) => {
   const [filteredItems, setFilteredItems] = useState(groceryList);
 
   const handleInputValueChange = (value: string) => {
-    const newFilteredItems = groceryList.map((group) => {
-      return {
-        ...group,
-        options: group.options.filter((item: Produce) =>
+    const newFilteredItems = groceryList
```
Lelith, line 151: @gui-santos @Chaoste here is the needed change for the filter function. In this way the whole group entry gets removed and if there are no matches at all items is going to be completely empty and work just like any other "no items left in the list" case, no matter the shape of the items.

Lelith: Here would be another way of solving the filer which is maybe more elegant?
```
  const handleInputValueChange = (value: string) => {
    const newFilteredItems: GroceryList[] = [];
    groceryList.forEach((group) => {
      const filteredOptions = group.options.filter((item: Produce) =>
        item.name.toLowerCase().includes(value.toLowerCase()),
      );
      if (filteredOptions.length > 0) {
        newFilteredItems.push({
          groupTitle: group.groupTitle,
          options: filteredOptions,
        });
      }
    });
    setFilteredItems(newFilteredItems);
  };
```

gui-santos: I didn't understand the problem

Why can't something like this work anymore?
```tsx
{!isLoading && items.length === 0 && (
  <div className={cx(styles.item, styles.disabled)}>
     {noMatchesMessage}
  </div>
)}
```

packages/components/autocomplete/src/Autocomplete.tsx
```diff
@@ -275,13 +275,18 @@ function _Autocomplete<ItemType>(
 
           {!isLoading && items.length === 0 && (
             <div className={cx(styles.item, styles.disabled)}>
-              {noMatchesMessage}
+              <Subheading margin="spacingXs" className="styles.NoMatchesTitle">
```
gui-santos, line 278:
```suggestion
              <Subheading margin="none" className={styles.noMatchesTitle}>
```

We are using camelCase for class names, that's why I rename the class to "noMatchesTitle"
And we are always passing it using styles as an object


I also suggest using `margin="none"` so the message does not have an inconsistent spacing around it
And we only get the spacing coming from the Popover itself

gui-santos (changes requested):
Looks good to me!

I made one suggest to make match our naming convention and the way we are using emotion in the repo

Besides that, I think you still need to update the default value of the `noMatchesMessage` prop to "No matches found" as @domarku suggested

packages/components/autocomplete/stories/Autocomplete.stories.tsx
Chaoste: I believe I fixed the reason. If the client only passes empty lists, we show the no matches string now.

packages/components/autocomplete/src/Autocomplete.tsx
```diff
@@ -158,6 +158,10 @@ function _Autocomplete<ItemType>(
       )
     : items;
 
+  const isShowingNoMatches = isUsingGroups(isGrouped, items)
```
Lelith, line 161: 😲 ofc that is also a way to solve it!

gui-santos:
I still think this is not looking like the image in @domarku comment

Screenshot 2021-11-04 at 13 47 54

Lelith:
@domarku @gui-santos i tried now to reproduce the exact same css as suggested, i do not know why the font still does look lighter?
image
image

massao:
> @domarku @gui-santos i tried now to reproduce the exact same css as suggested, i do not know why the font still does look lighter? image image

Its getting the `disabled` styling from the parent div, that has a `opacity: 0.5`

packages/components/autocomplete/src/Autocomplete.tsx
```diff
@@ -273,15 +277,20 @@ function _Autocomplete<ItemType>(
               </div>
             ))}
 
-          {!isLoading && items.length === 0 && (
+          {!isLoading && isShowingNoMatches && (
             <div className={cx(styles.item, styles.disabled)}>
```
massao, line 281:
```suggestion
            <div className={styles.item}>
```
This should fix the styling on the text

Lelith:
@massao thank you! that seems to do the trick.
image

gui-santos (approved):

PR #1468, opened 2021-11-01
fix(tabs component): fix issues and bugs for controlled tabs component

thebesson:
# Purpose of PR

the initial focused index based on default/current tab;
manual activation for controlled and uncontrolled tabs component;

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/packages/components/tabs/src/Tabs.tsx
+++ b/packages/components/tabs/src/Tabs.tsx
@@ -1,6 +1,16 @@
 import React from 'react';
 import type { CommonProps } from '@contentful/f36-core';
 import { TabsContextProvider } from './tabsContext';
+
+const calculateIndexMap = (children) => {
+  const tabListArray = React.Children.toArray(children).find(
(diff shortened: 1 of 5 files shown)


packages/components/tabs/src/Tabs.tsx
```diff
@@ -1,6 +1,16 @@
 import React from 'react';
 import type { CommonProps } from '@contentful/f36-core';
 import { TabsContextProvider } from './tabsContext';
+
+const calculateIndexMap = (children) => {
+  const tabListArray = React.Children.toArray(children).find(
```
suevalov, line 6: I'm not really sure about the solution. I would question why do we need to have an initially focused item at all.

The tabs can be rendered anywhere on the page and some other elements can be focused by default.

suevalov: I'd say that nothing should be really focused on when the component is rendered.

PR #1469, opened 2021-11-01
docs: folder structure guide

gui-santos:
# Purpose of PR

While working on the Next.js experiment, I noticed that we are not following any convention on how to organise our files in the repo. This was noticeable because the website tries to generate pages from the .mdx files, but they are not always in the same places or with the same name

I created a guide that proposes a structure for our files similar to what we have in the "features" folder in the webapp project

Please comment, suggest things, and fix my grammar!

## Note on the naming of the files

Right now, in v4, we always have two MD files in a folder because one markdown has the documentation of that component and the other has a warning for people who is using it during our "alpha" phase

We are almost releasing the beta version of v4 and all the documentation of every component will be moved to README.mdx files

That's another reason why this is the RIGHT moment to come up with a folder structure!

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/docs/folder-structure.md
+++ b/docs/folder-structure.md
@@ -52,7 +52,7 @@ In these cases, the folder structure of the package will look more like this:
     index.ts
     Button.tsx
   /stories
-    # stories for storybook
+    # stories for storybook of each component inside the package
   /test
     # test files
   README.mdx
--- a/docs/folder-structure.md
+++ b/docs/folder-structure.md
@@ -1,0 +1,96 @@
+# F36 Folder Structure
+
+To make it easy to maintain our files and folders we ask every contributor to follow this folder structure.
+Every package has its own folder, except for components which should be kept in the folder `/components` (every component must have its own folder).
+
+## Table of contents
+
+- [Components](#components)
+  - [Packages with more than one component](#packages-with-more-than-one-component)
+  - [Packages with utility functions](#packages-with-utility-functions)
+  - [Why do we use MDX?](#why-do-we-use-mdx)
+  - [Why is every `.mdx` file named "README"](#why-is-every-mdx-file-named-readme)
+
+## Components
+
+The basic folder structure of a component looks like this:
+
+```
+/my-component
+  /src
+    index.ts
+    MyComponent.tsx
+    MyComponent.styles.ts
+  /stories
+    # stories for storybook
+  /test
+    # test files
+  README.mdx
+  package.json
+```
+
+### Packages with more than one component
+
+Sometimes the package will include a "base" component and some variants that are build from the base component (e.g. Button).
+In these cases, the folder structure of the package will look more like this:
+
+```
+/button
+  /src
+    /ButtonGroup
+      README.mdx
+      index.ts
+      ButtonGroup.tsx
+    /IconButton
+      README.mdx
+      index.ts
+      IconButton.tsx
+    /ToggleButton
+      README.mdx
+      index.ts
+      IconButton.tsx
+    index.ts
+    Button.tsx
+  /stories
+    # stories for storybook
+  /test
+    # test files
+  README.mdx
+  package.json
+```
+
+### Packages with utility functions
+
+Another possible case is that the package has a component and some utility functions. The folder structure will loke like this:
+
+```
+/datetime
+  /src
+    /RelativeDateTime
+      README.mdx
+      index.ts
+      RelativeDateTime.tsx
+    /utils
+      README.mdx
+      index.ts
+      formatDateTimeUtils.ts
+    index.ts
+    DateTime.tsx
+  /stories
+    # stories for storybook
+  /test
+    # test files
+  README.mdx
+  package.json
+```
+
+### Why do we use MDX?
+
+We use `.mdx` file format instead of just `.md` because these files are also used
+by our website to generate its pages.
+You can read more about the benefits of mdx [here](https://mdxjs.com/).
+
+### Why is every `.mdx` file named "README"?
+
+The markdown files must be named "README" because we want them to show up in GitHub
+for the people that navigate through the folders of the repo looking for documentation.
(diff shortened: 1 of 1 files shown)


suevalov (approved):

docs/folder-structure.md
```diff
@@ -0,0 +1,96 @@
+# F36 Folder Structure
+
+To make it easy to maintain our files and folders we ask every contributor to follow this folder structure.
+Every package has its own folder, except for components which should be kept in the folder `/components` (every component must have its own folder).
```
burakukula, line 4: I am not sure if I understand this sentence. 🤔 
Maybe we could split it somehow, to clearly distinguish the concepts - packages vs components? For example:
"Every package has its own folder. _Here we could give some examples maybe?_" 
"New components should be kept in the folder `/components` and has its own folder as well."

docs/folder-structure.md
```diff
@@ -0,0 +1,96 @@
...
+      index.ts
+      IconButton.tsx
+    /ToggleButton
+      README.mdx
+      index.ts
+      IconButton.tsx
+    index.ts
+    Button.tsx
+  /stories
+    # stories for storybook
```
burakukula, line 55: f
```suggestion
    # stories for the storybook for all the components from the package
```

docs/folder-structure.md
```diff
@@ -0,0 +1,96 @@
...
+  /test
```
burakukula, line 56: I am wondering if maybe the tests folder should live in each component folder? but maybe it's overkill to have folder `test` in every component?

docs/folder-structure.md
gui-santos: What I meant was that:

package = every folder with a package.json file basically... and that we publish in npm as a package

Inside `/packages` we have: core, f36-website, f36-tokens, etc

So every package we create we create a folder for it in `/packages`, except when it's a react ui component
then we create a folder inside of `/packages/components`. (e.g. button, accordion, autocomplete, forms)

docs/folder-structure.md
gui-santos: # stories for storybook of each component inside the package
```suggestion
    # stories for storybook of each component inside the package
```

wdyt?

docs/folder-structure.md
```diff
@@ -0,0 +1,96 @@
...
+
+- [Components](#components)
+  - [Packages with more than one component](#packages-with-more-than-one-component)
+  - [Packages with utility functions](#packages-with-utility-functions)
+  - [Why do we use MDX?](#why-do-we-use-mdx)
+  - [Why is every `.mdx` file named "README"](#why-is-every-mdx-file-named-readme)
+
+## Components
+
+The basic folder structure of a component looks like this:
```
massao, line 16: maybe also point out that the `yarn generate component` would already create the folder with the files in the correct place?

docs/folder-structure.md
gui-santos: I like when we have the test file close to the component file, like:

```
Button.tsx
Button.test.ts
```

But since we were creating a `test` folder, I put in the document the way we do it now
Let's see what everyone thinks

docs/folder-structure.md
gui-santos: How can I explain that ☝️ better?

docs/folder-structure.md
massao: We can have the `.test.tsx` files in each component folder if needed.

docs/folder-structure.md
massao: Maybe just be more verbose? 🤔 
Something like: 
Every package that should be published should have it's own folder, with the exception of UI components, that should have their folder and files created inside the `components` directory.

docs/folder-structure.md
gui-santos: good point!

docs/folder-structure.md
burakukula: yup! Sounds good!

docs/folder-structure.md
burakukula: Yeah, ok. I guess we need to keep it as it is (do we? 😁). But I'm with you, I feel like tests belong to each component.

docs/folder-structure.md
```diff
@@ -0,0 +1,96 @@
...
+### Why do we use MDX?
+
+We use `.mdx` file format instead of just `.md` because these files are also used
+by our website to generate its pages.
+You can read more about the benefits of mdx [here](https://mdxjs.com/).
+
+### Why is every `.mdx` file named "README"?
+
+The markdown files must be named "README" because we want them to show up in GitHub
+for the people that navigate through the folders of the repo looking for documentation.
```
burakukula, line 96: hmm, now when I think about it, when I will look for the README for the button, I would need to go through the all REAME's that was found and check the file source to open the correct one. but I guess that's a common practice to call it README.mdx instead of Button.mdx, right?

docs/folder-structure.md
gui-santos: I will open a poll in the slack channel
Since, you, @massao,  and I want to change this already hahahaha

docs/folder-structure.md
gui-santos: They need to be called "README" otherwise you won't see anything when you go here in Github
https://github.com/contentful/forma-36/tree/forma-v4/packages/components/button

gui-santos: And, look at this:

Screenshot 2021-11-02 at 14 10 23

burakukula (approved):

massao (approved):
LGTM

docs/folder-structure.md
gui-santos: I will merge this now and we change this part about the tests once everyone has the chance to express their opinion

PR #1470, opened 2021-11-02
fix(menu-item): isDisabled prop

denkristoffer:
# Purpose of PR

Adds `isDisabled` prop to `MenuItem` component.

Which leaves me with the question, do we want to and if so is there any way to map standard HTML element props to our own names following our style guide? I.e. don't allow `disabled` but always map it to `isDisabled`.

### DIFF
--- a/packages/components/menu/src/MenuItem/MenuItem.tsx
+++ b/packages/components/menu/src/MenuItem/MenuItem.tsx
@@ -16,6 +16,9 @@ interface MenuItemInternalProps extends CommonProps {
   children?: React.ReactNode;
   as?: 'a' | 'button';
 
+  /** Renders item as disabled */
+  isDisabled?: boolean;
+
   /**
    * Sets focus on item
    */


packages/components/menu/src/MenuItem/MenuItem.tsx
```diff
@@ -16,6 +16,9 @@ interface MenuItemInternalProps extends CommonProps {
   children?: React.ReactNode;
   as?: 'a' | 'button';
 
+  /** Renders item as disabled */
+  isDisabled?: boolean;
```
suevalov, line 20: is it only missing typescript property? don't we need to add an implementation that takes this property into consideration?

denkristoffer: Yes. I mostly opened this PR to have the discussion around if we can and want to map these properties somehow. Maybe I should move it elsewhere?

bgutsol: Menu items has disabled property, it's inherited from standart `button` attrs. There are usage example in docs and storybook. https://github.com/contentful/forma-36/blob/52a7aa67ed9211cb4016eb8b3fc16de06adf305c/packages/components/menu/stories/Menu.stories.tsx#L72

bgutsol (changes requested):
Menu items has disabled property, it's inherited from standart `button` attrs. There are usage example in docs and storybook. https://github.com/contentful/forma-36/blob/52a7aa67ed9211cb4016eb8b3fc16de06adf305c/packages/components/menu/stories/Menu.stories.tsx#L72

packages/components/menu/src/MenuItem/MenuItem.tsx
denkristoffer: Yes but what I'm saying is that the "regular" button prop names don't align with our naming scheme `disabled` vs. `isDisabled` 🙂

Lelith: I see @denkristoffer point - but, even though native attributes do not follow our naming conventions, I think `disabled` should be expectable as a spread property on every component that is "disable-able". For me it would be much more confusing if i have to check every time if a component is now expecting a native `disabled` or the "conventional-correct" `isDisabled`

bgutsol: Yes, that was my idea not to change the default html attributes. Because I do nothing with this disabled prop inside component, it just goes straight to html tag props. 
But I also agree that it breaks consistency between our components 🤷‍♂️

gui-santos:
Can we close this PR? and maybe move the discussion somewhere else if it's still being discussed

denkristoffer:
Yes, but I think we should discuss this. I can see the idea of being able to spread props without changing them, but that also means that if I set it manually I have use a different name than everywhere else.

PR #1474, opened 2021-11-03
fix: do not focus menu trigger on mouse click outside

mshaaban0:
# Purpose of PR

Don't focus menu trigger on blur, and default to browser behaviour

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/components/popover/src/Popover.tsx
+++ b/packages/components/popover/src/Popover.tsx
@@ -213,7 +213,7 @@ export function Popover(props: PopoverProps) {
             return;
           }
 
-          closeAndFocusTrigger();
+          onClose();
         },
         onKeyDown: (event: React.KeyboardEvent<HTMLDivElement>) => {
           if (_props.onKeyDown) {
@@ -237,6 +237,7 @@ export function Popover(props: PopoverProps) {
       popoverElement,
       triggerElement,
       closeAndFocusTrigger,
+      onClose,
     ],
   );
 
--- a/packages/components/popover/src/Popover.tsx
+++ b/packages/components/popover/src/Popover.tsx
@@ -25,7 +25,7 @@ export interface PopoverProps {
   /**
    * Callback fired when the popover closes
    */
-  onClose?: Function;
+  onClose?: () => void;
 
   /**
    * Determines the preferred position of the Popover. This position is not
@@ -213,7 +213,7 @@ export function Popover(props: PopoverProps) {
             return;
           }
 
-          onClose();
+          onClose?.();
         },
         onKeyDown: (event: React.KeyboardEvent<HTMLDivElement>) => {
           if (_props.onKeyDown) {
--- a/packages/components/popover/src/PopoverContent/PopoverContent.styles.ts
+++ b/packages/components/popover/src/PopoverContent/PopoverContent.styles.ts
@@ -11,6 +11,8 @@ export const getPopoverContentStyles = (isOpen: boolean) => ({
     zIndex: tokens.zIndexDropdown,
     '&:focus': {
       outline: 'none',
+    },
+    '&:focus-visible': {
       boxShadow: tokens.glowPrimary,
     },
   }),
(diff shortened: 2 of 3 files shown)


suevalov (changes requested):
I think it's a questionable change. Many modern libraries focus on trigger after the close (Chakra, Radix) and it's considered a good practice.

bgutsol (changes requested):
Can you please elaborate please why do we want to make it this way? 
If it's only one specific case, we can add a flag prop

mshaaban0:
@suevalov @bgutsol So basically it's a feedback I got from Amr and other designers, their example was Google when opening a menu and clicking away with mouse the focus moves to whatever element is clicked, even if it not an interactive element the browser virtually focuses it and the next tab focuses the right interactive element.

I can also see it causing issues if the menu isn't closed and the user clicks an interactive element, example: 

https://user-images.githubusercontent.com/6163988/140058615-fac9cab1-8423-461b-a5de-0eed79e6a749.mov

Just to be clear, this change still keeps the trigger focus when using keyboard `tab` or `escape` it's only for mouse clicks

bgutsol:
Good catch @mshaaban0 
TBH Chakra doesn't have such problem with taking focus from interactive elements when you click on them 🤔

I agree that as a quick fix we can just remove the logic with focus on blur. But I would investigate later how we can improve it and make it Chakra like.

@mshaaban0 Can I ask also to change this in the `Popover`, please:
https://github.com/contentful/forma-36/blob/6b18c19a0e9b5e44f2f2462ac95bdcb85a38335f/packages/components/popover/src/Popover.tsx#L216

packages/components/popover/src/PopoverContent/PopoverContent.styles.ts
```diff
@@ -9,5 +9,9 @@ export const getPopoverContentStyles = (isOpen: boolean) => ({
     borderRadius: tokens.borderRadiusMedium,
     boxShadow: tokens.boxShadowDefault,
     zIndex: tokens.zIndexDropdown,
+    '&:focus': {
+      outline: 'none',
+      boxShadow: tokens.glowPrimary,
```
bgutsol, line 14: I remember @domarku didn't want that shadow around dropdown. Could you double check with him, please?

packages/components/popover/src/Popover.tsx
```diff
@@ -213,7 +213,7 @@ export function Popover(props: PopoverProps) {
             return;
           }
 
-          closeAndFocusTrigger();
+          onClose();
```
bgutsol, line 216:
```suggestion
          onClose?.();
```
`onClose` prop can be undefined.
I see that TS doesn't warn about this, it's because of a `Function` type in props `onClose?: Function`
Could you change it please to `() => void` 🙇‍♂️

packages/components/popover/src/PopoverContent/PopoverContent.styles.ts
mshaaban0: hey @domarku the popover content had the browser default outline, here is how it look before and after:

**Before**
Screenshot 2021-11-05 at 15 20 39

**After**
Screenshot 2021-11-05 at 15 20 10

bgutsol (approved):
Thanks @mshaaban0 🏅

suevalov (approved):

packages/components/popover/src/PopoverContent/PopoverContent.styles.ts
domarku: Oh yeah, that makes sense! Let's use our styling. Just want to make sure it's for keyboard focus only, not when using the mouse to open the popover / menu.

mshaaban0: at the moment it's for both mouse and keyboard on popover, menu is a different behaviour as it focuses the first element inside the menu

bgutsol: maybe we can use here `:focus-visible`, wdyt @mshaaban0 ?

mshaaban0: yeah nice, that'll work 👍🏼

PR #1477, opened 2021-11-03
fix: fix card issues

burakukula:
Small fixes for keyboard navigation for Card.
Removing `isFocused` property - are there any cases where we actually need to control this property?

### DIFF
--- a/packages/components/card/src/base-card/BaseCard.styles.ts
+++ b/packages/components/card/src/base-card/BaseCard.styles.ts
@@ -23,7 +23,7 @@ export const getBaseCardStyles = () => {
       borderTopLeftRadius: tokens.borderRadiusMedium,
       cursor: 'grab',
     }),
-    root: ({ hasHeader, isFocused, isHovered, isSelected }) => {
+    root: ({ hasHeader, isHovered, isSelected }) => {
       const styles: ObjectInterpolation<undefined> = {
         backgroundColor: tokens.colorWhite,
         borderColor: tokens.gray300,
@@ -40,6 +40,12 @@ export const getBaseCardStyles = () => {
         textDecoration: 'none',
         transition: `border-color ${tokens.transitionDurationDefault} ${tokens.transitionEasingDefault},
     box-shadow ${tokens.transitionDurationShort} ${tokens.transitionEasingDefault}`,
+
+        '&:focus': css({
+          borderColor: tokens.colorPrimary,
+          boxShadow: tokens.glowPrimary,
+          outline: '0',
+        }),
       };
 
       if (!hasHeader) {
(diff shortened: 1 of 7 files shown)


packages/components/card/src/base-card/BaseCard.styles.ts
```diff
@@ -40,6 +40,12 @@ export const getBaseCardStyles = () => {
         textDecoration: 'none',
         transition: `border-color ${tokens.transitionDurationDefault} ${tokens.transitionEasingDefault},
     box-shadow ${tokens.transitionDurationShort} ${tokens.transitionEasingDefault}`,
+
+        '&:focus': css({
```
massao, line 44: shouldn't we do the same for `&:hover`?

packages/components/card/src/base-card/BaseCard.tsx
```diff
@@ -66,7 +66,6 @@ function _BaseCard<E extends React.ElementType = typeof BASE_CARD_DEFAULT_TAG>(
   forwardedRef: React.Ref<HTMLElement>,
 ) {
   const styles = getBaseCardStyles();
-  const [isFocused, setIsFocused] = useState(isFocusedProp ?? false);
```
massao, line 69: We can also remove the `isFocusProp` from the props list, if it's not used anywhere else.

packages/components/card/src/base-card/BaseCard.styles.ts
burakukula: hover is actually depending on the fact if the card is interactive or not, so if it's a button. Maybe it's easier to keep it as it is in that case?

denkristoffer:
My note here would be that without this (and other states) we lose the ability to check them with Chromatic.

burakukula:
> My note here would be that without this (and other states) we lose the ability to check them with Chromatic.

ahhh, ok, I understand, but should we keep it for this purpose?

denkristoffer:
Don't know 🤷 I found it helpful when building the components but there's no denying that it complicates something simple.

massao:
> Don't know 🤷 I found it helpful when building the components but there's no denying that it complicates something simple.

We can also check if there isn't any addon for storybook that would help with that, the one I know of is this one: https://storybook.js.org/addons/storybook-addon-pseudo-states
but it would required a separate story for different states, as far as I remember.

burakukula:
> > Don't know 🤷 I found it helpful when building the components but there's no denying that it complicates something simple.
> 
> We can also check if there isn't any addon for storybook that would help with that, the one I know of is this one: https://storybook.js.org/addons/storybook-addon-pseudo-states but it would required a separate story for different states, as far as I remember.

@massao @denkristoffer maybe let's figure this out separately since that's something that might also apply to other components like Buttons?
If we agree I will clean up stories and remove the focus state from the overview.

massao (approved):

PR #1480, opened 2021-11-03
refactor(website): Improve navigation on menu and submenu rendering

massao:
# Purpose of PR

Make it so the submenus area always rendered even when not expanded, so the docsearch crawler can correctly find the links to be indexed.
Improve keyboard navigation on the menu

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/forma-36-website/src/components/Navigation.js
+++ b/packages/forma-36-website/src/components/Navigation.js
@@ -104,6 +104,12 @@ const MenuListItem = React.forwardRef(
       setIsExpanded(!isExpanded);
     };
 
+    const handleKeyDown = (event) => {
+      if (event.nativeEvent.code === 'Enter') {
+        handleToggle(event);
+      }
+    };
+
     const itemOffset = { paddingLeft: `${1 + hierarchyLevel}rem` };
 
     return (
@@ -113,6 +119,9 @@ const MenuListItem = React.forwardRef(
             <div
               css={[styles.link, styles.linkGroup, itemOffset]}
               onClick={handleToggle}
+              onKeyDown={handleKeyDown}
+              // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
+              tabIndex={0}
             >
               {isCategory ? (
                 <SectionHeading>{item.name}</SectionHeading>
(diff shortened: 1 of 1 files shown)


suevalov (approved):

packages/forma-36-website/src/components/Navigation.js
```diff
@@ -161,15 +173,20 @@ MenuListItem.defaultProps = {
...
+}) => {
   const activeRef = useRef(null);
   useEffect(() => {
     if (activeRef.current) {
       activeRef.current.scrollIntoView({ block: 'center' });
     }
   }, []);
   return (
-    <ul css={styles.list}>
+    <ul css={[styles.list, isHidden && styles.hidden]}>
```
Lelith, line 189: would it not make more sense to solve this via emotion cx?

Lelith (changes requested):

packages/forma-36-website/src/components/Navigation.js
gui-santos: ☝️ +1

massao: It's essentially the same, only different approach, `cx` would generate the classes to be passed to `className` prop, and the `css` prop is also from emotion.
but we could, but would be kind of merging an array, and passing the merged array to a function that can merge the array if the parameter is an array, and since we won't be doing composition here as we do on the components itself  not sure if needed 🤔

Lelith: @massao you are right - effect wise, it is no difference. but i think we should stick with one dedicated pattern and use the full power of emotion, if we already have committed  to it.

on the other hand, as i am no big fan of css-in-js anyways, i would clearly join in every argument that would move us away from emotion and use css-modules (with sass) only. but i think the mix and match of the patterns makes it harder to understand the applied styles.

Lelith: Sorry i also did not recognise that this is a v3 fix 🙈

Lelith (approved):

packages/forma-36-website/src/components/Navigation.js
gui-santos: I think, before emotion, we always preferred to use classNames
since it's the only way to use css modules

but when emotion was introduced, we started to get different ways of doing things just because emotion allow us to do it

the problem that I see is that when a team has too many ways of doing things, it's hard to onboard new developers to the codebase

let's keep like this then until the team decides on how we want to use emotion 🤷

PR #1513, opened 2021-11-11
fix: remove overflow hidden from BaseCard

gui-santos:
# Purpose of PR

BaseCard had a `overflow: hidden` that was cropping the focus glow of the things inside the card

Before:
card_focus_bug

After:
card_focus_fix

It's possible that we introduced this `overflow: hidden` issue in other places, some people spotted the focus glow getting cropped in Compose and Launch
So let's take care when using `overflow: hidden`, I would say that most of the time we do not need it

## PR Checklist

- [ ] Tests are passing


burakukula (approved):

denkristoffer:
This was definitely added for a reason, did you check that nothing breaks after this change?

PR #1516, opened 2021-11-12
fix: set nav and migration guide in the alpabetical order

burakukula:


bgutsol (approved):

packages/forma-36-website/gatsby-config.js
```diff
@@ -361,6 +331,36 @@ module.exports = {
...
+              {
+                name: 'Subheading',
+                link: '/components/subheading/',
+              },
+              {
+                name: 'DisplayText',
+                link: '/components/display-text/',
+              },
+            ],
+          },
```
gui-santos, line 363: I think there are more stuff to be sorted here, look:

Screenshot 2021-11-12 at 15 38 08

for example: Popover is after DateTime, but before EntityList

gui-santos (changes requested):

packages/forma-36-website/gatsby-config.js
burakukula: ah, true! sorry, didn't notice. on my way.

burakukula: hmm, now when I think about it, maybe Popover should be a subnav of Menu?

gui-santos: I only gave one example, there are a bunch of components that are not in the right alphabetical order

IMO the way Popover and Menu are now is fine

bgutsol: I would not make Popover as a subnav of Menu

burakukula: I guess should be ok now?
Screenshot 2021-11-12 at 16 11 44

massao (approved):

gui-santos (approved):

PR #1528, opened 2021-11-16
fix(autocomplete): fix dropdown props type

thebesson:
# Purpose of PR

This PR is a reaction to feedback about Autocomplete component - passing the `dropdownProps` prop causes a type error, because the `children` prop is required, but not passed. 

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/packages/forma-36-react-components/src/components/Dropdown/Dropdown.tsx
+++ b/packages/forma-36-react-components/src/components/Dropdown/Dropdown.tsx
@@ -86,7 +86,7 @@ export interface DropdownProps {
   /**
    * Child nodes to be rendered in the component
    */
-  children?: React.ReactNode;
+  children: React.ReactNode;
   /**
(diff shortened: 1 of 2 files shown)


packages/forma-36-react-components/src/components/Dropdown/Dropdown.tsx
```diff
@@ -86,7 +86,7 @@ export interface DropdownProps {
   /**
    * Child nodes to be rendered in the component
    */
-  children: React.ReactNode;
+  children?: React.ReactNode;
```
gui-santos, line 89: Is this correct?

I think we do want to make children mandatory in Dropdown
A person can't/shouldn't do this:

```
<Dropdown />
```

right?

gui-santos (changes requested):

gui-santos (approved):

PR #1532, opened 2021-11-16
docs: use double quotes when prop is string

gui-santos:
# Purpose of PR

I propose to add this convention to or code style guide

When the value of the prop is a string, pass it directly with double quotes. There is no need to wrap it with curly braces.

```
// bad
<Button variant={'primary'} >Add entry</Button>

// good
<Button variant="primary" >Add entry</Button>
```

This PR also replaces all the places where this convention is not being done now

This should be something that our eslint and/or prettier should warn but I didn't have time to check why it's not throwing warns when linting the code

## PR Checklist

- [ ] Tests are passing


burakukula (approved):
👌

PR #1533, opened 2021-11-18
fix: add step by step guide on how to migrate your project

burakukula:


MIGRATION.md
```diff
@@ -70,37 +75,93 @@
 
 ## How to migrate your packages to v4
 
-Install a package that contains all of the components from Forma 36. Tree-shaking will take care of your build, so it will include only components that you use.
-We separate only icons from the main package. If you need to use our icons, install them separately, by running the following commands:
-:
+### Step 1 - install new packages
```
burakukula, line 78: Potentially we will have just one codemod command to do everything. We still would need to have this kinda step by step guide anyway.

MIGRATION.md
```diff
@@ -70,37 +75,93 @@
...
+yarn remove @contentful/f36-tokens
+yarn remove @contentful/f36-fcss
+```
+
+Now find the following imports and remove them as well:
+
+```tsx
+import '@contentful/forma-36-react-components/dist/styles.css';
+import '@contentful/forma-36-fcss/dist/styles.css';
+import '@contentful/forma-36-tokens/dist/css/index.css';
```
bgutsol, line 164: Should we also add that user needs to add  our new`GlobalStyles` component that has normalize and some basic styles?

burakukula: AA! Good point!

MIGRATION.md
```diff
@@ -70,37 +75,93 @@
...
+npm install @contentul/f36-button
+```
+
+For YARN
+
+```bash
+yarn add @contentul/f36-button
+```
+
+Note: We don't recommend using it this way, it is much easier to [install full package of components](#Step-1-install-new-packages). Tree-shaking will take care of your build
```
gui-santos, line 128:
```suggestion
<Note> We don't recommend using it this way, it is much easier to install full package of components. Tree-shaking will take care of your build </Note>
```
You can use React components in mdx (It's how the Props table is rendered) so what do you think about using Note here? https://v4-forma-36.netlify.app/components/note/
You would have to use TexLink too for that link

gui-santos: Another option would be to use Markdown quote like:

> This is a quote

but I think our MDXRenderer is not prepare to render one of those so we would need to work a bit on it
(Maybe that's a job for another PR 🤷 )

gui-santos (changes requested):
This looks great!

I have only one suggestion regarding how we show that "Note"

gui-santos (approved):

MIGRATION.md
burakukula: uuu! I like the idea about the Note component!

MIGRATION.md
```diff
@@ -70,37 +76,113 @@
...
+
+```tsx
+import '@contentful/forma-36-react-components/dist/styles.css';
+import '@contentful/forma-36-fcss/dist/styles.css';
+import '@contentful/forma-36-tokens/dist/css/index.css';
+```
+
+### Step 4 - use GlobalStyles component
+
+Control default browser styles with the `GlobalStyles` component. The GlobalStyles component uses the Global component from [Emotion](https://emotion.sh/docs/globals) under the hood. Import GlobalStyles somewhere at the beginning of your project like in the example below:
```
thebesson, line 170:
```suggestion
Control default browser styles with the `GlobalStyles` component. The GlobalStyles component uses the Global component from [Emotion](https://emotion.sh/docs/globals) under the hood. Import GlobalStyles somewhere at the root of your project like in the example below:
```

thebesson (approved):

MIGRATION.md
thebesson: oh no I was too slow :)

burakukula: 😬 so sorry. I will fix it in a separate PR with the full review comments from Alexandra.

thebesson: oh, not a worry! it's not a big deal

PR #1537, opened 2021-11-18
feat: add option to run all codemods

burakukula:
Extend to have an option to run all existing codemods.

In the future:
 - cleanup package.json
 - remove CSS imports (there is function for that that is commented out)

### DIFF
--- a/packages/forma-36-codemod/bin/cli.js
+++ b/packages/forma-36-codemod/bin/cli.js
@@ -28,10 +28,10 @@ function checkGitStatus(force) {
     }
   }
 
-  if (!clean) {
+  if (!clean && !force) {
     console.log(
       chalk.yellow(
-        'Before we continue, please stash or commit your git changes.',
+        'Before we continue, please stash or commit your git changes or use the --force flag to override this check',
       ),
     );
     process.exit(1);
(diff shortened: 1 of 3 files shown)


packages/forma-36-codemod/bin/cli.js
```diff
@@ -28,10 +28,10 @@ function checkGitStatus(force) {
     }
   }
 
-  if (!clean) {
+  if (!clean && !force) {
     console.log(
       chalk.yellow(
-        'Before we continue, please stash or commit your git changes.',
+        'Before we continue, please stash or commit your git changes or use the --force flag to override this check',
```
gui-santos, line 34:
```suggestion
        'Before we continue, please stash or commit your changes.\nIf you want to override this check, use the --force flag',
```

Wdyt about doing this? I'm suggesting it so we avoid repeating "or" that much

This should render in two lines like:

```
Before we continue, please stash or commit your changes.
If you want to override this check, use the --force flag
```

packages/forma-36-codemod/transforms/v4-all.js
```diff
@@ -0,0 +1,74 @@
...
+
+//   return j(source)
+//     .find(j.ImportDeclaration)
+//     .forEach((path) => {
+//       if (path.value.source.value.match(/^@contentful\/forma-36.+?\.css$/)) {
+//         j(path).remove();
+//       }
+//     })
+//     .toSource();
+// }
```
gui-santos, line 44: leftover?

gui-santos (approved):
Cool!

packages/forma-36-codemod/transforms/v4-all.js
burakukula: just something that we will use later when we extend the "all in one" codemod - it will cleanup the package.json + this, so remove CSS imports.

suevalov (approved):

packages/forma-36-codemod/transforms/v4-all.js
massao: This is related to this point on the description:
> * remove CSS imports (there is function for that that is commented out)

We want to have it in the future so we left the codemod there

PR #1544, opened 2021-11-19
fix: update links for storybook for v4

burakukula:


bgutsol (approved):
☀️

massao (commented):
Should we consider having a subdomain for it too?? similar to the documentation website?

burakukula:
> Should we consider having a subdomain for it too?? similar to the documentation website?

@massao yeah, I thought of that, but it will take us a long time - need to request it in the infra team. Maybe we can just follow up?

PR #1545, opened 2021-11-22
feat(codemod): add option to update packages on codemod

massao:
This PR adds possibility to update packages through the codemod.
Also adds option to pass or select multiple codemods.

### DIFF
--- a/packages/forma-36-codemod/bin/inquirer-choices.js
+++ b/packages/forma-36-codemod/bin/inquirer-choices.js
@@ -1,3 +1,5 @@
+const inquirer = require('inquirer');
+
 const PARSER_CHOICES = [
   {
     name: 'JavaScript',
@@ -38,6 +40,7 @@ const TRANSFORMS_CHOICES = [
       'color-tokens-to-new-tokens: Converts deprecated color tokens to the new ones',
     value: 'color-tokens-to-new-tokens',
   },
+  new inquirer.Separator(),
   {
     name: 'v4-text-link: Converts TextLink component from Forma v3 to v4',
     value: 'v4-text-link',
--- a/packages/forma-36-codemod/bin/inquirer-choices.js
+++ b/packages/forma-36-codemod/bin/inquirer-choices.js
@@ -14,17 +14,16 @@ const PARSER_CHOICES = [
 const SETUP_CHOICES = [
   {
     name:
+      'run-all-v4: Update package json with new packages and remove old ones, remove v3 CSS imports and run all possible codemods for components.',
+    value: 'run-all-v4',
+  },
+  {
+    name:
       'update-package-json: Updates package.json file with correct packages',
     value: 'update-package-json',
   },
   {
-    name:
-      'update-package-json-all-v4: Updates package.json file with correct packages and run all existing v4 codemods',
-    value: 'update-package-json-all-v4',
-  },
-  {
-    name:
-      'migrate-all-components-to-v4: Run all exising codemods, and clean css imports',
+    name: 'migrate-all-components-to-v4: Run all exising codemods',
     value: 'migrate-all-components-to-v4',
   },
   {
--- a/packages/forma-36-codemod/bin/inquirer-choices.js
+++ b/packages/forma-36-codemod/bin/inquirer-choices.js
@@ -19,7 +19,7 @@ const SETUP_CHOICES = [
   },
   {
     name:
-      'update-package-json: Updates package.json file with correct packages',
+      'update-package-json: Updates package.json file with correct packages, and remove v3 CSS imports',
     value: 'update-package-json',
   },
   {
--- a/packages/forma-36-codemod/bin/inquirer-choices.js
+++ b/packages/forma-36-codemod/bin/inquirer-choices.js
@@ -11,7 +11,18 @@ const PARSER_CHOICES = [
 
 const SETUP_CHOICES = [
   {
-    name: 'migrate-all-components-to-v4: Run all exising codemods',
+    name:
+      'update-package-json: Updates package.json file with correct packages',
+    value: 'update-package-json',
+  },
+  {
+    name:
+      'update-package-json-all-v4: Updates package.json file with correct packages and run all existing v4 codemods',
+    value: 'update-package-json-all-v4',
+  },
+  {
+    name:
+      'migrate-all-components-to-v4: Run all exising codemods, and clean css imports',
     value: 'migrate-all-components-to-v4',
   },
   {
@@ -149,6 +160,10 @@ const TRANSFORMS_CHOICES = [
   {
     name: 'v4-inline-entry-card: Converts InlineEntryCard from Forma v3 to v4',
     value: 'v4-inline-entry-card',
+  },
+  {
+    name: 'v4-clean-css: Removes the imports for v3 css files',
+    value: 'v4-clean-css',
   },
   // Add extra codemods - do not remove
 ];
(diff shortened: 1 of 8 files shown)


packages/forma-36-codemod/bin/updateDependencies.js
```diff
@@ -0,0 +1,133 @@
...
+        !semverSatisfies(packages.PACKAGES_UPGRADE[pkg], allDeps[pkg]),
+    )
+    .map((pkg) => `${pkg}@${packages.PACKAGES_UPGRADE[pkg]}`);
+
+  // Select package manager based on lock file, fallback to npm if no lock file is found
+  pkgManager = [
+    (fs.existsSync(path.resolve(pkgDirname, 'package-lock.json')) && 'npm') ||
+      (fs.existsSync(path.resolve(pkgDirname, 'yarn.lock')) && 'yarn') ||
+      'npm',
+  ];
```
bgutsol, line 86: Didn't get why u used an array here? It will have only one item always, isn't it?

And another comment is that I would move `(fs.existsSync(path.resolve(pkgDirname, 'yarn.lock')) && 'yarn')` before check with `npm`, since if there both lock files, we should give a priority to `yarn`. We have such a case in our `field-editors`.

packages/forma-36-codemod/bin/updateDependencies.js
```diff
@@ -0,0 +1,133 @@
...
+        name: 'packages',
+        message: 'Do you want the packages to be automatically updated?',
+        default: false,
+      },
+    ])
+    .then((answers) => {
+      if (answers.packages) {
+        const cmd = [
+          getCmd(pkgManager, 'uninstall', removePackages),
+          getCmd(pkgManager, 'install', newPackages),
```
bgutsol, line 120: `pkgManager` is array

packages/forma-36-codemod/bin/inquirer-choices.js
```diff
@@ -11,7 +13,18 @@ const PARSER_CHOICES = [
...
   {
-    name: 'migrate-all-components-to-v4: Run all exising codemods',
+    name:
+      'update-package-json: Updates package.json file with correct packages',
+    value: 'update-package-json',
+  },
+  {
+    name:
+      'update-package-json-all-v4: Updates package.json file with correct packages and run all existing v4 codemods',
+    value: 'update-package-json-all-v4',
```
bgutsol, line 23: It will also remove the imports for v3 css files, right?
Can we add it as well to the `name` (description)? Because for me it was confusing. I had to check the code to be sure that this command will actually do everything.

packages/forma-36-codemod/bin/inquirer-choices.js
```diff
@@ -11,7 +13,18 @@ const PARSER_CHOICES = [
...
+  },
+  {
+    name:
```
burakukula, line 26: maybe we would need to have just `Run all exising codemods` option, without cleanup of the CSS imports

burakukula (commented):
Looks great and it works! 👏
I think we might want to improve the flow in the prompt a bit. Maybe we could go with something like:
 - run all - update package json with new packages and remove old once, remove CSS imports and run all possible codemods for components. Everything in one command.
 - clanup package json - update your package json to use new packages and remove old once + remove CSS imports
 - run codemods for all components - open question: do we need to have an option to run the codemod for components separately? Would this work when you add it to the command when you run the codemod?
 
 One more call out not related to this PR - when you run it in the app bootstrap there is a Workbench component, we would need to add step in the docs to add this package separately.

bgutsol (commented):
Great work 👏

One more general comment would be that script with packages update will not work with our monorepos, since we have to pass `-W` flag to install a dependency. Though I don't think we need to fix it, since it's only for our internal projects. wdyt?

packages/forma-36-codemod/bin/updateDependencies.js
massao: The array is mostly so we don't need to check the data type of the pkgManager on the getOutput function, we just use a forEach, and it will print just one or both package manager commands.

We can move, but I would still fallback to `npm` in case no lock file is found, as it's more probable that the user would have npm installed. WDYT??

packages/forma-36-codemod/bin/updateDependencies.js
massao: This should not be a problem, the `getCmd` function uses the value of pkgManager as an key of the object, and by default if you pass a non primitive value to `obj[nonPrimitive]`, it calls the toString on that value, and `['anyString'].toString()` equals 'anyString'.
But I will make it more verbose to make it clearer anyway

massao:
> Great work 👏
> 
> One more general comment would be that script with packages update will not work with our monorepos, since we have to pass `-W` flag to install a dependency. Though I don't think we need to fix it, since it's only for our internal projects. wdyt?

Yeah, but since the codemod isn't specifically target to internal use, I think making it as less specific as possible is better, also that is one of the reasons I put the confirmation before running the `install/uninstall` through the codemod, so the user can just say `No` and would get a list of what must be removed or installed.

packages/forma-36-codemod/bin/updateDependencies.js
bgutsol: Yes, of course, fallback to `npm` is totally valid.

massao:
@burakukula quick question, since this is for v4 only, should we remove the codemod for the color tokens? It's already updated on the v4, right??

packages/forma-36-codemod/bin/inquirer-choices.js
```diff
@@ -10,6 +12,16 @@ const PARSER_CHOICES = [
...
 
 const SETUP_CHOICES = [
+  {
+    name:
+      'run-all-v4: Update package json with new packages and remove old ones, remove v3 CSS imports and run all possible codemods for components.',
+    value: 'run-all-v4',
+  },
+  {
+    name:
+      'update-package-json: Updates package.json file with correct packages',
```
burakukula, line 22: I think this option should also remove CSS imports? Clean up of the package.json only does not help much I think? WDYT?

massao: true, if we are removing the packages from the packages.json there is no point in having the css still on the code

packages/forma-36-codemod/bin/updateDependencies.js
```diff
@@ -0,0 +1,160 @@
...
+        !allDeps[pkg] ||
+        !semverSatisfies(packages.PACKAGES_UPGRADE[pkg], allDeps[pkg]),
+    )
+    .map((pkg) => `${pkg}@${packages.PACKAGES_UPGRADE[pkg]}`);
+
+  // Select package manager based on lock file, fallback to npm if no lock file is found
+  const pkgManager =
+    (fs.existsSync(path.resolve(pkgDirname, 'yarn.lock')) && 'yarn') ||
+    (fs.existsSync(path.resolve(pkgDirname, 'package-lock.json')) && 'npm') ||
+    'npm';
```
bgutsol, line 112: Minor 💅
you can probably just leave:
```suggestion
    (fs.existsSync(path.resolve(pkgDirname, 'yarn.lock')) && 'yarn') ||
    'npm';
```

bgutsol (approved):
Beautiful 💎

packages/forma-36-codemod/bin/inquirer-choices.js
burakukula: cool! Now just the naming left I think. Maybe "update-setup" or something like this?

thebesson (approved):

PR #1555, opened 2021-11-23
docs(Button): show example for as property

Lelith:
This updates the documentation to show the variation of Buttons as links.

Screenshot at Nov 23 09-37-57
Screenshot at Nov 23 09-45-45

### DIFF
--- a/packages/components/button/Button.mdx
+++ b/packages/components/button/Button.mdx
@@ -67,11 +67,9 @@ Contentful buttons are available in 3 different sizes:
 Contentful buttons can be rendered as a-tag which allows also use them with a href.
 
 ```jsx
-<Stack>
-  <Button as="a" href="#">
-    As a with href tag
-  </Button>
-</Stack>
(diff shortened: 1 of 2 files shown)


packages/components/button/Button.mdx
```diff
@@ -62,6 +62,18 @@ Contentful buttons are available in 3 different sizes:
 </Stack>
 ```
 
+### Rendered as different Elements
+
+Contentful buttons can be rendered as a-tag which allows also use them with a href.
+
+```jsx
+<Stack>
```
burakukula, line 70: Looking good. Just one thing that I am not sure about: why do we need `<Stack>` in this example?

Lelith: haha, yeah i was just lazy copy&pasting from the other examples!

burakukula (approved):

massao (approved):

thebesson (approved):

PR #1558, opened 2021-11-23
docs(migration): add migration guide for List element

massao:


MIGRATION.md
```diff
@@ -1981,6 +1983,38 @@ import { Grid, GridItem } from '@contentful/f36-components';
...
+  <ListItem>ListItem2</ListItem>
+</List>;
+```
+
+to this:
+
+```tsx
+import { List, ListItem } from '@contentful/f36-components';
+
+<List as="ul" className="className" testId="testId">
```
burakukula, line 2012: hmm, maybe it should compound component at some point? 🤔

burakukula (approved):

MIGRATION.md
massao: Yes, I think so, but we can do as a minor change and deprecate the `ListItem` later on, WDYT?

PR #1564, opened 2021-11-24
chore(codemod): update which version of packages the codemod install

massao:
Make the codemod install the latest version, since we are still releasing fixes and changes, it's better than to have a fixed version


burakukula (approved):
Make sense. 🙂

PR #1571, opened 2021-11-29
docs(globalstyles): add documentation for GlobalStyles

massao:
Add documentation page for GlobalStyles core component

### DIFF
--- a/packages/core/src/GlobalStyles/GlobalStyles.mdx
+++ b/packages/core/src/GlobalStyles/GlobalStyles.mdx
@@ -1,0 +1,52 @@
+---
+title: 'GlobalStyles'
+type: 'component'
+slug: /components/globalstyles/
+github: 'https://github.com/contentful/forma-36/tree/forma-v4/packages/core/src/GlobalStyles'
+typescript: ./GlobalStyles.tsx
+v4Doc: true
+---
+
+import { Stack, TextLink } from '@contentful/f36-components';
+import { Props } from '@contentful/f36-docs-utils';
+
+GlobalStyles is a component that set the common style, and also normalize the browser styles.
+
+## Table of contents
+
+- [Import](#import)
+- [Examples](#examples)
+- [Props](#props-api-reference)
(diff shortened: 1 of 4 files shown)


packages/core/src/GlobalStyles/GlobalStyles.mdx
```diff
@@ -0,0 +1,52 @@
...
+    <GlobalStyles />
+    <App />
+  </>,
+  rootElement,
+);
+```
+
+## Props (API reference)
+
+<Props of="GlobalStyles" />
```
bgutsol, line 52: Could you add a description for the `withNormalize` prop type here https://github.com/contentful/forma-36/blob/3a30cdc0deaec8ad8c2c5f443a8bc7a0103c1396/packages/core/src/GlobalStyles/GlobalStyles.tsx#L9
Because now it has an empty description in the Props table.

bgutsol (approved):
All good 👍
One minor comment

burakukula (approved):
Looks good! Maybe just a small thing that we could also do in this PR. in the "getting starter" page we are showing how to use this component. Maybe we could link this docs from there? https://v4.f36.contentful.com/getting-started

packages/core/src/GlobalStyles/GlobalStyles.tsx
```diff
@@ -6,6 +6,9 @@ import tokens from '@contentful/f36-tokens';
 export const GlobalStyles = ({
   withNormalize = true,
 }: {
+  /**
+   * Determines if the browser common style should be normalized
+   */
```
bgutsol, line 11: Your description might sound like it's our custom internal thing. I would make it as clear as possible that it's just commonly used `normalise.css.` So if the users already have one in their project, it's clear for them that they need to pass `false` here.  WDYT?
```suggestion
  /**
   * Determines if normalise.css should be added, [more info](https://github.com/necolas/normalize.css)
   * 
   * @default true
   */
```

PR #1577, opened 2021-11-30
docs(utils): remove TabFocusTrap and add docs for utils

thebesson:
# Purpose of PR

- Removed TabFocusTrap
- Added utils documentation - useKeyboard, Portal, getStringMatch
- Moved the `useArrowKeyNavigation` hook to the `Menu` component folder - it was originally moved there to use it in the `Tabs` component, it's no longer needed in a global package, I think. 

## Note

I wasn't able to make examples work with demos, I'm not sure what the problem is. I checked them in codesandbox, so I know it should work, but on the documentation it looks like plane code.

## PR Checklist

- [x] Tests are passing


packages/components/utils/src/getStringMatch/getStringMatch.mdx
```diff
@@ -0,0 +1,92 @@
...
+import { getStringMatch } from '@contentful/f36-utils';
+```
+
+## Examples
+
+### Basic usage
+
+You pass the base string as a first argument, and a string to match it with as a second argument. The function will return an object containing the match, what comes before the match, and what comes after the match in the base string.
+
+```jsx static=true
```
burakukula, line 33: @thebesson I am not sure if that's intended here, but if you would like to display examples, they can not have `static=true`. Or you add this because they were broken and it was the only way to show them?

thebesson: yes, without `static=true` and without imports there was an error. and imports cannot be used without `static=true`. That's the problem

burakukula: hmmm, and is the `@contentful/f36-utils` installed? maybe it's missing in the dependencies?

packages/components/utils/src/Portal/Portal.mdx
```diff
@@ -0,0 +1,49 @@
...
+title: 'Portal'
+type: 'component'
+status: 'beta'
+slug: /utils/portal/
+typescript: ./Portal.tsx
+---
+
+import { Props } from '@contentful/f36-docs-utils';
+
+`Portal` component can be used to render children in to a DOM node outside of the parent component. It can be useful for rendering components, you dont want to be hidden, clipped by parent component. For example, models, popovers, dropdowns.
```
gui-santos, line 11:
```suggestion
`Portal` component can be used to render children in a DOM node outside of its parent component. It can be useful for rendering components that you don’t want to be clipped by their parent’s component. For example: modals, popovers, dropdown, etc.
```

packages/components/utils/src/Portal/Portal.mdx
```diff
@@ -0,0 +1,49 @@
...
+      style={{
+        backgroundColor: 'gray',
+      }}
+    >
+      Hi! I'm first in the document flow, but will be rendered at the end of the
+      document!
+    </Box>
+  </Portal>
+  Parent component content
+</Box>
```
gui-santos, line 44: I don't understand the Box component here

Shouldn't it be a Fragment or a Flex?

packages/components/utils/src/getStringMatch/getStringMatch.mdx
```diff
@@ -0,0 +1,92 @@
...
+      <Heading as="h4">Results</Heading>
+      <Paragraph element="p">{`{ before: ${before}, match: ${match}, after: ${after} }`}</Paragraph>
+    </Stack>
+  );
+};
+```
+
+### Example with other components
+
+You can find an example of using `getStringMatch` in [`Autocomplete` component documentation](https://v4.f36.contentful.com/components/autocomplete/#highlighting-an-item-with-getstringmatch)
```
gui-santos, line 88:
```suggestion
You can find an example of using `getStringMatch` in [`Autocomplete` component documentation](/components/autocomplete#highlighting-an-item-with-getstringmatch)
```

since the link is in the same website, I think this would be enough, no?

gui-santos (changes requested):
This is looking good, I got just some small suggestions

packages/components/utils/src/getStringMatch/getStringMatch.mdx
thebesson: it's inside the `@contentful/f36-utils`, so I don't know

packages/components/utils/src/Portal/Portal.mdx
```diff
@@ -0,0 +1,48 @@
+---
+title: 'Portal'
+type: 'component'
+status: 'beta'
+slug: /utils/portal/
+typescript: ./Portal.tsx
+---
```
gui-santos, line 7: I notice that you are already using the new format of the documentation in the body of this page
and that's great!

But I think I merged some changes to the plop template while this PR is in review and now I think it would be a good idea to rebase and implement some changes to the new pages

the first one is that in the frontmatter of these pages we need to pass `v4Doc: true` and we still didn't designed where to show the status badge so I think it can be removed for now. This is how the frontmatter will look like:

```suggestion
---
title: 'Portal'
type: 'component'
slug: /utils/portal/
typescript: ./Portal.tsx
v4Doc: true
---
```

Another change is the "Props (API reference)" title
I will add the comment with the changes in that line (to be continued)

packages/components/utils/src/Portal/Portal.mdx
```diff
@@ -0,0 +1,48 @@
...
+    >
+      Hi! I'm first in the document flow, but will be rendered at the end of the
+      document!
+    </Flex>
+  </Portal>
+  Parent component content
+</Flex>
+```
+
+## Props
```
gui-santos, line 46: We now have a "Open in Storybook" button aligned with this title
and because of that we need to use some react components here

the [Accordion doc](https://github.com/contentful/forma-36/blob/forma-v4/packages/components/accordion/Accordion.mdx) has a good example of what needs to be done, but for the Portal page the code looks like this:

```suggestion
<Stack justifyContent="space-between" marginTop="spacing2Xl" marginBottom="spacingL">
  <Heading id="props-api-reference" as="h2" marginTop="none" marginBottom="none">
    Props (API reference)
  </Heading>

  <TextLink
    href="https://v4-f36-storybook.netlify.app/?path=/story/utilities-portal"
    target="_blank"
    alignIcon="end"
    icon={<ExternalLinkIcon />}
  >
    Open in Storybook
  </TextLink>
</Stack>
```

As you can see, it's necessary to import Stack, Heading, and TextLink on the top of the MDX file or this will fail

gui-santos (changes requested):
Sorry @thebesson !
I merged some changes to the template for the new format of documentation and now I think this PR need some updates

But after those are done I think it will be perfect!

packages/components/utils/src/Portal/Portal.mdx
thebesson: ok, no problem

packages/components/utils/src/Portal/Portal.mdx
thebesson: but if there is no storybook page? should I add it?

gui-santos: Good point!

I guess in that case, you can just do `## Props (API reference)` if they have any props

packages/components/utils/src/getStringMatch/getStringMatch.mdx
gui-santos: You will need to pass this to the scope prop of LiveProvider
https://github.com/contentful/forma-36/blob/c08926fd6734339eab43fe98af5fbd947495d7b9/packages/forma-36-website/src/components/ComponentSource.js#L51-L57

as you can see we are already passing our components, icons, and tokens. If you remove icons, for example, you will see that some pages will break and that happens because react-live tries to render your code but does not have the dependencies it needs

thebesson: makes sense 👍

packages/components/utils/src/Portal/Portal.mdx
```diff
@@ -0,0 +1,48 @@
...
+    >
+      Hi! I'm first in the document flow, but will be rendered at the end of the
+      document!
+    </Flex>
+  </Portal>
+  Parent component content
+</Flex>
+```
+
+## Props (API reference)
```
gui-santos, line 46: I think Portal has a story in storybook, no?

Isn't this the same Portal that this doc is about?
https://v4-f36-storybook.netlify.app/?path=/story/utilities-portal

packages/components/utils/src/getStringMatch/getStringMatch.mdx
```diff
@@ -0,0 +1,81 @@
...
+};
+```
+
+### Example with other components
+
+You can find an example of using `getStringMatch` in [`Autocomplete` component documentation](/components/autocomplete#highlighting-an-item-with-getstringmatch)
+
+## Props (API reference)
+
+<Props of="getStringMatch" />
```
gui-santos, line 81: Screenshot 2021-12-07 at 15 06 22

It looks like `react-docgen-typescript` is not generating the props of this function correctly, but I think this bug we should fix in another PR/story

I will write a ticket for that

gui-santos (approved):
Nice! The example works now

I'm still confused about the Portal's storybook I thought this documentation was about the same Portal that we have in storybook (the story seems to be broken btw) but maybe it's not

If it is the same, I suggest adding the storybook button to the page

packages/components/utils/src/Portal/Portal.mdx
thebesson: yes, you are right!

PR #1579, opened 2021-11-30
[BAU-421]: feat: implement new design to the Props visualization component

gui-santos:
# Purpose of PR

Following up the feedback we got around our documentation, we want to change the way we show our props in the documentation:

new design:
Screenshot 2021-11-30 at 14 40 22

The only page with this new design is `components/accordion`

## Props file and folder structure

I moved a lot of the components that were in Props.tsx to their own files and I moved a bunch of files around so we follow our [new folder structure convention](https://github.com/contentful/forma-36/blob/forma-v4/docs/folder-structure.md)

## Props context refactor

When we first implemented this component in `f36-doc-utils` we did not have any convention on how to implement Context, so this component had a very different implementation of a context compared to other components like `Menu`
Since I had to touch the component in this PR, I refactored our implementation to have the same approach we have in the rest of the library
And we that, we can remove `lodash` and `constate` dependencies

## Props section title + "Open in Storybook" link

To achieve this:
Screenshot 2021-11-30 at 14 25 16

It was easier to use React components in our `.mdx` file. I know we don't use this feature of MDX a lot, but it is quite handy to use React in markdown files to achieve better layouts

I also made this change in the template used by plop to create new documentation

And since the storybook url is only used in that link, I removed storybook from the frontmatter of our MDX template

But if you think this is not relevant I can revert the changes

## Required props

Based on @domarku design I was not sure how to show required props, so I did this:

Screenshot 2021-11-30 at 14 49 33

They will always show on the top of the list and in alphabetical order

## Other changes to the list of props

- props are always shown in alphabetical order
- we show all the props now, there is no "Show all" props button
- Values of the props are always shown, for example this is how the `variant` prop of a button will render:

before:
Screenshot 2021-11-30 at 14 53 48
after:
Screenshot 2021-11-30 at 14 54 31

# Things to do in the future

- A lot of props do not have description because they are inherited from polymorphic props or they are native HTML attributes. Maybe in the future we extend the `Props` component to render descriptions to these props
- A lot of values of these props do not help the reader to understand what they should do, this happens because they are the types we get from React or something like that. Maybe we need to think on a way to overwrite these values. For example:
Screenshot 2021-11-30 at 14 59 49

## PR Checklist

- [x] Tests are passing


gui-santos:
@domarku 
I used the typography and color styles that I thought were the "correct" ones
And I used a TextLink for the "Open in Storybook" thing

We can change everything if we need 👌

suevalov:
I like the visual changes, but I don't really agree with 

> we show all the props now, there is no "Show all" props button

Some props are basically inherited props from HTML elements and the properties table becomes huge and inconvenient. People would not need this data in 99% of cases.

suevalov:
@gui-santos you'll need to calculate the list dynamically: https://github.com/contentful/forma-36/pull/1579/files#diff-7a48127f7ae7b9d92b5e13395cd672c07205bfdf6829599c6e654413d4502d91L31-L36

gui-santos:
> @gui-santos you'll need to calculate the list dynamically: https://github.com/contentful/forma-36/pull/1579/files#diff-7a48127f7ae7b9d92b5e13395cd672c07205bfdf6829599c6e654413d4502d91L31-L36

@suevalov I've just pushed a commit with a filtering solution, wdyt?
I based this solution on these lines https://github.com/contentful/forma-36/pull/1579/files#diff-7a48127f7ae7b9d92b5e13395cd672c07205bfdf6829599c6e654413d4502d91L44-L54

Do you still think we should filter out all of the props that have @types/react/index.d.ts as parent? I'm thinking about props like `className` that people use a lot

packages/f36-docs-utils/src/Props/PropsContext.ts
```diff
@@ -0,0 +1,52 @@
...
+  const propsContext = usePropsContext();
+
+  if (!propsContext[componentName]) {
+    return [];
+  }
+
+  const { props } = propsContext[componentName];
+  const allProps = Object.values(props)
+    .sort((a, b) => a.name.localeCompare(b.name))
+    .filter((prop) => !propsToHide.includes(prop.name));
```
gui-santos, line 36: @suevalov this is the filter I mentioned in the other comments

packages/f36-docs-utils/src/Props/PropsContext.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  const { props } = propsContext[componentName];
+  const allProps = Object.values(props)
+    .filter(
+      // We filter out the props that come from @types/react definitions
+      (prop) => !prop.parent?.fileName.includes('@types/react/index.d.ts'),
+    )
+    .filter((prop) => !propsToHide.includes(prop.name))
+    .sort((a, b) => a.name.localeCompare(b.name));
```
gui-santos, line 40: @suevalov this is the filter I mentioned in the other comments

packages/f36-docs-utils/src/Props/PropsContext.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  const { props } = propsContext[componentName];
+  const allProps = Object.values(props)
+    .filter(
+      // We filter out the props that come from @types/react definitions and/or are HTML elements
+      (prop) =>
+        !prop.parent?.fileName.includes('@types/react/index.d.ts') &&
+        !propsToHide.includes(prop.name),
+    )
+    .sort((a, b) => a.name.localeCompare(b.name));
```
gui-santos, line 41: @suevalov this is the filter I mentioned in the other comments

packages/f36-docs-utils/src/Props/PropsContext.ts
```diff
@@ -1,6 +1,10 @@
 import { createContext, useContext } from 'react';
 import type { PropComponentDefinition } from './types';
 
+// List of props that we do not need to show
+// because they are, basically, inherited props from HTML elements or React
+const propsToHide = ['ref', 'key', 'style', 'children'];
```
suevalov, line 6: I'd still leave `children`

suevalov (approved):

PR #1585, opened 2021-12-06
[WIP] code blocks code from separate files

bgutsol:
# Purpose of PR

_Don't merge! It's just an example of implementing it in the Next.js website version._

### Changes:
1. LiveEditor and LivePreview were removed since they didn't work well with the `import` and `export` statements that we have in the examples code. Instead, we have just a static markdown code block and component preview rendered by the MDX itself.
1. Any errors in the examples file will immediately cause the build break in the CLI/website page in the browser, with a decent error description. No hidden issues with broken components previews on production.
1. Now when you change anything in the example files, the website will be rebuilt and hot reloaded in a couple of seconds. No minutes of waiting for rebuilding the whole website with cleaning gatsby cache. 

### As the next steps:
1. Add the `Copy source code` button.
2. Add the `Open in Codesandbox` button. As a replacement for the LiveEditor with LivePreview.
3. I would also add a link to the storybook. Since we usually have the corresponding examples in the storybook for each one in the markdown.

### OR

1. Use the [Sandpack from Codesandbox](https://codesandbox.io/post/sandpack-announcement) that is able to render the whole example source file (with the `import` and `export`). And can handle basically everything that I described above.

**@contentful/team-bauhaus Please, play a little with it, make some errors on the examples source code and let me know WDYT? We can adjust everything as we want.**


bgutsol:
> @bgutsol The question here would be "how performant is Sandpack?" We thought about using CodeSandbox in every page but it is kind of slow

@gui-santos Yes, good question. I'd suggest to do a quick test for one component docs and check how good it is. If it's slow, we will stick to our custom solution.

thebesson:
Just based on my previous experience with Codesandbox, there sometimes can be problems with dependencies versions (when @burakukula made a sandbox for v4 testing, for example), I'm not sure how Sandpack will do on this part?

bgutsol (commented):
The following PR
https://github.com/contentful/forma-36/pull/1652

PR #1592, opened 2021-12-06
chore: pin node and yarn versions 

suevalov:
## Purpose

Pin exact NPM and Node.js versions for tool manager: https://volta.sh/

> With Volta, you can select a Node engine once and then stop worrying about it. You can switch between projects and stop having to manually switch between Nodes. You can install npm package binaries in your toolchain without having to periodically reinstall them or figure out why they’ve stopped working.


package.json
```diff
@@ -141,5 +141,9 @@
   "engines": {
     "node": ">=14.5.0",
     "yarn": ">=1.21.0"
+  },
+  "volta": {
+    "node": "14.5.0",
+    "yarn": "1.21.0"
```
gui-santos, line 147: Could we use version 1.19.0 instead?

version 1.21.0 has some bugs and yarn team already said they won't fix or maintain this version anymore because they are working on v2

https://github.com/yarnpkg/yarn/issues/7807

https://contentful.slack.com/archives/C029BE16E/p1628608153028200?thread_ts=1628607388.028100&cid=C029BE16E

suevalov: sure thing, done

gui-santos (approved):

PR #1597, opened 2021-12-07
fix: fix issue with padding for cards

burakukula:
So there was an issue with paddings for cards reported. 👀 

Screenshot 2021-12-07 at 15 33 56

In general styling for cards seems complex and for some reason, I have a feeling it might bring more issues in the future, if we would like to extend API for example. Maybe we should have a look at the API of cards and think of a slightly different approach. I don't have anything specific in mind, but maybe it's worth having a deeper look. (Forma 36 4.1 ?)
This PR might fix the issue at the moment, I am not sure if I am not forgetting about some use cases 🤔 Maybe some examples are not added in the overviews for each card. dunno. WDYT?

Also, interesting thing brought by @gui-santos , this issue is right now not visible in storybook for EntryCard, which is very confusing, since we can see it in prod, or when you reproduce the component on the website. which brings me to the question: why we couldn’t rely on the storybook for that case?

### DIFF
--- a/packages/components/card/src/asset-card/AssetCard.styles.ts
+++ b/packages/components/card/src/asset-card/AssetCard.styles.ts
@@ -21,9 +21,6 @@ export const getAssetCardStyles = () => {
             : `calc(1rem * (300 / ${tokens.fontBaseDefault}))`,
         padding: 0,
         textAlign: 'center',
-        '[data-card-part="content"]': {
-          padding: 0,
-        },
       };
 
       return css(styles);
--- a/packages/components/card/src/base-card/BaseCard.styles.ts
+++ b/packages/components/card/src/base-card/BaseCard.styles.ts
@@ -7,9 +7,6 @@ export const getBaseCardStyles = () => {
     contentBody: css({
       gridColumn: 'content',
       gridRow: 'content',
-      paddingBottom: tokens.spacingM,
-      paddingLeft: tokens.spacingM,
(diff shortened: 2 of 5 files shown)


gui-santos (approved):

gui-santos:
> In general styling for cards seems complex and for some reason, I have a feeling it might bring more issues in the future, if we would like to extend API for example. Maybe we should have a look at the API of cards and think of a slightly different approach. I don't have anything specific in mind, but maybe it's worth having a deeper look. (Forma 36 4.1 ?) This PR might fix the issue at the moment, I am not sure if I am not forgetting about some use cases 🤔 Maybe some examples are not added in the overviews for each card. dunno. WDYT?

I agree that the cards became too complex and now they are very hard to maintain. The polymorphic strategy we adopted is also not helping the situation, on the contrary.
I'm not sure how much we will be able to change in v4.1, but we could try of course

But, IMO, the solution is in Forma 36 stopping allowing so many ways of doing things so we decrease the number of cases that every component has. The card componentS, for example, could have: padding, no padding, small padding, href, onClick, three dot menu, status badge, etc...

> Also, interesting thing brought by @gui-santos , this issue is right now not visible in storybook for EntryCard, which is very confusing, since we can see it in prod, or when you reproduce the component on the website. which brings me to the question: why we couldn’t rely on the storybook for that case?

It's hard to say because the card component has too much complexity

PR #1601, opened 2021-12-08
fix: restructure menu readme

burakukula:
There is an issue with the broken website for the menu component caused by the example with the menu open by default. in order to fix it, I would need to add a property to not focus on the first item I guess (something like `itemIsNotFocused`), but I don't think we actually need this property. I am also not sure why we need the property `defaultIsOpen`, it is possible to control by using `isOpen` property I guess.

Anyway, since I was already there, I moved the README to the new format.

### DIFF
--- a/packages/components/menu/Menu.mdx
+++ b/packages/components/menu/Menu.mdx
@@ -1,10 +1,14 @@
 ---
 title: 'Menu'
 slug: /components/menu/
-github: 'https://github.com/contentful/forma-36/tree/master/packages/components/menu'
+github: 'https://github.com/contentful/forma-36/tree/forma-v4/packages/components/menu'
 typescript: ./src/Menu.tsx,./src/MenuItem/MenuItem.tsx,./src/MenuList/MenuList.tsx
 v4Doc: true
 ---
+
+import { Heading, Stack, TextLink } from '@contentful/f36-components';
+import { ExternalLinkIcon } from '@contentful/f36-icons';
+import { Props } from '@contentful/f36-docs-utils';
 
 Menu is a component that offers a list of choices to the user, such as a set of actions or links.
 Menu uses Popover component as a base.
@@ -24,10 +28,7 @@ Menu uses Popover component as a base.
   - [With React Router links](#with-react-router-links)
   - [With predefined focused menu item](#with-predefined-focused-menu-item)
   - [With submenu](#with-submenu)
-- [Props](#props)
-  - [Menu props](#menu-props)
-  - [Menu Item props](#menu-item-props)
-  - [Menu List props](#menu-list-props)
+- [Props](#props-api-reference)
 - [Content guidelines](#content-guidelines)
 - [Accessibility](#accessibility)
 - [Help improve this page](#help-improve-this-page)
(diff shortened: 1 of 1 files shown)


packages/components/menu/Menu.mdx
```diff
@@ -1,24 +1,46 @@
 ---
 title: 'Menu'
-type: 'component'
-status: 'stable'
 slug: /components/menu/
 github: 'https://github.com/contentful/forma-36/tree/master/packages/components/menu'
```
gui-santos, line 4: Could you also replace this link?

there is no Menu component in master branch, that's way this link takes the user to a 404

packages/components/menu/Menu.mdx
```diff
@@ -331,19 +332,6 @@ But you can redefine them on any submenu by passing those props directly to that
...
-## Accessibility
-
-- When the menu is opened, focus is moved to the first menu item. If you set `autoFocus`to `false`, it will not move the focus.
-- When the menu is closed, focus returned back to the trigger element.
-- You can use arrow keys (`ArrowUp` and `ArrowDown`) to navigate through menu items.
-- When the menu is open, click on `Esc` key will close the popover. If you set `closeOnEsc` to `false`, it will not close.
-- When the menu is open, click outside menu or blurring out will close the menu. If you set `closeOnBlur` to `false`, it will not close.
-- All the necessary a11y attributes for `Menu.List`, `Menu.Trigger` and `Menu.Item` are provided.
-
 ## Props
```
gui-santos, line 335: Did you use the plop templates to make this convertion?

This title here is supposed to be some react code because we need both a title and a textlink aligned

The code would be something like this:
```suggestion
<Stack justifyContent="space-between" marginTop="spacing2Xl" marginBottom="spacingL">
  <Heading id="props-api-reference" as="h2" marginTop="none" marginBottom="none">
    Props (API reference)
  </Heading>

  <TextLink
    href="https://v4-f36-storybook.netlify.app/?path=/story/components-menu"
    target="_blank"
    alignIcon="end"
    icon={<ExternalLinkIcon />}
  >
    Open in Storybook
  </TextLink>
</Stack>
```

But of course you also need to import these components on the top of this MDX files or the page will break
[Here is an example of mdx using this approach](https://github.com/contentful/forma-36/blob/forma-v4/packages/components/accordion/Accordion.mdx)

gui-santos (changes requested):
Looking good

I'm just suggesting to adopt the latest way of the doc with the storybook link
you can check how to do it in the plop template and in the accordion documentation page

packages/components/menu/Menu.mdx
burakukula: I didn't use the plop template 😬 I will. thanks!

massao (approved):

bgutsol:
> There is an issue with the broken website for the menu component caused by the example with the menu open by default. in order to fix it, I would need to add a property to not focus on the first item I guess (something like `itemIsNotFocused`), but I don't think we actually need this property. I am also not sure why we need the property `defaultIsOpen`, it is possible to control by using `isOpen` property I guess.
> 
> Anyway, since I was already there, I moved the README to the new format.

Yeah, we definitely don't need `itemIsNotFocused` prop. `defaultIsOpen` is handy when you don't want to make Menu fully controlled component but need to have it open on the first render. I would leave the example but make it `static=true`

gui-santos (approved):
🍰

bgutsol (approved):

PR #1606, opened 2021-12-09
feat: use next-mdx-remote for mdx parsing

suevalov:
# Purpose of PR

Uses https://github.com/hashicorp/next-mdx-remote for loading MDX content.
Now website picks up not only `components` but also all other MDX content.


packages/website/utils/content.ts
```diff
@@ -86,18 +44,56 @@ async function fetchFiles(targetPath: string) {
...
+      // return only those mdx files which have 'slug' set in meta section
+      return data.frontMatter.data.slug;
+    });
   return mdxData;
 }
 
-export { geltAllMDX, getPageBySlug };
+const allMdxSources = [
+  path.resolve('../../packages/components'),
+  path.resolve('../../packages/forma-36-website/src/content'),
```
gui-santos, line 67: I think we don't need this

We will move all this pages to next js package in the future
I think it will be easier to maintain them this way

wdyt?

packages/website/utils/content.ts
```diff
@@ -86,18 +44,57 @@ async function fetchFiles(targetPath: string) {
...
-  // Saving a map of slugs and the filepath to their mdx in a JSON file
-  const mdxFilepathBySlug = mdxData.reduce((acc, mdx) => {
-    return { ...acc, [mdx.slug]: mdx.filepath };
-  }, {});
-  const data = JSON.stringify(mdxFilepathBySlug, null, 2);
-  fs.writeFileSync('utils/mdxFilepathBySlug.json', data);
+  for (const path of paths) {
+    const newFiles = await fetchFiles(path);
+    allMDX = allMDX.concat(newFiles);
+  }
```
gui-santos, line 53: I think I didn't understand this change

Why do we need to initiate an empty array and then concat the fetched files?
Instead of just fetching the files directly like it was before

packages/website/utils/content.ts
```diff
@@ -86,18 +44,57 @@ async function fetchFiles(targetPath: string) {
...
 
+  const mdxData = allMDX
+    .map((filepath) => getDataByFilepath(filepath))
+    .filter((data) => {
+      // return only those mdx files which have 'slug' set in meta section
+      return data.frontMatter.data.slug;
+    });
```
gui-santos, line 60: Is it possible to move this filter to `getDataByFilepath`?

The way it is now, it looks like we are getting unnecessary data in the `map` method

packages/website/utils/content.ts
```diff
@@ -86,18 +44,57 @@ async function fetchFiles(targetPath: string) {
...
 
-export { geltAllMDX, getPageBySlug };
+const allMdxSources = [
+  path.resolve('../../packages/components'),
+  path.resolve('../../packages/core'),
+  path.resolve('../../packages/forma-36-website/src/content'),
+];
+
+async function getMdxSourceBySlug(slug: string[]) {
+  const mdxFiles = await getAllMdx(allMdxSources);
```
gui-santos, line 71: Doesn't this mean that now every time I go to a page I will get allMdx?

In the previous approach with the json file, we didn't need to do that
when you go to a page you just read the json to discover where the file of your slug is

it seems unnecessary to call getAllMdx for each page

packages/website/utils/content.ts
```diff
@@ -86,18 +44,57 @@ async function fetchFiles(targetPath: string) {
...
+  const joinedSlug = slug.join('/');
+  return mdxFiles.find(
+    (item) =>
+      item.frontMatter.data.slug === `/${joinedSlug}/` ||
+      item.frontMatter.data.slug === `/${joinedSlug}`,
+  );
```
gui-santos, line 77: Why do we need this? I thought the slugs where already  a string like "/components/button"

packages/website/pages/[...slug].tsx
```diff
@@ -0,0 +1,81 @@
...
+      remarkPlugins: [require('remark-code-titles')],
+      rehypePlugins: [mdxPrism, rehypeSlug, rehypeAutolinkHeadings],
+    },
+    scope: frontMatter.data,
+  });
+
+  return {
+    props: {
+      source: mdxSource,
+      frontMatter: frontMatter.data,
```
Comment (gui-santos) [inline packages/website/pages/[...slug].tsx:70]: That's what I meant in the other comment about destructuring the result of matter()

here we would avoid this `frontMatter.data`

packages/website/utils/content.ts
```diff
@@ -86,18 +44,57 @@ async function fetchFiles(targetPath: string) {
...
+    })
+    .map((slug) => {
+      const sanitizedSlug = slug.split('/').filter((item) => item);
+
+      return {
+        params: {
+          slug: sanitizedSlug,
+        },
+      };
+    });
```
gui-santos, line 95: Isn't it possible to combine both map methods?

```suggestion
    .map((page) => {
      const { slug } page.frontMatter.data;
      const sanitizedSlug = slug.split('/').filter((item) => item);

      return {
        params: {
          slug: sanitizedSlug,
        },
      };
    });
```

gui-santos (changes requested):
This is looking good!

I got a few questions and suggestions :)

packages/website/pages/[...slug].tsx
```diff
@@ -0,0 +1,81 @@
...
+};
+
+type ComponentPageProps = {
+  source: MDXRemoteSerializeResult;
+  frontMatter: {
+    title: string;
+  };
+};
+
+export default function ComponentPage(props: ComponentPageProps) {
```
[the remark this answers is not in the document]
Comment (suevalov) [reply packages/website/pages/[...slug].tsx:31]: because slug can have more than one segment: `/components/skeleton/skeleton-image/`

`[...slug]` means that slug is an array. in this case `["components", "skeleton", "skeleton-image"]`

packages/website/pages/index.tsx
```diff
@@ -50,16 +40,20 @@ const Home: NextPage<HomeProps> = (props) => {
...
-    if (a.slug > b.slug) {
-      return 1;
-    }
-    return 0;
-  });
+  const mdxData = await getMdxPaths();
+  const sortedComponentsList = mdxData
+    .map((data) => {
+      return { slug: data.params.slug.join('/') };
+    })
```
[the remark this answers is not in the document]
suevalov: yep! answered above

packages/website/utils/content.ts
suevalov: do you mean that will have a copy of them now? I'm fine with that but it works this way for now.

packages/website/utils/content.ts
```diff
@@ -2,55 +2,13 @@ import fs from 'fs';
...
-}
-
-function getSlugByFilepath(filepath: string) {
+function getDataByFilepath(filepath: string) {
   const fileContents = fs.readFileSync(filepath, 'utf8');
-  const { data } = matter(fileContents);
-
-  const slugArray = data.slug.split('/');
-  const slug = slugArray[slugArray.length - 2];
+  const frontMatter = matter(fileContents);
```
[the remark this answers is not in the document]
suevalov: `frontMatter` is passed down to props later, but I don't think this is important. I can do it either way here.

packages/website/utils/content.ts
suevalov: it's fetching from several folders, not from one only

packages/website/utils/content.ts
suevalov: `getDataByFilepath` cannot filter things out because it's used to transform data in `map` method

packages/website/utils/content.ts
suevalov: It's called once during the build and doesn't affect performance. No runtime implications either.

packages/website/utils/content.ts
suevalov: sometimes we have trailing slash and sometimes we don't :/

packages/website/utils/content.ts
suevalov: updated

packages/website/pages/[...slug].tsx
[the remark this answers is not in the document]
Comment (gui-santos) [reply packages/website/pages/[...slug].tsx:31]: ohhhhhhhhh I totally missed that! Thx

packages/website/utils/content.ts
gui-santos: yep! We are even going to remove some of these pages as we talked in power hour

This also reminded me that we will need to separate them from the sidemenu code
otherwise we won't have "getting started guide" on the top of the menu always

packages/website/utils/content.ts
gui-santos: Ahhh you mean that now it fetches from `/components` and `/core`

got it!

And we need to concat because both "sources" will return an array of files, right?

suevalov: yep

packages/website/package.json
```diff
@@ -15,16 +15,20 @@
...
     "next": "^12.0.7",
+    "next-mdx-remote": "^3.0.8",
     "react": "16.14.0",
     "react-dom": "16.14.0",
     "react-live": "^2.3.0",
+    "rehype-autolink-headings": "^5.0.1",
+    "rehype-slug": "^4.0.1",
     "remark": "^14.0.2",
-    "remark-html": "^15.0.1"
+    "remark-code-titles": "^0.1.1"
```
gui-santos, line 27: oh I totally missed this in my first review
how much do we trust in a package in v0.1.1? hahahaha

gui-santos: Btw, I used remark-html because it was recommend by next.js team in this blog post
https://nextjs.org/blog/markdown#data-source

packages/website/utils/content.ts
gui-santos: Ah! of course! 
And you can't return a null or something like that
Got it

In this case, I have another argument in favor of destructuring that frontMatter

**please tell me if I'm miss understanding this**

In getDataByFilepath, you return:

```
{
    content: fileContents,
    frontMatter: { data },
    filepath,
};
```

right?

and then here you need to do that filter, but if you destructure that you could do:
```suggestion
    .filter((data) => {
      // return only those mdx files which have 'slug' set in meta section
      return data.frontMatter.slug;
    });
```

even better:
```suggestion
    .filter(({ frontMatter }) => {
      // return only those mdx files which have 'slug' set in meta section
      return slug;
    });
```

no? Am I seeing things?

gui-santos (approved):
Thanks for explaining everything!

I understand how things are working now, great stuff!

I only posted some nitpicky suggestion, no blockers

massao (approved):

PR #1608, opened 2021-12-09
fix(modal): add props to modal examples and add focus state for tabs

thebesson:
# Purpose of PR

- add 'shouldCloseOnOverlayClick' and 'shouldCloseOnEscapePress' props to Modal examples
- Add visual focused state for Tab (focus state was missing on the active state)

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/packages/components/tabs/src/Tabs.styles.ts
+++ b/packages/components/tabs/src/Tabs.styles.ts
@@ -20,7 +20,7 @@ const styles = {
     border: 'none',
     cursor: 'pointer',
     outline: 'none',
-    '&:focus-visible &:focus': {
+    '&:focus-visible': {
       boxShadow: tokens.glowPrimary,
     },
     '&:before': {
--- a/packages/components/tabs/src/Tabs.styles.ts
+++ b/packages/components/tabs/src/Tabs.styles.ts
@@ -20,7 +20,7 @@ const styles = {
     border: 'none',
     cursor: 'pointer',
     outline: 'none',
-    '&:focus-visible': {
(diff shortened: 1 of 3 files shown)


burakukula:
@thebesson I have a question regarding focus state. I guess the solution should work in a way, that focus is only visually visible when you use a keyboard to navigate? Is that correct? I thought that this is why we have `focus-visible` added there. but seems like it doesn’t work this way now. 
If we want to have a standard focus, I guess we could just use `:focus` instead of `&:focus-visible, &[aria-selected="true"]
If we could like to have focus visible only on keyboard navigation, we would need to fix it I guess 🤔 I don’t understand why it's not working now. Or am I missing something?

thebesson:
@burakukula you are right, changed it

packages/components/tabs/src/Tabs.styles.ts
```diff
@@ -20,7 +20,7 @@ const styles = {
     border: 'none',
     cursor: 'pointer',
     outline: 'none',
-    '&:focus-visible': {
+    '&:focus-visible &:focus': {
```
bgutsol, line 23: We can't have both a `&:focus-visible` and a `&:focus` here. We should keep only one of them to make it work. 
- with `&:focus` the `boxShadow` will appear always when the item is in the focus state. Either it gets focused by keyboard or by mouse click. 
- with `&:focus-visible` the `box shadow` will appear only when the item is focused by a keyboard.

I would vote for `&:focus-visible` since it is how native `button` for example works. And it looks better 💅

bgutsol (approved):

PR #1617, opened 2021-12-13
chore: add turborepo

suevalov:
# Purpose of PR

Use https://turborepo.com/ for build command.

### DIFF
--- a/packages/website/utils/content.ts
+++ b/packages/website/utils/content.ts
@@ -64,7 +64,7 @@ async function getAllMdx(paths: string[]) {
 const allMdxSources = [
   path.resolve('../../packages/components'),
   path.resolve('../../packages/core'),
-  path.resolve('../../packages/forma-36-website/src/content'),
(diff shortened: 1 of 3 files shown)


packages/website/utils/content.ts
```diff
@@ -64,7 +64,7 @@ async function getAllMdx(paths: string[]) {
 const allMdxSources = [
   path.resolve('../../packages/components'),
   path.resolve('../../packages/core'),
-  path.resolve('../../packages/forma-36-website/src/content'),
+  // path.resolve('../../packages/forma-36-website/src/content'),
```
suevalov, line 67: We temporary disable picking up content from Gatsby website, because each item will be copied later one by one.

gui-santos: Maybe add this comment to the code, wdyt?

gui-santos (approved):
Code looks good to me

massao (approved):

PR #1619, opened 2021-12-13
[BAU-489] feat: implement a layout grid for the next js website

gui-santos:
# Purpose of PR

Implements this Grid for the new website. Should be easy to place future components and pages inside those areas 👌 

Screenshot 2021-12-13 at 15 51 42

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/packages/website/pages/_app.tsx
+++ b/packages/website/pages/_app.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import type { AppProps } from 'next/app';
+import Head from 'next/head';
 import { css } from 'emotion';
 import { GlobalStyles, Grid } from '@contentful/f36-components';
 
@@ -9,7 +10,13 @@ const styles = {
     overflow: 'hidden',
   }),
   gridItem: css({
+    display: 'flex',
+    flexDirection: 'column',
     overflow: 'auto',
+  }),
+  footer: css({
+    position: 'sticky',
+    top: '100vh',
   }),
 };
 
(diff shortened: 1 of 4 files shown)


packages/website/pages/_app.tsx
```diff
@@ -1,12 +1,58 @@
...
+      <Head>
+        <title>Forma 36 - The Contentful Design System</title>
+        <meta
+          name="description"
+          content="Forma 36 - The Contentful Design System"
+        />
+        <link rel="icon" href="/favicon.png" />
+      </Head>
+
+      <Grid
```
suevalov, line 38: I think it would be good to move it to `website/components/Layout.tsx`

gui-santos: What exactly? The Grid wrapping the Grid.Items?

I thought this way would be easier to understand what is happening
easier than:
```
<Layout>
  <Grid.Item>Topbar</Grid.Item>
  <Grid.Item>Sidemenu</Grid.Item>
  <Grid.Item>Content</Grid.Item>
</Layout>
```

gui-santos: Or are you suggesting to also move `Grid.Item` to separate components like `Topbar`, `Sidemenu`, etc.?

suevalov: I'd only leave `<Layout><Component {...pageProps} /></Layout>`

suevalov: Because at some point we'll probably have per page layouts: https://nextjs.org/docs/basic-features/layouts#per-page-layouts

suevalov: As this website grows out of only `Forma36` documentation and will have `field-editors` section and so on.

gui-santos: I don't get it, this is the same structure we have right now
the content of the pages have different layouts you can see it by comparing the Homepage, "Getting started guide", and any other component page

They use a "Layout" component but those pages still fall into the "content" area

This PR is just making sure that we will render the things around those content even when we change pages

And if we really feel the necessity to use `getLayout` in the future we do it in the future
no need to solve a problem we do not have now

gui-santos: Let me push this abstraction and you tell me If I understood the suggestion correctly or if I'm missing something

suevalov (approved):

PR #1620, opened 2021-12-13
feat: Sandpack POC

suevalov:
# Purpose of PR

An example of using Sandpack. 

**Benefits**

* Everything that codesandbox can do, but in our documentation.
* Each example can be opened as Codesandox
* Loads and compiles fast 

**Cons**

* Looks like there's no way to link a local copy of Forma. Basically the examples will be run with **the latest published** version of Forma.

### DIFF
--- a/packages/website/components/SandpackRenderer.tsx
+++ b/packages/website/components/SandpackRenderer.tsx
@@ -1,0 +1,50 @@
+import React from 'react';
+import { Sandpack } from '@codesandbox/sandpack-react';
+
+const indexFile = `import React, { StrictMode } from "react";
+import ReactDOM from "react-dom";
+import { GlobalStyles } from "@contentful/f36-components";
+import "./styles.css";
+
+import App from "./App";
+
+const rootElement = document.getElementById("root");
+ReactDOM.render(
+  <StrictMode>
+    <GlobalStyles />
+    <App />
+  </StrictMode>,
+  rootElement
+);`;
+
+const stylesFile = `body {
+  padding: 10px;
+}`;
+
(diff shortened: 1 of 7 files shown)


gui-santos:
> **Cons**
> 
> * Looks like there's no way to link a local copy of Forma. Basically the examples will be run with **the latest published** version of Forma.

Isn't this already a big problem?
It means that when we open a PR the preview website won't have the correct version and the examples may fail
previews of the website won't be reliable basically

packages/website/components/SandpackRenderer.tsx
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import { Sandpack } from '@codesandbox/sandpack-react';
+
+const indexFile = `import React, { StrictMode } from "react";
+import ReactDOM from "react-dom";
+import { GlobalStyles } from "@contentful/f36-components";
+import "./styles.css";
```
gui-santos, line 7: I think this file does not exist

gui-santos (changes requested):
It looks cool, but the only Con issue worries me a lot

packages/website/components/SandpackRenderer.tsx
suevalov: it exists when but not visible for edit

suevalov:
@gui-santos yeah, I think it's a big problem so my suggestion would be not to proceed with the solution

bgutsol (commented):
Looks great and works fine. Love it! 👏 
Curious if there will be any performance drawdowns if we have more than 10 examples on the same page, like the` Menu` component has

PR #1623, opened 2021-12-14
feat: nextjs props types context

bgutsol:
# Purpose of PR

1. Logic for providing context with components props types info.
2. Props filtering logic was moved from `PropsContext` to the script where we gather and prepare all those props. To decrease the size of data transferred to the client.

### DIFF
--- a/packages/website/utils/propsMeta.ts
+++ b/packages/website/utils/propsMeta.ts
@@ -1,0 +1,84 @@
+import path from 'path';
+import {
+  PropComponentDefinition,
+  PropDefinition,
+} from '@contentful/f36-docs-utils';
+const docgen = require('react-docgen-typescript');
+
+// List of props that we do not need to show
+// because they are, basically, inherited props from HTML elements or React
+const PROPS_TO_HIDE = ['ref', 'key', 'style'];
+
+const getPreparedComponent = (
+  component: PropComponentDefinition,
+): PropComponentDefinition => {
+  const resultComponent: PropComponentDefinition = {
+    description: component.description,
+    displayName: component.displayName,
+    props: {},
+  };
+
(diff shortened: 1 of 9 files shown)


suevalov (approved):

massao (approved):

packages/website/utils/propsMeta.ts
```diff
@@ -0,0 +1,84 @@
...
+      },
+    );
+
+    const components: PropComponentDefinition[] =
+      tsConfigParser.parse(sourcePath) || [];
+
+    const result: { [key: string]: PropComponentDefinition } = {};
+    components.map((component) => {
+      result[component.displayName] = getPreparedComponent(component);
+    });
```
gui-santos, line 57: Wdyt about using reduce here?

```suggestion
    const result: { [key: string]: PropComponentDefinition } = components.reduce((acc, component) => {
      return { ...acc, [component.displayName]: getPreparedComponent(component) };
    }, {});
```

packages/website/utils/propsMeta.ts
```diff
@@ -0,0 +1,84 @@
...
+    return {};
+  }
+}
+
+function getPropsMetadata(filePath: string, sourcesPaths?: string) {
+  if (!sourcesPaths) {
+    return null;
+  }
+
+  const propsMetadata: { [key: string]: PropComponentDefinition } = {};
```
gui-santos, line 72: You used this `{ [key: string]: PropComponentDefinition }` twice in the file

I suggest defining an interface in this file so we have it in one place only

gui-santos (approved):
LGTM

I made two suggestions but they are not blockers, only small improvements to the code

PR #1625, opened 2021-12-14
feat: styling content nextjs and copy header component

suevalov:
# Purpose of PR

Start of styling MDX pages for Next.js website.
Copies existing styles for the Header component.

### DIFF
--- a/packages/website/pages/[...slug].tsx
+++ b/packages/website/pages/[...slug].tsx
@@ -2,7 +2,6 @@ import React from 'react';
 import { useRouter } from 'next/router';
 import ErrorPage from 'next/error';
 import Head from 'next/head';
-import Link from 'next/link';
 import rehypeAutolinkHeadings from 'rehype-autolink-headings';
 import rehypeSlug from 'rehype-slug';
 import { serialize } from 'next-mdx-remote/serialize';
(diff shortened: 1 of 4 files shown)


gui-santos (approved):

packages/website/pages/[...slug].tsx
```diff
@@ -40,12 +39,10 @@ export default function ComponentPage(props: ComponentPageProps) {
...
         <title>Forma 36 - {props.frontMatter.title}</title>
       </Head>
-
       <article className={styles.root}>
-        <Link href="/">Back</Link>
         <h1>{props.frontMatter.title}</h1>
         <div>
-          <MDXRemote {...props.source} />
+          <MdxRenderer source={props.source} />
         </div>
```
Comment (gui-santos) [inline packages/website/pages/[...slug].tsx:46]: I aproved already but I forgot that we need to change this

We need to remove this `div` wrapping the MDXRemote/Renderer

And the root styles needs to be:
```
const styles = {
  root: css({
    display: 'flex',
    flexDirection: 'column',
  }),
};
```

Or the content will overlapp the footer of the website

Comment (suevalov) [reply packages/website/pages/[...slug].tsx:46]: @gui-santos I'll do this in the follow-up PR, thanks for the suggestion

PR #1626, opened 2021-12-14
feat: more styling of PageContent

suevalov:
# Purpose of PR

More styling of page content.
Copied Footer and Sidebar styles.
Copied images and icons.


packages/website/components/PageContent.tsx
```diff
@@ -0,0 +1,162 @@
...
+    width: 960px;
+    margin: 0 auto;
+    padding-top: ${tokens.spacing2Xl};
+  `,
+  header: css`
+    width: 960px;
+    margin: 0 auto;
+    padding-bottom: ${tokens.spacingXl};
+    margin-bottom: ${tokens.spacingXl};
+    border-bottom: 1px solid ${tokens.gray300};
```
gui-santos, line 28: the documentation pages with `v4Doc: true` should not have this border

In those pages we should only show the DisplayText

packages/website/pages/index.tsx
```diff
@@ -23,7 +19,7 @@ function Home({ componentsList }: HomeProps) {
           );
         })}
       </ul>
-    </>
+    </main>
```
gui-santos, line 22: I think adding a `main` tag here will duplicate it unnecessarily since this component is rendered inside PageContent (line 158) which means that in the DOM this is how it will show:

```html
<main>
  <main>
    <h1>Welcome...</h1>
    <ul>....</ul>
  </main>
</main>
```

gui-santos (changes requested):
This is looking good, but I think v4Doc pages won't look the same as they look in the current version and I there is an extra main tag

packages/website/components/PageContent.tsx
suevalov: shouldn't we delete `v4Doc` attributes now? Why not to have border-bottom after all titles?

packages/website/pages/index.tsx
suevalov: `index.tsx` is not rendered inside `PageContent`. `PageContent` is used only for MDX based pages

packages/website/components/PageContent.tsx
gui-santos: Another way to make this question is: why having them? hahahah

Anyway, it was just how @domarku designed them
if he is happy re-introducing the border-bottom, I will be fine as well

gui-santos: and the idea of the `v4Doc` attribute is because we are in this transition period where some docs follow the new format and some docs don't

suevalov: I think we can take each page separately lately, we have tasks for them, right?

packages/website/pages/index.tsx
gui-santos: Screenshot 2021-12-15 at 14 52 23

I've just took this picture in vercel link

packages/website/components/PageContent.tsx
gui-santos: We have tasks to adapt each page to the new format yes

packages/website/pages/index.tsx
gui-santos: There is also a main tag inside `Layout` (that's why I didn't want to abstract the Grid to another component)

the `Grid.Item` wrapping pages has `as="main"`

gui-santos (changes requested):
Everything looks good, but I think we should not have duplicated `main` tag

and the Title of each page should be `h1` because every page needs to have an h1 for SEO reasons

packages/website/components/PageContent.tsx
```diff
@@ -0,0 +1,162 @@
...
+function PageHeader(props: {
+  title?: string;
+  storybook?: string;
+  github?: string;
+  status?: string;
+}) {
+  return (
+    <header className={styles.header}>
+      {props.title && (
+        <DisplayText marginBottom="spacingL">{props.title}</DisplayText>
```
gui-santos, line 72:
```suggestion
        <DisplayText as="h1" marginBottom="spacingL">{props.title}</DisplayText>
```

gui-santos (changes requested):

packages/website/pages/index.tsx
gui-santos: And I forgot to mention, the advantage of removing this main here and keeping the one in Layout is that we make sure the title of each page is going to be inside of the main tag and screen readers and SEO like when the main content of a website has an `h1` inside of it

suevalov: got it, changed to article in `PostContent` and inde.tsx will be taken care of when working on Home screen

packages/components/collapse/src/Collapse.tsx
```diff
@@ -17,17 +17,13 @@ export interface CollapseProps extends CommonProps {
    * string for additional classNames
    */
   className?: string;
-  /**
-   * testId
-   */
-  'data-test-id'?: string;
```
gui-santos, line 23: 👀 

thanks for removing it

gui-santos (approved):

PR #1629, opened 2021-12-15
fix: fix Button to elipsis with longer content

burakukula:
So we would like to solve issues like this one:

image (8) (1)

We can add elipsis, so we get something like:

Screenshot 2021-12-15 at 10 41 10

I added quite random max-width of 250px. Maybe @domarku @fabe @damann you have an opinion around how big buttons should be when it comes to width?

The only corner case is the case where we have a loading, it seems strange to have this 2x dots:
Screenshot 2021-12-15 at 10 38 19


suevalov (approved):

thebesson (approved):

domarku:
An ellipsis at the end and `max-width: 250px` sounds good! What do you think about a max-width of `240px` to be somewhat consistent with the `medium` width input field?

For the loading state I think it would be better to replace the whole content of a loading button with just the `spinner`

PR #1631, opened 2021-12-15
fix: truncate entry card description

suevalov:
# Purpose of PR

Truncate long descriptions in EntryCard using CSS.


gui-santos (approved):
Oh that's what `isTruncated` does hahaha
#TIL

PR #1632, opened 2021-12-15
chore: small adjustements for Next.js

suevalov:
# Purpose of PR

* Moves all MDX components to a separate folder, so we can separate them from the components that are used to build the website itself.
* Updates metadata to match the Gatsby version


burakukula (approved):
LGTM

PR #1633, opened 2021-12-15
[BAU-492] docs: add documentation for spacing and typography tokens

gui-santos:
# Purpose of PR

Screenshot 2021-12-15 at 19 20 56

Screenshot 2021-12-15 at 19 22 32

## PR Checklist

- [ ] Tests are passing


packages/website/components/GlobalStyles.tsx
```diff
@@ -63,6 +63,10 @@ export const GlobalStyles = () => {
           padding: 0;
         }
 
+        img {
```
suevalov, line 66: I'm not sure if this style should go to global styles. I'd say it should go to PageContent only

gui-santos: when I tried to add this to MdxRenderer, next js asked me to use next/Image
but it has height and width as mandatory props
https://nextjs.org/docs/api-reference/next/image#required-props

wdyt?

gui-santos: the height I will have to add it manually

packages/website/components/MdxRenderer.tsx
```diff
@@ -41,12 +42,15 @@ const components = {
...
-  td: (props) => <TableCell {...props} />,
+  thead: (props) => <Table.Head {...props} />,
+  tbody: (props) => <Table.Body {...props} />,
+  tr: (props) => <Table.Row {...props} />,
+  th: (props) => <Table.Cell style={{ textAlign: 'left' }} {...props} />,
+  td: (props) => <Table.Cell {...props} />,
+  ...f36Components,
   ...MdxComponents,
+  Props,
+  PropsHeading,
```
bgutsol, line 53: We already spreading `MdxComponents` with `Props` and `PropsHeading`

gui-santos: Oops!

packages/website/components/GlobalStyles.tsx
bgutsol: Not sure I get it right. If we don't know width and height then we should use a simple img not `next/Image` What exect image we want to import with `next/Image`?

gui-santos: the typography page has two images

you can check my change in MdxRenderer.tsx now, I had to add to eslint-disable rules
and one of them is there because next/js warns that I'm using a `img` tag and they recommend using their component

but their component needs width, height, and src
and when I put, for example, `width="100%" height={340}` for the first image
the image get distorted 
`layout="fill"` makes it the size of the screen
and so on

so I decided to use `img` and disable the rules

bgutsol (approved):

thebesson (approved):

PR #1634, opened 2021-12-16
Fix asset image fit

thebesson:
# Purpose of PR

This fix should prevent the image on Asset preview from being distorted. 

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/packages/components/asset/src/Asset.styles.ts
+++ b/packages/components/asset/src/Asset.styles.ts
@@ -15,6 +15,7 @@ export function getAssetStyles() {
       height: '100%',
       maxWidth: '100%',
       maxHeight: '100%',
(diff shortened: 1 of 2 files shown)


packages/components/asset/src/Asset.styles.ts
```diff
@@ -15,6 +15,7 @@ export function getAssetStyles() {
       height: '100%',
       maxWidth: '100%',
       maxHeight: '100%',
+      objectFit: 'contain',
```
suevalov, line 18: I think it looks good, but it would be nice to see a story with an image that would be distorted without this fix. So we have a test for this case and so we can compare prev/now

suevalov (approved):

PR #1635, opened 2021-12-16
fix: allow to pass usePortal to Autocomplete

suevalov:
# Purpose of PR

We need an ability to pass truthy `usePortal` value in Autocomplete component as discovered while integrating v4 in Launch application.


gui-santos (approved):
Code looks good

But I suggest testing it with a screen reader because I remember I had to add `usePortal={false}` for a11y reasons

suevalov:
@gui-santos okay, I'll take a look 👍🏼

PR #1638, opened 2021-12-20
fix: fix dragHandle issue

burakukula:
Temporary fix to the issue that we face in the field editors.
The library used by field editors (react-sortable-hoc) is canceling events for buttons for some reason (https://github.com/clauderic/react-sortable-hoc/blob/d94ba3cc67cfc7d6d460b585e7723bdb50015e53/src/SortableContainer/defaultShouldCancelStart.js).
We should change the implementation on the field editors at some point. We will have a look at it while working on the keyboard nav for dragHandle (inspiration from user interface, where it's already done)

I added the example in the storybook,but skipped the website, we don't want to encourage anyone to use react-sortable-hoc, but maybe it's still worth keeping the example in the storybook?

### DIFF
--- a/packages/components/drag-handle/src/DragHandle.tsx
+++ b/packages/components/drag-handle/src/DragHandle.tsx
@@ -30,6 +30,10 @@ export type DragHandleInternalProps = CommonProps & {
    * is for screen readers only
    */
   label: string;
+  /**
+   * Set type button for div element
+   */
+  type?: string;
 };
 
 export type DragHandleProps = PropsWithHTMLElement<
--- a/packages/components/drag-handle/src/DragHandle.tsx
+++ b/packages/components/drag-handle/src/DragHandle.tsx
@@ -6,7 +6,7 @@ import React, {
   MouseEventHandler,
 } from 'react';
 import { cx } from 'emotion';
-import type { PropsWithHTMLDivElement } from '@contentful/f36-core';
+import type { PropsWithHTMLElement } from '@contentful/f36-core';
 import type { CommonProps, ExpandProps } from '@contentful/f36-core';
 import { DragIcon } from '@contentful/f36-icons';
 import { getStyles } from './DragHandle.styles';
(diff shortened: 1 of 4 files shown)


packages/components/drag-handle/src/DragHandle.tsx
```diff
@@ -107,7 +107,9 @@ export const DragHandle = forwardRef<
     );
 
     return (
-      <button
+      <div
```
suevalov, line 110: 🤯

suevalov (approved):

packages/components/drag-handle/src/DragHandle.tsx
burakukula: 🤷‍♀️

packages/components/drag-handle/src/DragHandle.tsx
```diff
@@ -6,7 +6,7 @@ import React, {
   MouseEventHandler,
 } from 'react';
 import { cx } from 'emotion';
-import type { PropsWithHTMLElement } from '@contentful/f36-core';
+import type { PropsWithHTMLDivElement } from '@contentful/f36-core';
```
bgutsol, line 9:
```suggestion
import type { PropsWithHTMLElement } from '@contentful/f36-core';
```

packages/components/drag-handle/src/DragHandle.tsx
```diff
@@ -32,13 +32,13 @@ export type DragHandleInternalProps = CommonProps & {
   label: string;
 };
 
-export type DragHandleProps = PropsWithHTMLElement<
+export type DragHandleProps = PropsWithHTMLDivElement<
```
bgutsol, line 35:
```suggestion
export type DragHandleProps = PropsWithHTMLElement<
```

bgutsol (commented):
Small typo

packages/components/drag-handle/src/DragHandle.tsx
bgutsol: I'm curious why u didn't get an TS error here

burakukula: fixed, yeah, dunno why it didn't catch it. Maybe I need to restart. Anyway, fixed. Thx!

bgutsol (approved):

PR #1644, opened 2021-12-20
fix: proper inline styles for next.js

suevalov:
# Purpose of PR

Inline critical styles for Next.js website.


bgutsol (approved):
Nice! 
Love how easy it is to implement such not trivial optimizations.

PR #1649, opened 2021-12-21
feat: automatic generation of table of contents

suevalov:
# Purpose of PR
Screenshot 2021-12-21 at 12 53 45

Table of Content is generated automatically for each MDX file.


MIGRATION.md
```diff
@@ -1,98 +1,5 @@
 # Migration
 
-## Table of contents
```
bgutsol, line 3: If I'm not mistaken, this file can be viewed only within Github, there is no related page on the website. So I'd keep the TOC here.

packages/website/components/TableOfContent.tsx
```diff
@@ -0,0 +1,73 @@
...
+    marginBottom: tokens.spacingL,
+    padding: tokens.spacingL,
+    marginTop: `calc(-1 * ${tokens.spacingL})`,
+    '@media (min-width: 1400px)': {
+      width: '350px',
+      marginLeft: '50px',
+      marginRight: '-100px',
+      display: 'inline-block',
+      background: 'white',
+      float: 'right',
```
bgutsol, line 20: float ❤️
At first glance on the website, I couldn't get how you did that magic with layout. But then I found that lovely `float` 😄 

I guess, in near future, we gonna make this TOC sticky, so it stays with scroll, right?

bgutsol (approved):
👏

MIGRATION.md
suevalov: ah, good point

packages/website/components/TableOfContent.tsx
suevalov: yeah, we can make it sticky later

suevalov: old school tricks 👴🏼 🪄

burakukula: WOW 😮

burakukula (approved):
👏

PR #1652, opened 2021-12-21
feat: next code blocks

bgutsol:
# Purpose of PR

Move docs examples to separate files.

### DIFF
--- a/packages/components/menu/Menu.mdx
+++ b/packages/components/menu/Menu.mdx
@@ -38,19 +38,59 @@ You can use following components to build a menu:
 ### Basic usage
 
 ```tsx file=examples/BasicMenu.tsx
-
+<Menu>
+  <Menu.Trigger>
+    <IconButton
+      variant="secondary"
+      icon={<MenuIcon />}
+      aria-label="toggle menu"
+    />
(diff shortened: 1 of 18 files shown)


suevalov (commented):
It looks as for me. Should we wait till the end of the week when we remove Gatsby website and point `https://v4.f36.contentful.com/` to Vercel?

When it's done, we can start moving examples component by component.

bgutsol:
I'd say that the PR is not blocked by pages migration. Since I did it in a way that doesn’t break the build process for other not migrated pages.

suevalov:
@bgutsol yeah, but if we merge it now, all examples in Menu will not be visible in `v4.f36.contentful.com`, right?

bgutsol:
> @bgutsol yeah, but if we merge it now, all examples in Menu will not be visible in `v4.f36.contentful.com`, right?

ah, yes, you are right! But as a temporary solution, we can keep both, like:
image
So for nextJs it will get example from a file, but for Gatsby from codeblock. @suevalov  WDYT?

suevalov:
@bgutsol works for me, but we need to tell the rest of the team

suevalov (approved):

PR #1654, opened 2021-12-22
feat: add Segment to nextjs

burakukula:


packages/website/pages/_document.tsx
```diff
@@ -1,11 +1,26 @@
...
+    const opts = {
+      apiKey: SEGMENT_KEY,
+      page: true,
+    };
+
+    if (NODE_ENV === 'production') {
+      return snippet.max(opts);
+    }
+
+    return '';
```
burakukula, line 21: I guess we need to always return string since `__html` expects that?

packages/website/pages/_document.tsx
```diff
@@ -20,6 +35,7 @@ export default class MyDocument extends Document<{
             data-emotion-css={this.props.ids.join(' ')}
             dangerouslySetInnerHTML={{ __html: this.props.css }}
           />
+          <script dangerouslySetInnerHTML={{ __html: this.renderSnipet() }} />
```
suevalov, line 38: can we do `async` here?

suevalov: I'd rather not render `<script>` at all if `__html` is empty

burakukula: sure, make sense.

packages/website/pages/_document.tsx
```diff
@@ -20,6 +35,9 @@ export default class MyDocument extends Document<{
             data-emotion-css={this.props.ids.join(' ')}
             dangerouslySetInnerHTML={{ __html: this.props.css }}
           />
+          {this.renderSnipet() && (
```
suevalov, line 38: call render `this.renderSnippet()` once and put it into a variable, use this variable to check the case and assign it to `__html`

suevalov (approved):

PR #1665, opened 2021-12-28
feat: playground page

suevalov:
# Purpose of PR

Implements a `/playground` page where users can play around with components in a fullscreen mode.

Supported features:
* sharing code as URLs with your peers
* opening your experiments as CodeSandbox
* each example in Documentation can be opened in Playground with one click 
   * only `Menu` works property for now
   * We need to extract all examples to separate files to make it work
* playground link is placed in navbar, but it will be moved after a visual refresh


burakukula (approved):
💪
Maybe we could iterate a bit on the layout in the playground. Footer is very big, maybe we could make it smaller and 100%, but we can do it later if we decide so.

suevalov:
@burakukula yep, I agree. Layout and design can be changed in a separate PR.

PR #1668, opened 2021-12-29
feat: Box & Flex documentation

suevalov:
# Purpose of PR

Updated `Box` documentation:

* Improved examples
* Examples are moved to separate files (and checked that they work in the Playground)
* Fixed broken links


packages/core/src/Box/examples/BoxPolymorphicExample.tsx
```diff
@@ -0,0 +1,33 @@
...
+        as="a"
+        display="inline"
+        href="https://contentful.com"
+        target="_blank"
+        marginRight="spacingL"
+      >
+        I am native link
+      </Box>
+      <Box
+        as={TextLink}
```
burakukula, line 20: I don't think we should encourage using examples like this one. In the documentation, we should show only examples that we actually recommend, not all the possible options. Why someone should use `Box` to create a `TextLink` instead of using `TextLink`?

packages/core/src/Box/examples/BoxPolymorphicExample.tsx
```diff
@@ -0,0 +1,33 @@
...
+        display="inline"
+        href="https://contentful.com"
+        target="_blank"
+        marginRight="spacingL"
+      >
+        I am link as TextLink component
+      </Box>
+      <Box as={Button} marginRight="spacingL">
```
burakukula, line 28: Same here, I don't understand why we recommend using `Box` like this.

packages/core/src/Box/examples/BoxPolymorphicExample.tsx
suevalov: good point, updated

packages/core/src/Box/examples/BoxPolymorphicExample.tsx
suevalov: good point, updated

packages/core/src/Flex/examples/FlexExample.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import { Flex } from '@contentful/f36-components';
+import tokens from '@contentful/f36-tokens';
+
+export default function FlexExample() {
+  const ColorBox = ({ backgroundColor, children }) => (
+    <Flex
+      as="div"
```
burakukula, line 8: I guess this is default?

burakukula (approved):
🙌

thebesson (approved):

PR #1674, opened 2022-01-03
feat: improve CopyButton examples

suevalov:
# Purpose of PR

Improves `CopyButton` documentation.

### DIFF
--- a/packages/components/copybutton/CopyButton.mdx
+++ b/packages/components/copybutton/CopyButton.mdx
@@ -5,57 +5,43 @@ status: 'stable'
 section: 'buttonComponents'
 slug: /components/copy-button/
 github: 'https://github.com/contentful/forma-36/tree/forma-v4/packages/components/copybutton'
-storybook: 'https://v4-f36-storybook.netlify.app/?path=/story/components-copybutton--default'
 typescript: ./src/CopyButton.tsx
 ---
 
-import { Props } from '@contentful/f36-docs-utils';
-import { CopyButton } from '@contentful/f36-copybutton';
+The `CopyButton` is a styled button that copies text into the user's clipboard.
 
-The CopyButton is a styled button that copies text into the user's clipboard. This is useful in combination with for example tables or inputs that show values that the user is likely to use elsewhere, such as automatically generated passwords.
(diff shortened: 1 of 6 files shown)


thebesson (approved):

packages/components/copybutton/CopyButton.mdx
```diff
@@ -5,57 +5,43 @@ status: 'stable'
...
-// import { CopyButton } from '@contentful/f36-copybutton';
+```jsx file=examples/CopyButtonExample.tsx
 
-<CopyButton value="Lorem Ipsum" />
 ```
 
-An example with a custom tooltip:
+### Changing the tooltip
+
+You can modify tooltip text and its placement by using the following properties:
```
gui-santos, line 29: Which following properties? the ones in the example? If yes I suggest:

```suggestion
You can modify tooltip text and its placement by using the `tooltipCopiedText`, `tooltipText`, and `tooltipProps` properties
```

And on that note, I think it would be probably better to explain `tooltipProps` in detail in a separate example
you could even add a link to the tooltip page and send the user there for further explanations

I would suggest splitting the example

packages/components/copybutton/CopyButton.mdx
```diff
@@ -5,57 +5,43 @@ status: 'stable'
...
-  tooltipText="Copy this text to clipboard"
-/>
 ```
 
-An example with custom tooltip placement and custom aria-label.
+### Real-world example
 
-```jsx
-// import { CopyButton } from '@contentful/f36-copybutton';
+```jsx file=examples/CopyButtonWithInputExample.tsx
```
gui-santos, line 37: Thanks for adding a real world example

gui-santos (approved):

packages/components/copybutton/CopyButton.mdx
suevalov: 👍🏼

packages/components/copybutton/CopyButton.mdx
suevalov: I updated the copy in this paragraph and added a link to `tooltip` documentation without splitting the example into 2 separate ones.

PR #1683, opened 2022-01-03
feat: IconButton examples

suevalov:
# Purpose of PR

Improves `IconButton` examples and documentation.

### DIFF
--- a/packages/components/button/src/IconButton/README.mdx
+++ b/packages/components/button/src/IconButton/README.mdx
@@ -28,17 +28,13 @@ You need to pass an [Icon component](../components/icon) you want to render to t
 
 ### As toolbar actions
 
-Usually `IconButton` should be used in a toolbars, action bars and in all other space restricted containers where rendering `Button` is not possible.
+In toolbars, action bars, and in all other space restricted containers we suggest using `IconButton` instead of `Button` component.
 
-We suggest to use [Tooltip](../components/tooltip) as a wrapper of `IconButton` to help user with understanding button's purpose.
+For these cases, it would be also beneficial to use [Tooltip](../components/tooltip) as a wrapper of `IconButton` to help users with understanding the button's purpose.
 
 ```jsx file=../../examples/IconButtonToolbarExample.tsx
 
 ```
-
-## Accessibility
-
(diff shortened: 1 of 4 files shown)


packages/components/button/src/IconButton/README.mdx
```diff
@@ -5,33 +5,44 @@ status: 'stable'
...
-// import { PreviewIcon } from '@contentful/f36-icons';
+You need to pass an [Icon component](../components/icon) you want to render to the `icon` prop, you may also provide `variant` and `size` to the icon.
+
+```jsx file=../../examples/IconButtonExample.tsx
+
+```
+
+### As toolbar actions
+
+Usually `IconButton` should be used in a toolbars, action bars and in all other space restricted containers where rendering `Button` is not possible.
```
burakukula, line 31: it suggests that this is the most common use case. I don't think that's true. Maybe we could do something like:

"In toolbars, action bars, and in all other space restricted containers we suggest using `IconButton` component. For those cases, it would be also beneficial to use [Tooltip](../components/tooltip) as a wrapper of `IconButton` to help users with understanding the button's purpose."

suevalov: A great suggestion

packages/components/button/src/IconButton/README.mdx
```diff
@@ -5,33 +5,44 @@ status: 'stable'
...
-/>
 ```
 
-## Props
+## Accessibility
 
-<Props of="IconButton" />
+We enforce `aria-label` property for `IconButton` component to ensure that these buttons are fully accessible for screen readers.
+
+## Props (API reference)
```
bgutsol, line 43: According to our docs pages mockup, the `Props` section should be before `Accessibility` 
image. 
Don't know how important it is, but at least it's good for consistency.

packages/components/button/src/IconButton/README.mdx
suevalov: done

packages/components/button/src/IconButton/README.mdx
suevalov: 🦅 👁️ - updated!

bgutsol (approved):

burakukula (approved):

PR #1684, opened 2022-01-03
feat: better ButtonGroup docs and fixed spaced implementation

suevalov:
# Purpose of PR

- Better documentation for the `ButtonGroup` component with more real-world examples.
- Fixes wrong `spaced` implementation (there was 2 * spacing space between items) by using `Stack` component in the implementation.
- Fixes broken `withDivider` implementation

### DIFF
--- a/packages/components/button/examples/ButtonGroupCollapsedExample.tsx
+++ b/packages/components/button/examples/ButtonGroupCollapsedExample.tsx
@@ -9,7 +9,7 @@ import { ChevronDownIcon, PlusIcon } from '@contentful/f36-icons';
 
 export default function ButtonGroupCollapsedExample() {
   return (
-    <Stack>
+    <Stack flexDirection="column">
       <ButtonGroup>
         <Button variant="secondary" size="small">
           Save
--- a/packages/components/button/examples/ButtonGroupCollapsedExample.tsx
+++ b/packages/components/button/examples/ButtonGroupCollapsedExample.tsx
@@ -1,0 +1,34 @@
+import React from 'react';
+import {
+  Stack,
+  ButtonGroup,
+  Button,
+  IconButton,
+} from '@contentful/f36-components';
+import { ChevronDownIcon, PlusIcon } from '@contentful/f36-icons';
+
+export default function ButtonGroupCollapsedExample() {
+  return (
+    <Stack>
+      <ButtonGroup>
+        <Button variant="secondary" size="small">
+          Save
+        </Button>
+        <IconButton
+          variant="secondary"
+          size="small"
+          aria-label="Add more"
+          icon={<PlusIcon />}
+        />
+      </ButtonGroup>
+      <ButtonGroup withDivider>
+        <Button variant="positive">Publish changes</Button>
+        <IconButton
+          variant="positive"
+          aria-label="Open dropdown"
+          icon={<ChevronDownIcon />}
+        />
+      </ButtonGroup>
+    </Stack>
+  );
+}
--- a/packages/components/button/src/ButtonGroup/README.mdx
+++ b/packages/components/button/src/ButtonGroup/README.mdx
@@ -5,61 +5,44 @@ status: 'stable'
 section: 'buttonComponents'
 slug: /components/button-group/
 github: 'https://github.com/contentful/forma-36/tree/forma-v4/packages/components/button/src/ButtonGroup'
-storybook: 'https://v4-f36-storybook.netlify.app/?path=/story/components-button-buttongroup--basic'
 typescript: ./ButtonGroup.tsx
 ---
 
-import { Props } from '@contentful/f36-docs-utils';
-import { Button } from '@contentful/f36-button';
-import { ChevronDownIcon, PlusIcon } from '@contentful/f36-icons';
+`ButtonGroup` should be used to group buttons whose actions are related, with an option to collapse them together or split them with an equal amount of free space.
 
-## Code examples
+## Import
 
-Example of usage:
-
-```jsx
-// import { ButtonGroup, Button } from '@contentful/f36-button';
-
-<ButtonGroup testId="group-button">
-  <Button variant="secondary">Button</Button>
-  <Button variant="secondary">Button</Button>
-  <Button
-    variant="positive"
-    aria-label="Open dropdown"
-    startIcon={<ChevronDownIcon />}
-  />
-</ButtonGroup>
+```js static=true
+import { ButtonGroup } from '@contentful/f36-components';
 ```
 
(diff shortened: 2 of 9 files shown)


packages/components/button/examples/ButtonGroupCollapsedExample.tsx
```diff
@@ -0,0 +1,34 @@
...
+  Stack,
+  ButtonGroup,
+  Button,
+  IconButton,
+} from '@contentful/f36-components';
+import { ChevronDownIcon, PlusIcon } from '@contentful/f36-icons';
+
+export default function ButtonGroupCollapsedExample() {
+  return (
+    <Stack>
```
gui-santos, line 12:
```suggestion
    <Stack flexDirection="column">
```

Since this two groups are different examples (I'm assuming this because of the text before them)
I would split them or show them in a column otherwise it feels to me that they are part of the same "thing"

gui-santos (approved):
Only a small suggestion

packages/components/button/src/ButtonGroup/README.mdx
```diff
@@ -5,61 +5,44 @@ status: 'stable'
...
-    variant="positive"
-    aria-label="Open dropdown"
-    startIcon={<ChevronDownIcon />}
-  />
-</ButtonGroup>
-```
+### Collapsed
 
-It's possible to add a divider for buttons that don't have border:
+The most common example is to group primary button with a secondary button or with a button that opens a dropdown with secondary actions.
```
burakukula, line 23: hmm, we say here that it is a common case to group primary and secondary button, but we don't show it in the example. Is it really that common btw?

packages/components/button/src/ButtonGroup/README.mdx
```diff
@@ -5,61 +5,44 @@ status: 'stable'
...
-  <Button variant="secondary">Button</Button>
-  <Button variant="secondary">Button</Button>
-  <Button
-    variant="positive"
-    aria-label="Open dropdown"
-    startIcon={<ChevronDownIcon />}
-  />
-</ButtonGroup>
-```
+### Collapsed
```
burakukula, line 21: Why we call it collapsed? 🤔

packages/components/button/src/ButtonGroup/README.mdx
```diff
@@ -5,61 +5,44 @@ status: 'stable'
...
-    Button
-  </Button>
-  <Button variant="positive" startIcon={<ChevronDownIcon />} />
-</ButtonGroup>
 ```
 
-Example of Group Button with spaced variant
+### Spaced
+
+Instead of collapsing buttons all together you can use `variant="spaced"` to make them spaced and control an empty space with `spacing` property.
```
burakukula, line 33: hmm, I don't get why we say collapsing buttons. but maybe it is my English hehehe. I always thought collapsing might be relevant to something like an accordion, menu, something that opens and closes.

burakukula: We could say just grouped (without space, if we want to be very specific)? dunno. maybe it's just me.

packages/components/button/src/ButtonGroup/README.mdx
```diff
@@ -5,61 +5,44 @@ status: 'stable'
...
-```
+### Collapsed
 
-It's possible to add a divider for buttons that don't have border:
+The most common example is to group primary button with a secondary button or with a button that opens a dropdown with secondary actions.
 
-```jsx
-// import { ButtonGroup, Button } from '@contentful/f36-button';
-// import { ChevronDownIcon, PlusIcon } from '@contentful/f36-icons';
+Also, it is possible to add a divider for buttons that don't have border.
```
burakukula, line 25: What do you mean by that? which buttons don’t have borders?

packages/components/button/examples/ButtonGroupCollapsedExample.tsx
suevalov: 👍🏼

packages/components/button/src/ButtonGroup/README.mdx
suevalov: I'm going to rephrase it to `One of the common use cases...` and we actually show this in the example below, aren't we?

packages/components/button/src/ButtonGroup/README.mdx
suevalov: Because this variant is called `collapsed` 🤔

packages/components/button/src/ButtonGroup/README.mdx
suevalov: Hmm... I'm questioning it now too, but I guess it's too late to change the name before the release. The better name would be `merged` or something like this. WDYT?

packages/components/button/src/ButtonGroup/README.mdx
suevalov: updated with an explanation

packages/components/button/src/ButtonGroup/README.mdx
burakukula: yeah, that sounds better to me. Collapsed... hmmm. why we called the prop collapsed? 😄

packages/components/button/src/ButtonGroup/README.mdx
```diff
@@ -5,61 +5,44 @@ status: 'stable'
...
 slug: /components/button-group/
 github: 'https://github.com/contentful/forma-36/tree/forma-v4/packages/components/button/src/ButtonGroup'
-storybook: 'https://v4-f36-storybook.netlify.app/?path=/story/components-button-buttongroup--basic'
 typescript: ./ButtonGroup.tsx
 ---
 
-import { Props } from '@contentful/f36-docs-utils';
-import { Button } from '@contentful/f36-button';
-import { ChevronDownIcon, PlusIcon } from '@contentful/f36-icons';
+`ButtonGroup` should be used to group buttons whose actions are related, with an option to collapse them together or split them with an equal amount of free space.
```
burakukula, line 11:
```suggestion
`ButtonGroup` should be used to group buttons whose actions are related, with an option to merge them together or split them with an equal amount of free space.
```

burakukula (approved):

packages/components/button/src/ButtonGroup/README.mdx
suevalov: I have no idea 😢 but I think it's too late to make a breaking change. Should we create a new variant `merged` and keep `collapsed` as a synonym of `merged` for backward compatibility? 

We can remove `collapsed` in v5.

thebesson (approved):

PR #1686, opened 2022-01-04
fix: fix examples for layout components

burakukula:


packages/core/src/Box/examples/BoxPolymorphicExample.tsx
```diff
@@ -7,15 +7,6 @@ export default function BoxSpacingExample() {
       <Box as="span" display="inline" marginRight="spacingL">
         I am span
       </Box>
-      <Box
-        as="a"
```
burakukula, line 11: Too controversial 😄. We shouldn’t recommend using Box to create links, users should use `TextLink`.

packages/core/src/Flex/examples/FlexExample.tsx
```diff
@@ -1,28 +1,40 @@
...
-  );
-
   return (
     <Flex justifyContent="space-between" alignItems="center">
-      <ColorBox backgroundColor={tokens.green500}>Green</ColorBox>
-      <ColorBox backgroundColor={tokens.red500}>Red</ColorBox>
-      <ColorBox backgroundColor={tokens.blue500}>Blue</ColorBox>
+      <Box
+        justifyContent="center"
+        alignItems="center"
```
bgutsol, line 10: `Box` doesn’t have flex styles, same comment for all other places
```suggestion
```

bgutsol (changes requested):

packages/core/src/Flex/examples/FlexExample.tsx
suevalov: Exactly, I'd keep ColorBox component here, but we can use `<div>` and center text differentely

packages/core/src/Stack/examples/VerticalStackExample.tsx
```diff
@@ -1,28 +1,40 @@
 import React from 'react';
-import { Stack, Flex } from '@contentful/f36-components';
+import { Stack, Box } from '@contentful/f36-components';
 import tokens from '@contentful/f36-tokens';
 
 export default function VerticalStackExample() {
-  const ColorBox = ({ backgroundColor, children }) => (
```
suevalov, line 6: I think having a component here is actually good and it allows to understand `<Stack>` code easier without stealing too much attention to children.

suevalov (changes requested):

packages/core/src/Box/examples/BoxPolymorphicExample.tsx
suevalov: Maybe we need to replace this example with something else to keep 3 items in a row?

packages/core/src/Flex/examples/FlexExample.tsx
burakukula: sure, overlooked. sorry. 
But I think we don't need text inside boxes at all.

packages/core/src/Stack/examples/VerticalStackExample.tsx
burakukula: Well, I don't really agree. The only code to understand here is `<Stack flexDirection="column" spacing="spacingS">`. In other examples, for Flex or Grid, we use Boxes directly. I think it is actually not needed to have this abstraction here and it makes it more consistent with other docs.

packages/core/src/Box/examples/BoxPolymorphicExample.tsx
burakukula: Why do we need 3 items?

suevalov: just to have more than 2 examples :)

suevalov (approved):

PR #1688, opened 2022-01-04
feat: updates ToggleButton documentation and improves ToggleButton props

suevalov:
# Purpose of PR

Improves `ToggleButton` documentation and updates new properties in ToggleButton props

### DIFF
--- a/packages/components/button/examples/ToggleButtonBasicExample.tsx
+++ b/packages/components/button/examples/ToggleButtonBasicExample.tsx
@@ -1,0 +1,18 @@
+import React, { useState } from 'react';
+import { ToggleButton } from '@contentful/f36-components';
+import { PreviewIcon } from '@contentful/f36-icons';
+
+export default function ToggleButtonBasicExample() {
+  const [isActive, setIsActive] = useState(false);
+  return (
+    <ToggleButton
+      isActive={isActive}
+      onToggle={() => {
+        setIsActive(!isActive);
+      }}
+      icon={<PreviewIcon />}
+    >
+      Toggle Button
+    </ToggleButton>
+  );
+}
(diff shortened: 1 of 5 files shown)


thebesson (approved):

packages/components/button/examples/ToggleButtonBasicExample.tsx
```diff
@@ -0,0 +1,18 @@
...
+  return (
+    <ToggleButton
+      isActive={isActive}
+      onToggle={() => {
+        setIsActive(!isActive);
+      }}
+      icon={<PreviewIcon />}
+    >
+      Toggle Button
+    </ToggleButton>
```
gui-santos, line 16: Is it possible to find a real ToggleButton in the webapp and use its label instead of a generic "Toggle Button"?

packages/components/button/examples/ToggleButtonStatesExample.tsx
```diff
@@ -0,0 +1,15 @@
...
+export default function ToggleButtonStatesExample() {
+  return (
+    <Stack>
+      <ToggleButton isActive onToggle={() => {}}>
+        Active Toggle Button
+      </ToggleButton>
+      <ToggleButton isDisabled onToggle={() => {}}>
+        Disabled Toggle Button
+      </ToggleButton>
+    </Stack>
```
gui-santos, line 13: Same comment as the one in the first example

suevalov: Updated

packages/components/button/examples/ToggleButtonBasicExample.tsx
suevalov: All web app examples are rather weird, so I came up with another example of `Like` button here

PR #1691, opened 2022-01-04
feat: display text documentation

suevalov:
# Purpose of PR

Updates `DisplayText` documentation


gui-santos (approved):
I guess my only comment is the same I did for @bgutsol in [#1698](https://github.com/contentful/forma-36/pull/1698)

Instead of using "I love Forma 36", we could use the cases that we have in the webapp

I found a couple of uses of DisplayText in there

I will approve because I know the team wants to push these tasks, but IMO we needed more real cases in the examples

suevalov:
@gui-santos I'd disagree that we need specific examples for Typography components. IMO, using the same text for all Typography examples helps to compare typography elements, one to another, their relative size, etc.

domarku:
I would agree with @gui-santos on using real world copy and compositions in **every** component example.
We may reduce the need for the user to compare between typography components with real examples.

For DisplayText, I found some in the web app:

`
<DisplayText>Speedy delivery…</DisplayText>
<Subheading>We’re setting up your space. This could take up to a minute.</Subheading>
`

gui-santos:
Suggestion of a "Composition" example with DisplayText in some context

```jsx
<Flex flexDirection="column" alignItems="center">
  <DisplayText marginBottom="spacingL">
        Nice one!{' '}
        <span role="img" aria-label="Shopping bag">
          🛍
        </span>
  </DisplayText>
  <Paragraph>
    You successfully purchased the Medium space My Blog.
  </Paragraph>
</Flex>
```

It should show this
Screenshot 2022-01-05 at 11 53 17

This is what the webapp shows the user when they purchase a new space

Again, I don't think this is a blocker and you do not need to add this example or anything, just trying to follow the feedback

packages/components/typography/examples/DisplayTextExample.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { Flex, DisplayText } from '@contentful/f36-components';
+
+export default function DefaultTextExample() {
```
bgutsol, line 4:
```suggestion
export default function DisplayTextExample() {
```

packages/components/typography/examples/DisplayTextMarginsExample.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { Flex, DisplayText } from '@contentful/f36-components';
+
+export default function DefaultTextMarginsExample() {
```
bgutsol, line 4:
```suggestion
export default function DisplayTextMarginsExample() {
```

bgutsol (approved):

thebesson (approved):

PR #1695, opened 2022-01-04
feat: reshuffle folders structure

burakukula:
Main changes:
 - moving test files next to Component itself
 - renaming `styles` files to follow `Component.styles.ts`
 - renaming foldes from `component-name` to `ComponentName`
 
 If the package contains more than one component (Button, Badge, Forms, Cards), I added Folders for each component (where there is a README for the component) + REAMDE.md, which contains a list of components from this package.


packages/components/badge/src/index.ts
```diff
@@ -1,5 +1,5 @@
-export { Badge } from './Badge';
-export type { BadgeProps } from './Badge';
-export { EntityStatusBadge } from './EntityStatusBadge';
-export type { EntityStatusBadgeProps } from './EntityStatusBadge';
+export { Badge } from './Badge/Badge';
+export type { BadgeProps } from './Badge/Badge';
```
gui-santos, line 2: I thought these two would stay in the src folder, no?

burakukula: but we have 2 separate README files for them. how do you think we should handle those cases?

packages/components/accordion/src/Accordion.test.tsx
```diff
@@ -3,7 +3,7 @@ import { render, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { axe } from '@/scripts/test/axeHelper';
 
-import { Accordion } from '../src';
+import { Accordion } from '.';
```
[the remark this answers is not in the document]
burakukula: hmm, we actually want to import from index I guess, to be able to use all the compound components.

packages/components/badge/src/index.ts
gui-santos: Ah I sent this messages before we talked this morning

they are not relevant anymore :)

packages/components/accordion/src/Accordion.test.tsx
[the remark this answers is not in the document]
gui-santos: uhmmm I thought this rule of importing from index was only for code outside of the package and importing it

But I guess it does not matter, it's just my personal preference to not have imports from "dots" hahahah

gui-santos (approved):
🏆 🏆 🏆 🏆 🏆 🏆 🏆 

thanks for your patience 🙏

suevalov (approved):

PR #1716, opened 2022-01-06
docs(date-time-utils): update documentation

thebesson:
# Purpose of PR

Updates documentation and removes examples for un-exported utils


suevalov (approved):

bgutsol (approved):
The only comment is that I would split and move import examples to the corresponding sections. So the structure would be:

- formatDateAndTime
  - import
  - examples

- formatMachineReadableDateTime
  - import
  - examples
  
But not a blocker

PR #1733, opened 2022-01-10
fix(Collapse): enrich props type and add display none for collapsed state

Chaoste:
# Purpose of PR

1. In order to extend my Collapse element with properties like ID to link it via ARIA to my DragHandle button, we should allow passing through native properties to the DOM node.

2. When the element is collapsed (and the transition is done), it adds `display: none` to the region. Thereby, the elements within are not reachable with `Tab` anymore. The style is removed when the expanding transition starts.

- feat: pass native props like ID to Collapse node
- fix: hide collapsed content with display none

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/components/collapse/src/Collapse.tsx
+++ b/packages/components/collapse/src/Collapse.tsx
@@ -19,10 +19,7 @@ interface CollapseInternalProps extends CommonProps {
   className?: string;
 }
 
-export type CollapseProps = PropsWithHTMLElement<
-  CollapseInternalProps,
-  'div'
->;
+export type CollapseProps = PropsWithHTMLElement<CollapseInternalProps, 'div'>;
 
 export const Collapse = ({
   children,
@@ -51,30 +48,26 @@ export const Collapse = ({
 
     const handleTransitionEnd = () => {
       if (current) {
-        current.style.height = 'auto';
+        if (isExpanded) {
+          current.style.removeProperty('height');
+        } else {
+          current.style.setProperty('display', 'none');
+        }
       }
     };
 
     if (current) {
-      if (isExpanded) {
-        current.addEventListener('transitionend', handleTransitionEnd);
+      current.addEventListener('transitionend', handleTransitionEnd);
+      const fromHeight = isExpanded ? '0px' : getPanelContentHeight();
+      const toHeight = isExpanded ? getPanelContentHeight() : '0px';
+      requestAnimationFrame(function () {
+        current.style.removeProperty('display');
+        current.style.setProperty('height', fromHeight);
 
         requestAnimationFrame(function () {
-          current.style.height = '0px';
-
-          requestAnimationFrame(function () {
-            current.style.height = getPanelContentHeight();
-          });
+          current.style.setProperty('height', toHeight);
         });
-      } else {
-        requestAnimationFrame(function () {
-          current.style.height = getPanelContentHeight();
-
-          requestAnimationFrame(function () {
-            current.style.height = '0px';
-          });
-        });
-      }
+      });
     }
 
     return () => {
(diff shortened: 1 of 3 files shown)


packages/components/collapse/src/Collapse.tsx
```diff
@@ -46,30 +48,26 @@ export const Collapse = ({
 
     const handleTransitionEnd = () => {
       if (current) {
-        current.style.height = 'auto';
+        if (isExpanded) {
+          current.style.removeProperty('height');
```
bgutsol, line 52: Since we have `height: 0` in the default styles, when you remove the height prop, it will be collapsed (back with 0 height).

packages/components/collapse/src/Collapse.tsx
```diff
@@ -46,30 +48,26 @@ export const Collapse = ({
...
-      if (isExpanded) {
-        current.addEventListener('transitionend', handleTransitionEnd);
-
-        requestAnimationFrame(function () {
-          current.style.height = '0px';
+      current.addEventListener('transitionend', handleTransitionEnd);
+      const fromHeight = isExpanded ? '0px' : getPanelContentHeight();
+      const toHeight = isExpanded ? getPanelContentHeight() : '0px';
+      requestAnimationFrame(function () {
+        current.style.removeProperty('display');
```
bgutsol, line 64: We have to remove property `display: none` before we get PanelContentHeight, because otherwise, it will be always `0`

bgutsol (commented):
Since we are removing the height prop in the end of the transition can we get rid of this complication with `px` to `rem` conversion in the `getPanelContentHeight` fn?

packages/components/collapse/src/Collapse.tsx
Chaoste: Oh, I totally oversaw the styles file. I'm adding auto height back to the event listener.

Chaoste:
> Since we are removing the height prop in the end of the transition can we get rid of this complication with `px` to `rem` conversion in the `getPanelContentHeight` fn?

Mmmh, why did we need it in the first place? I'm not sure about the dis-/advantages of using rem for this kind of transition.

bgutsol:
> > Since we are removing the height prop in the end of the transition can we get rid of this complication with `px` to `rem` conversion in the `getPanelContentHeight` fn?
> 
> Mmmh, why did we need it in the first place? I'm not sure about the dis-/advantages of using rem for this kind of transition.

I guess before we needed `rem` because that height value was staying on the DOM elem until it's open. So in case, there is any change to the base font size the height will adjust respectively. But now I see that we are not keeping that value, it's only needed to play the animation and for animation it doesn’t matter what type of units to use. So I'd rather keep it simple and use `px`.

packages/components/collapse/src/Collapse.tsx
Chaoste: I'm moving it below so the display is reset to the default before calculating the height. We don't need to wait for another animation frame to do so, do we? I'm not sure but it worked for me without waiting for another frame.

Chaoste:
@bgutsol Thanks for feedback, I added all your feedback and added the following two things:

1. `pointer-events: none` during the collapse animation (but don't block early interaction while expanding)
2. Per default, set `display: none` (next to `height: 0`) to avoid a jumping animation for an initially collapsed container. See this example where it starts collapsed but still runs the transition:

https://user-images.githubusercontent.com/9327071/148803848-bcc79902-9d56-464c-81b4-b137e9516cc8.mov

bgutsol (approved):
Nice 👍
Thanks for addressing all the comments.

bgutsol (commented):
Regarding failed ci check-links, it's on our side. We will fix it in master and I let you know to rebase.

bgutsol:
> Regarding failed ci check-links, it's on our side. We will fix it in master and I let you know to rebase.

fix is in master

PR #1735, opened 2022-01-10
Fix/component page structure

gui-santos:
# Purpose of PR

- Fix typography
- Fix spacing between elements
- Fix layout grid
- implement grid in component page
- refactor PageHeader
- add "intro" and "extra" in frontmatter to be added as content of the PageHeader

Screenshot 2022-01-10 at 16 31 29

## PR Checklist

- [ ] Tests are passing


suevalov (approved):
💯

packages/website/components/MdxRenderer.tsx
```diff
@@ -16,66 +15,38 @@ const {
...
-  h3: (props) => <Subheading as="h3" marginTop="spacingL" {...props} />,
+  h2: (props) => <Heading as="h2" marginTop="spacing2Xl" {...props} />,
+  h3: (props) => <Subheading as="h3" marginTop="spacingXl" {...props} />,
   h4: (props) => <Subheading as="h4" {...props} />,
   h5: (props) => <Subheading as="h5" {...props} />,
   h6: (props) => <Subheading as="h6" {...props} />,
-  p: (props) => (
-    <Paragraph {...props} className={styles.increesedSizeElement} />
-  ),
+  p: (props) => <Paragraph {...props} />,
```
suevalov, line 30: Didn't we want to have an increased size for all text elements?

bgutsol (commented):
Why does it have all the commits from `v4-branch`?

packages/website/components/MdxRenderer.tsx
gui-santos: No
we decided to go back with the typography system that we have and, in the future, we upgrade/fix the whole system if needed

gui-santos:
> Why does it have all the commits from `v4` branch?

@bgutsol 
because I started this branch before we merged v4 into master
and rebasing it was almost impossible so I merged master to this branch and now git history exploded
but I will squash all commits before merging it

packages/website/components/PageContent.tsx
```diff
@@ -1,194 +1,199 @@
...
+            View on Github
+          </TextLink>
+        )}
+      </Flex>
+
+      {showNote && (
+        <Flex flexDirection="column" marginBottom="spacingXl">
+          <Note variant="negative" title="Deprecated component">
+            {title} was deprecated in February 2022. It will be deleted from the
+            repository in 6 months.
```
bgutsol, line 124: What about making it more clear?
```suggestion
            {title} was deprecated in v4. It will be deleted from the
            repository on 12th July 2022.
```

packages/components/button/src/Button/README.mdx
```diff
@@ -1,15 +1,14 @@
 ---
 title: 'Button'
+description: 'Start actions, such as “Add”, “Cancel” or “Save”.'
```
bgutsol, line 3: Could you please elaborate on why we decided to change the long description (that we had in the content) with this one? Do we plan to do the same for each component page?

Also noticed that if the description is more than one line, it becomes hard to read
image

bgutsol (commented):
Couple of comments

packages/components/button/src/Button/README.mdx
gui-santos: >Also noticed that if the description is more than one line, it becomes hard to read

Oh! Good catch 
I used `Text` and didn't put any lineHeight there, I will fix it thanks!

>Could you please elaborate on why we decided to change the long description (that we had in the content) with this one? Do we plan to do the same for each component page?

@domarku designed this new page header that I think he didn't have time to show to the team yet
but the idea is to give short descriptions of what is the use of the component
I did the change just to in Button because I wanted to test the CSS and forgot to revert it before opening the PR, my bad
I will revert it and we can implement it later

and yes, the idea is that all components would follow this pattern

again, let me revert this change in Button.mdx and if we go with this we can simply add it later

Thanks for catching these leftovers and bugs 🙌

packages/website/components/PageContent.tsx
gui-santos: I liked the idea

cc @domarku

bgutsol (approved):

PR #1727, opened 2022-01-10
docs: fix imports in README files

burakukula:

PR #1738, opened 2022-01-11
docs: update ModalConfirm documentation

suevalov:
# Purpose of PR

Updates `ModalConfirm` documentation


packages/components/modal/src/ModalConfirm/README.mdx
```diff
@@ -5,48 +5,65 @@ status: 'stable'
...
 
-`ModalConfirm` is a convenience wrapper on top of `Modal`. It is intended to use for showing confirmation dialogs.
-
-### Import
+## Import
 
 ```js static=true
 import { ModalConfirm } from '@contentful/f36-components';
-// or
-import { ModalConfirm } from '@contentful/f36-modal';
```
bgutsol, line 21: Recently we discovered that there are apps where tree shaking is not working, like UI extensions (if I remember the name right). So we decided to keep the examples with separate imports.

bgutsol (approved):

packages/components/modal/src/ModalConfirm/README.mdx
suevalov: ah, do we have two options in all examples now?

bgutsol: yes, we did a follow up PR yesterday: https://github.com/contentful/forma-36/pull/1727

PR #1742, opened 2022-01-11
fix: allow icon colors for transparent IconButton

burakukula:


suevalov (commented):
By looking at the Chromatic diff, it looks like Note and Notification close button became darker and should be manually adjusted to a previous value

suevalov (approved):

PR #1745, opened 2022-01-12
fix: Update README

burakukula:


suevalov (commented):
Let's make all the links to `f36.contentful.com` and we also need to update README of the tokens package.

thebesson (approved):

burakukula:
@suevalov yeah, I need to also update main README, now I noticed that is also talking about old packages.

burakukula:
@suevalov I think before we do the switch I am not able to point everything to `https://f36.contentful.com/` - check links will fail for some of them, since they only exists in nextjs version (like contribution)

packages/forma-36-tokens/README.md
```diff
@@ -2,46 +2,46 @@
...
 ```css
-@import '@contentful/forma-36-tokens/dist/css/index.css';
+@import '@contentful/f36-tokens/dist/css/index.css';
 ```
 
 #### SCSS
 
 ```scss
-@import '@contentful/forma-36-tokens/dist/scss/index.scss';
+@import '@contentful/f36-tokens/dist/scss/index.scss';
```
suevalov, line 38: We don't have SCSS anymore

suevalov (approved):

PR #1747, opened 2022-01-12
docs: update promo text and show version number

suevalov:
# Purpose of PR

Updates promo copy on v3 website and adds a current version number


packages/forma-36-website/src/components/Header.js
```diff
@@ -24,6 +26,11 @@ const styles = {
     font-size: ${tokens.fontSizeXl};
     margin-left: ${tokens.spacingL};
   `,
+  versionText: css`
```
burakukula, line 29: Why do we need this class?

burakukula (approved):

PR #1766, opened 2022-01-13
fix: counter alignment

### LINKED ISSUE #1408: 🐛 Bug - Changing character count for TextField can make help text widths change too
# Forma 36 bug report

## Summary

The character count for the `TextField` component changes in width as the number changes. This can cause any help text to suddenly break:

https://user-images.githubusercontent.com/219017/137343057-0a97968d-5689-43b4-8014-932ef3c744b0.mov

## Proposal

Perhaps we want to make the counter take up the full possible width that it can take, based on the field's maximum length? We should combine that [font-variant-numeric](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-numeric) to avoid any changing widths because of different widths of numbers.

suevalov:
# Purpose of PR

Fixes #1408


gui-santos (approved):
🍕

massao (approved):

PR #1773, opened 2022-01-14
fix: order components alphabetically

burakukula:


packages/website/components/Topbar.tsx
```diff
@@ -92,7 +92,7 @@ export const Topbar = () => (
           <TopbarLink href="/tokens/color-system" label="Tokens" />
         </li>
         <li>
-          <TopbarLink href="/components/box" label="Components" />
+          <TopbarLink href="/components/accordion" label="Components" />
```
suevalov, line 95: Hmmm, I'm not a biggest fan of `/accordion` being the first page 🤔

burakukula: It's the first one when we order items alphabetically 🙂

gui-santos (approved):

PR #1774, opened 2022-01-14
docs(icons): update documentation

thebesson:
# Purpose of PR

Extract examples into separate files.

### DIFF
--- a/packages/components/icon/README.mdx
+++ b/packages/components/icon/README.mdx
@@ -68,7 +68,7 @@ Another way to render custom icons is to wrapping SVG paths in the `Icon` compon
 
 ### Built-in icons
 
-<PropsTable of="ArrowDownIcon" />
+<PropsTable of="ArrowDown" />
 
 ### Custom icons
 
--- a/packages/components/icon/README.mdx
+++ b/packages/components/icon/README.mdx
@@ -3,6 +3,7 @@ title: 'Icons'
 type: 'component'
 slug: /components/icon/
 github: 'https://github.com/contentful/forma-36/tree/master/packages/components/icon'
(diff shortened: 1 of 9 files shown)


packages/components/icon/examples/IconAsPropExample.tsx
```diff
@@ -0,0 +1,7 @@
+import * as React from 'react';
+import { Icon } from '@contentful/f36-components';
+import { MdAccessAlarm } from 'react-icons/md';
+
+export default function IconAsPropExample() {
+  return <Icon as={MdAccessAlarm} variant="secondary" />;
+}
```
gui-santos, line 7: Unfortunately because we remove the imports, this example is not very clear in the page

Screenshot 2022-01-14 at 16 07 25

I only understood where the `MdAccessAlarm` was coming from after I read the code in the PR

I don't know how to solve this tbh, maybe a comment in the example? maybe a SVG instead of `react-icons`?

🤷

packages/components/icon/README.mdx
```diff
@@ -178,3 +73,11 @@ const CustomIcon = (props) => {
 ### Custom icons
 
 <PropsTable of="Icon" />
```
gui-santos, line 75: Screenshot 2022-01-14 at 16 08 01

There's something wrong with the props here, could you take a look?

gui-santos (changes requested):
Looks great! I have only two comments

packages/components/icon/examples/IconAsPropExample.tsx
thebesson: there is an example with SVG icons. I will add a comment

gui-santos (approved):

PR #1777, opened 2022-01-17
docs: refactor Accordion readme to the new format

burakukula:


massao (approved):

packages/components/accordion/examples/AccordionExample.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { Accordion, Text } from '@contentful/f36-components';
+
+export default function IconButtonExample() {
+  return (
+    <Accordion>
+      <Accordion.Item title="What payment methods do you accept?">
```
suevalov, line 7: Can we have more than one accordion.item in the example?

packages/components/accordion/examples/AccordionWithOtherComponentExample.tsx
```diff
@@ -0,0 +1,33 @@
...
+  Text,
+  SectionHeading,
+  Stack,
+  Button,
+} from '@contentful/f36-components';
+
+export default function IconButtonExample() {
+  return (
+    <Accordion>
+      <Accordion.Item
```
suevalov, line 13: the same as above

packages/components/accordion/examples/AccordionExample.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { Accordion, Text } from '@contentful/f36-components';
+
+export default function IconButtonExample() {
+  return (
+    <Accordion>
```
suevalov, line 6: is there a way to control which items are opened programmatically? (controlled vs uncontrolled)

suevalov (changes requested):
I think we need to have a bit more sophisticated examples here

PR #1789, opened 2022-01-18
fix(button): set color explicitly for hover state

denkristoffer:
# Purpose of PR

Explicitly set a `color` on buttons. We need to do this for buttons rendered `as="a"`, because legacy styles are overriding `a:hover` 😕


Lelith (approved):
Looks good to me!

Issue #1769, opened 2022-01-14
💡 Proposal - Re-implement date and time type in TextInput


william-segel:
# Forma 36 contribution proposal

## The problem
In Forma36 V3, the TextInput component had **time** and **date** as input types but they seem to be missing in Forma36 V4. Since date and time inputs are being used in my projects its absence in V4 is more prevalent than ever.

## The proposed solution
Re-implement 'date' and 'time' types in the TextInput props along with styling and functionality according to V4 standard.

gui-santos:
Hi @william-segel 

I made a test using the playground in our website and this is what I got
https://bit.ly/3qsIx6R

the `type` prop of the TextInput component accepts [the same values that the `input` html tag does](https://www.w3schools.com/html/html_form_input_types.asp)

Is this what you need?

william-segel:
Hello @gui-santos!
Well, your playground code was enough to convince me that Forma V4 had the time and date types implemented so I did some further testing. 

It seems that the TextInput component only takes the union of "number" | "text" | "url" | "email" | "search" | "file" | "password" so the TypeScript file won't compile with "time" or "date" in the type property. But, when I put //ts-nocheck on the file using the component everything seems to work fine and the input shows the same UI as provided via your playground example.

Is it the component library (f36-components) that is missing the "time" and "date" types in the union? Or is it me that is accidentally using some outdated library somewhere else? I'm kinda new to TypeScript and ReactJS so I can't quite put my finger on it 🤔

gui-santos:
Hey @william-segel 

Sorry for the delay, I missed Github's notification in my emails

Anyway, you are right! And we found the culprit
For some reason we are overwriting the default types for `input` here:
https://github.com/contentful/forma-36/blob/master/packages/components/forms/src/BaseInput/types.ts#L26-L29

I will open a PR to fix it right away! Thanks for spotting this 🙏

PR #1793, opened 2022-01-19
fix: allows all native HTML values to type prop of an input

gui-santos:
# Purpose of PR

We were overwriting the default values for the `type` property of an input and because of that TS was complaining when you did something like `<TextInput type="date" />`

This will close https://github.com/contentful/forma-36/issues/1769

## PR Checklist

- [ ] Tests are passing


gui-santos:
cc: @william-segel

burakukula (approved):
💪 I think we did it to exclude date and time, but I guess it doesn’t make much sense.

bgutsol (approved):

suevalov (approved):

PR #1794, opened 2022-01-19
docs(card): update documentation

thebesson:


packages/components/card/examples/Card/WithLinkAndTargetExample.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { Card, Text } from '@contentful/f36-components';
+
+export default function WithLinkAndTargetExample() {
+  return (
+    <Card href="https://f36.contentful.com" target="_blank">
```
thebesson, line 6: TS linter yells at "target", I don't understand why. BaseCard accepts it. Maybe someone who's savvier with TS will be able to tell?

bgutsol: Yeah, not a 100% clear tslint message. By default Card is rendered as `article`, so you have to pass  `as="a"` prop in order to make it work.

bgutsol (approved):

PR #1796, opened 2022-01-20
docs: new format for EntityList

bgutsol:
# Purpose of PR

- new format for EntityList component

### DIFF
--- a/packages/components/entity-list/README.mdx
+++ b/packages/components/entity-list/README.mdx
@@ -8,91 +8,34 @@ storybook: 'https://v4-f36-storybook.netlify.app/?path=/story/components-entity-
 typescript: ./src/EntityList.tsx,./src/EntityListItem/EntityListItem.tsx
 ---
 
-# EntityList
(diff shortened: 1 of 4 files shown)


thebesson (approved):

packages/components/entity-list/README.mdx
```diff
@@ -8,91 +8,34 @@ storybook: 'https://v4-f36-storybook.netlify.app/?path=/story/components-entity-
...
-  />
-  <EntityList.Item
-    title="Entry 2"
-    description="Description"
-    contentType="My content type"
-    status="draft"
-    withDragHandle
-  />
-</EntityList>
+```jsx file=examples/EntityListWithDragHandleExample.tsx
```
burakukula, line 37: I am not sure if we want to do it, but I think dragging functionality should work in forma 36 docs as well. I mean, it is a bit confusing that we show how to use prop, but we don't show how to implement the dragging functionality itself

PR #1804, opened 2022-01-21
fix: allow unique IDs for collapse component

YvesRijckaert:
# Purpose of PR

Currently the `Collapse` component typescript interface does not allow `id` to be set with a prop.

Even though in the story file of the `Collapse` component an ID has been set: https://github.com/contentful/forma-36/blob/050169cac56863191c0dd8a564300d595605b710/packages/components/collapse/stories/Collapse.stories.tsx#L39

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/components/collapse/src/Collapse.tsx
+++ b/packages/components/collapse/src/Collapse.tsx
@@ -16,6 +16,10 @@ interface CollapseInternalProps extends CommonProps {
    * string for additional classNames
    */
   className?: string;
+  /**
+   * A unique id of the collapse component
+   */
+  id?: string;
 }
 
 export type CollapseProps = PropsWithHTMLElement<CollapseInternalProps, 'div'>;


gui-santos (approved):

denkristoffer:
Shouldn't this get added automatically by `PropsWithHTMLElement`? 🤔

YvesRijckaert:
> Shouldn't this get added automatically by `PropsWithHTMLElement`? 🤔

That doesn't seem to work for me in Compose..

```
Type '{ children: Element[]; id: string; isExpanded: boolean; "aria-labelledby": string; className: string; }' is not assignable to type 'IntrinsicAttributes & CollapseProps'.

Property 'id' does not exist on type 'IntrinsicAttributes & CollapseProps'
```

bgutsol (changes requested):
@denkristoffer you are right, it should be provided by `PropsWithHTMLElement`.
@YvesRijckaert could you mention pls what version of Forma you are using in Compose? It should be `4.0.3` or higher. 

As I see right now Compose using `4.0.2` so probably, @YvesRijckaert you just need to bump `f36-components`

PR #1814, opened 2022-01-24
fix: add link to submit the idea in GitHub for components

burakukula:
So maybe something like this?

Screenshot 2022-01-24 at 14 50 50

 + remove dead blog from the footer


packages/website/components/SidebarLink.tsx
```diff
@@ -100,14 +101,30 @@ export function SidebarLink({
...
+          titleStyles.clickable,
+          titleStyles.sidebarItem,
+        ])}
+        {...linksProps}
+      >
+        {children}
+        {isExternal ? <ExternalLinkIcon variant="muted" /> : null}
+      </a>
+    </Link>
+  ) : (
```
gui-santos, line 127: Why is this code necessary?

This link should work the same way that external links in the "Introduction" sidebar work

I don't think any change to SidebarLink is necessary, you should only need to add something like this to the Sidebar at line 140

```jsx
<SidebarSection
  links={[{
    title: 'Submit an idea in GitHub',
    slug:
      'https://github.com/contentful/forma-36/discussions/new',
    type: 'link',
  }]} 
  currentPage={currentPage} 
/>
```

This is all you need to do and it will have the same styles and everything

gui-santos: On the same matter, I think the copy of this link should be something shorter, look what happens in my screen (I have a mouse and Chrome shows a scrollbar in our Sidebar)

Screenshot 2022-01-25 at 16 52 29

gui-santos (changes requested):
Hey thanks for taking that task

I think you are writing more conde than we need, I added a suggestion of how I think you can add a link without creating new logic to Sidebarlink

ping me if you want to pair on this

packages/website/components/SidebarLink.tsx
burakukula: ok, I thought we might not necessarily want to render it as list element, but ok, sure, on my way to fix it.

packages/website/components/Sidebar.tsx
```diff
@@ -143,6 +144,16 @@ export function Sidebar({ currentPage = '/' }: Props) {
 
       {activeSection === WEBSITE_SECTION.COMPONENTS && (
         <>
+          <SidebarSection
+            links={[
+              {
+                title: 'Proposal in GitHub',
```
burakukula, line 150: @gui-santos dunno, maybe something like this? or better. "Idea in GitHub" ? WDYT?

gui-santos: looks good to me, I'm not the best with copy

I tried this as well

Screenshot 2022-01-26 at 11 14 14

But we can always change copy later, I just told you because the scrollbar was braking the layout

gui-santos (approved):
🦙

gui-santos:
Btw, I don't understand why it's saying that the links in this PR are broken 🤷 

it's complaining about `https://www.contentful.com/legal/` but I thought we were ignoring this link

Maybe it's complaining because you removed a link from the footer???? I don't know

gui-santos:
@burakukula are we merging this PR?

burakukula:
> @burakukula are we merging this PR?
> 
> I think it's worth rebasing it

On it, sorry, forgot about this PR 🤦‍♀️

PR #1819, opened 2022-01-25
docs(pill): update docs

thebesson:


packages/components/pill/README.mdx
```diff
@@ -8,58 +8,41 @@ storybook: 'https://v4-f36-storybook.netlify.app/?path=/story/components-pill--b
 typescript: ./src/Pill.tsx
 ---
 
-import { InfoCircleIcon } from '@contentful/f36-icons';
-
 Pills represent a set of one or multiple objects which can be entered or changed by the user. By default, pills can be removed by the user. When necessary, the order of pills can be changed by the user using drag and drop. They are used to represent complex entities in small blocks such as contact details. Contentful Pills are are used to render transition locales and tag elements.
```
bgutsol, line 11:
```suggestion
Pills represent a set of one or multiple objects which can be entered or changed by the user. By default, pills can be removed by the user. When necessary, the order of pills can be changed by the user using drag and drop. They are used to represent complex entities in small blocks such as contact details. Contentful Pills are used to render transition locales and tag elements.
```

Could you rephrase this description a bit? For example, first three sentences don't really make sense to me

packages/components/pill/README.mdx
```diff
@@ -8,58 +8,41 @@ storybook: 'https://v4-f36-storybook.netlify.app/?path=/story/components-pill--b
...
 
-## Code examples
+## Examples
 
-Example of usage with draggable and close option
+### Using with draggable and close option
 
-```jsx
-// import { Pill } from '@contentful/f36-pill';
+```jsx file=./examples/DraggableAndClosePillExample.tsx
```
bgutsol, line 25: Don't know what is a better way to improve those two examples, but they feel kind of useless. Especially the second one with custom dragHandleComponent. WDYT, can we improve them somehow?

bgutsol (commented):
Couple of comments

packages/components/pill/README.mdx
gui-santos: I agree with you, it seems that the second one could be removed

Maybe there are some use cases in webapp

thebesson: I decided to keep those basic examples, just in case someone would need to just copy a basic one, and I added an example with react-sortable-hoc. let me know what do you think :)

packages/components/pill/examples/IntegrationPillExample.tsx
```diff
@@ -0,0 +1,80 @@
...
+      useDragHandle
+      axis="xy"
+      distance={10}
+      onSortEnd={({ oldIndex, newIndex }) => {
+        swapItems({ oldIndex, newIndex });
+      }}
+    >
+      {items.map((item, index) => (
+        <SortablePill
+          key={item + index}
```
bgutsol, line 72: why do you need an index here? It's kind of good practise not to use the index as keys for items that can change their order.

bgutsol (approved):
All good 👍
Just one minor comment

packages/components/pill/examples/IntegrationPillExample.tsx
thebesson: We want it to be re-rendered, right?

bgutsol: Yes, but in our case item is unique, so we can use only it for the `key` prop, right?

PR #1822, opened 2022-01-26
feat: implement short description in page content header

gui-santos:
# Purpose of PR

- Moves all components related to PageContent to a separated folder in `/components` and it exports only the thing used by files outside of that folder
- It implements a functionality to get every text before the first heading (##) in a mdx file and uses that as the "short intro"
- Makes the first paragraph of the "short intro" bigger like in the designs

Screenshot 2022-01-26 at 12 42 43

I managed to create a system to edit the header of the content easily without us losing any features of MDX

Basically, during the creation of the page, we will extract from the MDX file everything that comes before a heading (a sentence in mdx that starts with `#`)

Screenshot 2022-01-26 at 12 43 00

In this example, everything after the frontmatter and before "## Import" will be extracted and stored in a prop to be used later by `PageContentHeader`

```tsx
// PageContent.tsx line 77
<PageContentHeader title={title} github={github} status={status}>
  {source.shortIntro && <MdxRenderer source={source.shortIntro} />}
</PageContentHeader>
```

If the page has a "shortIntro" we will render another MdxRenderer in the PageContentHeader with the text we selected earlier

This system is good because it will work immediately with every page and it does not change how we write the pages

... but that's also a problem, when I say it works in every page I meant EVERY PAGE

Screenshot 2022-01-26 at 12 43 23

E-V-E-R-Y P-A-G-E

Screenshot 2022-01-26 at 12 43 42

That's why I'm going trough every page and adding some headings or moving the content to avoid this huge headers with tables and lists

WDYT?

## PR Checklist

- [ ] Tests are passing


bgutsol (approved):
Looks good! A bit hacky, but if we want to have it working out of the box, there are not many other ways.

My only concern is that it could be hard to come up with a short and concise description for every component. Sometimes one line sentence is not enough.

massao (approved):

PR #1824, opened 2022-01-26
docs(workbench): improve documentation format of Workbench

massao:


massao:
@contentful/team-bauhaus Should we add any examples??
I didn't add, because I remember that we didn't want to encourage the usage of this component. 🤔

gui-santos (approved):
You're right, we don't need examples maybe

We can add later if we decide to do it

thebesson (approved):

PR #1831, opened 2022-01-28
fix(workbench-header): allow ReactNode as title

denkristoffer:
# Purpose of PR

We're currently using elements in the WorkbenchHeader `title` in some places, the current types don't allow for that.

Also adds some other fixes needed for the content model editor:
- `icon` prop takes an Element, which is needed for passing things like `variant` or `size` on product icons.


denkristoffer:
I think this might be too simple a solution, since the current Header wraps `title` in a `Heading` component 🤔

packages/components/workbench/src/WorkbenchHeader/WorkbenchHeader.tsx
```diff
@@ -10,7 +11,7 @@ import { getWorkbenchHeaderStyles } from './WorkbenchHeader.styles';
 
 export interface WorkbenchHeaderProps extends CommonProps {
   /** This is the title that will be shown inside the Header component */
-  title: string;
+  title: ReactNode;
```
gui-santos, line 14: Maybe you could include the change in line 61 in this PR

Basically, if title is a string we render the way we render it now, but if is a ReactNode we just render the ReactNode

wdyt?

gui-santos (dismissed):

packages/components/workbench/src/WorkbenchHeader/WorkbenchHeader.tsx
gui-santos: @denkristoffer any chance of following up with this PR changes?

bgutsol (commented):
Great, looks good!
Can you also add a small description and examples to the README file? With the example of title and icon passed as a ReactElement.

denkristoffer:
@bgutsol Sure, but is the Workbench from Forma deprecated or not? I see that there's no existing examples or much documentation for it.

bgutsol:
> @bgutsol Sure, but is the Workbench from Forma deprecated or not? I see that there's no existing examples or much documentation for it.

@denkristoffer we have [docs page](https://f36.contentful.com/components/workbench), here is the [readme](https://github.com/contentful/forma-36/blob/607301d57a2e83190d2aa298120ddb8493e8c429/packages/components/workbench/README.mdx). 
But you are right there are no usage examples. @gui-santos do you know why we decided not to add any examples.

@denkristoffer Not to block this PR, maybe let's just add examples in the storybook, at least for future us. WDYT?

bgutsol (approved):
Thanks @denkristoffer

thebesson (approved):
@denkristoffer Workbench is being deprecated, this is why it's not included in the main package, just FYI 👍

gui-santos:
> 
> But you are right there are no usage examples. @gui-santos do you know why we decided not to add any examples.
> 

Because we were supposed to remove Workbench from everywhere, but I don't think we will do that any time soon so let's add examples to the docs!

PR #1843, opened 2022-01-31
fix: remove min-height from GlobalStyles

bgutsol:
# Purpose of PR

- Remove min-height from GlobalStyles. Because it breaks the auto resizer of our App SDK in the iframes.


burakukula (approved):
would this not break anything else?

bgutsol:
> 

I've checked Compose and Launch, and it won't break them. As well as our new format website.

PR #1844, opened 2022-02-01
fix: use replaceState to avoid reloading the playground page

gui-santos:
# Purpose of PR

use replaceState to avoid reloading the playground page while the user is writing their code

## PR Checklist

- [ ] Tests are passing


thebesson (approved):
👏

packages/website/components/Playground/useUrlSync.tsx
```diff
@@ -1,20 +1,24 @@
...
+    if (code !== router.query.code) {
+      const href = `/playground?code=${code}`;
+      window.history.replaceState('', '', href);
+      setCodeUrl(window.location.origin + href);
     }
-    const href = `/playground?code=${code}`;
-    router.push(href, undefined, { shallow: true });
   }, [debouncedCode]);
+
+  return { router, codeUrl };
```
massao, line 23: Since you don't use the `router` returned here, can't it be removed?

massao (approved):

packages/website/components/Playground/useUrlSync.tsx
gui-santos: good catch!

PR #1845, opened 2022-02-01
feat(pagination): add Pagination component

massao:
# Purpose of PR

PR to contain all other PRs about pagination component before being merged.

## PR Checklist

- [x] Tests are passing


packages/components/pagination/src/Pagination.tsx
```diff
@@ -0,0 +1,123 @@
...
+    // Active page is at the start of the pagination page count
+    return baseRange;
+  }
+  if (page > total - neighboursCount) {
+    // Active page is at the end of the pagination page count
+    return baseRange.map((i) => i + total - PAGINATION_RANGE - 1);
+  }
+  // Active page is in the middle of the pagination count
+  return baseRange.map((i) => i + page - neighboursCount - 1);
+}
```
gui-santos, line 50: Shouldn't this be in utils as well?

gui-santos (approved):
🍨 

Amazing!

I have only on small suggestion

packages/components/pagination/src/Pagination.tsx
gui-santos: Well, I don't know

I think keeping it here and not exporting it is a good idea as well ✌️

PR #1847, opened 2022-02-01
refactor: update children type definition on ButtonGroup

massao:
# Purpose of PR

Update type definition to use also accept `falsy` values so conditional rendering can be used

### DIFF
--- a/packages/components/button/stories/ButtonGroup.stories.tsx
+++ b/packages/components/button/stories/ButtonGroup.stories.tsx
@@ -33,6 +33,7 @@ export default {
 export const basic: Story<ButtonGroupProps> = (args) => {
   return (
     <ButtonGroup {...args}>
+      {false && <Button variant="secondary">Button</Button>}
       <Button variant="secondary">Button</Button>
       <Button variant="secondary">Button</Button>
       <Button variant="secondary">Button</Button>
--- a/packages/components/button/src/ButtonGroup/ButtonGroup.tsx
+++ b/packages/components/button/src/ButtonGroup/ButtonGroup.tsx
@@ -43,9 +43,12 @@ function _ButtonGroup(
         if (!child) {
           return null;
         }
-        return React.cloneElement(child, {
+        return React.cloneElement(child as React.ReactElement, {
           key,
-          className: cx(styles.groupContent, child.props.className),
+          className: cx(
+            styles.groupContent,
+            (child as React.ReactElement).props.className,
+          ),
         });
       })}
     </Box>
--- a/packages/components/button/src/ButtonGroup/types.ts
+++ b/packages/components/button/src/ButtonGroup/types.ts
@@ -1,5 +1,6 @@
 import { CommonProps } from '@contentful/f36-core';
 import type { SpacingTokens } from '@contentful/f36-tokens';
+import { ReactElement } from 'react';
 
 export type ButtonGroupVariants = 'spaced' | 'merged' | 'collapsed';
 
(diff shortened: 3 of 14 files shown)


bgutsol (commented):
We should also add a props check everywhere, otherwise, if you pass `string` or `number` as a child it will break render. Since these primitives don't have a `props` field like `ReactElement`.
https://f36.contentful.com/playground?code=N4Igxg9gJgpiBcICWBbADhATgFwAQCUYBDMPAM0whVwHJNjSaBuAHQDtUMddh3dcAythIBrADR9cAIQCu2bBDYBxSjLQS2-WfMUb+ASUhttCthoC+uClVoABI9hhtsZGQBsA9GQDMANgC0kOiKTtgAzszsnFh4wLgAwgAWMABulGwAIhAA7myGurgACm4yYfmaltbUNPaKjs6unj4BSEYRrGzsMAAeXHiwZETu5DJspEiK0nKmKhBqALIwmADmMFAAot1E6G4wABQAlDyS9NgymJp7kvwAPEKiVrvdGUj044oAvCzgECUonSAAHzXfi4G4mRSzNTAzSguHAGhQMJQGjmEFw8HTSYpIiYJBEZxfEBhGBGKC4gCe31wYSQAC8YESwigiG43N8YXCuYIiCkYOjQTcPBC2JzuWDyiKBXCcXiCdgmaTFOTMFSQNLQbSGUyWWzvhr+Lj8f43EQAEYwNxEgCCUCguBQWH56th4twrU+wBuxVK5VwHkBaNdXIDAqFIqhaDFtw89zAIjFByYQZA5iAA

bgutsol (commented):
I’m taking my words back regarding changing everything to ReactNode. After a closer look, I understood that for some components it has to be ReactElement only. For example IconButton. It doesn’t make sense to render it without an icon. I think we have to evaluate each case separately. WDYT?

[example](https://f36.contentful.com/playground?code=N4Igxg9gJgpiBcICWBbADhATgFwAQCUYBDMPAM0whVwHJNjSaBuAHQDtUMddhcBJSGwBCAV2zYIbADS4AythIBrXAF9cFKrQACg7DDbYyIgDYB6MgGYAbAFpI6SfuwBnZu05Y8vAMJFj+qCJMAUlVdUpqGh1JPQMjM0tbJEFXVjZ2GAAPLjxYMiITchE2UiRQkOExCTYAUUyidH8ACgBKHnZcXHpsEUw2XCaOztwAHnkldX9MgBEkelLJAF4WcAhjERR0kAA+IeHR8bBFXf794ZGK0XFJPbPOgDcgpCIDZZBnGEFAzABPFdu7k8iDZjEQAEYwYxvWSQz54bAACxguECen+pzunWSS2A+WMHxUAOGphOdxGpkOx1u5MppNwLTShLYIBUQA)

gui-santos (approved):
🏅

packages/components/button/stories/ButtonGroup.stories.tsx
```diff
@@ -33,6 +33,7 @@ export default {
 export const basic: Story<ButtonGroupProps> = (args) => {
   return (
     <ButtonGroup {...args}>
+      {false && <Button variant="secondary">Button</Button>}
```
bgutsol, line 36: Do you want to leave it, or just forgot to remove?

bgutsol (approved):

packages/components/button/stories/ButtonGroup.stories.tsx
massao: to serve as example, and also if kind of test that it's working 😄

PR #1851, opened 2022-02-04
fix(card): don't omit withDragHandle type

denkristoffer:
# Purpose of PR

Reinstates the `withDragHandle` boolean on `Card` component. This prop is required to be able to use `dragHandleRender`.

### DIFF
--- a/packages/components/card/examples/Card/WithCustomDragHandle.tsx
+++ b/packages/components/card/examples/Card/WithCustomDragHandle.tsx
@@ -41,7 +41,7 @@ export default function WithCustomDragHandle() {
   const SortableCard = SortableElement(({ label }) => {
     return (
       <Card
-        dragHandleRender={() => <SortableDragHandle item={label} />}
+        dragHandleRender={() => <SortableDragHandle />}
         padding="none"
         withDragHandle
       >
--- a/packages/components/card/examples/Card/WithCustomDragHandle.tsx
+++ b/packages/components/card/examples/Card/WithCustomDragHandle.tsx
@@ -1,0 +1,76 @@
+import React from 'react';
+import { Box, Card, DragHandle, Flex } from '@contentful/f36-components';
+import { css } from 'emotion';
+import {
+  SortableContainer,
+  SortableElement,
+  SortableHandle,
+} from 'react-sortable-hoc';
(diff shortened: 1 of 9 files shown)


bgutsol (changes requested):
Pls, check this discussion first 
https://github.com/contentful/forma-36/pull/1583#discussion_r761977175

I guess, if we need a Card with a drag handle we should provide styling for that first.

bgutsol (commented):
Three comments here:
1. If you just pass `withDragHandle` prop, without passing `padding` prop, you won't get any typescript errors that padding should be `none`. As a result, it will be rendered like this:
image

2. If you pass both props with `padding="none"` you still will get a kind of raw card (no padding for content). Is it what we want?
image

3. I would also add some examples with descriptions to component docs and storybook.

gui-santos:
Any news on this PR?

denkristoffer:
I will work on this tomorrow 🙏

denkristoffer:
I'm not sure how to fix the remaining type error. React is complaining about `Ref<any>` not matching the `HtmlDivElement` type from `const DRAG_HANDLE_DEFAULT_TAG = 'div'`. Best I can think of is not using the `Element` constant [here](https://github.com/contentful/forma-36/pull/1851/files#diff-a88dc7ed1ba105783c1298bdecbd105429a43c93d5f24081ee6fa6eea7b9828aR127) and instead rendering either a button or a div in an if statement.

@massao @bgutsol any ideas?

massao:
> I'm not sure how to fix the remaining type error. React is complaining about `Ref<any>` not matching the `HtmlDivElement` type from `const DRAG_HANDLE_DEFAULT_TAG = 'div'`. Best I can think of is not using the `Element` constant [here](https://github.com/contentful/forma-36/pull/1851/files#diff-a88dc7ed1ba105783c1298bdecbd105429a43c93d5f24081ee6fa6eea7b9828aR127) and instead rendering either a button or a div in an if statement.
> 
> @massao @bgutsol any ideas?

I think that dropping the `Element` constant and rendering only one or the other is a good solution, because even if we may be able to fix the typing, it could become too complex to maintain later.

bgutsol (approved):

Issue #1848, opened 2022-02-02
🐛 Bug - Note styles override link button colours


denkristoffer:
# Forma 36 bug report

## Summary

<img width="453" alt="Screenshot 2022-02-02 at 14 16 09" src="https://user-images.githubusercontent.com/219017/152161366-4dab0dce-f5a9-4741-a90c-0d06532f3f6c.png">

If you render a `<Button as="a" />` in a Note, the Note variant styles are currently overriding the button's colour because of this line: https://github.com/contentful/forma-36/blob/0e528bed1942e8b4268ec68f2e2fad67f3389f42/packages/components/note/src/Note.styles.tsx#L14-L16

Is there a way we can select something like `a:not(our-button-component)`?

gui-santos:
Is this a bug or a design implementation issue? @domarku 
my question is: do we want a link that looks like a button inside a note?

Also, just looking by the code, it seems that the intention here was to change the color of the link depending on the variant of the Note

I would say the easiest way to "fix" this is to change the styles in the Button component to enforce it to keep the white color when it has `as="a"`

denkristoffer:
>  my question is: do we want a link that looks like a button inside a note?

I would say we do, we use it quite a few places where we want an obvious primary action inside a note, like these EAP signups. I don't think a TextLink is eye-catching enough for that.

denkristoffer:
@gui-santos 

> I would say the easiest way to "fix" this is to change the styles in the Button component to enforce it to keep the white color when it has as="a"

How would you do this?

denkristoffer:
@gui-santos Ping 🙂

Lelith:
@denkristoffer should we close this? I do not think we can fix it without removing the general rule for links in the note.styles this is one of the big downsides of css-in-js imho. If we would have proper class names we could prevent the button style from beeing overwrittin by hooking on the generic button class :(

PR #1856, opened 2022-02-07
fix(button): enforce color properties with !important

denkristoffer:
# Purpose of PR

Sets `color: ... !important`on buttons.

Yes, this is bad form, but currently buttons that are rendered `as="a"` are rendered with the wrong colours because of `a:link` and `a:hover` styling in our legacy CSS. This is a suggested temporary fix until we can get rid of legacy styles.

See also #1848 for a similar selector issue albeit with a different reason.

### DIFF
--- a/packages/components/button/src/Button/Button.styles.ts
+++ b/packages/components/button/src/Button/Button.styles.ts
@@ -24,6 +24,8 @@ const variantToStyles = (variant: ButtonVariant): CSSObject => {
   switch (variant) {
     case 'primary':
       return {
+        // !important is a temporary necessity until we get rid of legacy styles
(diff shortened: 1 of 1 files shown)


packages/components/button/src/Button/Button.styles.ts
```diff
@@ -24,13 +24,12 @@ const variantToStyles = (variant: ButtonVariant): CSSObject => {
   switch (variant) {
     case 'primary':
       return {
-        color: tokens.colorWhite,
+        color: `${tokens.colorWhite} !important`,
```
burakukula, line 27: Maybe you could leave a comment over here that this is a temporary fix? WDYT?

burakukula (approved):

PR #1870, opened 2022-02-08
fix: playground top bar

bgutsol:
# Purpose of PR

- Fix the case when the playground subheader is collapsed because of the code area that is bigger than the page height.
image


massao (approved):
🔥

PR #1871, opened 2022-02-08
fix: menu item focus styles

bgutsol:
# Purpose of PR

- Make Menu item `hover` state styles different from `focus` state.

@domarku pls take a look 🖼


domarku (approved):
J'APPROVE

PR #1872, opened 2022-02-08
feat(website): add link to open Github issue

massao:
Add link on PageContentHeader to give about the Component.
* The link will use the `component-feedback.md` issue template.
* Add the component name to the title


packages/website/components/PageContent/PageContentHeader.tsx
```diff
@@ -57,15 +69,21 @@ export function PageContentHeader({
       </DisplayText>
 
       {github && (
```
gui-santos, line 71: Shouldn't we remove this condition now?

This prop is not used by the things that are rendered if the condition is true

gui-santos (approved):
Looks great! I just have one question

bgutsol (approved):
👍

packages/website/components/PageContent/PageContentHeader.tsx
massao: Nice catch, forgot to remove it

PR #1873, opened 2022-02-08
feat: by default render Popover only when it's open

bgutsol:
# Purpose of PR

- Rollback the logic when Popover is rendered in the DOM only when it's open and make it a default behavior. 
But with the option to turn it off.  Because there is a problem with our `Autocomplete` and Downshift library if the popover is not initially rendered. See the error below and more details in the [related PR](https://github.com/contentful/forma-36/pull/1355)

image


gui-santos (approved):
🏆

PR #1874, opened 2022-02-08
fix: improved collapsable examples

burakukula:
In the next PR we will provide a custom template for examples, it will be a bit darker.

### DIFF
--- a/packages/website/components/GlobalStyles.tsx
+++ b/packages/website/components/GlobalStyles.tsx
@@ -28,8 +28,8 @@ export const GlobalStyles = () => {
         body {
           color: ${tokens.gray800};
           font-family: ${tokens.fontStackPrimary};
-          font-size: ${tokens.fontSizeL};
-          line-height: ${tokens.lineHeightL};
+          font-size: ${tokens.fontSizeM};
+          line-height: ${tokens.lineHeightM};
           min-width: 1280px;
         }
 
--- a/packages/website/components/LiveEditor/ComponentSource.tsx
+++ b/packages/website/components/LiveEditor/ComponentSource.tsx
@@ -120,8 +120,8 @@ const styles = {
     border-radius: 0 0 ${tokens.borderRadiusMedium} ${tokens.borderRadiusMedium};
   `
 };
-const ChevronRightIcon = f36icons.PreviewIcon;
-const ChevronDownIcon = f36icons.PreviewOffIcon;
+const ComponentPreviewIcon = f36icons.PreviewIcon;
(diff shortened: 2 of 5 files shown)


packages/website/components/GlobalStyles.tsx
```diff
@@ -28,8 +28,8 @@ export const GlobalStyles = () => {
         body {
           color: ${tokens.gray800};
           font-family: ${tokens.fontStackPrimary};
-          font-size: ${tokens.fontSizeL};
-          line-height: ${tokens.lineHeightL};
+          font-size: ${tokens.fontSizeM};
+          line-height: ${tokens.lineHeightM};
```
bgutsol, line 32: Do we really want to have code in the code blocks smaller? For me, it's a bit inconvenient right now to read it.

packages/website/components/LiveEditor/ComponentSource.tsx
```diff
@@ -118,51 +147,78 @@ export function ComponentSource({
...
+              {showSource && (
+                <Flex>
+                  <CopyButton
+                    tooltipProps={{ placement: 'top' }}
+                    className={styles.copyButton}
+                    value={children}
+                    size="small"
+                  />
+                  {isExampleFromFile && (
+                    <Flex marginLeft="spacing2Xs">
```
gui-santos, line 200: we don't need this Flex I think

we can use `gap` in the Flex containing these buttons

packages/website/components/LiveEditor/StaticSource.tsx
```diff
@@ -3,7 +3,7 @@ import React from 'react';
 import Highlight, {
   defaultProps as HighlightDefaultProps,
 } from 'prism-react-renderer';
-import theme from 'prism-react-renderer/themes/github';
+import github from 'prism-react-renderer/themes/github';
```
gui-santos, line 6: Same comment as before

gui-santos (approved):
Wow! Looks really cool

I have a couple of suggestions but they are not blockers 🙌

gui-santos:
@burakukula 

Screenshot 2022-02-09 at 10 12 06

bgutsol (approved):
Love the styles of hidden code and an 👁 button ⭐️

Maybe we can have them all collapsed by default since the part of the code is visible even in collapsed state, wdyt? I remember we had all collapsed before v4

packages/website/components/GlobalStyles.tsx
burakukula: Yeah, that was the requirement from design.

PR #1875, opened 2022-02-09
fix(website): override style so what you select is what you edit

massao:
**Before**
image

**After**
image


gui-santos (approved):
UHmmm is this conflicting with @burakukula fix?

gui-santos:
Anyway, we should go for the solution with less code hahahaha

massao:
> UHmmm is this conflicting with @burakukula fix?

Different approach, #1876 adds padding to the wrapper, and the textarea is smaller, with this approach it makes so the textarea and pre tags use the entire space of the wrapper.
Up to @domarku to decide, since the main difference would be visual

burakukula:
@massao let's go with this solution, it looks better that my hack.

PR #1876, opened 2022-02-09
fix: hack paddings

burakukula:
I didn't come up with anything other that this hack, but it does not look perfect 😕

### DIFF
--- a/packages/website/components/LiveEditor/ComponentSource.tsx
+++ b/packages/website/components/LiveEditor/ComponentSource.tsx
@@ -98,15 +98,19 @@ const styles = {
   playgroundButton: css`
     font-family: ${tokens.fontStackPrimary};
   `,
-  // !important is needed to overwrite the react live setup.
+  // setting up background to match editor theme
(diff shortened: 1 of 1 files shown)


packages/website/components/LiveEditor/ComponentSource.tsx
```diff
@@ -98,15 +98,19 @@ const styles = {
   playgroundButton: css`
     font-family: ${tokens.fontStackPrimary};
   `,
-  // !important is needed to overwrite the react live setup.
+  // setting up background to match editor theme
+  editorWrapper: css`
+    padding: ${tokens.spacingL} ${tokens.spacingL} ${tokens.spacingXl};
+    background-color: #f5f8fa;
```
gui-santos, line 104: No token?

gui-santos (approved):
It looks great! It's not hacky it's CSS 🆗 

I think you need to check your VSCode settings:
Screenshot 2022-02-09 at 12 24 33

PR #1883, opened 2022-02-11
fix(Badge): disable newlines

Chaoste:
# Purpose of PR

We have an example where there is not much space for the badge. Still, the text in the badge should never break the word for a new line.

image

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/components/badge/src/Badge/Badge.styles.ts
+++ b/packages/components/badge/src/Badge/Badge.styles.ts
@@ -79,6 +79,7 @@ export const getBadgeStyles = ({
     letterSpacing:
       '0.06rem' /*move to tokens or update wide letter spacing token*/,
     borderRadius: `${tokens.borderRadiusSmall}`,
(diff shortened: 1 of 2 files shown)


ronaldronson (approved):

denkristoffer:
In that case it might be good to also hide the overflow with an ellipsis? @domarku what do you think?

Chaoste:
In our case, it was good to have it like it is now, so the text next to the badge would be shortened with an ellipsis.

denkristoffer:
Sounds like it's about `flex-grow` and `flex-shrink` then, is that wrong?

Chaoste:
You're probably right about the flex configuration, I will test it after the changes for Badge are live.

I adjusted the styles to show an ellipsis and added an example to the story:
Screenshot 2022-02-15 at 09 47 22

denkristoffer (approved):

PR #1884, opened 2022-02-11
fix: allow children to be passed to InlineEntryCard

gui-santos:
# Purpose of PR

Screenshot 2022-02-14 at 15 09 55

We've made a mistake in the past. I think we assumed the `InlineEntryCard` only needed `title`, but because of that we created a bug in Rich Text editor where the title is being rendered as the content of the card:

image

The title was supposed to be a prop to just render the native HTML tooltip, so now I updated it to show our Tooltip (see first image)

This PR fixes the bug and updates the documentation

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/packages/components/card/src/EntryCard/EntryCard.types.ts
+++ b/packages/components/card/src/EntryCard/EntryCard.types.ts
@@ -10,7 +10,7 @@ export type EntryCardInternalProps = Omit<
   'badge' | 'header' | 'padding' | 'ref' | 'type' | 'title'
 > & {
   /**
-   * The title of the entry, it will be used as the value for the native HTML attribute "title"
+   * The title of the entry, it will be used as the value for the tooltip
    */
   title?: string;
   /**
--- a/packages/components/card/src/EntryCard/EntryCard.types.ts
+++ b/packages/components/card/src/EntryCard/EntryCard.types.ts
@@ -10,7 +10,7 @@ export type EntryCardInternalProps = Omit<
   'badge' | 'header' | 'padding' | 'ref' | 'type' | 'title'
 > & {
   /**
-   * The title of the entry
+   * The title of the entry, it will be used as the value for the native HTML attribute "title"
    */
   title?: string;
   /**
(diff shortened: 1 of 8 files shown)


z0al (approved):
🥇

z0al:
> wouldn't we want to use a `Tooltip` instead?

I think the point here is to maintain the existing functionality: having `props.children` for body and `title` for the on-hover text.

I don't have an opinion here and I also don't think my team would mind it if F36 used the Tooltip component internally instead of the native HTML title. Maybe double-check with Design?

gui-santos:
@denkristoffer 

After talking to @damann 
I updated the code to use our Tooltip! 🙌

packages/components/card/src/InlineEntryCard/InlineEntryCard.tsx
```diff
@@ -44,14 +46,18 @@ export const InlineEntryCard = ({
     );
   }
 
+  console.log({ title, children });
```
denkristoffer, line 49: Left over 🙂

denkristoffer (changes requested):

packages/components/card/src/InlineEntryCard/InlineEntryCard.tsx
gui-santos: oops! hahahah

packages/components/card/src/EntryCard/EntryCard.types.ts
```diff
@@ -10,7 +10,7 @@ export type EntryCardInternalProps = Omit<
   'badge' | 'header' | 'padding' | 'ref' | 'type' | 'title'
 > & {
   /**
-   * The title of the entry
+   * The title of the entry, it will be used as the value for the native HTML attribute "title"
```
massao, line 13:
```suggestion
   * The title of the entry, it will be used as the value for the tooltip
```

massao (approved):

denkristoffer (approved):

PR #1891, opened 2022-02-16
feat: new property to typography components

burakukula:
Introducing new prop to typography components to manage cases like this one:

card-overflow2

WDYT?


denkristoffer:
Sorry, misclick 😬

massao (approved):
LGTM

thebesson (approved):

burakukula:
> Sorry, misclick 😬
> 
> My initial thought is that this kind of thing should be handled in the different components, not in the text elements.

@denkristoffer yeah, that was my initial thought, but I noticed that we already has to handle it in a couple of places in cards, I assume we might have the same issue in Note, and maybe more. Maybe in that case extending the API of typography might be a good idea.

PR #1916, opened 2022-03-01
fix(menu-item): keyboard-only focus style

denkristoffer:
# Purpose of PR

Makes MenuItem focus styling keyboard-only.


denkristoffer:
Am I doing the changeset thing right?

gui-santos (approved):
I don't know if the changeset is already working, but the rest looks good

mshaaban0 (approved):
💅

PR #1918, opened 2022-03-02
refactor: add import/no-extraneous-dependencies rule check

massao:
Add check for imports of packages not set as dependencies on package.json., filtering out test and examples files.

Adjusted imports and renamed `spec.*` files to `test.*` to keep it consistent.

----
For the changes on the build script of tokens, I compared the previous built files, and the new ones with `diff`, and they show no differences at all.


packages/website/package.json
```diff
@@ -11,25 +11,34 @@
...
     "@contentful/f36-components": "^4.0.0",
-    "@contentful/f36-pagination": "^4.1.0-beta",
     "@contentful/f36-core": "^4.0.0",
+    "@contentful/f36-docs-utils": "^4.0.0",
+    "@contentful/f36-icon": "^4.0.0",
     "@contentful/f36-icons": "^4.0.4",
+    "@contentful/f36-pagination": "^4.1.0-beta",
+    "@contentful/f36-tokens": "^4.0.0",
+    "@contentful/f36-utils": "^4.0.0",
+    "@emotion/core": "^10.0.17",
```
gui-santos, line 21: two emotions? 👀

gui-santos (approved):
LGTM!

I just found weird that we had to install another emotion package in the website

packages/website/package.json
massao: This is because `emotion@10` packages were a mess, basically `@emotion/core` had some functions to interact with the jsx, they even renamed it to `@emotion/react` on the v11.

But the `cx` function that we use, comes from `emotion`, that is the framework agnostic package, so we import it from there, but the `Global` we use for the `GlobalStyles`, comes from `@emotion/core`, so we import from both packages.

And to make things even weirder, both have the `css` method, but the `Global` and the `cx` aren't on both 😓 

Other solution would be to replace the `Global` with the `injectGlobal` and we would be only using the `emotion` package. 🤔

packages/forma-36-tokens/tools/build.js
```diff
@@ -1,7 +1,35 @@
...
+}
+
+function pickBy(obj, predicate) {
+  return Object.fromEntries(Object.entries(obj).filter(predicate));
+}
+
+function findKey(obj, predicate) {
+  const [key] = Object.entries(obj).find((entry) => predicate(entry[1]));
+  return key || undefined;
+}
```
gui-santos, line 32: 👏  👏  👏  👏  👏  👏 

with a couple of simple functions we just avoid importing a big package

gui-santos (approved):

packages/website/package.json
gui-santos: uhmmmmm maybe we another PR we try to consolidate that

PR #1930, opened 2022-03-03
fix: fix entry card description

burakukula:
Fixing the issue with card description
Screenshot 2022-03-03 at 19 21 39

Fix for https://contentful.atlassian.net/browse/ZEND-1898

### DIFF
--- a/packages/components/card/src/EntryCard/EntryCard.tsx
+++ b/packages/components/card/src/EntryCard/EntryCard.tsx
@@ -56,7 +56,7 @@ function EntryCardDescription({
   const truncatedDescription = truncate(description, 95, {});
 
   return (
-    <Paragraph marginBottom="none" isTruncated>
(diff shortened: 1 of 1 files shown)


denkristoffer (approved):

packages/components/card/src/EntryCard/EntryCard.tsx
```diff
@@ -56,7 +56,7 @@ function EntryCardDescription({
   const truncatedDescription = truncate(description, 95, {});
```
denkristoffer, line 56: Is this not working or what?

burakukula: it is, but the problem is actually `isTruncated` prop. as far as I understand they are doing the same thing, and we need to break-word to handle this case.

PR #1938, opened 2022-03-09
fix(note): allow rendering children in div

denkristoffer:
# Purpose of PR

The current `Note` component doesn't allow us to use `<Paragraph />` children because you aren't supposed to wrap `<p />` elements inside each other. I think we should allow this, so this is a proposal on how to achieve that. The tradeoff is that if we only pass a string as child it doesn't get wrapped in a paragraph element.

tl;dr: I want to be able to do this:

```tsx
<Note>
  <Paragraph>This is not allowed</Paragraph>
  <Paragraph>Because Note already renders a p element</Paragraph>
</Note>
```


burakukula:
@denkristoffer yeah, I am wondering about the impact of changing all the existing `Note` content to be actually a `div` not paragraph. I guess it's not a huge problem. From a11y stand point it is ok as far as I know, so I guess we should be fine with the change 🤔

denkristoffer:
@burakukula 

> yeah, I am wondering about the impact of changing all the existing Note content to be actually a div not paragraph. I guess it's not a huge problem.

I can't think of any.

gui-santos (approved):

denkristoffer:
I can add a check with `typeof children === 'string'` and then wrap children in a `Paragraph` if that's `true`, do you feel that's sufficient?

denkristoffer:
"Broken" is a bit strong here, I'd say 😉 But I pushed a proposal with the previous idea.

PR #1939, opened 2022-03-09
docs: use placeholder images from placeholder.pics

denkristoffer:
# Purpose of PR

The current placeholder service isn't working with Chromatic at the moment. I found this one that might be useful instead, although I'm not sure how much load it can handle: https://placeholder.pics/

The service insists on adding a border to the image, which I'm not a fan of. People might confuse it for a border in the library. You can see what it looks like on Chromatic. Thoughts?


denkristoffer:
placeholder.com is working again 😌

PR #1959, opened 2022-03-18
docs: add changesets usage to contributing guide

massao:
# Purpose of PR

Updating `Commits & Release` section on Readme and `Contributing Guide` as we now use changeset and only release new versions of packages that contain a changeset with the new version and notes


thebesson (approved):
👏

PR #1973, opened 2022-03-23
docs: integrate contentful to F36 website

burakukula:
Adding contentful to Forma 36 website:
 - setup
 - moving content of introduction and guidelines to Contentful
 
 TODO in the future:
  - dynamically generate side navigation based on category and subcategory
  - move tokens content to Contentful as well


packages/website/lib/api.ts
```diff
@@ -0,0 +1,94 @@
...
+  );
+  return extractArticle(entry);
+}
+
+export async function getAllArticles(preview = false) {
+  const entries = await fetchGraphQL(
+    `query {
+      kbAppArticleCollection(where: { slug_exists: true }) {
+        items {
+          ${ARTICLE_GRAPHQL_FIELDS}
```
burakukula, line 87: I'm gonna use simpler query over here. only thing we need here is slug, but on the separate PR.

packages/website/pages/[...slug].tsx
```diff
@@ -58,76 +53,142 @@ export default function ComponentPage({
...
-          {
-            nav: false,
-            headings: ['h1', 'h2', 'h3'],
-            customizeTOC: (t) => {
-              toc = transformToc(t) as any;
-              return false;
+// [WIP]: I started adding stronger types to this function but I could not finish
+// that's why I renamed some stuff here, I was trying to follow this:
+// https://nextjs.org/docs/basic-features/typescript#static-generation-and-server-side-rendering
+export const getStaticProps = async (context: {
```
Comment (bgutsol) [inline packages/website/pages/[...slug].tsx:89]: I would separate some parts into functions because right now it's a bit hard to read and understand. For example the part with shortIntroText can be isolated

bgutsol (approved):
Nice work 👏
No blockers, the only comment is that I would consider refactoring `[...slug].tsx` page as a follow-up.

Lelith (approved):
Looks good to me!

PR #1980, opened 2022-04-01
docs(controlled-accordion): remove types

denkristoffer:
# Purpose of PR

Remove types from ControlledAccordion example. I wish we could add types but they are breaking the example 😅


burakukula (approved):
sure, curious why they are breaking 🤔

PR #1985, opened 2022-04-05
fix(workbench-header): support undefined icon prop

denkristoffer:
# Purpose of PR

The `WorkbenchHeader` doesn't require the `icon` prop, but the current implementation tries to render a JSX element from the prop even when it's undefined. This PR fixes it by explicitly checking for `undefined` first.


Lelith (approved):
Looks good to me!

burakukula (approved):

PR #1994, opened 2022-04-13
fix(modal): don't scale modal with size="zen"

denkristoffer:
# Purpose of PR

### DIFF
--- a/.changeset/long-onions-behave.md
+++ b/.changeset/long-onions-behave.md
@@ -1,0 +1,5 @@
+---
+"@contentful/f36-modal": patch
+---
+
+fix(modal): fix scalign issues for modals with size "zen"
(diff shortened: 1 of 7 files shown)


Lelith (approved):
Looks good to me 👍  
But this might be a really stupid question, why is it named zen? as like in "zen buddhism"?

denkristoffer:
>  But this might be a really stupid question, why is it named zen? as like in "zen buddhism"?

Great question! I don't know, it was before my time 😆 I've always assumed it was about providing a zen state, something meditative, uninterrupted or focused.

.changeset/long-onions-behave.md
```diff
@@ -0,0 +1,5 @@
+---
+"@contentful/f36-modal": patch
+---
+
+fix(modal): fix scalign issues for modals with size "zen"
```
massao, line 5:
```suggestion
fix(modal): fix scaling issues for modals with size "zen"
```

massao (approved):
Small typo but looks good

PR #2003, opened 2022-04-22
feat: generate links for the sidebar from contentful

gui-santos:
# Purpose of PR

This will start generating a json file with the links that come from Contentful so we can add them at the sidebar automagically, see this (I've wrote this "Adding icons" page recently in Contentful):

Screenshot 2022-04-22 at 16 55 24

I also added the "What’s new" link to the Introduction section

Screenshot 2022-04-22 at 16 55 07

## PR Checklist

- [ ] Tests are passing


packages/website/utils/contentfulSidebarLinks.json
```diff
@@ -0,0 +1,46 @@
+{
```
denkristoffer, line 1: Should this file be gitignored then?

denkristoffer:
Awesome!!

Am I understanding this right that you want to allow links from both Contentful and a local file? If so, could you explain the rationale of not going with just one of them?

packages/website/utils/contentfulSidebarLinks.json
gui-santos: I didn't ignore it because the other one was not ignored

Should I ignore both of them?

gui-santos:
> Awesome!!
> 
> Am I understanding this right that you want to allow links from both Contentful and a local file? If so, could you explain the rationale of not going with just one of them?

@denkristoffer 
We have pages that come from Contentful and pages that come from mdx files in the repo, so we need to extract sidebar links from both sources

packages/website/utils/contentfulSidebarLinks.json
denkristoffer: I'd say so, it should be generated on build, right?

gui-santos: Screenshot 2022-04-25 at 15 13 26

we were wrong, we need these files

denkristoffer (approved):

Lelith (approved):
Looks good to me!

PR #2008, opened 2022-04-25
fix: fix wrong type definition

massao:
# Purpose of PR

Fixes types being generated wrongly

## Testing
Checkout branch and run `yarn build`, and check that the generated `d.ts` files don't have a `declare export namespace`

### DIFF
--- a/.changeset/giant-seas-deny.md
+++ b/.changeset/giant-seas-deny.md
@@ -1,0 +1,12 @@
+---
+'@contentful/f36-autocomplete': patch
+'@contentful/f36-card': patch
+'@contentful/f36-entity-list': patch
+'@contentful/f36-list': patch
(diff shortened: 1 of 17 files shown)


denkristoffer (approved):

Lelith:
Thank you for taking care. As i am a TS noob, can you maybe explain / link something that describes the difference between having them as vanilla functions vs arrow functions?

massao:
> Thank you for taking care. As i am a TS noob, can you maybe explain / link something that describes the difference between having them as vanilla functions vs arrow functions?

There shouldn't be any difference, is just that for some reason, when we were using the function declaration, the built type definition was created wrongly.
This is probably an issue with the build configuration itself, but it can take a while to figure out, this is a quick fix so we have the types correctly generated.

PR #2009, opened 2022-04-25
feat: promote pagination to stable

massao:
# Purpose of PR

* Remove private flag from `Pagination` so it's published with other packages.
* Add it to the umbrella pacakge `@contentful/f36-components`.
* Update usage on website.


Lelith (approved):
Looks good to me!

PR #2044, opened 2022-05-05
refactor: [] [Card] don't need to pass `withDragHandle` when `dragHandleRender` is provided

loweisz:
# Purpose of PR

It is more of a suggestion. When providing a `dragHandleRender` to the `Card` you also have to provide `withDragHandle`. 
But can we not assume that we want to render a drag handler when providing the render function? 

So I flipped the condition so that when providing the `dragHandleRender` you don't need to pass the `withDragHandle` anymore

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/packages/components/card/src/BaseCard/BaseCard.tsx
+++ b/packages/components/card/src/BaseCard/BaseCard.tsx
@@ -191,10 +191,10 @@ function _BaseCard<E extends React.ElementType = typeof BASE_CARD_DEFAULT_TAG>(
       testId={testId}
       title={title}
     >
-      {withDragHandle
-        ? dragHandleRender
-          ? dragHandleRender({ drag, isDragging })
-          : drag
+      {dragHandleRender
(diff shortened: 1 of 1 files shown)


bgutsol:
I think it's a breaking change. Because, if someone uses it in the way where they always pass `dragHandleRender` and  only change `withDragHandle` flag to enable/disable dragging, introducing your change will keep dragging enabled all the time, even when  `withDragHandle={false}`. 

Also, IMO it hasn't enough value to introduce as a breaking change. And on the other side, in such a way we have a single-responsibility principle. So flag `withDragHandle` is responsible for enabling/disabling and `dragHandleRender` is responsible for rendering custom Drag or default one.

@loweisz what do you think?

loweisz:
@bgutsol That is actually a good point, I agree. However I think we should the adjust the type of the `Card` component that it accepts the `withDragHandle` to be passed to it. 

But I will close this PR then :)

PR #2050, opened 2022-05-12
style: fix autocomplete input with button styling [HOMER-856]

Lelith:
### Description
This fixes the styling of the autocomplete input. 
- input placeholder and typed in text was flowing underneath the icon button
- the icon button was having too much spacing around and so it did not look as integrated part of the autocomplete input.

**Before**
before_hover
Too much space around the button makes it look hovering weird and distached from the input field, instead of beeing part of it

before_long_inputtext
Long input texts would overflow underneath the X button and look broken

before_clearstate
long input texts would just dissapear undernath the button and be cut off.

**After**
after_dropdown
Button is now tucked fitting into the input field with no visual space

after_clearstate
Long selections and palceholders get overflown via ellipsis

after_long_inputtext
Long input texts stop before the button

### DIFF
--- a/packages/components/autocomplete/src/Autocomplete.styles.ts
+++ b/packages/components/autocomplete/src/Autocomplete.styles.ts
@@ -10,7 +10,7 @@ export const getAutocompleteStyles = (listMaxHeight: number) => ({
     position: 'relative',
   }),
   inputField: css({
-    paddingRight: '32px',
+    paddingRight: tokens.spacingXl,
     textOverflow: 'ellipsis',
     whiteSpace: 'nowrap',
   }),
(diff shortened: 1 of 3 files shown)


giotiskl (approved):

packages/components/autocomplete/src/Autocomplete.styles.ts
```diff
@@ -9,12 +9,18 @@ export const getAutocompleteStyles = (listMaxHeight: number) => ({
   combobox: css({
     position: 'relative',
   }),
+  inputField: css({
+    paddingRight: '32px',
```
mshaaban0, line 13: This could use `spacingXl` token it is `32px`

mshaaban0 (approved):
Looks good. 👍🏼

PR #2051, opened 2022-05-12
fix: workbench header icon alignments

mshaaban0:
# Purpose of PR

Align Workbench header icons properly
**Before**
Screenshot 2022-05-12 at 16 40 16

**After**
Screenshot 2022-05-12 at 16 39 52

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/components/button/src/Button/Button.tsx
+++ b/packages/components/button/src/Button/Button.tsx
@@ -82,7 +82,7 @@ function _Button<E extends React.ElementType = typeof BUTTON_DEFAULT_TAG>(
     <>
       {startIcon && iconContent(startIcon)}
       {children && (
-        <Box as="span" display="inline-flex" className={styles.buttonContent}>
+        <Box as="span" display="block" className={styles.buttonContent}>
           {children}
         </Box>
       )}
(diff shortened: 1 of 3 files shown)


packages/components/button/src/Button/Button.tsx
```diff
@@ -82,7 +82,7 @@ function _Button<E extends React.ElementType = typeof BUTTON_DEFAULT_TAG>(
     <>
       {startIcon && iconContent(startIcon)}
       {children && (
-        <Box as="span" display="block" className={styles.buttonContent}>
```
bgutsol, line 85: Is this fix also related to Workbench fix?

bgutsol (commented):
Workbench fix looks good. But I would ask you to revert the Button change

packages/components/button/src/Button/Button.tsx
mshaaban0: @bgutsol Yes it is related. SVGs inside block elements adds extra space "browser behaviour" display `inline-flex` is one way to fix it

bgutsol: But we don't pass SVGs as children to the button, we have `startIcon/endIcon` props for that. What exactly you are trying to fix with it?

mshaaban0: Reverted: Back button in workbench was using button and injecting an icon as a children. Instead I refactored it to use IconButton which uses `flex` 🤷🏼

bgutsol: Awesome, thank you 🙌

bgutsol (approved):

PR #2052, opened 2022-05-13
feat: make native tooltip title optional []

MohammedAl-Rowad:
# Purpose of PR

Make the [title tooltip](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title) optional in the `<Pill />` component 

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/components/pill/src/Pill.tsx
+++ b/packages/components/pill/src/Pill.tsx
@@ -15,6 +15,12 @@ export type PillInternalProps = CommonProps & {
    * Text that will be shown on the pill
    */
   label: string;
+
+  /**
+   * A boolean that tells if the label should be used as a native tooltip title
+   * @default true
+   */
+  useLabelAsTitle?: boolean;
   /**
    * Function that handles when the close icon is clicked. Close icon visibility depends on if this property is set.
    */
@@ -43,6 +49,7 @@ export const Pill = React.forwardRef<HTMLDivElement, ExpandProps<PillProps>>(
       onClose,
       testId = 'cf-ui-pill',
       onDrag,
+      useLabelAsTitle = true,
       className,
       dragHandleComponent,
       variant = 'idle',
@@ -67,7 +74,7 @@ export const Pill = React.forwardRef<HTMLDivElement, ExpandProps<PillProps>>(
               <DragIcon className={styles.icon} variant="muted" />
             </span>
           ))}
-        <span title={label} className={styles.label}>
+        <span title={useLabelAsTitle ? label : null} className={styles.label}>
           {label}
         </span>
         {onClose && (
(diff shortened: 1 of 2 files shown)


z0al (commented):
I'd defer to F36 folks for the final say but what about we simply override the `title` implementation to use the `Tooltip` component without an additional option instead?

MohammedAl-Rowad:
> I'd defer to F36 folks for the final say but what about we simply override the `title` implementation to use the `Tooltip` component without an additional option instead?

I'm fine with this, let's hear what the F36 team have to say

Lelith:
Hey, i am not sure if i understand the purpose of this change. Whats the reason behind not having the label double as tooltip? Is this not exactly what the linked ticket would like you to do? Showing a shortened text on hover as the original text?

MohammedAl-Rowad:
> Hey, i am not sure if i understand the purpose of this change. Whats the reason behind not having the label double as tooltip? Is this not exactly what the linked ticket would like you to do? Showing a shortened text on hover as the original text?

I was adding a [Tooltip](https://f36.contentful.com/components/tooltip) over a [Pill](https://f36.contentful.com/components/pill)

```jsx
      <Tooltip maxWidth="none" isDisabled={!textIsEllipsis} content={tag.label}>
        <Pill
          ref={trackRefChange}
          className={styles.tag}
          key={tag.value}
          label={tag.label}
          onClose={isDisabled ? undefined : () => onTagPillClose(tag.value)}
        />
      </Tooltip>
```

which will result in this 

Screenshot 2022-05-16 at 10 42 05

So we thought we could make the pill use the forma tooltip by default or make the [usage of the title optional](https://github.com/contentful/forma-36/blob/2b6d93caaf3f870919bb60745634462e3456a615/packages/components/pill/src/Pill.tsx#L70)

denkristoffer:
Cool, in that case we should move to the Tooltip component 👍

MohammedAl-Rowad:
> Cool, in that case we should move to the Tooltip component 👍

I've used the tooltip component, but I don't think we should always show it, maybe when there is ellipses in the text

that's what I did, what do you think?

Lelith (approved):

packages/components/pill/src/Pill.tsx
```diff
@@ -50,6 +52,18 @@ export const Pill = React.forwardRef<HTMLDivElement, ExpandProps<PillProps>>(
     } = props;
 
     const styles = getPillStyles(variant);
+    const [textIsEllipsis, setTextIsEllipsis] = React.useState(false);
```
denkristoffer, line 55: I'd prefer a clearer name here so that someone who's new to the project could understand what it means. Something like `textIsTruncated` maybe, or `textOverflowsContainer`.

denkristoffer (approved):

MohammedAl-Rowad:
replaced by 
https://github.com/contentful/forma-36/pull/2057

Issue #2046, opened 2022-05-09
🐛 Bug - collapsable component example in the documentation is slightly broken


burakukula:
https://user-images.githubusercontent.com/4272331/167413958-ccb57df6-5d6e-437c-ad64-d008e4779b4d.mov

Lelith:
will be fixed by : https://github.com/contentful/forma-36/pull/2062

PR #2062, opened 2022-05-31
fix: collapse example

Lelith:
the button was jumping when opening the collapsable element.


bgutsol (approved):
🙌

burakukula (approved):

burakukula:
closes https://github.com/contentful/forma-36/issues/2046

PR #2065, opened 2022-06-01
fix: add local zIndex scope for buttonGroup

bgutsol:
# Purpose of PR

- Add stacking context for the ButtonGroup and InputGroup components. It should fix the bug when the group elements are stacking not within the Group’s context, but within the context of the parent element (because the ButtonGroup/InputGroup itself didn't create a stacking context)


denkristoffer (approved):
Should we do the same for `InputGroup`?

mshaaban0 (approved):

PR #2087, opened 2022-06-13
feat: add Datepicker component [BAU-690]

bgutsol:
# Purpose of PR

- add Detepicker component
- calendar types adjustments

## Open questions:
- should we provide **input mask** like `dd/mm/yyyy` so when you type it automatically adds slashes `/`. (e.g. [MUI has it](https://mui.com/x/react-date-pickers/date-picker/))


packages/components/datepicker/src/Datepicker.tsx
```diff
@@ -1,27 +1,184 @@
...
+import { Popover } from '@contentful/f36-popover';
+import { DayPickerSingleProps } from 'react-day-picker';
+import FocusLock from 'react-focus-lock';
+import { TextInput, TextInputProps } from '@contentful/f36-forms';
+import { IconButton } from '@contentful/f36-button';
+import { CalendarIcon } from '@contentful/f36-icons';
+import { PopoverProps } from '@contentful/f36-popover/src';
 
-export type DatepickerProps = CommonProps;
+const DATE_FORMAT = 'dd/MM/yyyy';
```
denkristoffer, line 21: Does this format match how we're showing dates in products? We don't use `–` or `.`?

packages/components/datepicker/src/Datepicker.tsx
```diff
@@ -1,27 +1,184 @@
...
 
-function _Datepicker(props: DatepickerProps, ref: React.Ref<HTMLDivElement>) {
+// Props comparison to our current datepickers:
+// Current      DayPicker
+// value        selected
+// onChange     onSelect
+// min          fromDate
+// max          toDate
```
denkristoffer, line 28: Is there a reason to keep this in both the README and here? I prefer README, so we don't have to update/remove two places.

packages/components/datepicker/src/Datepicker.tsx
```diff
@@ -1,27 +1,184 @@
...
+            id={id}
+            isInvalid={isInvalid || !isDateValid(parseInputDate(inputValue))}
+            isDisabled={isDisabled}
+            isRequired={isRequired}
+            isReadOnly={isReadOnly}
+          />
+          <IconButton
+            aria-label="Choose Date"
+            variant="secondary"
+            icon={<CalendarIcon aria-label="Choose Date" variant="muted" />}
```
denkristoffer, line 158: I don't think we need both labels

packages/components/datepicker/src/Datepicker.tsx
```diff
@@ -1,27 +1,184 @@
...
+            value={inputValue}
+            onChange={handleInputChange}
+            id={id}
+            isInvalid={isInvalid || !isDateValid(parseInputDate(inputValue))}
+            isDisabled={isDisabled}
+            isRequired={isRequired}
+            isReadOnly={isReadOnly}
+          />
+          <IconButton
+            aria-label="Choose Date"
```
denkristoffer, line 156:
```suggestion
            aria-label="Choose date"
```

packages/components/menu/src/MenuItem/MenuItem.tsx
```diff
@@ -33,7 +33,7 @@ function _MenuItem<E extends React.ElementType = typeof MENU_ITEM_DEFAULT_TAG>(
 ) {
   const { testId, className, as, isInitiallyFocused, ...otherProps } = props;
 
-  const id = useId(null, 'menu-item');
+  const id = useId(undefined, 'menu-item');
```
denkristoffer, line 36: What are these changes about? Is there a type problem?

packages/components/datepicker/src/Datepicker.tsx
```diff
@@ -1,27 +1,184 @@
...
+    // we want to run this hook only when `selected` prop changes
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [selected]);
+
+  const handleInputChange: ChangeEventHandler<HTMLInputElement> = useCallback(
+    (e) => {
+      setInputValue(e.currentTarget.value);
+
+      const date = parseInputDate(e.currentTarget.value);
+      console.log('handleInputChange', { date });
```
denkristoffer, line 111:
```suggestion
```

packages/components/datepicker/src/Datepicker.tsx
```diff
@@ -1,27 +1,184 @@
...
+      if (isDateValid(date)) {
+        console.log('isDateValid');
```
denkristoffer, line 113:
```suggestion
```

packages/components/datepicker/src/Datepicker.tsx
```diff
@@ -1,27 +1,184 @@
...
+        onSelect(date);
+      } else {
+        console.log('NOT isDateValid');
```
denkristoffer, line 116:
```suggestion
```

packages/components/datepicker/README.mdx
```diff
@@ -13,6 +13,14 @@ Short component description covering when it's used.
...
 Optional: Most common misuse and what to use instead.
 
+Props comparison to our current datepickers:
+| Current | DayPicker |
+|---|---|
+| value | selected |
+| onChange | onSelect |
+| min | fromDate |
+| max | toDate |
+
```
denkristoffer, line 23: This doesn't work well on the website, maybe we can do a proper markdown table instead?

Screenshot 2022-06-13 at 15 42 49

denkristoffer (commented):
This looks AMAZING! I'm really happy about it 😄 I have a few comments though.

And I tried it on the website, there's an issue at the moment where the field renders its initial state as invalid before the user has touched it: https://forma-36-nnbkkdi2j.colorfuldemo.com/components/datepicker#example-title-variant-name=

packages/components/datepicker/src/Datepicker.tsx
bgutsol: Mostly we use such format: 
image

But I agree that we have to double-check it with the design team. @damann what do you think?

packages/components/datepicker/README.mdx
bgutsol: I believe @mshaaban0 will adjust it while working on docs.

packages/components/menu/src/MenuItem/MenuItem.tsx
bgutsol: yes, for some reason the build for the website was failing because of the types mismatch. Although I have no idea why it started failing only on this particular PR.

bgutsol:
> This looks AMAZING! I'm really happy about it 😄 I have a few comments though.
> 
> And I tried it on the website, there's an issue at the moment where the field renders its initial state as invalid before the user has touched it: https://forma-36-nnbkkdi2j.colorfuldemo.com/components/datepicker#example-title-variant-name=

@denkristoffer thanks for the feedback, the issue on the website should be fixed now

denkristoffer:
Looks great. One thought, what's the logic behind rendering the month and year with a select when there's a minimum or maximum date, but making it static in other cases?

denkristoffer (approved):
Okay that makes sense, but it still gives me a weird "new" experience based on that prop. I'd consider removing in a follow-up if we decide on that 🙂 

Last thing, it would be cool if the datepicker was open by default for the stories so that Chromatic can pick up on changes.

bgutsol:
> Okay that makes sense, but it still gives me a weird "new" experience based on that prop. I'd consider removing in a follow-up if we decide on that 🙂
> 
> Last thing, it would be cool if the datepicker was open by default for the stories so that Chromatic can pick up on changes.

Do you think it’s worth to introduce controlled props for Datepicker (possibility to control open/closed from outside)? Otherwise we can provide `defaultIsOpen` prop

packages/components/datepicker/src/Datepicker.tsx
```diff
@@ -1,27 +1,177 @@
...
+ * Provides functionality for date selection.
+ */
+export function Datepicker(props: DatepickerProps) {
   const styles = getStyles();
+  const {
+    //common props
+    testId = 'cf-datepicker',
+    className,
+    style,
+    // input props
```
mshaaban0, line 45: Do you think we should group these? 
```js
{
  //...
  inputProps: { id, isDisabled, ...otherInputProps },
  popoverProps: { ... },
  calendarProps: { ... },
  //...
}
```
Users might need to pass some more of the components props then they'll have to add it here first

packages/components/datepicker/src/Datepicker.tsx
```diff
@@ -1,27 +1,177 @@
...
+    isRequired,
+    // popover props
+    usePortal,
+    // calendar props
+    selected,
+    onSelect,
+    fromDate,
+    toDate,
+    locale,
+    ...restCalendarProps
```
mshaaban0, line 59: nit: we use `otherProps` across our components

packages/components/datepicker/src/Datepicker.tsx
```diff
@@ -1,27 +1,177 @@
...
 
   return (
-    <div
-      data-test-id={props.testId}
-      ref={ref}
-      className={cx(styles.datepicker, props.className)}
+    <Popover
+      isOpen={isPopoverOpen}
+      onClose={() => setIsPopoverOpen(false)}
+      usePortal={usePortal}
```
mshaaban0, line 130: this will be undefined by default, so we might need to just to spread the props here.

packages/components/datepicker/src/Datepicker.tsx
bgutsol: I don't like the idea of having these separate prop objects for each component we use inside.
For example props like `selected`, `onSelect`, `fromDate`, `toDate` - are not specifically calendar Props they are used to define input behavior as well. Same about input props like `isDisabled`, `isInvalid`, they are not about input only, but about the whole Datepicker.
If we want it to be such a flexible component we should might consider make it a compound one, or smth similar with render functions.
Otherwise, users should see it as one complete component and they don't care what it's built from, IMO.

What do you think? Maybe you can provide some references where such pattern is used well, the one with grouped props.

packages/components/datepicker/src/Datepicker.tsx
```diff
@@ -1,27 +1,177 @@
...
+    (value: string) => parse(value, DATE_FORMAT, new Date(), { locale }),
+    [locale],
+  );
+
+  const [isPopoverOpen, setIsPopoverOpen] = useState(false);
+  const [inputValue, setInputValue] = useState<string>(() =>
+    selected ? format(selected, DATE_FORMAT) : '',
+  );
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
bgutsol: Unfortunately no, because selected can be changed outside as well.

packages/components/datepicker/src/Datepicker.tsx
bgutsol: I made it like this to prevent bleeding it to the `Calendar` component through `otherProps`

packages/components/datepicker/src/Datepicker.tsx
bgutsol: done

bgutsol:
> > Okay that makes sense, but it still gives me a weird "new" experience based on that prop. I'd consider removing in a follow-up if we decide on that 🙂
> > Last thing, it would be cool if the datepicker was open by default for the stories so that Chromatic can pick up on changes.
> 
> Do you think it’s worth to introduce controlled props for Datepicker (possibility to control open/closed from outside)? Otherwise we can provide `defaultIsOpen` prop

@denkristoffer 👆

denkristoffer:
I'd say `defaultIsOpen` is good for now, leaves it open for us to add controlled props at a later point if we ever need to.

packages/components/datepicker/src/Datepicker.tsx
denkristoffer: I think we should split anything going directly to subcomponents. We used to have a style guide to define this pattern, but I can't find it at the moment. @mshaaban0 do you have it?

mshaaban0 (approved):
Thanks for addressing the comments. Tested and looking good.

packages/components/datepicker/src/Datepicker.tsx
mshaaban0: Just to be clear, this can be in another PR once we agree on an approach

Issue #2091, opened 2022-06-14
🐛 Bug - F36 v4 codemod migration SelectField missing selectProps


andreascful:
# Forma 36 bug report

## Summary

When trying to migrate a `SelectField` via the codemod migration the `selectProps` prop is not being migrated to the newly created `Select` component. 
Screenshot shows an example of what happens in such a case.

![image](https://user-images.githubusercontent.com/93705529/173535247-8e125975-8dbf-4769-8bc6-2cd75673b22a.png)

## Expected results

The `selectProps` prop should be migrated to the newly created `Select` component.

PR #2092, opened 2022-06-14
fix: improve TextField and SelectField codemods

### LINKED ISSUE #2086: 🐛 Bug - F36 v4 codemod migration TextField conditional validation
# Forma 36 bug report

## Summary

When trying to migrate to Forma36 v4 via the provided codemods the following error will occur when there is a TextField component that contains inline conditional logic: 

```
Transformation error (undefined does not match field "name": string of type Identifier)
Error: undefined does not match field "name": string of type Identifier
```

Example of the TextField:

```
<TextField
        labelText="ID"
        required
        name={`field.${fields.id.name}`}
        id={`field.${fields.id.name}`}
        type="text"
        value={fields.id.value}
        onChange={(ev) => {
          SetFieldValue({ name: fields.id.name, value: ev.target.value });
        }}
        validationMessage={
          fields.id.errors.length > 0
            ? fields.id.errors.map((error) => error.message).join('\n')
            : ''
        }
      />
```

massao:
# Purpose of PR

* Make TextField codemod work with binary conditions for validation message.
* Add transformer for selectProps on the SelectField codemod.

Closes #2086 #2091

### DIFF
--- a/.changeset/dull-ads-breathe.md
+++ b/.changeset/dull-ads-breathe.md
@@ -1,0 +1,6 @@
+---
+'@contentful/f36-codemod': minor
+---
+
+- Add support for binary conditional on TextField validation mesasge
+- Add tranformer to selectProps for SelectField codemod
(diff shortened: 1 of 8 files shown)


andreascful (approved):
Looks great. Thank you for the improvements!

bgutsol (approved):
magic 🧙‍♂️

PR #2093, opened 2022-06-16
fix: enable keeping the dropdown open after select

Lelith:
Issue description:
For multi-selects,clicking again on the input did not open the drawer again, because the element was still in focus. Which makes sense from the keyboards perspective. So improve the UX around multi-selects allow to keep the dropdown open after selection.

### DIFF
--- a/packages/components/autocomplete/src/Autocomplete.tsx
+++ b/packages/components/autocomplete/src/Autocomplete.tsx
@@ -76,6 +76,11 @@ export interface AutocompleteProps<ItemType>
    */
   clearAfterSelect?: boolean;
   /**
+   * If this is set to `true` the dropdown menu will stay open after the select
+   * @default false
+   */
+  closeAfterSelect?: boolean;
+  /**
    * This is the value will be passed to the `placeholder` prop of the input.
    * @default "Search"
    */
@@ -130,6 +135,7 @@ function _Autocomplete<ItemType>(
     id,
     className,
     clearAfterSelect = false,
+    closeAfterSelect = true,
     defaultValue = '',
     selectedItem,
     items,
@@ -217,6 +223,9 @@ function _Autocomplete<ItemType>(
           }
           if (clearAfterSelect) {
             handleInputValueChange('');
+          }
+          if (!closeAfterSelect) {
+            toggleMenu();
           }
           break;
         default:
--- a/packages/components/autocomplete/examples/AutocompleteMultiSelectionExample.tsx
+++ b/packages/components/autocomplete/examples/AutocompleteMultiSelectionExample.tsx
@@ -37,6 +37,7 @@ export default function AutocompleteMultiSelectionExample() {
         renderItem={(item) => item.name}
         // When this prop is `true`, it will clean the TextInput after an option is selected
         clearAfterSelect
+        closeAfterSelect={false}
       />
 
       <span>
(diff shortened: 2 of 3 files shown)


andreascful (approved):
Lgtm! Thank you.
I think this is a convenience feature that users would really miss if we'd implement it.

packages/components/autocomplete/src/Autocomplete.tsx
```diff
@@ -75,6 +75,11 @@ export interface AutocompleteProps<ItemType>
    * @default false
    */
   clearAfterSelect?: boolean;
+  /**
+   * If this is set to `true` the dropdown menu will stay open after the select
+   * @default false
```
mshaaban0, line 80: true or false? :)

Lelith: sorry, default will be true, will update the text

packages/components/autocomplete/README.mdx
```diff
@@ -74,6 +76,14 @@ Besides the correct shape of the object the Autocomplete component needs to rece
 
 ```
 
+### Fully controlled selection from outside
+
+In order to use proper form validation you need to be able to control the actual input field from the autocomplete, because the search query value is not the actual selection. This is done via the selectedItem property.
+
+```jsx file=./examples/ControlledAutocomplete.tsx
```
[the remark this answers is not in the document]
Lelith: yes, i wrote the example, but forgot to link it in the readme :)

packages/components/autocomplete/src/Autocomplete.tsx
```diff
@@ -75,6 +75,11 @@ export interface AutocompleteProps<ItemType>
    * @default false
    */
   clearAfterSelect?: boolean;
+  /**
+   * If this is set to `true` the dropdown menu will stay open after selecting an item
```
denkristoffer, line 79:
```suggestion
   * If this is set to `false` the dropdown menu will stay open after selecting an item
```

I think my suggestion got messed up.

denkristoffer, line 79:
```suggestion
   * If this is set to `true` the dropdown menu will close after selecting an item
```

This is what I wanted to do. The question is if it makes more sense to talk about the case where the menu stays open or the one where it closes.

Lelith: So my thoughts around this where:
I wanted to keep the prop name close to the "clearAfterSelect" one as well as not introducing a breaking change.
Which is why i named it "closeAfterSelect" we could also put the whole variable vice versa to have false as the default and name it "keepOpenAfterSelect" and then the default is false (no breaking change) and setting it to true is keeping the dropdown open.
So might be easier to read and use?

denkristoffer (approved):

.changeset/happy-fans-share.md
```diff
@@ -0,0 +1,5 @@
+---
+'@contentful/f36-autocomplete': patch
+---
+
+Enable controlling the dorpdown menu open and close state after select with new property "closeAfterSelect"
```
andreascful, line 5: I don't know if that is relevant but minor typo "dorpdown" :)

Lelith: 🧐 thanks

denkristoffer:
```suggestion
Enable controlling the menu open and close state after select with new property "closeAfterSelect"
```

PR #2095, opened 2022-06-17
docs: add Calendar and Datepicker components

mshaaban0:
# Purpose of PR

Add documentation for `Calendar` and `Datepicker` components on F36 Website

## PR Checklist

- [x] Tests are passing


packages/components/datepicker/README.mdx
```diff
@@ -6,22 +6,20 @@ slug: /components/datepicker/
...
-value       selected
-onChange    onSelect
-min         fromDate
-max         toDate
--->
+Use Datepicker to ask user's for date input in forms, modals, or filtering.
+
+<Flex>
+  <Badge variant="featured">Beta</Badge>
+</Flex>
```
denkristoffer, line 22: Not sure how we usually handle this, but I think this would look nicer in the title. What do you think?

Screenshot 2022-06-17 at 13 41 27

packages/components/datepicker/examples/Datepicker/Format.tsx
```diff
@@ -0,0 +1,22 @@
...
+  const [selectedDay, setSelectedDay] = useState(new Date());
+
+  return (
+    <>
+      <Datepicker
+        dateFormat="do LLL yyyy"
+        selected={selectedDay}
+        onSelect={setSelectedDay}
+      />
+      <br />
```
denkristoffer, line 14: Let's put these in a `Flex` instead for some increased niceness 😎

packages/components/datepicker/src/Calendar/README.mdx
```diff
@@ -0,0 +1,78 @@
+---
+title: 'Calendar'
+status: 'beta'
```
massao, line 3: If we want to show the `note` on the documentation page this should be set to Alpha, as we only check for Alpha and deprecated:
https://github.com/contentful/forma-36/blob/0b200eb7d815647873889df31368de9a9de94274/packages/website/components/PageContent/PageContentHeader.tsx#L61-L63

packages/components/datepicker/examples/Datepicker/DefaultOpen.tsx
```diff
@@ -0,0 +1,14 @@
+import React, { useState } from 'react';
+import { Datepicker } from '@contentful/f36-datepicker';
+
+export default function BasicExample() {
+  const [selectedDay, setSelectedDay] = useState(new Date());
+
+  return (
+    <Datepicker
+      defaultIsOpen
```
denkristoffer, line 9: This example is not open by default for me on the website 🤔

packages/components/datepicker/src/Calendar/README.mdx
massao: Or we could create a new note for Beta components, just need a copy for that, and we can have the same styling in general

packages/components/datepicker/README.mdx
massao: If we add the note we wouldn't need the badge, if I remember correctly we didn't add the badge for the pagination component when it was added, only the note

denkristoffer: That works for me.

packages/components/datepicker/examples/Datepicker/DefaultOpen.tsx
denkristoffer: You're right, I probably clicked in another example. I think we can keep it.

packages/components/datepicker/src/Calendar/README.mdx
mshaaban0: Thanks, I was looking for that 👍🏼

packages/components/datepicker/README.mdx
mshaaban0: Done, now it shows the alpha note

packages/components/datepicker/src/Calendar/README.mdx
```diff
@@ -0,0 +1,78 @@
...
+isNew: true
```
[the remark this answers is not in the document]
mshaaban0: Thanks, done.

packages/components/datepicker/README.mdx
```diff
@@ -1,27 +1,19 @@
 ---
 title: 'Datepicker'
-status: 'beta'
+status: 'alpha'
```
burakukula, line 3: Out of curiosity: Why this component is in alpha? Do we use `alpha`? I thought that we only use `beta` and then we move component to public release.

packages/components/datepicker/README.mdx
```diff
@@ -1,27 +1,19 @@
...
 isNew: true
```
burakukula, line 4: even though `isNew: true` is added I don't see the `<Badge>New</Badge>` in the sidebar navigation. It should be displayed the same way like it is for Pagination right?

burakukula:
This is not really related to Datepicker, probably more to the group, but just to validate with you. This is not exactly how we would like this to look, right? 🧐 
Screenshot 2022-06-20 at 09 52 52
Screenshot 2022-06-20 at 09 52 57

packages/components/datepicker/README.mdx
massao: This is because we had designs for the Alpha note but not for the Beta one, but we could use both actually.
But would need to have a copy for the Beta version note to display on the page

packages/components/datepicker/README.mdx
massao: This was because the prop wasn't being passed to items on subnavigations on the sidebar, it is fixed now

massao (approved):
LGTM

packages/components/datepicker/README.mdx
mshaaban0: Try again, Vercel didn't build the last commit.

packages/components/datepicker/README.mdx
mshaaban0: I couldn't find any guidelines or docs in the repo for releases/deployments. We will need to work on our release processes, and document it. for now we will stay with what we have `alpha` and as it's also make sense in [Software Release Life Cycle](https://en.wikipedia.org/wiki/Software_release_life_cycle)

burakukula: Both `alpha` and `beta` are aligned with [Software Release Life Cycle](https://en.wikipedia.org/wiki/Software_release_life_cycle), I don't mind really, I just think we should be consistent. We didn't use alpha for v4, this is where my questions is coming from. (We actually didn’t use it for `Pagination` neither as far as I remember? 🤔 ).

burakukula (approved):

PR #2099, opened 2022-06-22
test: add datepicker tests

massao:
# Purpose of PR

Add datepicker tests.

### DIFF
--- a/packages/components/datepicker/src/Datepicker.test.tsx
+++ b/packages/components/datepicker/src/Datepicker.test.tsx
@@ -57,7 +57,7 @@ describe('Datepicker', function () {
     });
   });
 
-  it('renders the calendar initialy open and close when esc key is pressed', async () => {
+  it('renders the calendar initialy open', () => {
     render(
       <Datepicker selected={testDate} onSelect={jest.fn()} defaultIsOpen />,
     );
@@ -67,6 +67,14 @@ describe('Datepicker', function () {
 
     expect(datepicker).toHaveAttribute('aria-expanded', 'true');
     expect(popoverContent).toBeInTheDocument();
+  });
+
+  it('should close the calendar when esc is pressed', async () => {
+    render(
+      <Datepicker selected={testDate} onSelect={jest.fn()} defaultIsOpen />,
+    );
+    const popoverContent = screen.getByTestId('cf-ui-popover-content');
+    const datepicker = screen.getByTestId('cf-ui-datepicker');
 
     act(() => {
       fireEvent.keyDown(document.activeElement, {
--- a/packages/components/datepicker/src/Datepicker.test.tsx
+++ b/packages/components/datepicker/src/Datepicker.test.tsx
@@ -10,6 +10,7 @@ import userEvent from '@testing-library/user-event';
 import { Datepicker } from './Datepicker';
 import { format } from 'date-fns';
 import { act } from 'react-dom/test-utils';
+import { axe } from 'jest-axe';
 
 describe('Datepicker', function () {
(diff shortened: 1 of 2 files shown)


packages/components/datepicker/src/Datepicker.test.tsx
```diff
@@ -1,11 +1,101 @@
...
+
+  it('renders with default date set', () => {
+    render(<Datepicker selected={testDate} onSelect={jest.fn()} />);
+
+    expect(screen.getByTestId('cf-ui-datepicker-input')).toHaveValue(
+      format(testDate, 'dd LLL yyyy'),
+    );
+  });
+
+  it('opens calendar when button is clicked with', async () => {
```
bgutsol, line 42:
```suggestion
  it('opens calendar when button is clicked', async () => {
```

packages/components/datepicker/src/Datepicker.test.tsx
```diff
@@ -1,11 +1,101 @@
...
+      });
+    });
+
+    await waitFor(() => {
+      expect(datepicker).toHaveAttribute('aria-expanded', 'false');
+      expect(popoverContent).not.toBeInTheDocument();
+    });
+  });
+
+  it('updates value and and trigger onSelect when clicking a day on calendar', async () => {
```
bgutsol, line 82:
```suggestion
  it('updates value and trigger onSelect when clicking a day on calendar', async () => {
```

packages/components/datepicker/src/Datepicker.test.tsx
```diff
@@ -1,11 +1,101 @@
...
+    act(() => {
+      userEvent.click(within(popover).getByText(newDate.getDay()));
+      rerender(
+        <Datepicker selected={newDate} onSelect={onSelect} defaultIsOpen />,
+      );
+    });
+    expect(screen.getByTestId('cf-ui-datepicker-input')).toHaveValue(
+      format(newDate, 'dd LLL yyyy'),
+    );
+    expect(onSelect).toHaveBeenCalledTimes(1);
```
bgutsol, line 99: I would leave here just an assertion for `onSelect` function, that it's been called once and with `newDate` as an argument.
Because when you check the input value on rerender it's basically the same test that you do in the `renders with default date set` above

bgutsol (commented):
Could you add also tests for:
- a11y with axe, for example like in [Popover tests](https://github.com/contentful/forma-36/blob/48a511bc48c17d3e4bf0cbfb8d16da8c3cbc29b2/packages/components/popover/src/Popover.test.tsx#L229)
- test on disabled datepicker
- test on datepicker with error

bgutsol (approved):
Beautiful 🏅

packages/components/datepicker/src/Datepicker.test.tsx
```diff
@@ -1,11 +1,131 @@
...
+      expect(screen.getByTestId('cf-ui-datepicker')).toHaveAttribute(
+        'aria-expanded',
+        'true',
+      );
+      expect(screen.getByTestId('cf-ui-popover-content')).toBeInTheDocument();
+      expect(screen.getByText(format(testDate, 'LLLL yyyy'))).toBeTruthy();
+    });
+  });
+
+  it('renders the calendar initialy open and close when esc key is pressed', async () => {
```
denkristoffer, line 60: I think this test should be split into two cases, one about `defaultIsOpen` and the other for closing on escape.

denkristoffer (commented):
One thought

denkristoffer (approved):

PR #2123, opened 2022-07-12
build(deps): remove webpack dependencies

denkristoffer:
# Purpose of PR

Looks to me like we can remove webpack as an explicit dependency. Let's see.


mshaaban0 (commented):
Nice, we have some docs in readme needs updating https://github.com/contentful/forma-36/blob/main/packages/forma-36-react-components/README.md#building

denkristoffer:
> Nice, we have some docs in readme needs updating [https://github.com/contentful/forma-36/blob/main/packages/forma-36-react-components/README.md#building](https://github.com/contentful/forma-36/blob/main/packages/forma-36-react-components/README.md?rgh-link-date=2022-07-12T20%3A30%3A55Z#building)

Wow, speak of outdated! Let me give that a go.

mshaaban0 (approved):
💅🏼

PR #2151, opened 2022-08-04
feat: calendar reexport custom components, provide docs example

bgutsol:
- Re-export some components from [react-day-picker](https://react-day-picker.js.org/guides/custom-components) lib that we use as a base.
- Provide an example with a custom component
- Update Parcel to the latest version (why not? 😄)


massao (approved):

burakukula:
I guess this will get fixed when we bump the datepicker package?

Screenshot 2022-08-04 at 15 18 53

bgutsol:
> I guess this will get fixed when we bump the datepicker package?
> 
> Screenshot 2022-08-04 at 15 20 01

Yes!

PR #2152, opened 2022-08-04
fix: update website with the component statuses [BAU-903]

burakukula:
Updated statuses that follows the release approach:
 - alpha then beta - gray `Badge` + gray `Note`
 - stable, but still new component (like Pagination) - blue `Badge`, we don't add `Note`
 - deprecated - red `Badge` + red `Note`
 
 Correct me if I am wrong, but I think the Note content will get updated as soon as the update README files will be merged:
Screenshot 2022-08-04 at 15 04 43

The content for the docs in here https://forma-36-git-cleanup-statuses.colorfuldemo.com/guidelines/component-statuses is coming from Contentful, if you have any feedback about the copy, please leave your comment here or maybe just adjust it in Contentful.


fabe (approved):

packages/website/utils/contentfulSidebarLinks.json
```diff
@@ -8,6 +8,10 @@
       "title": "Adding icons",
       "slug": "/guidelines/adding-icons"
     },
+    {
+      "title": "Forma 36 component statuses",
```
burakukula, line 12: Maybe it should be called differently? Something like "Components release statuses"? WDYT?

mshaaban0 (commented):
We don't have any components or release processes in `alpha` at the moment. I would recommend to ditch `alpha` completely as the discussed earlier and stick to `beta` only

massao:
> Correct me if I am wrong, but I think the Note content will get updated as soon as the update README files will be merged:
It should because it uses the source of the monorepo, worst case we would need to bump the lock file, not a big issue I think

burakukula:
> We don't have any components or release processes in `alpha` at the moment. I would recommend to ditch `alpha` completely as the discussed earlier and stick to `beta` only

From our discussions I understood that we want to keep it, sorry, I am totally confused now. 
From my perspective it is kinda make sense, if we ever gonna release in alpha we would have a badge and content in place for that. I know that @Lelith is thinking about potentially releasing Multiselect component in alpha, for example. 
I mean, it how exactly the process of the release of components look like right now and maybe it makes sense to have this transparency for users?

packages/website/components/SidebarLink.tsx
```diff
@@ -129,9 +136,21 @@ export function SidebarLink({
             {children}
             {isExternal && <ExternalLinkTrimmedIcon variant="muted" />}
           </span>
-          {isNew && (
-            <Badge className={styles.badge} variant="primary">
-              new
+          {(isNew || isDeprecated || isBeta || isAlpha) && (
```
massao, line 139: Will we still use `new` in some case?? or could we remove it?

burakukula: We would use it for components like "Pagination". - stable but new.

burakukula: I think it is kinda still valuable to indicate new components, that are just got out of beta.

packages/website/utils/content.ts
```diff
@@ -83,11 +83,19 @@ async function getMdxPaths() {
...
   const sidebarLinks = pages.reduce((acc, page) => {
-    const { title, slug, section, isNew = undefined } = page.frontMatter.data;
+    const {
+      title,
+      slug,
+      section,
+      isNew = undefined,
+      isBeta = undefined,
+      isAlpha = undefined,
+      isDeprecated = undefined,
```
bgutsol, line 93: why do we need `undefined` by default here?

bgutsol (approved):
lgtm

packages/website/utils/contentfulSidebarLinks.json
bgutsol: I would just call it `Component release phases` without Forma 36. Since you are already on the f36 website, you know that it's related to forma

packages/website/utils/content.ts
burakukula: Good question! 🙂 I didn’t investigate much here, just added new values on top of the ligic that was already in place. I can check it out.

bgutsol:
> We don't have any components or release processes in `alpha` at the moment. I would recommend to ditch `alpha` completely as the discussed earlier and stick to `beta` only

I'd say it makes sense to keep it. Even though we not gonna run every component through every stage, for some experimental components it makes sense to have an alpha phase.

massao (approved):

PR #2159, opened 2022-08-09
fix: bump datepicker in the website and fix Note

burakukula:


yarn.lock
```diff
@@ -1818,451 +1818,6 @@
   resolved "https://registry.npmjs.org/@contentful/browserslist-config/-/browserslist-config-2.0.0.tgz"
   integrity sha512-wx/zWaVfhxwiIwET/51+Zg5PzEeyMIfWbOIgyi1zR4IM+1gJhyDgMPzUsChJazakI0EUqn1mcr+UJpwVnZ71MQ==
 
-"@contentful/f36-accordion@^4.14.0":
```
bgutsol, line 1821: what happened to `yarn.lock`? 🤔 Why it has so many removed fields, could you pls run `yarn install` one more time, just to be sure that it's correct

bgutsol (approved):

yarn.lock
burakukula: yeah, just did. I am not sure really why this happens now. if you check out the main package.json there are no `@contentful/f36-` dependencies. but I guess lock should contain deps from different packages anyway. I am not really sure why this is happening now. 
I have run the `yarn install` in the root dir. Maybe those deps in the lock are left for a while?

burakukula: 🤔 but that's strange, since it's actually `v4.14.0` so the current one I think.

bgutsol: Not sure, but it's fine

burakukula: yeah? you think we are safe to merge?

bgutsol: yes

PR #2163, opened 2022-08-16
feat: xlarge size for Icon

burakukula:
In order to be able to provide xlarge size icon for ProductIcons,I would like to extend the Icon sizes, since we use it under the hood in experience-components package. WDYT?

### DIFF
--- a/packages/components/icon/examples/IconSizesExample.tsx
+++ b/packages/components/icon/examples/IconSizesExample.tsx
@@ -17,6 +17,9 @@ export default function IconSizesExample() {
       <Flex alignItems="center">
         <CalendarIcon size="large" /> <Text>Large</Text>
       </Flex>
+      <Flex alignItems="center">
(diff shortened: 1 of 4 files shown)


burakukula:
@fabe WDYT if we add this additional size?

fabe (approved):

packages/components/icon/examples/IconSizesExample.tsx
```diff
@@ -17,6 +17,9 @@ export default function IconSizesExample() {
       <Flex alignItems="center">
         <CalendarIcon size="large" /> <Text>Large</Text>
       </Flex>
+      <Flex alignItems="center">
+        <CalendarIcon size="xlarge" /> <Text>Large</Text>
```
denkristoffer, line 21:
```suggestion
        <CalendarIcon size="xlarge" /> <Text>Extra large</Text>
```

denkristoffer (approved):

PR #2164, opened 2022-08-24
feat(website): build navigation from CF

denkristoffer:
Moves much of the website navigation to Contentful, meaning we fetch it on build and pass it into our layout. This should make it possible to build the website for anyone with a Contentful token. Most importantly this should mean that external contributors can open PRs that should "just work".

Check it out at https://forma-36-git-fix-build-website-without-json-file.colorfuldemo.com/

Please have a look at the content model in the web app and give feedback on that as well!

# How it works
Fetches all top bar links and sidebar links for each page when building the page and passes these links as a prop to the `Layout` component, which further passes a `links` array to the `Sidebar` component when we want to.

# Possible improvements
In the future when Next.js implements their layouts RFC we can simplify some of this by moving the navigation fetching to the layout components themselves instead of having the same call on every page.


packages/website/components/SidebarSection.tsx
```diff
@@ -24,36 +30,25 @@ const styles = {
...
-  isBeta?: boolean;
-  isAlpha?: boolean;
-  isDeprecated?: boolean;
-};
-export type SidebarSectionType = {
-  title: string;
-  links: SidebarLinkType[];
-  type: 'section';
+const isLinkActive = (href, currentPage) => {
+  const [, , linkedPage] = href.split('/');
```
bgutsol, line 34: 💅
```suggestion
  const linkedPage = href.split('/')[2];
```

packages/website/lib/api.ts
```diff
@@ -95,26 +116,131 @@ export async function getAllArticles(preview = false) {
...
+        const category = {
+          links: item.links,
+          slug: `/${sectionSlug}`,
+          title: item.name,
+        };
 
-  // create a JSON file with sidebar links for pages that come from Contentful
-  const data = JSON.stringify(sidebarLinks, null, 2);
-  fs.writeFileSync('utils/contentfulSidebarLinks.json', data);
+        // Do we actually use this?
```
bgutsol, line 194: what's that? 🤔

bgutsol (commented):
Amazing work @denkristoffer  🏅

Questions regarding content types:
- The category type is not clear enough for me. For example: why do we have both category and section for the "Guidelines"?
- It's a bit confusing when we use category in the "Introduction" section to add a bunch of v3 links. Maybe we might consider a better naming instead of "Category" for such links. Just an open question 🤷‍♂️

denkristoffer:
> The category type is not clear enough for me. For example: why do we have both category and section for the "Guidelines"?

Starting from the end, good question! I have not cleaned up old content, I don't think we're actually using the "Guidelines" category.

The navigation structure is planned like this:

Screenshot 2022-08-26 at 13 05 23

packages/website/lib/api.ts
denkristoffer: We're fetching the `previewDescription` field on `main` branch as well but I'm not sure we actually use it.

bgutsol: let's try to remove it then?

bgutsol (approved):

PR #2168, opened 2022-08-26
chore(deps): update typescript

denkristoffer:
Next.js expects TypeScript version 4.3.2 or above. This upgrade also gives us some cool features like `type` imports in named imports, so that we don't have to split component and type imports.


bgutsol (approved):
👏

PR #2169, opened 2022-08-26
feat(skeleton): compound component

denkristoffer:
Make a `Skeleton` compound component. Docs could be improved a bit to show off the pattern more.


burakukula (approved):
Nice!

Lelith (approved):
Looks good to me!

denkristoffer:
@Lelith @burakukula 

I tried combining the Skeleton docs into one page: https://forma-36-git-feat-compound-skeleton.colorfuldemo.com/components/skeleton

What do you think, better or worse?

Lelith (approved):
i like the one page document better, its easier to read and understand.

massao (approved):
LGTM, would it be worth to keep a note or something saying that the previous way would still work?? so developers that are already using it, wouldn't keep from updating to this version?? 🤔

PR #2173, opened 2022-08-29
docs: add beta installation instructions

denkristoffer:
# Purpose of PR

Add installation instructions to docs for beta packages:

- https://forma-36-git-docs-beta-package-install.colorfuldemo.com/components/calendar
- https://forma-36-git-docs-beta-package-install.colorfuldemo.com/components/datepicker

### DIFF
--- a/packages/components/datepicker/README.mdx
+++ b/packages/components/datepicker/README.mdx
@@ -15,6 +15,18 @@ The Datepicker component combines text input, and calendar in dropdown.
 
 Use Datepicker to ask user's for date input in forms, modals, or filtering.
 
+## Install beta package
+
+```bash static=true
+npm install @contentful/f36-datepicker@beta
+```
+
+For yarn:
+
+```bash static=true
+yarn add @contentful/f36-datepicker@beta
+```
+
 ## Import
 
 ```jsx static=true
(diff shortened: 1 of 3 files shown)


packages/components/datepicker/README.mdx
```diff
@@ -15,6 +15,18 @@ The Datepicker component combines text input, and calendar in dropdown.
 
 Use Datepicker to ask user's for date input in forms, modals, or filtering.
 
+## Install beta package
+
+```bash static=true
+npm install @contentful/f36-datepicker@beta
+```
```
mshaaban0, line 22: I guess that won't solve it for other beta/alpha packages. Maybe it makes sense to have this section in [PageContent](https://github.com/contentful/forma-36/tree/main/packages/website/components/PageContent) and make use of `isBeta`, `isAlpha` flags from FrontMatter?

denkristoffer: I don't see it as something with much use (yet). Maybe we can look into it when @massao has finished work on how we want to do pre-releases?

mshaaban0: If this is a quick fix, then 👍🏼

Issue #2166, opened 2022-08-25
💡 Proposal - Add 'overlayClassName' prop to Modal component


maxcheremisin:
# Forma 36 contribution proposal

## The problem

Customise Modal overlay.

![image](https://user-images.githubusercontent.com/22265863/186731804-851096e7-aa53-4c1f-97f7-1e6b5d4c6d00.png)

## The proposed solution

Add an optional `overlayClassName` prop for Modal component to have a possibility to change the style of it.

## Breaking changes

No Breaking changes

denkristoffer:
Done.

PR #2175, opened 2022-08-29
feat(modal): add 'overlayProps' to Modal component

maxcheremisin:
# Purpose of PR

Add a way to style the Modal backdrop

This PR looks to resolve the issue brought up in proposal #2166

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/components/modal/src/Modal.tsx
+++ b/packages/components/modal/src/Modal.tsx
@@ -73,6 +73,11 @@ export interface ModalProps extends CommonProps {
   allowHeightOverflow?: boolean;
 
   /**
+   * Optional props to override overlay behaviour
+   */
+  overlayProps?: Pick<CommonProps, 'className' | 'style'>;
+
+  /**
    * Optional props to override ModalHeader behaviour
    */
   modalHeaderProps?: Partial<ModalHeaderProps>;
@@ -135,6 +140,7 @@ export const Modal = ({
     size,
     allowHeightOverflow,
     className: otherProps.className,
+    overlayClassName: otherProps.overlayProps?.className,
   });
 
   React.useEffect(() => {
@@ -184,6 +190,7 @@ export const Modal = ({
         content: {
           top: position === 'center' ? 0 : topOffset,
         },
+        overlay: otherProps.overlayProps?.style,
       }}
       className={{
         base: styles.base.root,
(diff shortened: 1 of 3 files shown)


denkristoffer:
Thanks for the PR! 🙌  Let's wait for design feedback on the proposal before moving on this.

Lelith (approved):
Looks good to me!

packages/components/modal/src/Modal.tsx
```diff
@@ -73,6 +73,11 @@ export interface ModalProps extends CommonProps {
    */
   allowHeightOverflow?: boolean;
 
+  /**
+   * CSS class to be appended to the overlay element
+   */
+  overlayClassName?: string;
```
denkristoffer, line 79: To align with our prop naming conventions this should go in an `overlayProps` object prop instead, so that you can use the modal like this:

```tsx
<Modal overlayProps={{ className: 'something' }} />
```

denkristoffer (changes requested):

packages/components/modal/src/Modal.tsx
maxcheremisin: should it be the `CommonProps` interface?

maxcheremisin: hmm, we can't pass `testId` to overlay element as a prop, in this case we would need to render it ourselves. so should i just declare it like `overlayProps: { className?: string}` or render a custom overlay?

denkristoffer: Hmm I think let's leave out the `testId` for now and go with just the `className` 🤔

denkristoffer:
Changes look good! Can you update from `main`?

maxcheremisin:
> Changes look good! Can you update from `main`?

Done!

denkristoffer (approved):

denkristoffer:
@massao any idea why changeset bot is seeing changes to 33 packages here?

massao:
> @massao any idea why changeset bot is seeing changes to 33 packages here?

Because we use fixed versions on the components packages, so whenever there a new version, all packages get a version bump, but we could have a discussion about this, as maybe it doesn't make too much sense to try keeping all of them in sync 🤔

denkristoffer:
All right, thanks! Let's merge.

denkristoffer:
@all-contributors please add @maxcheremisin for code

PR #2178, opened 2022-08-29
fix: Datepicker input value

bgutsol:
# Purpose of PR

- Update datepicker input value when selected is undefined

### DIFF
--- a/packages/components/datepicker/package.json
+++ b/packages/components/datepicker/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@contentful/f36-datepicker",
-  "version": "4.1.0-beta.6",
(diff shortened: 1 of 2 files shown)


denkristoffer:
Can you explain why this is needed? I see we already check the value during the initial setting of the state, so I guess this catches cases where the `selected` prop is used and then changed to `undefined`?

bgutsol:
> Can you explain why this is needed? I see we already check the value during the initial setting of the state, so I guess this catches cases where the `selected` prop is used and then changed to `undefined`?

@denkristoffer yes, exactly. We need it for example in the date-field-editor. There we have "Clear" button, which clears time input, date input and timezone input, it passes `undefined` value.

PR #2179, opened 2022-08-30
feat: require node 14.17

denkristoffer:
# Purpose of PR

Upgrades Docker images used in CI and bumps the required version of Node to 14.17 (which is necessary to land #2148 later on).


denkristoffer:
@massao can I add a changeset that bumps every package?

massao:
> @massao can I add a changeset that bumps every package without actually adding all of them manually?

You can run `yarn changeset` and select all packages, or at least the ones that you need to, and you add the same message as on the commit

denkristoffer:
> You can run yarn changeset and select all packages, or at least the ones that you need to, and you add the same message as on the commit

Okay, I've added every package as a minor. I guess it might not matter for consumers though, so maybe we should just not release anything? 🤔

burakukula (approved):

bgutsol (approved):

PR #2183, opened 2022-08-30
build: fix size-limit check and add GitHub action

denkristoffer:
# Purpose of PR

The size-limit check is broken in `main`, this branch fixes that and adds a GitHub action that should comment with size differences on pull requests 🙂 

FYI this PR will not go green until after it's been merged due to the way the action works.


package.json
```diff
@@ -149,17 +151,18 @@
...
     "semver": "^7.3.5",
+    "size-limit": "^8.0.1",
     "style-loader": "^2.0.0",
     "turbo": "^1.0.6",
     "typescript": "4.7.3"
   },
   "engines": {
     "node": ">=14.17.0",
-    "yarn": ">=1.19.0"
+    "yarn": ">=1.18.0"
```
mshaaban0, line 161: downgrade, more compatibility 👍🏼

packages/forma-36-react-components/package.json
```diff
@@ -13,8 +13,7 @@
   "browserslist": "extends @contentful/browserslist-config",
   "scripts": {
     "build:clean": "rm -rf dist",
-    "build": "yarn build:clean && microbundle src/index.ts && yarn size",
-    "size": "size-limit",
+    "build": "yarn build:clean && microbundle src/index.ts",
```
mshaaban0, line 16: Q: why are we using `microbundle` (Rollup) only for this package and not parcel like others?

.github/workflows/size-limit.yml
```diff
@@ -0,0 +1,16 @@
+name: 'size'
+
+on: pull_request
+  branches:
+    - main
```
mshaaban0, line 5: Heads up there is an error on this part, the action is not reporting back to the PR
https://github.com/contentful/forma-36/actions/runs/2962500861/workflow

packages/forma-36-react-components/package.json
denkristoffer: I've no idea, I tried to track it down and it comes from the original v4 pull request. I wanted to look into it in a separate PR 🤷

package.json
denkristoffer: I'm not sure why we required 1.19 in the first place, any idea?

bgutsol (approved):

.github/workflows/size-limit.yml
mshaaban0: @denkristoffer it's still failing to build 🤔

denkristoffer: @mshaaban0 It's not going to work for this PR, but you can see it working here: https://github.com/contentful/forma-36/runs/8136279821?check_suite_focus=true

The reason it fails here is that the action works by running `build` and then `npx size-limit --json` first on the branch of the PR, then on `main`, and it compares the two size results. Before this PR was merged the configuration for `size-limit` was broken on the `main` branch, so it couldn't compare the results and then exited with an error. It should work for future PRs, although it looks like the config needs to be tweaked.

PR #2186, opened 2022-09-01
fix: avoid shrinking the switch input to not cut if off [HOMER-1216]

Chaoste:
# Purpose of PR

As soon as the width of the component is limited and the description includes a line wrap, the switch input will be cut off.

Screen Shot 2022-09-01 at 11 59 58

Using the flex-shrink property, we tell the flexbox never to shrink this part but only the text block.
Screen Shot 2022-09-01 at 12 00 42

## PR Checklist

- [x] Tests are passing


denkristoffer (approved):
Awesome!

PR #2187, opened 2022-09-01
build: build with tsup and esbuild

denkristoffer:
This PR moves Forma 36 from Parcel to esbuild (through tsup)

Note that the **size limit GitHub action check is NOT going to go green on this PR** since the build output is different than our main branch.

## Why?

Mainly because we've been having issues with Parcel. Circle CI has often been throwing segmentation faults during the build process lately, requiring us to restart failed builds all the time. [esbuild is super fast](https://esbuild.github.io/) but it doesn't output types and [it's not going to](https://github.com/evanw/esbuild/issues/95).

Instead we use [tsup](https://github.com/egoist/tsup) which is a wrapper around esbuild that DOES let us build types. It's still really fast, although the type compilation slows us down a bunch. Without it this PR can build all our components in about 17 seconds on my machine ⚡ 

## What we might need to discuss

tsup builds everything to `dist` directories. I've currently set it to build a CommonJS file to `dist/index.js` and an ESM file to `dist/index.mjs`. esbuild does not let us build UMD. In my opinion that's fine, except we currently provide it, so to avoid breaking changes this PR still uses Microbundle to build the `@contentful/f36-components` package, which has the following output:

- `index.js` – CommonJS
- `index.modern.mjs`– Uses Microbundle's ["Modern mode"](https://github.com/developit/microbundle#-modern-mode-) which means it transpiles ESM with Babels – actually really cool 👍 
- `index.mjs` – ESM
- `index.umd.js` – UMD

It would be good to drop UMD support when we can.


packages/forma-36-react-components/src/index.ts
```diff
@@ -1,31 +1,31 @@
-export * from '@contentful/f36-collapse';
+export * from '@contentful/f36-accordion';
```
burakukula, line 1: A - Z 😁

tsup.config.ts
```diff
@@ -0,0 +1,14 @@
...
+export default defineConfig({
+  clean: true,
+  dts: true,
+  entryPoints: ['src/index.ts'],
+  external: ['react', 'react-dom', '@contentful/f36-*'],
+  format: ['cjs', 'esm'],
+  minify: true,
+  platform: 'browser',
+  sourcemap: true,
+  target: 'es2016',
```
denkristoffer, line 12: Maybe we can bump this, what do you think?

burakukula (approved):
LGTM

massao (approved):

bgutsol (approved):
NICE 🏅

tsup.config.ts
bgutsol: I'd say we either use `es6/es2015` - a kinda recommended standard for supporting most browsers, or we use any later version, but then there is no guarantee for support in all modern browsers.

denkristoffer: Our official stance is we support the latest 2 major versions of a browser. Other than Safari, this gives us a very modern base.

https://www.contentful.com/faq/about-contentful/#which-browsers-does-contentful-support

denkristoffer: And consumers can always transpile the library themselves if they want to support something else.

denkristoffer: According to [caniuse.com](https://caniuse.com/?feats=mdn-javascript_operators_optional_chaining%2Cmdn-javascript_operators_nullish_coalescing%2Cmdn-javascript_builtins_globalthis%2Ces6-module-dynamic-import%2Cbigint%2Cmdn-javascript_builtins_promise_allsettled%2Cmdn-javascript_builtins_string_matchall%2Cmdn-javascript_statements_export_namespace%2Cmdn-javascript_statements_import_meta) we can safely bump to es2020 – what do you think @bgutsol @massao ?

bgutsol: ok 👍

PR #2188, opened 2022-09-01
Feat: new component multiselect [WIP]

Lelith:
draft for work in progress 

## Open tasks

- [ ] select all shortcut
- [ ] resort selected elements to the top
- [ ] max selection
- [ ] information message
- [ ] write documentation
- [ ] write examples
- [ ] add tests


packages/components/multiselect/README.mdx
```diff
@@ -0,0 +1,118 @@
...
+import { Multiselect } from '@contentful/f36-components';
+// or
+import { Multiselect } from '@contentful/f36-multiselect';
+```
+
+## Examples
+
+### Basic usage
+
+The Multiselect requires 3 props to work:
```
Chaoste, line 23:
```suggestion
The Multiselect requires three props to work:
```

packages/components/multiselect/README.mdx
```diff
@@ -0,0 +1,118 @@
...
+
+- `items`: It’s an array containing the items that will be shown as selectable options when the user types something in the `TextInput`.
+- `onInputValueChange`: This function will be called every time the user types something in the input.
```
Chaoste, line 26: Since this is a public interface, I would abstract some internal knowledge away to make it more smooth/ easier to read.
For the naming, my proposal would be `onSearchValueChange` since the users types something to search for stuff - while input is a very general name. The MultiSelect itself is an input.

packages/components/multiselect/README.mdx
```diff
@@ -0,0 +1,118 @@
...
+  The component will pass the `item`, which the filter method is currently iterating over, and the `inputValue` prop of the `TextInput` component.
+- `onSelectItem`: This function is called when the user selects one of the options of the list. The component will pass the selected item as an argument to the function.
+
+An `Multiselect` with a list of spaces will look like this:
```
Chaoste, line 30:
```suggestion
A `Multiselect` with a list of spaces will look like this:
```

packages/components/multiselect/examples/MultiSelectBasicUsageExample.tsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import { Autocomplete, Stack, Paragraph } from '@contentful/f36-components';
+
+export default function AutocompleteBasicUsageExample() {
```
Chaoste, line 4:
```suggestion
export default function MultiSelectBasicUsageExample() {
```

packages/components/multiselect/README.mdx
```diff
@@ -0,0 +1,118 @@
...
+
+```jsx file=./examples/MultiselectBasicUsageExample.tsx
+
+```
+
+### Using objects as `items`
+
+We showed how to create an Multiselect with an array of string but it’s also possible to use other types of data as `items`.
+A very common way of using the Multiselect is with objects and for that, with a few changes to the previous example this can be done:
```
Chaoste, line 39:
```suggestion
We showed how to create a Multiselect with an array of strings but it’s also possible to use other data types as `items`.
A widespread way of using Multiselect is with objects and for that, with a few changes to the previous example this can be done:
```

_Suggestions by Grammarly_

packages/components/multiselect/src/Multiselect.tsx
```diff
@@ -0,0 +1,276 @@
...
+  /**
+   * Boolean to control whether or not to render the suggestions box in a React Portal.
+   * Rendering content inside a Portal allows the suggestions box to escape the bounds
+   * of its parent while still being positioned correctly.
+   * Defaults to `false`
+   */
+  usePortal?: boolean;
+}
+
+function _Multiselect(props: MultiselectProps, ref: React.Ref<HTMLDivElement>) {
```
Chaoste, line 108:
```suggestion
function _MultiSelect(props: MultiSelectProps, ref: React.Ref<HTMLDivElement>) {
```

This should be `MultiSelect` in this file as well for the file naming and the documentation.

packages/components/multiselect/README.mdx
```diff
@@ -0,0 +1,118 @@
...
+```jsx file=./examples/MultiselectWithObjectsExample.tsx
+
+```
+
+Both `itemToString` and `renderItem` are necessary when passing objects as items and they both will receive an "item" as an argument.
+
+If you are using Typescript, you can tell the Multiselect what is the type of your items to make these functions strongly typed.
+You can do that by writing the component like this `<Multiselect<ItemType> {...props}/>`
+
+### Highlighting an item with `getStringMatch`
```
Chaoste, line 50: Looks like the rest of the documentation isn't updated at all.

packages/components/multiselect/src/Multiselect.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+          value={`${fruit.id}`}
+          label={fruit.name}
+          onSelectItem={mockOnSelectItem}
+        />;
+      })}
+    </Multiselect>,
+  );
+}
+
+describe('Multiselect', () => {
```
Chaoste, line 50: Tests for proper working properties would be nice

packages/components/multiselect/src/Multiselect.tsx
```diff
@@ -0,0 +1,276 @@
...
+    listRef,
+    listWidth = 'auto',
+    listMaxHeight = 180,
+    isLoading = false,
+    usePortal = false,
+    testId = 'cf-multiselect',
+    children,
+  } = props;
+
+  const styles = getMultiselectStyles();
```
Chaoste, line 132: Is it correct to generate the styles on every render cycle even though nothing changes?

packages/components/multiselect/src/Multiselect.tsx
```diff
@@ -0,0 +1,276 @@
...
+  const searchInputRef = useRef(null);
+
+  const handleSearchChange = (event) => {
+    setSearchValue(event.target.value);
+    onSearchValueChange(event);
+  };
+
+  const resetSearch = () => {
+    // this looks a bit hacky, but is the official way of externally triggering the onChange handler for an input
+    // https://stackoverflow.com/questions/23892547/what-is-the-best-way-to-trigger-onchange-event-in-react-js/46012210#46012210
```
Chaoste, line 146:
```suggestion
    // https://stackoverflow.com/a/46012210/17269164
```

packages/components/multiselect/src/Multiselect.tsx
Lelith: Good catch! I think its not necessary in this way anymore, before there was a variable handover from props, will adjust the code here :)

packages/components/multiselect/src/Multiselect.test.tsx
Lelith: yes, tests are on my todo list!

packages/components/multiselect/README.mdx
Lelith: yes, still WIP :)

packages/components/multiselect/src/Multiselect.tsx
Lelith: so i just checked it, and it seems that in forma we have a linter rule that dissallowes the usage of emotion when its not generated on every render... i m going to ask in the chapter about this.

packages/components/multiselect/src/MultiselectOption.tsx
```diff
@@ -0,0 +1,64 @@
...
+  label,
+  value,
+  itemId,
+  onSelectItem,
+  searchValue,
+  isChecked = false,
+  isDisabled = false,
+  ...rest
+}: MulitselectOptionProps) => {
+  const styles = getMultiselectStyles();
```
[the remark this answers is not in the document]
Lelith: synced with the chapter, and we want to keep this logic. as the component grows there will also likely be more variables needed.

packages/components/multiselect/src/Multiselect.tsx
```diff
@@ -0,0 +1,278 @@
...
+                />
+              </>
+            )}
+            {isLoading && <ListItemLoadingState />}
+
+            {!isLoading && childrenLength > 0 && (
+              <ul className={styles.list} data-test-id="cf-multiselect-items">
+                {React.Children.map(children, (child) => {
+                  if (React.isValidElement(child)) {
+                    return React.cloneElement(child, {
```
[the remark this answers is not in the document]
Lelith: For the search input value if there is one, because it is used to highlight the search phrase in the found elements.

For optimisation i could only clone when there is a search and if not, just return the children without the enrichment. 
I have to change it also a bit, once i allow also grouping, because then there would not only be Options as child but also something for the group title.

There probaply should also be a more specific type for children then just a ReactNode then i guess. To make clear that it expects either an Option or a GroupTitle.

packages/components/multiselect/src/Multiselect.tsx
Lelith: Chapter says we keep the rule, and also the complexity of the styles will very likely increase with time.

packages/components/forms/src/Multiselect/Multiselect.tsx
```diff
@@ -0,0 +1,286 @@
...
+                    ) : (
+                      <SearchIcon variant="muted" />
+                    )
+                  }
+                  onClick={() => {
+                    if (searchValue) {
+                      resetSearch();
+                    }
+                  }}
+                  isDisabled={searchValue === ''}
```
mshaaban0, line 242:
```suggestion
                  isDisabled={!searchValue}
```

packages/components/forms/src/Multiselect/MultiselectOption.tsx
```diff
@@ -0,0 +1,64 @@
...
+  searchValue,
+  isChecked = false,
+  isDisabled = false,
+  ...rest
+}: MulitselectOptionProps) => {
+  const styles = getMultiselectStyles();
+
+  return (
+    <li {...rest}>
+      <label htmlFor={itemId} className={styles.item(isDisabled)}>
```
mshaaban0, line 31: You can probably add styles.disabled instead of rewriting the disabled styles in styles.item

mshaaban0 (approved):
Amazing work, really neat.

packages/components/forms/src/Multiselect/Multiselect.styles.ts
```diff
@@ -0,0 +1,74 @@
...
+      flexDirection: 'row',
+      alignItems: 'center',
+      opacity: isDisabled ? 0.5 : 1,
+
+      '&:focus, &:hover': {
+        backgroundColor: tokens.gray100,
+      },
+      '&:active': {
+        backgroundColor: tokens.gray200,
+      },
```
denkristoffer, line 65: Can we use the keyboard-only focus styling here? See the Menu component for example.

packages/components/forms/package.json
```diff
@@ -7,10 +7,15 @@
...
     "@babel/runtime": "^7.6.2",
+    "@contentful/f36-button": "^4.17.0",
     "@contentful/f36-core": "^4.17.0",
     "@contentful/f36-icons": "^4.17.0",
+    "@contentful/f36-skeleton": "^4.17.0",
     "@contentful/f36-tokens": "^4.0.1",
     "@contentful/f36-typography": "^4.17.0",
+    "@contentful/f36-popover": "^4.17.0",
+    "@contentful/f36-utils": "^4.17.0",
+
```
denkristoffer, line 18:
```suggestion
```

packages/components/forms/src/Multiselect/Multiselect.tsx
```diff
@@ -0,0 +1,286 @@
...
+   */
+  searchInputRef?: React.Ref<HTMLInputElement>;
+  /**
+   * Use this prop to get a ref to the toggle button of the component
+   */
+  toggleRef?: React.Ref<HTMLButtonElement>;
+  /**
+   * Use this prop to get a ref to the list of items of the component
+   */
+  listRef?: React.Ref<HTMLUListElement>;
```
denkristoffer, line 78: I'd like to challenge this, it's a reversal to the pattern we moved away from with the breaking changes in v3 to v4. I believe our code conventions show a clear usage of the `subComponentProps={{ ref }}` pattern.

packages/components/forms/src/Multiselect/Multiselect.tsx
```diff
@@ -0,0 +1,286 @@
...
+    // https://stackoverflow.com/a/46012210/17269164
+
+    const nativeInputValueSetter = Object.getOwnPropertyDescriptor(
+      window.HTMLInputElement.prototype,
+      'value',
+    ).set;
+    nativeInputValueSetter.call(internalSearchInputRef.current, '');
+    const forcedEvent = new Event('change', { bubbles: true });
+    internalSearchInputRef.current.dispatchEvent(forcedEvent);
+  };
```
denkristoffer, line 152: I'm not sure I understand why we can't do `setSearchValue('')` instead of this, would you mind explaining?

packages/components/forms/src/Multiselect/Multiselect.tsx
```diff
@@ -0,0 +1,286 @@
...
+  /**
+   * Use this prop to get a ref to the list of items of the component
+   */
+  listRef?: React.Ref<HTMLUListElement>;
+
+  /**
+   * It sets the width of the list
+   * @default "auto"
+   */
+  listWidth?: 'auto' | 'full';
```
denkristoffer, line 84: Elsewhere we use a `isFullWidth` boolean, seems like the same could apply here, or do we expect to extend this union at some point?

Based on the previous comment it could be combined into something like

```
listProps: {
  isFullWidth?: boolean;
  ref?: React.Ref<HTMLUListElement>;
}
```

packages/components/forms/src/Multiselect/Multiselect.tsx
```diff
@@ -0,0 +1,286 @@
...
+                  }}
+                  ref={mergeRefs(searchInputRef, internalSearchInputRef)}
+                />
+                <IconButton
+                  aria-label="Clear"
+                  className={styles.toggleButton}
+                  variant="transparent"
+                  icon={
+                    searchValue ? (
+                      <CloseIcon variant="muted" />
```
denkristoffer, line 232: Should these icons be configurable?

packages/components/forms/src/Multiselect/Multiselect.tsx
```diff
@@ -0,0 +1,286 @@
...
+          <>
+            {hasSearch && (
+              <>
+                <TextInput
+                  aria-label="Search"
+                  type="text"
+                  value={searchValue}
+                  className={styles.inputField}
+                  id={id}
+                  data-test-id="cf-multiselect-search"
```
denkristoffer, line 219:
```suggestion
                  testId="cf-multiselect-search"
```

denkristoffer (commented):
Thanks for working on this! I've added some thoughts.

packages/components/forms/src/Multiselect/Multiselect.tsx
Lelith: I copied these from the autocomplete property list, but i m happy to use the other reference convention.

packages/components/forms/src/Multiselect/Multiselect.tsx
Lelith: If we only clear searchValue, it will not call the callback function onSearchChange with the input change event, so the filter of the list will not be reset.
One big issue we had when using the autocomplete, was that, when the input changed, you never got the full event, but only the value of the input. Which lead down to a list of issues for onChange / onBlur validations.

This way the reset of the search value will trigger the expected onChange event. In the linked stackoverflow thread there is an explanation from Dan Abramov about why programmatically changing the value of an input is not firing the native event handler.

packages/components/forms/src/Multiselect/Multiselect.tsx
Lelith: As far as i talked with @FerdinandoDeMeo only the startIcon should be configurable, so the icons inside the search input shall be magnify glass and the close one.
But if needed, we can update those to be configurable too in the future.

packages/components/forms/src/Multiselect/MultiselectOption.tsx
Lelith: I don t understand the difference here, can you explain to me what is the benefit of combining two classes vs changing the style via the variable?  because i have seen both patterns across components.

mshaaban0: Combining classes will allow you to remove `opacity` and `cursor` from `item` styles because they already exist in `disabled` and avoid duplications

packages/components/forms/src/Multiselect/Multiselect.styles.ts
Lelith: do you mean the ':focus-visible' style? not sure what you mean

packages/components/forms/src/Multiselect/Multiselect.tsx
Lelith: i m not sure how the proper syntax would be, is there a component where we use this approach with subcomponent properties?

packages/components/forms/src/Multiselect/Multiselect.tsx
denkristoffer: Right, I see. I wish I caught them for the autocomplete then 😅

packages/components/forms/src/Multiselect/Multiselect.tsx
```diff
@@ -0,0 +1,285 @@
...
+    <div
+      data-test-id={testId}
+      className={cx(styles.multiselect, className)}
+      ref={ref}
+    >
+      <Popover
+        usePortal={usePortal}
+        isOpen={isOpen}
+        onClose={() => setIsOpen(false)}
+        isFullWidth={isFullWidth === 'full'}
```
denkristoffer, line 190:
```suggestion
        isFullWidth={isFullWidth}
```

packages/components/forms/src/Multiselect/Multiselect.tsx
denkristoffer: Datepicker for instance takes `inputProps`: https://f36.contentful.com/components/datepicker#visual-states

packages/components/forms/src/Multiselect/Multiselect.styles.ts
denkristoffer: Sorry, to be clear it's that selector yes, but we have to apply it in an annoying order to avoid breaking focus state for for browsers who don't implement it yet. Looks like this: https://github.com/contentful/forma-36/blob/a64a2ce00e1ae300228ff87d68d1fdebc61787e2/packages/components/button/src/Button/Button.styles.ts#L59-L67

packages/components/forms/src/Multiselect/Multiselect.tsx
Lelith: i'll put that on my list of breaking change updates for the autocomplete. 📝

packages/components/forms/src/Multiselect/Multiselect.tsx
Lelith: ahh thanks, that one did slip trough!

packages/components/forms/src/Multiselect/MultiselectOption.tsx
Chaoste: Is it also a possibility to do something like `input[disabled] + label` instead of doing it via JS?

packages/components/multiselect/stories/MultiSelect.stories.tsx
```diff
@@ -0,0 +1,137 @@
+import React, { useState } from 'react';
```
Chaoste, line 1: This file also needs a renaming to `Multiselect`.

btw, I really thought it should be `MultiSelect` because I consider them two separate words ("multi select" or "multi-select") but there are [examples](https://mdbootstrap.com/docs/standard/extended/multiselect/) on the internet that proofed my wrong 😐

packages/components/multiselect/README.mdx
```diff
@@ -0,0 +1,82 @@
...
+- isChecked (optional): controlls if a element is initially selected
+- isDisabled (optional): controlls if a element is changeable
+
+```jsx file=./examples/MultiselectBasicUsageExample.tsx
+
+```
+
+### Searchable options
+
+One of optional feature of the Multiselect component is to allow filtering the list via a search field.
```
Chaoste, line 57:
```suggestion
One optional feature of the Multiselect component is to allow filtering the list via a search field.
```

packages/components/multiselect/src/Multiselect.tsx
```diff
@@ -0,0 +1,278 @@
...
+  const styles = getMultiselectStyles();
+
+  const [searchValue, setSearchValue] = useState(defaultValue);
+  const [isOpen, setIsOpen] = useState(false);
+
+  const internalSearchInputRef = useRef(null);
+
+  const handleSearchChange = (event) => {
+    setSearchValue(event.target.value);
+    onSearchValueChange(event);
```
Chaoste, line 131:
```suggestion
    onSearchValueChange?.(event);
```

You're saying that this needs to be provided but still, we might wanna make sure that we don't throw a plain error in this case.

packages/components/multiselect/src/Multiselect.tsx
```diff
@@ -0,0 +1,278 @@
...
+  /**
+   * current Selected items, to be shown on the trigger button
+   */
+  currentSelection?: Array<string>;
+
+  /**
+   * If this is set to `true` there will be a search input inside the drawer
+   * @default false
+   */
+  hasSearch?: boolean;
```
Chaoste, line 47: Isn't this implicitly defined by passing `onSearchValueChange`? Or is there a case where you define only one of them?

Lelith: that is actually a very fair point, saves us one prop and makes the implementation more straigth forward!! 🚀

packages/components/multiselect/stories/MultiSelect.stories.tsx
Lelith: yes, same for me, i started out with MultiSelect just to discover that it should be Multiselect, and this file name slipped my renaming.

packages/components/multiselect/README.mdx
```diff
@@ -0,0 +1,81 @@
...
+import { Multiselect } from '@contentful/f36-components';
+// or
+import { Select } from '@contentful/f36-forms';
+```
+
+## Examples
+
+### Basic usage
+
+The Multiselect is a fully controlled from element and expects the selectable items as children via the compund component Multiselect.Option.
```
Chaoste, line 27:
```suggestion
The Multiselect is a fully controlled form element and expects the selectable items as children via the compound component Multiselect.Option.
```

packages/components/multiselect/README.mdx
```diff
@@ -0,0 +1,81 @@
...
+
+Properties are:
+
+for the multiselect itself:
```
Chaoste, line 31:
```suggestion
For the multiselect itself:
```

packages/components/multiselect/README.mdx
```diff
@@ -0,0 +1,81 @@
...
+<PropsTable of="Multiselect" />
+
+## Content guidelines
+
+- Multiselect label should be short, contain 1 to 3 words
+- Label should be written in a sentence case (the first word capitalized, the rest lowercase)
+
+## Accessibility
+
+- dismisses the dropdown when selecting with the enter key
```
Chaoste, line 81:
```suggestion
- Dismisses the dropdown when selecting with the enter key
```

packages/components/multiselect/examples/MultiselectBasicUsageExample.tsx
```diff
@@ -0,0 +1,51 @@
...
+    'Travel Blog',
+    'Finnance Blog',
+    'Fitness App',
+    'News Website',
+    'eCommerce Catalogue',
+    'Photo Gallery',
+  ];
+
+  // This `useState` is going to store the selected "space" so we can show it in the UI
+  const [selectedSpace, setSelectedSpace] = React.useState<Array<string>>([]);
```
Chaoste, line 16:
```suggestion
  // This `useState` is going to store the selected "spaces" so we can show it in the UI
  const [selectedSpaces, setSelectedSpaces] = React.useState<Array<string>>([]);
```

packages/components/multiselect/stories/Multiselect.stories.tsx
```diff
@@ -0,0 +1,137 @@
...
+      alignItems="start"
+    >
+      <Multiselect {...args} currentSelection={selectedFruits}>
+        {fruitStrings.map((item, index) => {
+          return (
+            <Multiselect.Option
+              value={item}
+              label={item}
+              onSelectItem={handleSelectItem}
+              key={`${item}-${index}`}
```
Chaoste, line 63: It's just an example code but still I would like to leave the feedback that key shouldn't use the index if possible:

See the [official docs](https://reactjs.org/docs/lists-and-keys.html).

> We don’t recommend using indexes for keys if the order of items may change. This can negatively impact performance and may cause issues with component state. Check out Robin Pokorny’s article for an [in-depth explanation on the negative impacts of using an index as a key](https://robinpokorny.com/blog/index-as-a-key-is-an-anti-pattern/). If you choose not to assign an explicit key to list items then React will default to using indexes as keys.

packages/components/multiselect/src/Multiselect.tsx
```diff
@@ -0,0 +1,273 @@
...
+                  aria-label="Search"
+                  type="text"
+                  value={searchValue}
+                  className={styles.inputField}
+                  id={id}
+                  testId="cf-multiselect-search"
+                  placeholder={searchPlaceholder}
+                  onChange={(event) => {
+                    handleSearchChange(event);
+                  }}
```
Chaoste, line 210: nitpick: This means that `TextInput` will be re-rendered on every re-rendering of `Multiselect`. We could optimise this
```suggestion
                  onChange={handleSearchChange}
```
(And using useCallback for handleSearchChange)

Chaoste (approved):
Except for typos and nitpicks, this looks good :)

denkristoffer (commented):
Shouldn't the source either go in the `forms` directory or not in that package?

packages/components/multiselect/README.mdx
```diff
@@ -0,0 +1,87 @@
...
+
+Multiselect allows users to make a multi selection from a list of options.
+It has multiple optional features like a search.
+
+## Import
+
+```jsx static=true
+import { Multiselect } from '@contentful/f36-components';
+// or
+import { Select } from '@contentful/f36-forms';
```
denkristoffer, line 20:
```suggestion
import { Multiselect } from '@contentful/f36-forms';
```

Lelith: ahh another one slipped trough. will leave this out for now completely until it is a stable part of the forms package.

Lelith:
@denkristoffer as i synced with  @massao i will keep the multiselect outside of the forms package until it is stable.

Chaoste (approved):

Chaoste:
FYI I removed the wrapping label component for the MultiselectOption. That's because there is already one rendered by Checkbox itself and thus, we currently had two labels.

Chaoste (approved):
LGTM!

packages/components/multiselect/src/Multiselect.test.tsx
```diff
@@ -51,20 +59,26 @@ function renderComponent(customProps: Partial<MultiselectProps>) {
...
-    userEvent.click(screen.getByRole('button', { name: 'Toggle Multiselect' }));
-    expect(screen.getByRole('list')).toBeInTheDocument();
-    userEvent.click(screen.getByRole('button', { name: 'Toggle Multiselect' }));
+  it('opens and closes the drawer', async () => {
+    const [{ user }] = renderComponent();
+
+    const toggleButton = screen.getByRole('button', {
+      name: 'Toggle Multiselect',
+    });
+    await waitFor(() => user.click(toggleButton));
```
Chaoste, line 71:
```suggestion
    await user.click(toggleButton);
```

Chaoste (changes requested):

packages/components/multiselect/src/Multiselect.test.tsx
Lelith: the test setup here seems different, if i do not use waitFor, the tests will not finish and get stuck.

PR #2190, opened 2022-09-02
refactor: test and lint fixes

denkristoffer:
# Purpose of PR

Test and lint fixes. More to come.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -102,9 +102,6 @@ module.exports = {
     {
       files: [
         '**/forma-36-react-components/tools/**/*',
-        '**/forma-36-react-components/**/Icon/svg/*',
-        '**/forma-36-react-components/**/ProductIcon/**/*',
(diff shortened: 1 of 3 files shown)


.eslintrc.js
```diff
@@ -125,5 +122,12 @@ module.exports = {
         'import/no-extraneous-dependencies': 'off',
       },
     },
+    {
+      // Example files and codemods are allowed to log to console
+      files: ['**/examples/**', '**/packages/forma-36-codemod/**/*'],
+      rules: {
+        'no-console': 'off',
```
mshaaban0, line 129: Should we go with warn? So at least VSCode can highlight it

mshaaban0 (approved):

.eslintrc.js
denkristoffer: It's not disabling the rule globally, just for our code examples — where we're using it in click handlers — and our codemod, which is supposed to write to the console 🙂 It's still a warning everywhere else.

PR #2191, opened 2022-09-02
build: remove assignees from issue templates

denkristoffer:
# Purpose of PR

Get rid of assignees from issue templates, I think it's clear we're not keeping them up to date and maybe we don't need them? 🙂


Lelith (approved):
Looks good to me!

burakukula (approved):

PR #2194, opened 2022-09-05
chore(deps): remove explicit webpack loaders

denkristoffer:
# Purpose of PR

Clean up some more dependencies by removing explicit Webpack loader dependencies.


mshaaban0 (approved):

mshaaban0:
There's 80% reduction in the size limit, that's too good to be true, are we sure we're good without these loaders? 🙂

denkristoffer:
@mshaaban0 The size limit check is not trustworthy yet, sorry 🙁

PR #2201, opened 2022-09-06
fix: fully specify JS imports

denkristoffer:
# Purpose of PR

I can't get size-limit to work otherwise, which is an indication other Webpack users would run into problems 🙁

### DIFF
--- a/packages/components/datetime/src/RelativeDateTime/RelativeDateTime.tsx
+++ b/packages/components/datetime/src/RelativeDateTime/RelativeDateTime.tsx
@@ -6,9 +6,9 @@ import type {
 } from '@contentful/f36-core';
 
 import dayjs from 'dayjs';
-import utcPlugin from 'dayjs/plugin/utc';
-import relativeTime from 'dayjs/plugin/relativeTime';
-import calendarPlugin from 'dayjs/plugin/calendar';
+import utcPlugin from 'dayjs/plugin/utc.js';
+import relativeTime from 'dayjs/plugin/relativeTime.js';
+import calendarPlugin from 'dayjs/plugin/calendar.js';
 dayjs.extend(utcPlugin);
 dayjs.extend(relativeTime);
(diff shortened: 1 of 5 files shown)


massao (approved):

massao:
Do you have a link or something with explanation why it doesn't work otherwise?? for context and sutff

denkristoffer:
@massao @mshaaban0 Sure. Let me know if this link doesn't work but here's the full error: https://github.com/contentful/forma-36/runs/8208682864?check_suite_focus=true#step:3:1495

Basically it looks like this:
> "error": "Error: ERROR in ./packages/components/datetime/dist/index.mjs 3:0-33\nModule not found: Error: Can't resolve 'dayjs/plugin/utc' in '/home/runner/work/forma-36/forma-36/packages/components/datetime/dist'\nDid you mean 'utc.js'?\nBREAKING CHANGE: The request 'dayjs/plugin/utc' failed to resolve only because it was resolved as fully specified\n(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '\"type\": \"module\"').\nThe extension in the request is mandatory for it to be fully specified.\nAdd the extension to the request.

As you can see one of the breaking changes in Webpack 5 is that it requires you to fully specify the file endings of imports in ESM JS files (like our `.mjs`). I had worked around this with [this change](https://github.com/contentful/forma-36/pull/2201/files#diff-6b43ff354d5b098f231b01acae212968bfb9b2826e1ddaddde9b1cdd233fd3e4) in our Storybook Webpack config, but that doesn't seem to work for size-limit, which also uses a Webpack check that we want to support. To my mind this is an indication that consumers could run into the same issue with their Webpack setup, and I don't want them to have to go through that override to use Forma 36, so what I decided to fully specify the file endings instead, like the error wants us to.

PR #2203, opened 2022-09-07
feat: Forma 36 website preview

bgutsol:
# Purpose of PR

- Add preview logic and api routes

### DIFF
--- a/packages/website/next.config.js
+++ b/packages/website/next.config.js
@@ -4,5 +4,5 @@ const withMDX = require('@next/mdx')({
 });
 
 module.exports = withMDX({
-  pageExtensions: ['tsx', 'mdx'],
+  pageExtensions: ['ts', 'tsx', 'mdx'],
 });
--- a/packages/website/pages/api/preview.ts
+++ b/packages/website/pages/api/preview.ts
@@ -1,0 +1,33 @@
+import { getSingleArticleBySlug } from '../../lib/api';
+
+export default async function preview(req, res) {
+  const { secret, slug } = req.query;
+
+  if (secret !== process.env.CONTENTFUL_PREVIEW_ACCESS_TOKEN || !slug) {
+    return res.status(401).json({ message: 'Invalid token' });
+  }
+
+  // Fetch the headless CMS to check if the provided `slug` exists
+  const post = await getSingleArticleBySlug(slug, true);
+
+  // If the slug doesn't exist prevent preview mode from being enabled
+  if (!post) {
+    return res.status(401).json({ message: 'Invalid slug' });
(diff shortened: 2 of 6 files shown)


packages/website/pages/api/preview.ts
```diff
@@ -0,0 +1,33 @@
...
+  if (secret !== process.env.CONTENTFUL_PREVIEW_ACCESS_TOKEN || !slug) {
+    return res.status(401).json({ message: 'Invalid token' });
+  }
+
+  // Fetch the headless CMS to check if the provided `slug` exists
+  const post = await getSingleArticleBySlug(slug, true);
+
+  // If the slug doesn't exist prevent preview mode from being enabled
+  if (!post) {
+    return res.status(401).json({ message: 'Invalid slug' });
```
mshaaban0, line 15: Shouldn't this be 404 since it's not found?

mshaaban0 (approved):
Works very well, thanks

Screenshot 2022-09-07 at 12 54 02

PS: I changed the preview url on Contentful to the preview link of this PR for testing. After merge I assume we will use the default url

packages/website/next.config.js
```diff
@@ -4,5 +4,5 @@ const withMDX = require('@next/mdx')({
 });
 
 module.exports = withMDX({
-  pageExtensions: ['tsx', 'mdx'],
+  pageExtensions: ['ts', 'tsx', 'mdx'],
```
denkristoffer, line 7: This PR doesn't add any `.ts` pages, are you sure this is necessary?

packages/website/pages/api/preview.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  // Enable Preview Mode by setting the cookies
+  res.setPreviewData({});
+
+  // Redirect to the path from the fetched post
+  // We don't redirect to req.query.slug as that might lead to open redirect vulnerabilities
+  // res.writeHead(307, { Location: `/${post.slug}` })
+  const url = `/${post.slug}`;
+  res.setHeader('Content-Type', 'text/html');
+  res.write(
```
denkristoffer, line 26: I think you can do something like (pseudo code) `res.status(200).redirect(307, url);` instead.

packages/website/next.config.js
bgutsol: yes, files under api folder are `.ts`

denkristoffer: Yeah but they don't render as pages anyway.

bgutsol: NextJs ignores files which extensions are not in this array

PR #2206, opened 2022-09-08
chore: add script to make versioning and publishing prereleases easier

massao:
# Purpose of PR

Add script to make the handling of prereleases packages easier.

The script will handle:
* bumping the version according to the selected prerelease (alpha, beta)
* adding to the ignore field on changeset configuration
* publish to npm

PS: kept with the `--dry-run` flag for now for testing, will remove it before merging into `main`

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -2,7 +2,12 @@
   "$schema": "https://unpkg.com/@changesets/config@1.6.4/schema.json",
   "changelog": [
     "@changesets/changelog-github",
-    { "repo": "contentful/forma-36" }
+    {
+      "repo": "contentful/forma-36"
+    }
+  ],
+  "ignore": [
+    "@contentful/f36-website"
   ],
   "commit": false,
   "fixed": [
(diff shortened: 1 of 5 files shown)


Lelith (approved):
thank you so much for taking care of this!

denkristoffer:
How, if at all, does this handle having both a stable and a prerelease version of a component? Is that even possible?

massao:
> How, if at all, does this handle having both a stable and a prerelease version of a component? Is that even possible?

It doesn't, it could be possible with a bit more tweaking, because we would need to know if it is a new prerelease our going back to a previous one have to store the information of the prerelease somehow, so there is no version conflict, e.g:

```
alpha -> stable fix -> back to alpha = `alpha.x`
or
alpha -> merge alpha -> new alpha = `4.x.x-alpha.0`
```

For that we could have a json file that store the prerelease versions that weren't merged into stable yet, but we would need to also add a command to mark that we will merge a prerelease package into stable. 🤔 

WDYT?

denkristoffer:
I think that's okay, but we should add documentation specifying that prerelease is only for new components. Or something like that. What do you think?

denkristoffer (approved):

PR #2217, opened 2022-09-12
feat: update react-day-picker to fix some minor issues, pin the version

bgutsol:
# Purpose of PR

- update react-day-picker to fix some minor issues (for example before it was possible to focus disabled days with keyboard navigation )
- pin `react-day-picker` version to prevent breaking changes in styles

### DIFF
--- a/packages/components/datepicker/package.json
+++ b/packages/components/datepicker/package.json
@@ -16,7 +16,7 @@
     "@contentful/f36-tokens": "^4.0.0",
     "emotion": "^10.0.17",
     "date-fns": "^2.28.0",
-    "react-day-picker": "8.2.1",
+    "react-day-picker": "^8.2.1",
     "react-focus-lock": "^2.9.1"
   },
   "peerDependencies": {
--- a/packages/components/datepicker/package.json
+++ b/packages/components/datepicker/package.json
@@ -16,7 +16,7 @@
     "@contentful/f36-tokens": "^4.0.0",
     "emotion": "^10.0.17",
     "date-fns": "^2.28.0",
-    "react-day-picker": "^8.0.7",
+    "react-day-picker": "8.2.1",
     "react-focus-lock": "^2.9.1"
   },
   "peerDependencies": {
--- a/packages/components/datepicker/src/Calendar/Calendar.styles.ts
+++ b/packages/components/datepicker/src/Calendar/Calendar.styles.ts
@@ -196,6 +196,7 @@ export const getStyles = (): ClassNames => {
       fontWeight: tokens.fontWeightDemiBold,
       textAlign: 'center',
       height: '32px',
+      color: tokens.gray600,
     }),
 
     tbody: css({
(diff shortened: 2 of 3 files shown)


mshaaban0 (approved):

packages/components/datepicker/package.json
```diff
@@ -16,7 +16,7 @@
     "@contentful/f36-tokens": "^4.0.0",
     "emotion": "^10.0.17",
     "date-fns": "^2.28.0",
-    "react-day-picker": "^8.0.7",
+    "react-day-picker": "8.2.1",
```
mshaaban0, line 19: Do we want to ignore the patches as well? I'm not sure about the pinning part

denkristoffer: I don't think we should pin.

bgutsol: classnames should not change, but there was already a case when they introduced a fix and therefore changed a pseudo-class from aria-disabled to disabled. What problem do you see with pinning this dependency?

bgutsol: I thought dependabot would still create for us PRs to bump this package, no?

mshaaban0: dependabot will stop bumping versions, i think we can cover such cases with tests

bgutsol: How would you cover it with tests?
And the additional problem is that you can get a newer version of `react-day-picker` in exp-packages for example and you will get broken styles there.

bgutsol: if I allow patch updates, will dependabot work?

mshaaban0: > How would you cover it with tests?

Basically check if the classnames you're overriding being rendered when the component mount, it should be fairly simple.

Is experience-packages consuming react-day-picker directly? Isn't it consuming datepicker from f36? So broken styles will never reach the packages if tests are failing in dependabot PRs

bgutsol: Btw, why do you think dependabot will stop bumping this package? 
It says that by default it will bump every dep defined in the package.json
image

denkristoffer: >  And the additional problem is that you can get a newer version of react-day-picker in exp-packages for example and you will get broken styles there. It is actually what happened to me right now.

If styles can break like that we should probably pin. But it seems brittle, maybe we can make the implementation more robust later on.

mshaaban0: Also chromatic visual diffs should break when styles change 🤔

denkristoffer: > Also chromatic visual diffs should break when styles change 🤔

But that's the problem if we don't pin, consumers can have a different version of react-day-picker than the one we've built with.

bgutsol: I'd say if the pinned dep is such a problem then we can unpin it back. Because I've already changed all the pseudo-classes to regular ones, except the `dropdown`. The regular class names should not change easily, it's gonna be major update and we should catch it with our styles types.

denkristoffer (approved):
LGTM.

We can revisit the pinning issue if it happens again.

PR #2227, opened 2022-09-14
fix: add button margin reset on GlobalStyles

massao:
# Purpose of PR
Safari has a default margin set to `2` on buttons, we reset all buttons to have no margin, to fix issues with spacing

**Before:**
image

**After:**
image


mshaaban0 (approved):

denkristoffer (approved):
Nice catch :+1:

PR #2232, opened 2022-09-14
fix(Notification): break long words into new lines

mshaaban0:
# Purpose of PR

- Long words in notifications bugs the component. hence, we introduce this change to break long words into new lines
- Fix Notifications triggers alignment in docs

**Before**
image

Screenshot 2022-09-14 at 16 45 35

**After**
Screenshot 2022-09-14 at 16 31 51
Screenshot 2022-09-14 at 16 46 23

### DIFF
--- a/packages/components/notification/src/NotificationItem/NotificationItem.styles.ts
+++ b/packages/components/notification/src/NotificationItem/NotificationItem.styles.ts
@@ -54,6 +54,7 @@ const titleStyle: CSSObject = {
 const contentStyle: CSSObject = {
   color: tokens.gray700,
   wordBreak: 'break-word',
+  hyphens: 'auto',
   '&:last-child': {
     marginBottom: 0,
   },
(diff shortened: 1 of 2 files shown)


packages/components/notification/src/NotificationItem/NotificationItem.styles.ts
```diff
@@ -53,6 +53,7 @@ const titleStyle: CSSObject = {
 
 const contentStyle: CSSObject = {
   color: tokens.gray700,
+  wordBreak: 'break-word',
```
denkristoffer, line 56: Do we want a `hyphens: auto` in there as well?

mshaaban0: I don't see it changing anything 🤷🏼 
Screenshot 2022-09-14 at 16 57 18

denkristoffer: It would be almost the same, but in some cases it would insert a hyphen which looks better and less broken:

> The browser is free to automatically break words at appropriate hyphenation points, following whatever rules it chooses.
> – https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens

massao (approved):

Issue #1857, opened 2022-02-07
🐛 Bug - TextInput in TextInput.Group is positioned on top of tooltips


denkristoffer:
# Forma 36 bug report

## Summary

If I add a `TextInput` and a `CopyButton` with a tooltip in a `TextInput.Group`, the input is positioned on top of the tooltip:

<img width="351" alt="Screenshot 2022-02-07 at 13 03 34" src="https://user-images.githubusercontent.com/219017/152784933-6ce184b5-07c5-4785-a1e5-0ee3e40b352c.png">

The only way to get around it is to set `usePortal` on the `Tooltip`, but there are situations where this is not great, such as when the tooltip is placed so close to the edge of the bounding box that using a portal would create overflow.

denkristoffer:
Only other thing I can think of is that the input is disabled 🤔 

```tsx
<TextInput.Group>
  <TextInput
    isDisabled
    value={contentTypeId}
  />
  <CopyButton value={contentTypeId} />
</TextInput.Group>
```

denkristoffer:
@contentful/team-bauhaus Can anyone have a look again please? 🙇

massao:
Hey @denkristoffer the cause of this one is because when the input is `disabled`, `invalid` or `focused` it has a higher z-index than the other elements, so the glow of the focus or incorrect is always upfront.
So this causes other problems, as if the user has the input selected, and hover on the button, the tooltip will also be behind it, and if we increase the z-index of the tooltip, the glow will be cut, here for example:
![image](https://user-images.githubusercontent.com/1071799/157060488-2a16a01e-48ff-4d31-a709-395cea5aff96.png)
![image](https://user-images.githubusercontent.com/1071799/157060694-ba8c5fc6-c353-44cf-b5fb-e49f4ed597e0.png)

I will create a PR to only have this behaviour of increasing `z-index` only when the field is focused instead, but I think it would be worth for the designers to take a look to see if we can have a different design that wouldn't cause this problem with parts of elements being overlapped, maybe not having the glow, or always having a spacing between elements, don't know.
cc @fabe @domarku

denkristoffer:
Looks like we've lost the error border in groups with this fix: https://www.chromatic.com/test?appId=5fd1dda724cc620021ace8c5&id=62289dd43214f9003aaff9aa

PR #2243, opened 2022-09-19
fix: handling of zIndex on TextInput.Group

### LINKED ISSUE #1978: 🐛 Bug - invalid inputs in InputGroup don't show right border
# Forma 36 bug report

## Summary

After fixing #1857 the invalid inputs in an InputGroup don't show the red border on the right side, the one bordering the button.
See https://www.chromatic.com/test?appId=5fd1dda724cc620021ace8c5&id=62289dd43214f9003aaff9aa
Screenshot 2022-03-30 at 17 05 43

massao:
# Purpose of PR

Fixes handling of zIndex on TextInput.Group, by giving `invalid` and `disabled` inputs a higher value, but if an element inside the group is `hovered` or `focused`, it gets an higher z-index in case it has a tooltip, it's not overlapped by the input itself (refer to #1857)

closes #1978 

> ~~**Known issues**~~
> ~~since we are changing the z-index in multiple places, some overlaps may happen:~~
> ~~If the input is focused~~
> ~~image~~
> ~~And then the button with a tooltip is hovered, it will appear like this~~
> ~~image~~
>
> ~~@fabe FYI~~

By removing the base z-index of all children elements on the InputGroup, and only updating the z-index of the input when it's invalid or disabled, seems to solve the issue:
image

You can check it on [Storybook](https://5fd1dda724cc620021ace8c5-gfctmvkyiq.chromatic.com/?path=/story/form-elements-inputgroup--overview)

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/.changeset/smooth-students-watch.md
+++ b/.changeset/smooth-students-watch.md
@@ -1,0 +1,5 @@
+---
+'@contentful/f36-forms': patch
+---
+
+fix zIndex handling on TextInput.Group
--- a/packages/components/forms/src/BaseInput/BaseInput.styles.ts
+++ b/packages/components/forms/src/BaseInput/BaseInput.styles.ts
@@ -17,11 +17,18 @@ const getSizeStyles = ({ size }): CSSObject => {
   };
 };
 
+const getZIndex = ({
+  isDisabled,
+  isInvalid,
+  zIndexBase = tokens.zIndexDefault,
+}) => (isDisabled || isInvalid ? zIndexBase + 1 : zIndexBase);
+
 const getStyles = ({ as, isDisabled, isInvalid, size, resize }) => ({
   rootComponentWithIcon: css({
     position: 'relative',
     display: 'flex',
     width: '100%',
+    zIndex: getZIndex({ isDisabled, isInvalid }),
   }),
   input: css({
     outline: 'none',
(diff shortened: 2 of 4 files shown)


denkristoffer:
Do we have any idea why we did it the other way before? I'm fearing some kind of edge case is going to slip by us here 🤔

massao:
> Do we have any idea why we did it the other way before? I'm fearing some kind of edge case is going to slip by us here 🤔
> 
> Played around with it a bit here, the behaviour with the active button is a bit weird.
> 
>  Screen.Recording.2022-09-20.at.15.18.55.mov

Not that I can remember 🤔 
Maybe we will need to rethink how we display it, I was checking the toggle-group of [radix](https://www.radix-ui.com/docs/primitives/components/toggle-group), that is kind of similar to this.
But in their case since it's not an input that can have different states, they handle the border, by faking the border with a background on the container, and having the children having a margin around them.
And using box-shadow to create the focus border, which doesn't cause the element changing sizes.

But I don't think that is applicable in our case, since we want only the input to have the red borders in case it's invalid,
we could remove the negative margin, but it would look weird in some cases, like the border being too thick: 
image

But we wouldn't need to play around with z-indexes  and would probably avoid other problems 🤔

fabe (approved):
The current state on the Chromatic preview looks good to me. We'll revisit this later to see if there are more fundamental changes we can make to this component to avoid these edge cases altogether. The double border we should definitely avoid!

PR #2245, opened 2022-09-20
fix(codemod): pass the name prop to the Select not FormControl

massao:
# Purpose of PR

Select codemod was passing the name prop to the FormControl instead of to the Select.


Lelith (approved):
lgtm 👍

denkristoffer (approved):

PR #2249, opened 2022-09-21
refactor(autocomplete): add prop to force showing the empty list when focused

massao:
# Purpose of PR

Add prop to force showing the list when focused, even if empty and no search value is set.
Instead of not rendering, we render a empty hidden div to pass the ref to it to avoid downshift error messages, for more context see [here](https://github.com/downshift-js/downshift/issues/1167#issuecomment-1088022842)

### DIFF
--- a/packages/components/autocomplete/src/Autocomplete.tsx
+++ b/packages/components/autocomplete/src/Autocomplete.tsx
@@ -384,6 +384,8 @@ function _Autocomplete<ItemType>(
               )}
           </Popover.Content>
         ) : (
+          // We need to render an empty hidden div, so we can pass the menuProps or downshift will show a warning about it
+          // https://github.com/downshift-js/downshift/issues/1167#issuecomment-1088022842
           <div {...menuProps} className={cx(styles.hidden)} />
         )}
       </Popover>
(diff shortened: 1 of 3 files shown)


denkristoffer:
Alternatively we could make the component take a prop to fully control the cutoff for showing the empty list message. `items.length > cutoff` with `cutoff` defaulting to `0`. I'm wondering if consumers want that level of control.

massao:
> Alternatively we could make the component take a prop to fully control the cutoff for showing the empty list message. `items.length > cutoff` with `cutoff` defaulting to `0`. I'm wondering if consumers want that level of control.

Hmm can you think of a use case where that would be necessary??
What I was also thinking is that in case you have lots of items, you may want to not show them, without a input value set, but that could be handled by using it as a controlled component, and providing the filtered items when needed 🤔

denkristoffer:
> Hmm can you think of a use case where that would be necessary??

I can't really think of anything for us, but it seems more flexible and could hopefully help us avoid coming back to this for a third time in the future. I'm not against landing this as it is though 🙂

packages/components/autocomplete/src/Autocomplete.tsx
```diff
@@ -377,6 +383,8 @@ function _Autocomplete<ItemType>(
                 />
               )}
           </Popover.Content>
+        ) : (
+          <div {...menuProps} className={cx(styles.hidden)} />
```
bgutsol, line 387: Could you add a comment here explaining why we need it, so we don't remove it later?

bgutsol (approved):
one minor comment, otherwise LGTM

PR #2251, opened 2022-09-22
build: switch to simple-git-hooks

denkristoffer:
# Purpose of PR

Husky v4 is unmaintained. We can either upgrade to Husky v8, which honestly seems pretty cool but [has dropped support for autoinstalling](https://blog.typicode.com/husky-git-hooks-autoinstall/), or we can move to something similar to v4. I've looked at [simple-git-hooks](https://github.com/toplenboren/simple-git-hooks) which seems like a nice, small alternative. If it doesn't work for us we can always try the larger move to Husky v8.


massao (approved):
LGTM

PR #2258, opened 2022-09-29
chore: add function to use NPM_TOKEN on CI

massao:
# Purpose of PR

Publishing to npm is failing on CI asking that the user should be logged into NPM.
We are adding a function to use NPM_TOKEN on the CI to avoid this problem

### DIFF
--- a/scripts/changesets/generate-releases.js
+++ b/scripts/changesets/generate-releases.js
@@ -54,20 +54,11 @@ const getReleasedPackages = async (csOutput, pkgs) => {
   }, []);
 };
 
-function setNpmToken() {
-  const env = process.env;
-  const registry = '//registry.npmjs.org/';
-  childProcess.execSync(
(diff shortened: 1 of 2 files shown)


scripts/changesets/generate-releases.js
```diff
@@ -54,11 +54,20 @@ const getReleasedPackages = async (csOutput, pkgs) => {
   }, []);
 };
 
+function setNpmToken() {
+  const env = process.env;
+  const registry = '//registry.npmjs.org/';
+  childProcess.execSync(
+    `npm config set ${registry}:_authToken=${env.NPM_TOKEN}`,
```
mshaaban0, line 61: I think that might expose the token to logs, I would do it in circleci config so they can hide it.
https://circleci.com/docs/deploy-to-npm-registry

mshaaban0 (approved):

PR #2264, opened 2022-10-04
feat(f36-codemod): ESM

denkristoffer:
# Purpose of PR

ESM codemod to unblock package upgrades.

### DIFF
--- a/.changeset/calm-eggs-lay.md
+++ b/.changeset/calm-eggs-lay.md
@@ -1,0 +1,5 @@
+---
+"@contentful/f36-codemod": minor
+---
+
+feat(f36-codemod): ESM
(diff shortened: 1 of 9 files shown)


massao (approved):

denkristoffer:
I tried this out locally and it seemed to work fine even with node 14. @massao are there any gotchas or specific situations you think I should test before merging?

massao:
> I tried this out locally and it seemed to work fine even with node 14. @massao are there any gotchas or specific situations you think I should test before merging?

maybe just running it against some other repos??

massao:
btw, should we mark this as a major change on the codemod?

denkristoffer:
I ran the wizard on a part of our web app. Will try out with arguments.

PR #2265, opened 2022-10-05
feat(badge): start and end icon

### LINKED ISSUE #2125: 💡 Proposal - Badge component: Pass Icon component

## The problem

In User Interface, there are some use cased where Badge is being used with an Icon / Image as a child component, however, this breaks the layout:

image

I understand it is also used in launch, with some color inconsistencies:

image

Currently, i'm having to add additional css to fix this, e/g give the Icon absolute positioning.
## The proposed solution

Add an `icon` prop to the Badge, similar to how `IconButton` works.

## Breaking changes

I don't believe so.

denkristoffer:
All credit goes to @Wake1st. I messed up the history of #2220, it's easier to pick it up here than try to fix it...

Closes #2125 

# TODO

~~Need to figure out how to get the `EntityStatusBadge` to build with the union types for `BadgeSize` and the icons. The API works pretty well other than that 👍~~


packages/components/badge/src/Badge/Badge.tsx
```diff
@@ -5,49 +5,83 @@ import {
...
-
-  children: React.ReactNode;
-}
+type BadgeSizeWithIconProps =
+  | {
+      /**
+       * Sets the size of the component
+       * @default default
+       */
+      size?: PickUnion<BadgeSize, Exclude<BadgeSize, 'small'>>;
```
bgutsol, line 20: If I got the logic correctly, we can simplify this:
```suggestion
      size?: Exclude<BadgeSize, 'small'>;
```

packages/components/badge/src/Badge/Badge.tsx
```diff
@@ -5,49 +5,83 @@ import {
...
+       * Expects any of the icon components. Renders the icon aligned to the end
+       */
+      endIcon?: React.ReactElement;
+    }
+  | {
+      /**
+       * Sets the size of the component
+       * @default default
+       */
+      size: PickUnion<BadgeSize, 'small'>;
```
bgutsol, line 35:
```suggestion
      size?: BadgeSize;
```

packages/components/badge/src/Badge/README.mdx
```diff
@@ -26,69 +26,82 @@ import { Badge } from '@contentful/f36-badge';
 
 Used for the entity that has been changed
 
-```jsx file=../../examples/PrimaryBadgeExample.tsx
+```tsx file=../../examples/PrimaryBadgeExample.tsx
```
bgutsol, line 29: Types are not allowed in the example files, otherwise, it will break the playground we use. 
Even though `tsx` lang type here won't break anything, I'd suggest rolling it back to `jsx` for the sake of consistency with other components examples.

I'd say we probably need to change all the example files to `.jsx` instead. But it's not this PR scope for sure.

bgutsol (commented):
Nice one 🏅
Just some nits 💅

packages/components/badge/src/Badge/Badge.tsx
massao: We can't do that sadly, because we want to avoid the usage of the icons when the size is set to `small`, which is why we are using this discriminated union type, so in case the user sets `size="small"` and tries to add an icon TS will show an error

mshaaban0: A comment here might be helpful for future human browsers

packages/components/badge/src/EntityStatusBadge/EntityStatusBadge.tsx
```diff
@@ -14,19 +14,18 @@ const statusMap: { [key in EntityStatus]: BadgeVariant } = {
 };
 
 export interface EntityStatusBadgeProps
-  extends Omit<BadgeProps, 'variant' | 'children'> {
-  size?: BadgeSize;
+  extends Omit<BadgeProps, 'variant' | 'children' | 'endIcon' | 'startIcon'> {
```
mshaaban0, line 17: Any reasons why are we omitting the icons here? Are we certain we don't want icons with EntityStatusBadge?

mshaaban0 (approved):
Other than EntityStatusBadge comment, it LGTM ✅

packages/components/badge/src/EntityStatusBadge/EntityStatusBadge.tsx
massao: I didn't actually touch this part, I'm trying to find a bit of context around this, to be sure 🤔

Wake1st: That was probably me; I asked @denkristoffer on the contentful slack channel, and he said that it probably didn't need to be added.
image

I believe the reason is that the status badges already have a predetermined design which doesn't include using icons. See the [storybook](https://f36-storybook.contentful.com/?path=/story/components-entitystatusbadge--overview) for and example of this. 

Of course, the icons can always be added in the future.

bgutsol (approved):
🎅 🎁 🎄

denkristoffer:
@all-contributors please add @Wake1st for code

PR #2269, opened 2022-10-06
fix: codemod issues text field select

massao:
* Fix issue with transforming objet to prop when value was boolean
* Fix issue with validationMessage when using memberExpression like: `fields.name.error`


0math (approved):
💯

PR #2270, opened 2022-10-06
fix(multiselect): show noMatchesMessage, closeOnBlur after clear [HOMER-1178]

Chaoste:
# Purpose of PR

- fix: pass through popover className to overwrite zIndex
- fix: scan through children hierarchy for transformation and correct count
- fix: focus list when clicking search to keep the focus for escOnBlur

## PR Checklist

- [x] Tests are passing


packages/components/multiselect/src/Multiselect.tsx
```diff
@@ -172,43 +215,33 @@ function _Multiselect(props: MultiselectProps, ref: React.Ref<HTMLDivElement>) {
...
-        }
-        return child;
-      });
+      return iterateOverChildren(
+        children,
+        (child) => child.type === MultiselectOption,
+        (child) => {
+          const onSelectItem = (even: React.ChangeEvent<HTMLInputElement>) => {
+            focusList();
+            child.props?.onSelectItem(even);
```
Lelith, line 235:
```suggestion
          const onSelectItem = (evt: React.ChangeEvent<HTMLInputElement>) => {
            focusList();
            child.props?.onSelectItem(evt);
```

idk if it was my typo but even is really confusing me as variable name in this context

packages/components/multiselect/src/Multiselect.tsx
```diff
@@ -86,14 +86,48 @@ export interface MultiselectProps extends CommonProps {
...
+    // equal to (if (child == null || typeof child == 'string'))
+    if (!React.isValidElement(child)) return child;
+    if (!filter(child))
+      return iterateOverChildren(child.props.children, filter, transform);
+    return transform(child);
+  });
+};
+
+// Scan through the whole hierachy to count the number of children where `filter` returns true
+const countMatchingChildren = (
```
Lelith, line 114: i m not 100% sold on not allowing people to show other children than options. We already had the case for autocomplete where someone wanted to show that the backend just did not respond anything to show without a search.

so maybe if we have children but none is an option we just show that as content but not inside the multiselect item list?

packages/components/multiselect/src/Multiselect.tsx
```diff
@@ -133,18 +175,19 @@ function _Multiselect(props: MultiselectProps, ref: React.Ref<HTMLDivElement>) {
     [onSearchValueChange, setSearchValue],
   );
 
-  const resetSearch = () => {
+  const resetSearchInput = useCallback(() => {
+    if (!searchValue) return;
```
Lelith, line 179: as the button is inactive when there is no search value, this is a little double trouble?

Lelith (changes requested):
lets talk again about when to not render which children.
also as the cloning is now so much more complex i think we need a little more test coverage on this.

packages/components/multiselect/src/Multiselect.tsx
Chaoste: True, I don't where this comes from ^^
I'd rather call it `event` since it doesn't save space but is more readable

packages/components/multiselect/src/Multiselect.tsx
Chaoste: This logic already existed before but below when calling`resetSearch`. I just wanted to put this in a `useCallback` and clean up the render method. We can also drop it if we don't need it anymore. Only case to have this is that we also enforce programmatically this behaviour while you can just remove the `disabled` tag via the devtools in the browser.

packages/components/multiselect/src/Multiselect.tsx
Lelith: i think we are talking about different things.
When I interpret this piece of code correctly, if you do not hand over an MultiselectOption, none of the children gets rendered.
But i want to keep the possibility to render an individual message that is different from the "no matches found" information, that should only be rendered when the search returned no results.

Lelith (approved):
Thanks for elevating the filter functionality for cloning the components.

Chaoste:
Can anyone help out with merging? It doesn't let me merge it

PR #2278, opened 2022-10-13
feat: auth protected pages [CFCX-4]

bgutsol:
- Add protected pages that require auth
- Fix rich text list and list item render

**Keep in mind:**
Auth won't work on the preview website because the redirect URL is configured for prod. But you can test it locally.

**UPDATE:**
- restricted node version to `>=14.18.0`, because version `14.17.0` doesn't support `import('node:path')` that we use [here](https://github.com/contentful/forma-36/blob/f06baa85c5dc72fada7999cc67d3ef72f4a3267d/packages/website/pages/%5B...slug%5D.tsx#L147).


packages/website/next.config.js
```diff
@@ -5,4 +5,18 @@ const withMDX = require('@next/mdx')({
...
   pageExtensions: ['ts', 'tsx', 'mdx'],
+  async redirects() {
+    return [
+      {
+        source: '/login',
+        destination: '/api/auth/signin',
+        permanent: true,
+      },
+      {
+        source: '/signin',
```
denkristoffer, line 16: Do we need both `/login` and `/signin`?

packages/website/components/Topbar/Topbar.tsx
```diff
@@ -99,6 +105,14 @@ export function Topbar({ links }: TopbarProps) {
...
         <Flex className={styles.docSearchContainer}>
           <DocSearch />
+          {session ? (
+            <Button
+              className={styles.signOut}
+              onClick={() => signOut({ callbackUrl: '/' })}
+            >
+              Log out
+            </Button>
+          ) : null}
```
denkristoffer, line 115: Do we want a log in button?

bgutsol: No, on the last Forma meeting with designers we agreed that we won't add any links/buttons for login. So it's just URL that you have to remember basically.

packages/website/next.config.js
bgutsol: I though that it would be nice to have both since we don't have any link/button to the login page. Do you think it's overkill?

massao: I think that keeping only one would be fine

massao (approved):
lgtm

packages/website/next.config.js
bgutsol: done, thanks for the feedback!

bgutsol (commented):
@denkristoffer pls take a look one more time. I've also fixed the sidebar active link logic in the last commit

denkristoffer (approved):

PR #2302, opened 2022-10-25
feat(button): allow icon variant for transparent buttons

denkristoffer:
# Purpose of PR

This PR allows you to set an icon variant for buttons with `variant="transparent"`.

It also fixes a bug with the icon button: We previously allowed setting `variant` for all icon buttons, even the other variants. This is now fixed to only allow `variant="transparent"` just like the regular button.

### DIFF
--- a/.changeset/eleven-turkeys-press.md
+++ b/.changeset/eleven-turkeys-press.md
@@ -1,0 +1,7 @@
+---
+"@contentful/f36-button": patch
+"@contentful/f36-datepicker": patch
+---
+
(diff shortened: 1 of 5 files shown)


bgutsol (approved):
lgtm 
The only thing I'd ask you to remove the `variant="muted` prop that we pass here: https://github.com/contentful/forma-36/blob/a64a2ce00e1ae300228ff87d68d1fdebc61787e2/packages/components/datepicker/src/Datepicker.tsx#L228
since it's not relevant anymore

PR #2304, opened 2022-10-27
fix: remove default button margin in safari

bgutsol:
By default Safari adds `2px` margin to `button`


denkristoffer:
Don't forget the changeset 🙂

Chaoste (approved):
🚀

denkristoffer (approved):
Looks right to me 👍 I'd say we don't need the comment in all these styles but that's just a matter of opinion.

PR #2305, opened 2022-10-27
docs: add note about non-interactive elements to popover

massao:
# Purpose of PR

Add note about non-interactive elements on popover documentation


denkristoffer:
I'm not sure I agree with this, we're explicitly saying in the line above that you should use an interactive element, and then we give a way to hack around that?

bgutsol (commented):
Agree with @denkristoffer it shouldn't be in the `Content guidelines` section

bgutsol:
> yeah, I was unsure where to put it too, I thought to add at the bottom of the page But could be weird to be after accessibility, so maybe at the bottom of the examples, WDYT?

Yes, let's put it there

denkristoffer:
Unrelated to this PR, but I can't see the change on the Vercel preview, probably because of caching. Would be nice to get that fixed.

denkristoffer (approved):
Thanks for fixing the build! It seems a bit weird to me to have a note below the examples, but to be honest I don’t have a better suggestion :shrug:

bgutsol (approved):
I've added a commit with the suggestion, used our Note component to make it more isolated from other content, WDYT?

denkristoffer:
Yeah, nice! Are the code tags supposed to work inside the Note?

Also, when we use the Note like this I think we can move it to content guidelines again, what do you think?

Issue #2043, opened 2022-05-04
💡 Proposal - hidden=until-found attribute to accordion / collapse component


YvesRijckaert:
# Forma 36 contribution proposal

## The problem

The find-in-page functionality does currently not work with collapsed content.

## The proposed solution

The new `hidden=until-found` attribute and `beforematch` event, auto-opens section on text find. (Currently only works in Chromium 102), but it is part of the HTML spec.
This would be a great progressive enhancement, so in browsers that do not support the new attribute and event, the components will still work.

I would be happy to make this change.

## More info

https://developer.chrome.com/blog/hidden-until-found/

PR #2311, opened 2022-10-31
fix: display none has negative impact on svgs used inside a collapse 

Lelith:
# Purpose of PR
This is a side-effect fix.

I have multiple collapsable containers using this collapse animation. Inside of these containers i have a colorful svg icon with a gradient in several of them.

Every time the first container that has the icon gets collapsed, the gradient disappears. [This is a known chromium bug since almost 10 years now](https://bugs.chromium.org/p/chromium/issues/detail?id=258029) The issue exists not only in chrome but also in firefox (it was also in webkit, but safari fixed it at some point)

What happens is, that the reference all of the svg files have to the gradient (via id) always points to its first defintion. When this first svg is now set to display:none this defintion stops to exist for all following instances of the same svg in the page.

The simplest to achieve fix is: do not use display:none but visibility:hidden to hide the content.

If there are no objections or other reasons why display:none is the important rule, please let me know and we can think about other solutions.

Maybe its also worth checking again the proposal here https://github.com/contentful/forma-36/issues/2043

### DIFF
--- a/.changeset/forty-experts-hug.md
+++ b/.changeset/forty-experts-hug.md
@@ -1,0 +1,5 @@
+---
+'@contentful/f36-collapse': patch
+---
+
+Fix negative side effect for svgs
--- a/package-lock.json
+++ b/package-lock.json
@@ -37879,14 +37879,14 @@
     },
     "packages/components/accordion": {
       "name": "@contentful/f36-accordion",
-      "version": "4.21.0",
+      "version": "4.21.3",
       "license": "MIT",
       "dependencies": {
-        "@contentful/f36-collapse": "^4.21.0",
-        "@contentful/f36-core": "^4.21.0",
-        "@contentful/f36-icons": "^4.21.0",
+        "@contentful/f36-collapse": "^4.21.3",
+        "@contentful/f36-core": "^4.21.3",
+        "@contentful/f36-icons": "^4.21.3",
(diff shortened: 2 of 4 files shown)


Chaoste (approved):

denkristoffer (commented):
I wonder if this could break use of `useEffect` in children components 🤔  Any idea?

denkristoffer:
> Every time the first container that has the icon gets collapsed, the gradient disappears. [This is a known chromium bug since almost 10 years now](https://bugs.chromium.org/p/chromium/issues/detail?id=258029) The issue exists not only in chrome but also in firefox (it was also in webkit, but safari fixed it at some point)

Do you have a reproduction of this I can try?

Lelith:
@denkristoffer here is a super quick codepen: https://codepen.io/Lelith/pen/rNKxJBG
if you change the css from display:none to visibility:hidden the svg looks good again.

denkristoffer:
I figured the switch might affect the DOM but it seems to be fine.

Lelith:
@denkristoffer i saw that too, and locally it got fixed once i rebuilt the packages. it seems to be stuck in a inbetween state, like updating the css rules but not the javascript (or vice versa?)

Lelith:
so i tried it in our app, but it breaks the behaviour of the dragable again, so will close it

PR #2312, opened 2022-11-02
feat: Entity list item isSelected prop

### LINKED ISSUE #2077: Proposal - EntityListItem selection
### Discussed in https://github.com/contentful/forma-36/discussions/1000

<div type='discussions-op-text'>

<sup>Originally posted by **Spring3** August  6, 2020</sup>

## The problem

Currently EntityList can only render the EntityListItems. It would be a huge help if it could support selection, just like Cards component does.

## The proposed solution

Support the `selected` or `isSelected` property for the EntityListItem, that will, if `true`, apply styles to the EntityListItem, highlighting it as a selected item.

[Please check the way Cards currently work](https://forma-36-react-components.netlify.app/?path=/story/components-card--default) and I think replicating the same behaviour for the EntityListItem, would be just perfect for this feature

## Breaking changes

No breaking changes</div>

mgueyraud:
# Purpose of PR

Closes #2077.

Add a prop isSelected to the entity list item, and the corresponding styles of it. Basically the same way as [the card works](https://forma-36-react-components.netlify.app/?path=/story/components-card--default).

So, to maintain the consistency of the EntityListItem component, I only override the `backgroundColor` and the `boxShadow` when it has the `isSelected` prop, so it looks like this by itself:

Screen Shot 2022-11-01 at 22 19 54

And when it's used within a EntityList component, that component is in charge to add more styles such as left, right and top borders.

## PR Checklist

- [x] Tests are passing


denkristoffer:
Thanks for the contribution @mgueyraud! I will be closing this until or if it's ever ready. That's not meant to pressure you into working on it at all by the way, but there's no need to keep it laying around in the current state 🙂

PR #2328, opened 2022-11-14
feat(Skeleton): add and expose raduis properties

mshaaban0:
# Purpose of PR

* Adds default border radius of `4px` to Skeleton components
* Exposes `radiusX` and `radiusY` property to Skeleton components

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/components/skeleton/src/SkeletonText/SkeletonText.tsx
+++ b/packages/components/skeleton/src/SkeletonText/SkeletonText.tsx
@@ -1,3 +1,4 @@
+import tokens from '@contentful/f36-tokens';
 import React, { useCallback } from 'react';
 
 type stringOrNumber = string | number;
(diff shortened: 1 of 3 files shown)


denkristoffer:
I think I prefer the skeletons for text without the radius, but I'll let @fabe decide 😄

packages/components/skeleton/src/SkeletonText/SkeletonText.tsx
```diff
@@ -23,6 +24,14 @@ export interface SkeletonTextProps {
...
    */
   marginBottom?: stringOrNumber;
+  /**
+   * X-axis border radius (in number)
+   */
+  radiusX?: stringOrNumber;
+  /**
+   * Y-axis border radius (in number)
+   */
+  radiusY?: stringOrNumber;
```
bgutsol, line 34: is there a use case for a different border-radius between X and Y. Can we just expose one property that will be used for both values?

bgutsol: ah, I see that we already had it as a separate props in the SkeletonImage

bgutsol (approved):

fabe:
> I think I prefer the skeletons for text without the radius, but I'll let @fabe decide 😄

With v4 we added border radius to pretty much every block element, I think we just forgot to do it in this case. But we can always reevaluate this later and see if it works better or not!

fabe (approved):

PR #2329, opened 2022-11-15
fix: fix tooltip show/hide logic with delays

bgutsol:
- fix: fix tooltip show/hide logic with delays
- regenerate pkg-lock.json, since I couldn't run `npm i` due to deps conflicts:
image


denkristoffer (commented):
Don't forget the changeset :)

massao (approved):
LGTM

PR #2332, opened 2022-11-17
fix(basecheckbox): remove label prop

massao:
# Purpose of PR

Remove label prop as it was renamed to aria-label but kept on the typing


denkristoffer (approved):

mshaaban0 (commented):
I'm wondering why it was explicitly added in the first place 🤔 doesn't feel right. #Yolo

PR #2334, opened 2022-11-21
chore: bump workbench

bgutsol:
- bump workbench and manually publish to have the correct latest version on npm
image


denkristoffer (commented):
Don't forget the changeset 👍

bgutsol:
> 

we don't need it, I've already published the version manually. I'd argue we need it in our changelog

denkristoffer (approved):

PR #2338, opened 2022-11-23
fix(menu): Safari issue on closing menu clicking on trigger

massao:
# Purpose of PR

Fix issue with Safari when trying to close menu by clicking on the trigger

**Obs:**
* Due to safari not correctly passing focus to the button, sometimes it causes the blue border to always show on the menu when opened.
* Another proposed [solution](https://github.com/facebook/react/issues/2011#issuecomment-654830955) around this issue is to add a div with `tabindex=0` inside the button and use it to trigger the events, which in our case would be tricky since the trigger could be a button or not.
* Downshift has another [approach](https://github.com/downshift-js/downshift/blob/4c3d3eab8a5d1a826f792d0b8d1302715400d7ed/src/downshift.js#L720-L758) but I couldn't make it work in the menu, the blur event was always being triggered after the `MouseDown`, so trying to set the focus on the `onClick` wasn't working. Which would means that we would have to manually focus on the events.


packages/components/menu/src/Menu.tsx
```diff
@@ -150,12 +150,26 @@ export function Menu(props: MenuProps) {
     [closeAndFocusTrigger, handleArrowsKeyDown],
   );
 
+  const isMouseDown = useRef<Boolean>(false);
```
bgutsol, line 153:
```suggestion
 // the comment that explains the issue in safari that we are trying to fix here. Because the one we have in the `onMouseDown` event is not clear enough for me. 
  const isTriggerMouseDown = useRef<Boolean>(false);
```
WDYT?

packages/components/menu/src/Menu.tsx
```diff
@@ -150,12 +150,26 @@ export function Menu(props: MenuProps) {
...
       getTriggerProps: (_props = {}, _ref = null) => ({
+        onMouseDown: (event) => {
+          // handle odd case for Safari which
+          // prevent triggering blur when MouseDown
+          isMouseDown.current = true;
+          _props.onMouseDown?.(event);
+        },
+        onMouseUp: (event) => {
+          // handle odd case for Safari
+          // set isMouseDown as false to enable blur
```
bgutsol, line 169: and then this comments we can delete I guess

bgutsol (approved):
Nice, I like the approach you chose 👍

I would just add more explanation around the issue in the code.

bgutsol (approved):

PR #2341, opened 2022-11-25
fix(text-link): darker muted variant

denkristoffer:
# Purpose of PR

- Darkens the colour used in the `muted` variant of the `TextLink` component.
- Removes the style unit tests. Let's keep it to Chromatic.


denkristoffer:
With this change the `muted` variant is unnecessarily close to the `secondary` one. Let's instead consider deprecating `muted` altogether.

PR #2342, opened 2022-11-28
feat(workbench): make WorkbenchHeader description accept ReactElement

massao:
# Purpose of PR

* Update WorkbenchHeader component to also accept description as a React Component as well as string.
* Update description text color

https://github.com/contentful/user_interface/pull/12598#discussion_r1024889006

### DIFF
--- a/packages/components/workbench/src/WorkbenchHeader/WorkbenchHeader.tsx
+++ b/packages/components/workbench/src/WorkbenchHeader/WorkbenchHeader.tsx
@@ -18,7 +18,7 @@ export interface WorkbenchHeaderProps extends CommonProps {
   /** This is the icon that will be shown on the left side of the title and it's possible to use Forma 36’s icons or external icons */
   icon?: IconComponent | ReactElement;
   /** This is the text that will be shown on the right side of the title in the Header component */
-  description?: string;
+  description?: string | ReactElement;
   /** It's possible to pass a ReactNode to be shown at the end of the Header */
   actions?: ReactNode;
   /** If a function is passed to the onBack prop the Header will show a back button that will call this function when clicked */
@@ -38,6 +38,23 @@ export const WorkbenchHeader = ({
   const styles = getWorkbenchHeaderStyles(hasBackButton);
   const iconComponent =
     Icon === undefined ? null : isValidElement(Icon) ? Icon : <Icon />;
+
+  const renderDescription = (
+    description: WorkbenchHeaderProps['description'],
+  ) => {
+    if (typeof description === 'string') {
+      return (
+        <Paragraph
+          className={styles.description}
+          marginBottom="none"
+          marginRight="spacingM"
+        >
+          {description}
+        </Paragraph>
+      );
+    }
+    return description;
+  };
 
   return (
     <header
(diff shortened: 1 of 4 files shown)


packages/components/workbench/stories/Workbench.stories.tsx
```diff
@@ -42,7 +42,7 @@ export default {
 
 interface Args {
   ['Header Title']: string;
-  ['Header Description']: string;
+  ['Header Description']: string | ReactElement;
```
mshaaban0, line 45: nit: `ReactNode` has `ReactElement | string | number`  by default

mshaaban0 (approved):
Looks good 👍🏼

packages/components/workbench/src/WorkbenchHeader/WorkbenchHeader.tsx
```diff
@@ -73,15 +90,7 @@ export const WorkbenchHeader = ({
...
-      {description && (
-        <Paragraph
-          className={styles.description}
-          marginBottom="none"
-          marginRight="spacingM"
-        >
-          {description}
-        </Paragraph>
-      )}
+      {renderDescription && renderDescription(description)}
```
bgutsol, line 93: I guess it is supposed to be this, right?
```suggestion
      {description && renderDescription(description)}
```

packages/components/workbench/src/WorkbenchHeader/WorkbenchHeader.tsx
```diff
@@ -18,7 +18,7 @@ export interface WorkbenchHeaderProps extends CommonProps {
   /** This is the icon that will be shown on the left side of the title and it's possible to use Forma 36’s icons or external icons */
   icon?: IconComponent | ReactElement;
   /** This is the text that will be shown on the right side of the title in the Header component */
-  description?: string;
+  description?: string | ReactElement;
```
bgutsol, line 21: it should be `ReactNode`, if it's `ReactElement` then it will have to be rendered like `<Description>` (like icon prop), but we don't need it for the description
```suggestion
  description?: string | ReactNode;
```

bgutsol (changes requested):

packages/components/workbench/stories/Workbench.stories.tsx
massao: I know,I kept as this since this is how this is used on the title on the component too, probably because we don't want numbers only? 🤔 
but I can change it anyway, WDYT??

bgutsol: I'd say that for description it doesn't make much sense, even a number should be allowed if someone needs it

bgutsol (approved):
👍

packages/components/workbench/stories/Workbench.stories.tsx
mshaaban0: yeah I don't see a reason to prevent numbers as well

massao: Yeah and actually the users can even hack it, doing something like "123" even for the title 😅

PR #2323, opened 2022-11-09
docs: website package internal only

benomatis:
Added a note that the website package is intended for internal use only.

See Slack discussion https://contentful-community.slack.com/archives/C5AJ45S6R/p1667895043907169?thread_ts=1667658184.124739&cid=C5AJ45S6R

PR #2343, opened 2022-11-30
docs: add Note to outside contributors for forma websites

bgutsol:
# Purpose of PR

Replicate https://github.com/contentful/forma-36/pull/2323 since that one is failing on CI


denkristoffer:
I think the better solution here is to provide a JSON export that users can load into their own space with the right content model, and then add some docs around getting that up and running instead.

massao (approved):

PR #2344, opened 2022-12-02
fix(base-check-box): align items to flex-start

denkristoffer:
# Purpose of PR

Aligns checkbox label to `flex-start` to avoid it going above automatic width.


Lelith (approved):
Looks good to me!

massao (approved):

PR #2345, opened 2022-12-02
Extend Multiselect: Select All component

Lelith:
# Purpose of PR

This extends the Multiselect component for another compount component "SelectAll". This component renders another MultiSelect Option that expects a callback function which will select all your elements. 

image

image

How it works:
Add the SelectAll component somewhere in your Multiselect List. Then create a function that will toggle all your available options, just as any other onSelectItem handler, it expects an event as property.
In order to recognize when all options have been selected, it is recommended to implement another function for that, which uses a useMemo hook to detect changes in the current selection, and then check if all available options are selected.
See also the Storybook and documentation for example implementations.

Also fixes some minor issues:
- design of storybook example
- makes the width dependent of size of surrounding container
- the element list has now the fixed size of the trigger Button
- the 'n more' text is now in gray600
- adjusted the size of the search bar to be small


bgutsol (commented):
Huge work @Lelith  💪
I can see that because `Multiselect` is a controlled component you had to introduce a lot of complexity and hacky logic for that `checkAll` functionality. Have you considered just adding an example for the users where they can easily build it on their own? And maybe just provide some compound component for styling purposes. Seems like it will reduce complexity, won't reduce performance (no need to call every option callback to check all of them), and leaves flexibility to the end user (in case they want to adjust label or tweak a bit "select all" logic or pass a tracking on the callback). WDYT?
[For example](https://f36.contentful.com/playground?code=N4Igxg9gJgpiBcICWBbADhATgFwAQCUYBDMPAM0whVwHJNjSaBuAHQDtUMddhcBlbCQDWuAL64KVWgAFIbbDHlkArgBsA9GQDMANgC0kdBDaLsAZ2btOWPLwCya7EjMxVMUmImVqNWcYVKapq6eiiOzq7u2JZs7DAAHlx4sGREjhLKbKRIxrgOqk4ubqQAQkRmSGAAqmZEAOYwAKLxROhuABQAlDzsuLhyZnhmaCQwZrgAvLgA2r19tAAqmEQAbq64JaoQdTQANHN9NABiSGxsRFkwG1s7+2zztCfYJmbjAIJoaHsHtAByMAB3cYAdRgACMKgpvvd5jQYABhKgoGCYMBXeFEQQ3ZQwaEPGgABQAFhBsBBcABxIiqNyYACeeNwAF1WLEYep1LgFkTnLgAAbKFwCTEwPm4Xl1CCnOq4Mm4QZYK7YIlXIpRGBQXAsEDDUba+XkgFXMAXeUkgHivCnWUq3BVACScwGeGmatIGr4IzRZl28pg2D4kXdUE9ozMTMmBAY2AAdIKYMKFO1pkzOqy+s6ZtTVIHigooL6XNg3jTc+qoBGpoQSLH44mYO1UqoXGn2HMOVyeeMVFknLkAUgabgwcbsxrcMY0TarvHMH68+NTUgFNRrcqrqpnHgIGQJ2g+2wzE7jINcESLlA3GXSPaV5H2jA1vJuhMAHw9GEZk+2foqsBCDVfRWakcU8KZH1MGNBEwBpsHTeYkF3dowD-ACoG6YAfj6Itr3zUNvXado0HoFZ6xfd9phjKjiMfesgJAmBU3gvpxFcFwPweL9DzwEwAVwj0vTGSM3TwwSzBjMhBwUTBCN1NFyPlQTcAAQgmKZgNUHFW0-eYcKDUSw3aXj+JDMTtIeUQ5lEVlj24+cohLVR7wg59JnfTCdMzXgUPcNCwNwFzY2g2DmL9YtS30jVkNQjVzIQpCfP-WKOM4vS8wEwy5LGToflY5srg8ziwpM-CxmTVNcqstsYXobBlEwe52h+AAeYV-wkNx4gAESQehsmMCZtUgTSUFiEBcGpJA6jYW8YBQMxBp1aDsG1V8sNwZr8kKfT1ozer6HkXCcjYCZgBEjLvUsnSHgwDA1kwAlKDQBbgF4ZwjjUVRgSQKBlXgWVMFA0QrqKtbrvmTbwhEmMAHl92O3aistOb7SgRbcMc7VEaKjScUW7MsfBpHcFUIgR1UU7sxM3AAH5aC6sZ9ImmkaFwf6aFw5nVBoEHOJYFa8GRlBUfR-TMZAdb+f5vpjFw2aUFOkTHN5h4pcF5x4RitHgCpyKoBV4ncHUMHibOsSYxQIg0FkwSFMKw3Mw04TBKgiAABkICNTAMRcLoY3oNBSbRdp1BYFgzHUOpfRoPQaDi4navqxrsYhraIjzWH4eMFOHgAulTr5PO9AAEmADTgb5A3DaFkXgD5LKS7L6kK6rw3cZgRXBNb4nSfJzvRm7pHZf0+XTvPNhLwTEeV0HoqNa1xW9dK8TTjATTYDMdoss6Wf5mNlP4840Qd-W5r1DTkSTYh9Q2qEK-tKukBRCAA)

packages/components/multiselect/README.mdx
```diff
@@ -68,6 +68,14 @@ To enable searching the following properties have to be set.
 
 ```
 
+### SelectAll shortcut
+
+To offer a shortcut for selecting and deselecting all options, use the property hasSelectAll. This will render a checkbox on top of the list which will trigger the selection and deselection of all elements.
```
Chaoste, line 73: Here you say "hasSelectAll" but in the code it's "hasCheckAll"

packages/components/multiselect/src/Multiselect.tsx
```diff
@@ -60,6 +61,12 @@ export interface MultiselectProps extends CommonProps {
    */
   searchInputName?: string;
 
+  /**
+   * Enables a select All shortcut checkbox
```
Chaoste, line 65:
```suggestion
   * Enables a shortcut checkbox to select or delect all items
```

packages/components/multiselect/src/SelectAllOption.tsx
```diff
@@ -0,0 +1,91 @@
...
+  checked: boolean,
+) => {
+  if (child.props?.isChecked !== checked && !child.props?.isDisabled) {
+    event.target.value = child.props?.value;
+    event.target.checked = checked;
+    child.props?.onSelectItem(event);
+  }
+};
+
+const areChildrenAllSelected = (
```
Chaoste, line 26:
```suggestion
const areAllChildrenSelected = (
```

Chaoste, line 26: Also, this looks like copied code from `Multiselect`. If there is time, I think it would b worth putting this into a util file and just import it twice

packages/components/multiselect/src/SelectAllOption.tsx
```diff
@@ -0,0 +1,91 @@
...
+  );
+
+  // checks if all children are true
+  React.useEffect(() => {
+    const allTrue = areChildrenAllSelected(
+      childNodes,
+      (child) => child.type === MultiselectOption,
+    );
+    setAllSelected(allTrue);
+  }, [childNodes]);
```
Chaoste, line 66: I might be missing a obstacle here but I would try to avoid these kinds of state updates. You listen to a (transformed version) of an input variable and then set a state variable. I think that a `UseMemo` would be more appropriate since it saves you one render cycle. Here you first have one cycle with the new prop and then another one to update the state.

Chaoste: If I'm not mistaken than this effet will get called when you change the `checked` state for a checkbox. This means that you don't need the state variable at all and can also save the call at line 73. By just updating the DOM nodes, this component rerenders and updates the memorized variable `allSelected` (instead of a controlled state)

Chaoste (commented):
I think this is a nice feature that saves us from writing the same logic multiple times. Thanks for implementing :)

packages/components/multiselect/src/SelectAllOption.tsx
Chaoste: I might be wrong here because Multiselect is controlled but just wanted to raise the possibility of reducing complexity. Dunno if it's possible.

Lelith: i will have another look at it, but without the effect, it has some weird double-clicking required issues.

packages/components/multiselect/src/SelectAllOption.tsx
Lelith: It is very similar to the "iterateOverChildren" functionality which we do to count them. I think i can find a abstraction so we can put that function into a utils file and probaply also merge it with the enriching function. All of them are recursively checking the child components and then do different things with the found options.

Chaoste: True, if the generalization is too complex, we can easily skip this. No need to write super intelligent utilities for the sake of DRY :)

packages/components/multiselect/src/SelectAllOption.tsx
Chaoste: I think it's also not a problem to have it this way. It's rather a micro-optimization that doesn't need to block this PR :)

Lelith: So i took another go at it and i don't need to call the method when initializing, its enough with the effect. In this case childNodes are props that change, not another state variable or anything inside. So useEffect is the correct hook over useCallback or useMemo.

packages/components/multiselect/src/SelectAllOption.tsx
Lelith: the only thing they have in common is that they work recursively over the children. which is neccesary because we don t have power over the child components in this component. which gives us some advantags as well as some problems.
maybe this would be the point to question if it is better to have the children as compound components or if we should take the render of the children in our own hands. but that leads then to the same limitations that the autocomplete has.

i think there is no great answer, but i also dislike the high complexity of the solution.

denkristoffer:
I agree with @bgutsol that this is a lot to go against the component's implementation. Is the problem here that the MultiSelect is controlled only? This seems to implement a sort of in-between, where we have an uncontrolled part of a controlled component. Am I misunderstanding that?

Lelith:
I totally get that this becomes overly complex, and shuts down a part of logic from the user. I am not happy with it either.

The main issue when implementing outside of the component is that you also have to recognise somehow when all options are already selected. For our story example that is quite simple done when you compare the list of available options with the list of selected options.

But when we go in the actual use case, there is most likely a form controller there, too. that has initially selected items e.g. or items that are not longer part of the available options, but are still selected, coming from the saved form.
So you have to then write yourself a function that is identifying which options are there, and if they are selected, and if the current selection state reflects one of "all elements selected"

Also setting a selection, in the real world, is not only done by adding it to a state, that is only a shallow copy of the available options. For having it selected in a controlled form, you need to trigger a real onChange handler, with a real event. So you still need to loop trough all options and call an onChange event for them.

You also need to implement it for every single multi-select which shall have the shortcut.

Having there the relation of children is kind of simpler, even tough with the possibility of having nested children, means we have to loop trough them recursively. 

But i will provide an outside implementation example and then we can see which one is suiting more, and maybe i can put the overly complex one in experience components?

Lelith:
@denkristoffer @bgutsol i refactored it and removed the complex logic, to only have a compund component and let the user decide themselve how they want to deal with toggling and detecting and updated storybook and the example accordingly.

I think ultimately delivering the lightweight solution for the majority of users will be better. And if i find myself fighting in a corner too much in the apps, i still can create a "with Select All" HOC

packages/components/multiselect/src/Multiselect.tsx
```diff
@@ -60,6 +60,12 @@ export interface MultiselectProps extends CommonProps {
    */
   searchInputName?: string;
 
+  /**
+   * Enables a shortcut checkbox to select or delect all items
+   * @default false
+   */
+  hasCheckAll?: boolean;
```
bgutsol, line 67: don't need this anymore

packages/components/multiselect/src/Multiselect.tsx
```diff
@@ -136,7 +142,7 @@ const countMatchingChildren = (
 
 function _Multiselect(props: MultiselectProps, ref: React.Ref<HTMLDivElement>) {
   const {
-    className,
+    className = '',
```
bgutsol, line 145: Do we really need it?

packages/components/multiselect/src/SelectAllOption.tsx
```diff
@@ -0,0 +1,31 @@
...
+  ...otherProps
+}: SelectAllOptionProps) => {
+  const styles = getMultiselectStyles();
+  const displayLabel = label || isChecked ? 'Deselect all' : 'Select all';
+  return (
+    <MultiselectOption
+      value="all"
+      label={displayLabel}
+      itemId="SelectAll"
+      onSelectItem={(event) => onSelectItem(event)}
```
bgutsol, line 25: Can we just pass a callback directly?
```suggestion
      onSelectItem={onSelectItem}
```

packages/components/multiselect/stories/Multiselect.stories.tsx
```diff
@@ -95,21 +95,20 @@ export const Basic = () => {
...
-
     const currentFruit = fruits.find((fruit) => fruit.value === value);
     if (checked) {
       setSelectedFruits((prevState) => [...prevState, currentFruit.name]);
     } else {
-      const newSelectedFruits = selectedFruits.filter(
-        (fruit) => fruit !== currentFruit.name,
-      );
-      setSelectedFruits(newSelectedFruits);
+      //its important to use prevState to avoid race conditions when using the state variable as reference.
```
bgutsol, line 102: nice 👍

bgutsol (approved):
Thanks for addressing our comments ❤️
only some small nits

packages/components/multiselect/src/Multiselect.test.tsx
```diff
@@ -249,3 +256,22 @@ describe('Multiselect with search', () => {
...
+      { id: 3, name: 'Avocado 🥑', isChecked: true, isDisabled: false },
+    ];
+    const [{ user }] = renderComponent({}, produce, true);
+    await user.click(
+      screen.getByRole('button', { name: 'Toggle Multiselect' }),
+    );
+    expect(
+      screen.queryByRole('checkbox', { name: 'Deselect all' }),
+    ).toBeInTheDocument();
+    await user.click(screen.queryByRole('checkbox', { name: 'Deselect all' }));
```
Chaoste, line 274:
```suggestion
    expect(
      screen.getByRole('checkbox', { name: 'Deselect all' }),
    ).toBeInTheDocument();
    await user.click(screen.getByRole('checkbox', { name: 'Deselect all' }));
```

packages/components/multiselect/stories/Multiselect.stories.tsx
Chaoste: That's definitely a good catch!

packages/components/multiselect/src/Multiselect.tsx
Lelith: oh that was just an experiment i forgot to remove again. thanks

Chaoste (approved):

PR #2375, opened 2023-01-13
feat(copy-button): async values

denkristoffer:
# Purpose of PR

- clean up CopyButton code so that we can more easily pass it values from asynchronous calls. See the example for an idea: https://forma-36-git-feat-copy-async-value.colorfuldemo.com/components/copy-button#using-with-promises
- Disable jest's global fake timers
  - I can't get fake timers to work with the tests written for this copy button. This PR disables them globally and uses them explicitly where it makes sense.
- Upgrades `turbo` package to `^1.7.0`. I had to do this to bust the cache and get the install to work properly on CircleCI.

There are slight changes to the styles since I switched to using our Button component as a base. It looks more like it does in Figma this way too. Please review @fabe 🙂 

## Learning

The [new example](https://github.com/contentful/forma-36/pull/2375/files#diff-5e54ca1b04c9eb591af07a11b8aef471e9149259bba3737bdfc10bc468878d60) didn't work until I changed the syntax of `loadValue` from `const loadValue = async () => { ... }` to `async function loadValue() { .... }`. It was complaining about the use of `await` being outside of an `async` function. It's probably related to the live render package. Just as a note if we ever have to do something similar again.

### DIFF
--- a/packages/components/copybutton/package.json
+++ b/packages/components/copybutton/package.json
@@ -6,12 +6,12 @@
     "build": "tsup"
   },
   "dependencies": {
+    "@contentful/f36-button": "^4.24.1",
     "@contentful/f36-core": "^4.24.1",
     "@contentful/f36-icons": "^4.23.2",
     "@contentful/f36-tokens": "^4.0.1",
     "@contentful/f36-tooltip": "^4.24.1",
-    "emotion": "^10.0.17",
-    "react-copy-to-clipboard": "^5.1.0"
+    "emotion": "^10.0.17"
   },
   "peerDependencies": {
     "react": ">=16.8",
--- a/packages/components/copybutton/src/CopyButton.tsx
+++ b/packages/components/copybutton/src/CopyButton.tsx
@@ -12,10 +12,7 @@ import { Button, type ButtonProps } from '@contentful/f36-button';
 import { getCopyButtonStyles } from './CopyButton.styles';
 
 function isPromiseLike<T>(x: T | PromiseLike<T>): x is PromiseLike<T> {
-  return (
-    (typeof x === 'object' || typeof x === 'function') &&
-    typeof (x as PromiseLike<T>).then === 'function'
-  );
+  return typeof (x as PromiseLike<T>).then === 'function';
 }
 
 type StringValue = {
--- a/packages/components/copybutton/src/CopyButton.tsx
+++ b/packages/components/copybutton/src/CopyButton.tsx
@@ -1,12 +1,13 @@
 import React, {
-  useState,
   useCallback,
-  useRef,
-  type MouseEventHandler,
   useEffect,
+  useRef,
+  useState,
+  type MouseEventHandler,
+  type FocusEventHandler,
 } from 'react';
 import { CopyIcon } from '@contentful/f36-icons';
-import type { CommonProps, ExpandProps } from '@contentful/f36-core';
+import type { ExpandProps } from '@contentful/f36-core';
 import { Tooltip, type TooltipProps } from '@contentful/f36-tooltip';
 import { Button, type ButtonProps } from '@contentful/f36-button';
 import { getCopyButtonStyles } from './CopyButton.styles';
@@ -15,30 +16,10 @@ function isPromiseLike<T>(x: T | PromiseLike<T>): x is PromiseLike<T> {
   return typeof (x as PromiseLike<T>).then === 'function';
 }
 
-type StringValue = {
-  preload?: never;
-  /**
-   * Value that will be copied to clipboard when the button is clicked
-   */
-  value: string;
-};
-
-type PromiseValue = {
-  /**
-   * Preload the value so that it can be copied as soon as possible
-   */
-  preload?: boolean;
-  /**
-   * Value that will be copied to clipboard when the button is clicked
-   */
-  value: () => Promise<string>;
-};
-
-export type CopyButtonProps = CommonProps & {
-  /**
-   * Props that are passed to the button component
-   */
-  buttonProps?: Omit<ButtonProps, 'size'>;
+export type CopyButtonProps = Omit<
+  ButtonProps,
+  'children' | 'endIcon' | 'startIcon' | 'isDisabed' | 'size'
+> & {
   /**
    * Function that gets called when the button is clicked
    */
@@ -59,7 +40,7 @@ export type CopyButtonProps = CommonProps & {
   tooltipProps?: Omit<TooltipProps, 'content' | 'children'>;
   /**
    * Label to be used on aria-label for the button
-   * @default Copy {value} to clipboard
+   * @default Copy to clipboard
    */
   label?: string;
   /**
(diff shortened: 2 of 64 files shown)


packages/components/copybutton/src/CopyButton.tsx
```diff
@@ -44,76 +75,106 @@ export interface CopyButtonProps extends CommonProps {
...
+        <Button
+          aria-label={
+            label ??
+            `Copy ${
+              isPromiseLike(resolvedValue.current)
+                ? ''
+                : `"${resolvedValue.current}" `
+            }to clipboard`
+          }
+          aria-live="assertive"
```
mshaaban0, line 162: Voice over isn't working properly. I think we also need to tell it something like (Value "..." has been copied to clipboard) once copied. Currently it just repeats the aria-label value

Screenshot 2023-01-23 at 12 48 33

packages/components/menu/src/Menu.test.tsx
```diff
@@ -80,6 +80,7 @@ describe('Menu', function () {
   });
 
   it('do call onClose when selecting list item', async () => {
+    jest.useFakeTimers();
```
mshaaban0, line 83: wouldn't `beforeEach` work here?

packages/components/tooltip/src/Tooltip.test.tsx
```diff
@@ -110,11 +131,12 @@ describe('Tooltip', () => {
     const tooltip = screen.getByRole('tooltip');
     expect(tooltip.getAttribute('id')).toBe('Tooltip');
     expect(tooltip.nodeName).toMatch(/span/i);
+
+    jest.runOnlyPendingTimers();
```
mshaaban0, line 135: same here `beforeEach` and `afterEach`?

packages/components/copybutton/src/CopyButton.tsx
```diff
@@ -44,76 +75,106 @@ export interface CopyButtonProps extends CommonProps {
...
+  const styles = getCopyButtonStyles({ size });
+  const [isLoading, setIsLoading] = useState(false);
   const [copied, setCopied] = useState(false);
   const button = useRef<HTMLButtonElement | null>(null);
-  const handleOnCopy = useCallback(
-    (e: string) => {
-      if (onCopy) {
-        onCopy(e);
+  const resolvedValue = useRef<string | Promise<string>>(
+    typeof value === 'function' ? value() : value,
```
mshaaban0, line 103: It's kinda tricky for me to logic about this one. 
Why do we need to call `value` when we're resolving it within the `preload` useEffect?
Do you think it makes sense to just preload all the time and not make it optional?

packages/components/menu/src/Menu.test.tsx
denkristoffer: We only need this for tests using user-event. Additionally, if we put it in `beforeEach` it will also be used with the axe tests, and there's an issue with `axe-core` where it doesn't work with jest's fake timers: https://github.com/dequelabs/axe-core/issues/3055

packages/components/copybutton/src/CopyButton.tsx
denkristoffer: Personally I was thinking about not calling the value here and only doing it with the effect for full control, but this behaviour replicates the API from the web app. I'm not sure which I prefer, what if it's a really expensive API call, would we want to trigger that just on render? The button might never be clicked 🤔

packages/components/copybutton/src/CopyButton.tsx
denkristoffer: Should we move it to the tooltip instead? I think that makes more sense.

bgutsol (commented):
The main comment is basically about what we discussed in DMs, that we probably can replace all the 
```
jest.useFakeTimers();
userEvent.setup({ advanceTimers: jest.advanceTimersByTime });
```
with just `userEvent.setup();` for component that don't have any timers logic.

packages/components/copybutton/src/CopyButton.tsx
```diff
@@ -44,76 +75,106 @@ export interface CopyButtonProps extends CommonProps {
...
+    load();
+  }, [preload, value]);
+
+  const handleClick = useCallback<
+    MouseEventHandler<HTMLButtonElement>
+  >(async () => {
+    if (isPromiseLike(resolvedValue.current)) {
+      setIsLoading(true);
+      resolvedValue.current = await resolvedValue.current;
+      setIsLoading(false);
```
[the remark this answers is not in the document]
denkristoffer: I don't think so. Would you want a button to insert things into your clipboard if you didn't click it? By the way, for the security reasons I don't even think it's possible to write to the clipboard outside of user event handlers.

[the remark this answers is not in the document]
mshaaban0: I wouldn't no, but I mean without preload, after the user clicks it should resolve the value and show some sort of success message or error if the promise fails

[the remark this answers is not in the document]
denkristoffer: I'm going with Bohdan's comments and removing the preload completely for now. Instead we can preload the value in whatever is rendering the button.

packages/components/copybutton/src/CopyButton.tsx
denkristoffer: I decided against that since it's not accessible either. What do you think about the new way I'm doing it? Only problem I'm seeing is that once you copy something it repeats itself and says "Value copied" twice 😅

packages/components/copybutton/src/CopyButton.tsx
```diff
@@ -44,76 +75,106 @@ export interface CopyButtonProps extends CommonProps {
...
-            type="button"
-            ref={button}
-            className={cx(styles.copyButton, {
-              [styles.copyButtonDisabled]: isDisabled,
-            })}
-            aria-label={`Copy ${value} to clipboard`}
-            disabled={isDisabled}
-            aria-live="assertive"
-          >
+    <div ref={ref} data-test-id={testId} className={className} {...otherProps}>
```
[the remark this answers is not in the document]
denkristoffer: I don't remember, but I think we initially wanted to wrap most components in a root so that the `ref` passed was consistent. I've switched it to the button.

bgutsol (approved):
Looks good! Thanks for the changes 🏅

packages/components/copybutton/package.json
```diff
@@ -6,12 +6,12 @@
     "build": "tsup"
   },
   "dependencies": {
+    "@contentful/f36-button": "^4.24.1",
     "@contentful/f36-core": "^4.24.1",
     "@contentful/f36-icons": "^4.23.2",
     "@contentful/f36-tokens": "^4.0.1",
     "@contentful/f36-tooltip": "^4.24.1",
-    "emotion": "^10.0.17",
-    "react-copy-to-clipboard": "^5.1.0"
```
mshaaban0, line 14: 🎉

mshaaban0 (approved):
Looks good 👍🏼

PR #2381, opened 2023-01-20
feat(timepicker): migration release

denkristoffer:
# Don't merge this!

## What is it?
We need a timepicker component using Forma 36 v4 if we want to get completely rid of v3 in our web app. This is a stopgap so that we have a component we can use until we extract and make a proper time picker. Is it great? No. Does it work? I think so.

This thing is HARD to test since all the pipelines are going to fail. Please help me.

There's a static storybook preview at https://www.chromatic.com/component?appId=5fd1dda724cc620021ace8c5&csfId=components-timepicker&buildNumber=4023&k=63ce7a41d66c7f57c931d8ad-1200-interactive-true&h=2&b=-1

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -99,7 +99,7 @@ jobs:
           command: yarn storybook:build
       - run:
           name: Chromatic
-          command: yarn chromatic
+          command: yarn chromatic --project-token=${CHROMATIC_PROJECT_TOKEN} --exit-zero-on-changes --storybook-build-dir ./dist-storybook
 
   release:
     docker:
--- a/packages/components/timepicker/package.json
+++ b/packages/components/timepicker/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@contentful/forma-36-react-timepicker",
-  "version": "0.5.0-alpha.101",
+  "version": "1.0.0",
   "license": "MIT",
   "files": [
     "dist",
@@ -23,11 +23,10 @@
     "build": "parcel build index.ts"
   },
   "peerDependencies": {
-    "react": ">=16.8"
+    "react": ">=16.8",
+    "react-dom": ">=16.8"
   },
   "devDependencies": {
-    "@contentful/forma-36-react-components": "^3.73.2",
-    "@contentful/forma-36-tokens": "^0.11.2",
     "@testing-library/jest-dom": "5.15.0",
     "@testing-library/react": "12.1.2",
     "@types/is-hotkey": "^0.1.1",
@@ -36,8 +35,9 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.6.2",
-    "@contentful/forma-36-react-components": "^3.100.7",
-    "date-fns": "^2.9.0",
+    "@contentful/f36-components": "^4.24.1",
+    "@contentful/f36-tokens": "^4.0.1",
+    "date-fns": "^2.29.3",
     "emotion": "^10.0.17",
     "is-hotkey": "^0.1.6",
     "lodash.orderby": "^4.6.0"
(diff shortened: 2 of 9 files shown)


packages/components/timepicker/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@contentful/forma-36-react-timepicker",
```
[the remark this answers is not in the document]
denkristoffer: I'm thinking we might want to save the name for a proper timepicker at some point but I don't know. What do you think?

packages/components/timepicker/src/Timepicker.test.tsx
```diff
@@ -58,56 +53,54 @@ describe('TimePicker', () => {
...
+      render(
+        <Timepicker
+          date={format(addYears(new Date(), 1), 'yyyy/MM/dd')}
+          onChange={onChange}
+          value="12:00"
+        />,
+      );
+
+      const timepicker = screen.getByTestId('time');
+      userEvent.type(timepicker, '{arrowdown}');
```
mshaaban0, line 100: I think this works, but wondering if you tried `userEvent.keyboard(...)` and if it worked?

.circleci/config.yml
```diff
@@ -99,7 +99,7 @@ jobs:
           command: yarn storybook:build
       - run:
           name: Chromatic
-          command: yarn chromatic
+          command: yarn chromatic --project-token=${CHROMATIC_PROJECT_TOKEN} --exit-zero-on-changes --storybook-build-dir ./dist-storybook
```
mshaaban0, line 102: Something is odd about the storybook of this PR, it seems like it's for V3 for some reason https://5fd1dda724cc620021ace8c5-xqpqgpuszm.chromatic.com/?path=/story/documentation-general--page

mshaaban0: Also vercel is failing because it can't find `website` package 🤔

mshaaban0: Ah ok, why do you want to merge this to v3 branch?

packages/components/timepicker/src/Timepicker.test.tsx
denkristoffer: Yeah I did it, it works, but `userEvent.type()` is easier here since you don't have to first use `userEvent.click()` to focus the input 😄

.circleci/config.yml
denkristoffer: I don't actually want to merge, just have this branch to release from. The package was deleted from v4.

packages/components/timepicker/package.json
```diff
@@ -23,23 +23,20 @@
...
-    "@types/is-hotkey": "^0.1.1",
-    "@types/jest": "25.2.3",
-    "@types/lodash.orderby": "^4.6.6"
+    "@types/jest": "25.2.3"
   },
   "dependencies": {
     "@babel/runtime": "^7.6.2",
-    "@contentful/forma-36-react-components": "^3.100.7",
-    "date-fns": "^2.9.0",
+    "@contentful/f36-components": "^4.24.1",
```
bgutsol, line 36: it's like "back to the future" component from V3 uses components from V4 😄

bgutsol (approved):
LGTM

packages/components/timepicker/package.json
mshaaban0: Doesn't that mean when the user interface uses a different version it will include both?
Why don't we move all of this to a new package on `main` v4 and mark it as `v1.0.0` or even `v4.0.0-alpha`

denkristoffer:
Released as `@contentful/forma-36-react-timepicker@1.0.0`: https://www.npmjs.com/package/@contentful/forma-36-react-timepicker

PR #2382, opened 2023-01-23
fix(Readme.mdx): fixed readme doc with correct import directory for custom icons

Lakesxo:
# Purpose of PR
- I realized that custom icons can not be imported from "@contentful/f36-components" rather they are imported from "@contentful/f36-icons" but the docs says otherwise. So I have decided to kindly point this out and improve the page with the right import directory ❤️

## PR Checklist

- [x] Tests are passing


denkristoffer:
👋 Hi @Lakesxo, thanks for the contribution! I see your point here, I had to go back and read it again myself to understand what is going on. The code import is actually the right one, but the comment above it is not explaining what we mean by "custom icons", which leads to confusion.

What the second import tries to tell us is how to render totally custom icons like something from a third-party library https://f36.contentful.com/components/icon#custom-icons-and-third-party-libraries – do you have a suggestion how we can phrase this better? Do you find `// or, for third-party icons` clearer?

Lakesxo:
> 👋 Hi @Lakesxo, thanks for the contribution! I see your point here, I had to go back and read it again myself to understand what is going on. The code import is actually the right one, but the comment above it is not explaining what we mean by "custom icons", which leads to confusion.
> 
> What the second import tries to tell us is how to render totally custom icons like something from a third-party library https://f36.contentful.com/components/icon#custom-icons-and-third-party-libraries – do you have a suggestion how we can phrase this better? Do you find `// or, for third-party icons` clearer?

Oh okay. Yeah i think the third-party icons phrase sounds better. 
But when i tried to import a forma custom icon e.g PlusIcon from "@contentful/f36-components" it didn't work until i imported from "@contentful/f36-icons";

denkristoffer:
@Lakesxo No you're right, that's the way to do it. Maybe we should change the first line to make it more clear too 🤔

Lakesxo:
@denkristoffer Yeah i think you should. Maybe something like:

// Import as an object :
`import * as icons from '@contentful/f36-icons';`
// Import single component as :
`import { PlusIcon, DoneIcon } from "@contentful/f36-icons";`
// Third party custom icons as :
`import { Icon } from '@contentful/f36-components';`


**Access import as an object:**
```
function IconExample() {
     return (
          <>
              {icons.PlusIcon}
              {icons.DoneIcon}
          </>
     )
}
```
OR

**Access import as a single component:**
```
function IconExample() {
     return (
          <>
              <PlusIcon />
              <DoneIcon />
          </>
     )
}
```

denkristoffer:
@Lakesxo That looks good to me 😄 Would you like to update the PR with that?

Lakesxo:
> @Lakesxo That looks good to me 😄 Would you like to update the PR with that?

Sure, will do

denkristoffer (approved):

denkristoffer:
Thanks for the contribution @Lakesxo! I'm sorry it took so long to get merged 🙇 Thanks again!

PR #2386, opened 2023-01-24
fix(copy-button): omit native HTML onCopy prop

denkristoffer:
# Purpose of PR

Omit native `onCopy` prop from CopyButton

### DIFF
--- a/packages/components/copybutton/src/CopyButton.tsx
+++ b/packages/components/copybutton/src/CopyButton.tsx
@@ -13,7 +13,7 @@ import { cx } from 'emotion';
 
 export type CopyButtonProps = Omit<
   ButtonProps,
-  'children' | 'endIcon' | 'startIcon' | 'isDisabed' | 'size'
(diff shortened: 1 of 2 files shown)


packages/components/copybutton/src/CopyButton.tsx
```diff
@@ -13,7 +13,7 @@ import { cx } from 'emotion';
 
 export type CopyButtonProps = Omit<
   ButtonProps,
-  'children' | 'endIcon' | 'startIcon' | 'isDisabed' | 'size'
+  'children' | 'endIcon' | 'onCopy' | 'startIcon' | 'isDisabed' | 'size'
```
mshaaban0, line 16: Seems like a valid event, 🤔 why removing it?

denkristoffer: The CopyButton is overwriting it with it's own `onCopy` that we're using.

mshaaban0 (approved):

PR #2388, opened 2023-01-24
chore(copy-button): omit more Button props

denkristoffer:
# Purpose of PR

Omit more props that should not be used on the CopyButton.

### DIFF
--- a/packages/components/copybutton/src/CopyButton.tsx
+++ b/packages/components/copybutton/src/CopyButton.tsx
@@ -18,7 +18,7 @@ export type CopyButtonProps = Omit<
   | 'onCopy'
   | 'onClick'
   | 'startIcon'
-  | 'isDisabed'
+  | 'isDisabled'
   | 'size'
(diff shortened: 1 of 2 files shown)


packages/components/copybutton/src/CopyButton.tsx
```diff
@@ -13,7 +13,15 @@ import { cx } from 'emotion';
 
 export type CopyButtonProps = Omit<
   ButtonProps,
-  'children' | 'endIcon' | 'onCopy' | 'startIcon' | 'isDisabed' | 'size'
+  | 'children'
+  | 'endIcon'
+  | 'onCopy'
+  | 'onClick'
+  | 'startIcon'
+  | 'isDisabed'
```
massao, line 21: typo? `isDisabled`

Lelith (approved):
can you double check the `isDisabled` vs `isDisabed`typo? otherwise looks good to me.

PR #2412, opened 2023-03-03
feat: Caption component

denkristoffer:
# Purpose of PR

Add a Caption typography component.

### DIFF
--- a/packages/components/typography/src/Caption/Caption.tsx
+++ b/packages/components/typography/src/Caption/Caption.tsx
@@ -1,0 +1,62 @@
+import React from 'react';
+import {
+  FontSizeTokens,
+  FontWeightTokens,
+  LineHeightTokens,
+} from '@contentful/f36-tokens';
+import type {
+  CommonProps,
+  MarginProps,
+  PolymorphicComponent,
+  PolymorphicProps,
+  ExpandProps,
+} from '@contentful/f36-core';
+import { Text } from '../Text';
+
+const CAPTION_DEFAULT_TAG = 'span';
+
+export interface CaptionInternalProps extends CommonProps, MarginProps {
+  children?: React.ReactNode;
+  fontWeight?: Extract<
+    FontWeightTokens,
+    'fontWeightNormal' | 'fontWeightMedium'
+  >;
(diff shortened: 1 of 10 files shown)


fabe (approved):

packages/components/typography/src/Caption/Caption.tsx
```diff
@@ -0,0 +1,58 @@
...
+import type {
+  CommonProps,
+  MarginProps,
+  PolymorphicComponent,
+  PolymorphicProps,
+  ExpandProps,
+} from '@contentful/f36-core';
+import { Text } from '../Text';
+
+const CAPTION_DEFAULT_TAG = 'span';
```
mshaaban0, line 12: > `Caption` is used to display secondary and tertiary information

if it's mainly tend to live standalone of sentences and paragraphs, I'd recommend `strong` for better semantics

denkristoffer: Why is that? `strong` is defined like this by Mozilla: "The `<strong>` HTML element indicates that its contents have strong importance, seriousness, or urgency". We want to use `Caption` to de-emphasise information.

massao (approved):
LGTM!
Just not sure if Caption is the best name for the component, since it's only a styling of text, and can be confused with the actual caption semantics, like the [caption](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/caption) or [figcaption](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figcaption) elements 🤔

packages/components/typography/src/Caption/Caption.tsx
mshaaban0: Would you say `<small />` fits the purpose? I just think span is meant to wrap inline text within text and not to display a standalone info

denkristoffer: I don't think `small` has semantic meaning either, does it? To be honest I think `span` is perfect here because it has no semantic meaning, which I don't think we should add. Visually de-emphasising with just CSS makes sense to me.

mshaaban0 (approved):

PR #2415, opened 2023-03-06
feat(badge): update Badge and EntityStatusBadge styling

massao:
# Purpose of PR

Update styling for Badge components:
* Use `Caption` component instead of span element;
* Cleanup style to remove unnecessary values;
* Update `EntityStatusBadge` to have Title Case, to have correct design without changes on API.

### DIFF
--- a/packages/components/badge/src/Badge/Badge.styles.ts
+++ b/packages/components/badge/src/Badge/Badge.styles.ts
@@ -49,17 +49,15 @@ const sizeToStyles = ({ size }: { size: BadgeSize }): CSSObject => {
   switch (size) {
     case 'small':
       return {
-        padding: `3px ${tokens.spacing2Xs}`,
-        fontSize: '0.625rem',
-        lineHeight: '0.625rem',
-        maxHeight: '16px',
+        padding: `0 ${tokens.spacing2Xs}`,
+        lineHeight: tokens.lineHeightS,
+        maxHeight: tokens.lineHeightS,
       };
     default:
       return {
         padding: `0 ${tokens.spacingXs}`,
-        fontSize: `calc(1rem * (12 / ${tokens.fontBaseDefault}))`,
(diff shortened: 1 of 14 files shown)


denkristoffer:
Would you update the labels used in the examples as well? https://forma-36-git-feat-cfcx-218.colorfuldemo.com/components/badge

packages/components/badge/src/Badge/Badge.styles.ts
```diff
@@ -83,9 +76,10 @@ export const getBadgeStyles = () => ({
...
-  badgeText: css({
-    whiteSpace: 'nowrap',
-    overflow: 'hidden',
-    textOverflow: 'ellipsis',
-  }),
+  badgeText: ({ size }: { size: BadgeSize }) =>
+    css({
+      color: 'currentcolor',
+      lineHeight: 'inherit',
+      fontSize: size === 'small' && '0.688rem',
```
denkristoffer, line 83: Is this magic number cleared with @fabe? I would prefer not to use it if we can.

denkristoffer (commented):
There are some leftovers from the Caption PR, I think this needs to be rebased.

packages/components/badge/src/Badge/Badge.styles.ts
massao: This is because for the small variant we need 11px for font-size, and we don't have that on the tokens, we could create a token for it, but not sure if it's used in other places, @fabe will we use it in more places? Or should we fallback to the 12px token?

fabe: Let's fall back to the 12px token then. I think it doesn't change the height of the element, so we can go for that!

denkristoffer (approved):
All right let's try to get Chromatic and Linux fixed separately 👍

mshaaban0 (approved):

PR #2417, opened 2023-03-06
feat(menu): new look and feel (#2417

denkristoffer:
# Purpose of PR

- New look and feel for `Menu`.
- Adds `isActive` prop to `Menu.Item`


packages/components/menu/src/MenuItem/MenuItem.styles.ts
```diff
@@ -1,44 +1,54 @@
...
+      },
 
       '&:hover': {
         backgroundColor: tokens.gray100,
       },
       '&:focus': {
         boxShadow: `inset ${tokens.glowPrimary}`,
         // just to make boxShadow with rounded corners
-        borderRadius: tokens.borderRadiusMedium,
+        borderRadius: tokens.borderRadiusSmall,
```
mshaaban0, line 43: Any reason for going with small border radius here?
I think it looks odd that it doesn't match the border radius of the element.

Screenshot 2023-03-09 at 14 15 10

denkristoffer: I will defer to @fabe, maybe I misunderstood the comment in Figma 🙂

fabe: Yeah, we need the small one here to get to the [correct nested border radii](https://www.30secondsofcode.org/articles/s/css-nested-border-radius). So outside is 8px, inside 4px. We could also use our regular border radius (6px), but then the container would need 10px. I think we don't have a token for that?

packages/components/menu/src/MenuItem/MenuItem.styles.ts
```diff
@@ -1,44 +1,54 @@
...
       },
       '&:focus:not(:focus-visible)': {
         boxShadow: 'unset',
         borderRadius: 'unset',
```
mshaaban0, line 47: I think we need to remove this here because it's messing with the active state 
Screenshot 2023-03-09 at 18 15 55

fabe (commented):
Thanks, just got a couple points!

1. Vertical padding on the item should be 6px, this is to get to 32px in height (as opposed to 28px right now)
2. Dividers should be gray-200 instead of 300.
3. Some of the items don't have any border radius, mostly the first item but also sometimes the lower ones. Is that the :focus problem?
4. Could we get rid of the hover state if the item is disabled?
5. Looks like we set a custom color to the scrollbar, could we try gray-200 there instead of 400?
6. The container of the scrollbar is overflowing the container (see corners).
7. We can remove the margin-right on the menu items if there's a scrollbar, to even out the spacing between the item, scrollbar, and container.
8. Could we remove the hover state when a menu item `isActive`? Background can stay the same when active

Let me know if anything is unclear!

denkristoffer:
@fabe

> Looks like we set a custom color to the scrollbar, could we try gray-200 there instead of 400?

There shouldn't be anything touching the scrollbars 🤔 Can you share screenshots of the problems you're seeing with the scrollbars?

> Could we remove the hover state when a menu item isActive? Background can stay the same when active

I think this differs from the space switcher then, should we make it the same there?

fabe:
> There shouldn't be anything touching the scrollbars 🤔 Can you share screenshots of the problems you're seeing with the scrollbars?

The scollbar here is exactly our gray-400 value, ideally that should be a lighter gray or a 20/30% transparent black.

image

> I think this differs from the space switcher then, should we make it the same there?

On environments, right? Yep, I'd say that should be the same in the space switcher.

denkristoffer:
This is what it looks like for me:
Screenshot 2023-03-10 at 14 05 23

packages/components/menu/src/MenuItem/MenuItem.styles.ts
denkristoffer: Removed. I wonder why it breaks it though 🤔

mshaaban0 (approved):
Looking fresh 💅

denkristoffer:
@fabe want to have another look?

fabe:
Looks good! Last thing I saw is about the spacing around the dividers. There's not enough above, and too much below: 

image

PR #2418, opened 2023-03-07
feat: update Table

bgutsol:
# Purpose of PR

- Refactor Table components structure to be aligned with our common pattern.
- Adjust TableCell styles.


massao (approved):
LGTM

bgutsol:
> 1. Let's always add a border-bottom for the last row (`embedded` layout),
> 2. In the `inline` layout, could we use 12px (spacing-s) as vertical padding for cols? It's 8px right now. For the embedded layout, 8px is fine.
> 3. Could we set `vertical-align: middle` already in this PR, or a follow up?

@fabe thanks for the review!

1. done ✅
2. I guess you meant horizontal padding, right? I've changed it for both embedded and inline layouts, since it doesn't require some complex logic to create (table context to pass props). But if it's critical to have different padding in these two layouts - let me know, we can do it.
3. done ✅

denkristoffer:
@fabe @bgutsol I'm not sure if we should default to `vertical-align: middle` in all our tables, look at a case like our own props table: https://forma-36-git-cfcx-220.colorfuldemo.com/components/table#table

Also @bgutsol, it looks like the props for the compound table components don't work anymore 🤔

fabe:
> I guess you meant horizontal padding, right?

Yes, looks good now!

> @fabe @bgutsol I'm not sure if we should default to `vertical-align: middle` in all our tables, look at a case with multiple lines of content like our own props table: https://forma-36-git-cfcx-220.colorfuldemo.com/components/table#table – compare the `css` row to https://f36.contentful.com/components/table#table

Good point, should we add a prop for this? We can keep the current behavior to not break anything, but we'll likely move into having more tables that are vertically centered.

denkristoffer:
> Good point, should we add a prop for this? We can keep the current behavior to not break anything, but we'll likely move into having more tables that are vertically centered.

Yeah let's do a `verticalAlignment` prop 👍

denkristoffer (approved):

Issue #1322, opened 2021-09-17
🐛 Bug - 


arthurbfcarvalho:
# Forma 36 bug report

## Actual results

Here are screenshots:

**Mac:**
![image](https://user-images.githubusercontent.com/15033988/133813577-b1e425ea-4bfb-4b8d-b62e-a530560fe3bf.png)

**Linux:** 
![image](https://user-images.githubusercontent.com/15033988/133813762-db8abd56-c29f-419a-bb9c-6068ea36934d.png)

suevalov:
@arthurbfcarvalho could you please check if it's any better with v4 release?

denkristoffer:
Current status from Ubuntu
![image](https://user-images.githubusercontent.com/219017/165052508-cf4f511f-05f2-4afb-a566-2479e16209c0.png)

denkristoffer:
So far, investigating this has led to the theory that some Linux distributions return a typeface where the weight we're trying to use simply doesn't exist, and it then falls back to the nearest existing weight. In the case of the screenshot the `400` seems to be rendered with `500`.

PR #2419, opened 2023-03-09
feat(font-stack): add system-ui

denkristoffer:
# Purpose of PR

Chrome on Linux doesn't seem to render our medium font weight properly. This seems to be caused by issues with the browser on that plaform: 
- https://bugs.launchpad.net/ubuntu/+source/fonts-ubuntu/+bug/1512111
- https://stackoverflow.com/questions/53027044/why-is-chrome-on-linux-displaying-the-wrong-font-weight
- https://stackoverflow.com/questions/22001035/font-weight-ignored-in-chrome

This is the same bug as #1322

This PR adds `system-ui` to our font stack to see if that helps.


denkristoffer:
Doesn't seem to fix the issue :(

PR #2420, opened 2023-03-09
fix(docs): remove flex styles from live preview wrapper

mshaaban0:
# Purpose of PR

Reason: [Skeleton](https://f36.contentful.com/components/skeleton#__next) isn't working on Safari

Currently we're using display flex on live preview wrapper to center all examples in the middle. This however sometimes breaks the default behaviour of components e.g. Skeletons works well when displayed in `div: block` but when within flex component the svg needs a fixed height.

IMO, we should display examples as it would be rendered in a normal div as a wrapper.

The cost here is that live examples won't be aligned to the center, which I believe is fine and even looks better to my eyes 

## PR Checklist

### DIFF
--- a/packages/website/components/LiveEditor/ComponentSource.tsx
+++ b/packages/website/components/LiveEditor/ComponentSource.tsx
@@ -75,12 +75,6 @@ const styles = {
     font-family: ${tokens.fontStackPrimary};
     border-radius: ${tokens.borderRadiusMedium} ${tokens.borderRadiusMedium} 0 0 !important;
   `,
-  previewWrapper: css({
-    display: 'flex',
-    flexDirection: 'column',
-    alignItems: 'center',
-    justifyContent: 'center',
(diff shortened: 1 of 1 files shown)


denkristoffer:
Would you mind adding a before/after screenshot?

mshaaban0:
> Would you mind adding a before/after screenshot?

### Before


#### Skeleton:
Screenshot 2023-03-09 at 18 22 21

#### Card:
Screenshot 2023-03-09 at 18 22 01

#### Accordion:
Screenshot 2023-03-09 at 18 20 06



### After

#### Skeleton:
Screenshot 2023-03-09 at 18 22 45

#### Card:
Screenshot 2023-03-09 at 18 21 50

#### Accordion:
Screenshot 2023-03-09 at 18 19 00

denkristoffer (approved):
Thanks!

PR #2436, opened 2023-03-20
feat: table sorting

denkristoffer:
# Purpose of PR

Table sorting props and an example.


packages/components/table/README.mdx
```diff
@@ -31,6 +31,14 @@ To achieve that result, you can iterate over the data and create `Table.Row` and
 
 ```
 
+### With sorting
+
+Table cells in the table header can be marked as sortable and sorted either ascending or descending.
```
giotiskl, line 36:
```suggestion
Table cells in the table header can be marked as sortable and sorted either in ascending or descending order.
```

packages/components/table/src/TableCell/TableCell.tsx
```diff
@@ -17,10 +22,20 @@ export const sortingDirections = {
   desc: 'desc',
 };
 
-export type TableCellSorting = keyof typeof sortingDirections | boolean;
+export enum TableCellSorting {
+  ascending = 'ascending',
```
giotiskl, line 26:
```suggestion
  Ascending = 'ascending',
  Descending = 'descending',
```
It's good practice to name enum keys with PascalCase (from the TS docs: https://www.typescriptlang.org/docs/handbook/enums.html)

packages/components/table/src/TableCell/TableCell.tsx
```diff
@@ -67,7 +98,31 @@ function _TableCell(
...
       testId={testId}
     >
-      {children}
+      {isSortable ? (
+        <button
+          aria-label={`Sort ${
+            isSorted === TableCellSorting.ascending
+              ? TableCellSorting.descending
+              : TableCellSorting.ascending
+          } by ${children}`}
```
giotiskl, line 107: what happens if `children` is not string and it is JSX? Won't this interpolation then show something like `by Function x()`?

bgutsol: wanted to ask the same, I guess we need to check the children type first and use it if it's `string` only

packages/components/table/examples/TableWithSortingPropsExample.tsx
```diff
@@ -0,0 +1,90 @@
...
+      id: '3',
+      name: 'Layout',
+      description:
+        'A page consisting of freely configurable and rearrangeable content modules.',
+      updatedAt: 'Nov 15, 2021',
+      status: 'published',
+    },
+  ];
+
+  const [sorting, setSorting] = useState(undefined);
```
bgutsol, line 29: Maybe it's better to provide some sorting (e.g. `TableCellSorting.Ascending`) by default because it was confusing that I have to click first on the Column heading to be able to see the sorting arrows in the example.

packages/components/table/examples/TableWithSortingPropsExample.tsx
```diff
@@ -0,0 +1,90 @@
...
+  const [sortedRows, setSortedRows] = useState(contentTypes);
+
+  const handleSort = (direction) => {
+    setSortedRows((rows) => {
```
bgutsol, line 33: Sorting logic doesn't work for me on the Table sorting example preview, it doesn't sort the column items in the right order.

bgutsol (commented):
Sorting itself doesn't work for me on the Table sorting example preview.
And also maybe it's better to provide some sorting (e.g. `TableCellSorting.Ascending`) by default because even for me it was confusing that I have to click first on the Column heading to be able to see the sorting arrows.

packages/components/table/examples/TableWithSortingPropsExample.tsx
denkristoffer: I missed a place for the enum rename. Fixed.

packages/components/table/examples/TableWithSortingPropsExample.tsx
denkristoffer: I actually don't think I agree here, it's seems quite common when I think of column sorting that the default is "unsorted". What we could do to fully support this is actually that on the third click we cycle back to this unsorted state.

denkristoffer:
I'm not sure what is causing the layout shift on Chromatic 😭 https://www.chromatic.com/pullrequest?appId=5fd1dda724cc620021ace8c5&number=2436&activeElementId=comparison-test-key-5ffd97971cab67e5f5687187-1200&view=changes

packages/components/table/examples/TableWithSortingPropsExample.tsx
denkristoffer: But I made the sorting a bit more understandable. I think.

bgutsol: The sortable example from Shopify is actually sorted by default by `Net sales` column, so we can see the sorting arrows when we open the page. That's what I meant we can do with our example as well. 
I don't think that cycling to the unsorted state is a good idea since that's not how sorting in real-world tables works.

But not a blocker.

bgutsol: And the example code is a bit hard to read and understand since it has a lot of ternary operators. I'd prefer to hide the complexity that users don't need to get an idea of how to add sorting to our table. Shopify did a good job on it by moving sorting logic to a separate function.
image

Also no a blocker.

bgutsol (approved):

mshaaban0:
> I'm not sure what is causing the layout shift on Chromatic 😭 https://www.chromatic.com/test?appId=5fd1dda724cc620021ace8c5&id=641da1974cd8e0fcd6bc1f00 @mshaaban0 any ideas?

@denkristoffer The layout shift is coming from the hidden icon, If you remove the svg from the button in `Organization role` column in Storybook example, size will match the current live example. I think it's fine to shipit 👍🏼

mshaaban0:
> I'm not sure what is causing the layout shift on Chromatic 😭 https://www.chromatic.com/test?appId=5fd1dda724cc620021ace8c5&id=641da1974cd8e0fcd6bc1f00 @mshaaban0 any ideas?

@denkristoffer The layout shift is coming from the hidden icon, If you remove the svg from the button in `Organization role` column in Storybook example, size will match the current live example. I think it's fine to shipit 👍🏼

PR #2438, opened 2023-03-22
fix(modal): remove modal initial focus timeout

maxcheremisin:
# Purpose of PR

- remove an arbitrary timeout before making initial focus
- check if the already focused element is a descendant of modal element (allows to use native `autoFocus` for initial focus)

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/components/modal/src/Modal.tsx
+++ b/packages/components/modal/src/Modal.tsx
@@ -143,19 +143,32 @@ export const Modal = ({
     overlayClassName: otherProps.overlayProps?.className,
   });
 
-  React.useEffect(() => {
-    if (props.isShown) {
-      setTimeout(() => {
-        if (props.initialFocusRef && props.initialFocusRef.current) {
-          if (props.initialFocusRef.current.focus) {
-            props.initialFocusRef.current.focus();
-          }
-        } else if (contentRef.current) {
-          focusFirstWithinNode(contentRef.current);
-        }
-      }, 100);
+  const onInitialFocus = () => {
+    const contentEl = contentRef.current;
+    if (contentEl) {
+      const activeEl = document.activeElement;
+      const isContentContainsActive =
+        contentEl !== activeEl && contentEl.contains(activeEl);
+
+      if (isContentContainsActive) {
+        return;
+      }
     }
-  }, [props.isShown, props.initialFocusRef]);
+
+    const initialFocusEl = props.initialFocusRef?.current;
+    if (initialFocusEl) {
+      initialFocusEl.focus?.();
+    } else if (contentEl) {
+      focusFirstWithinNode(contentEl);
+    }
+  };
+
+  const onAfterOpen: ModalProps['onAfterOpen'] = (...args) => {
+    if (props.onAfterOpen) {
+      props.onAfterOpen(...args);
+    }
+    onInitialFocus();
+  };
 
   const renderDefault = () => {
     return (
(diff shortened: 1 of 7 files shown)


bgutsol (approved):
Nice ⚡️ Thanks for improving it!

Do you think it makes sense to add a sentence in our docs that users can also leverage the native `autoFocus` attr to set the initial focus? Maybe we can add it to the existing [Setting initial focus example](https://github.com/contentful/forma-36/blob/7e46b96fd42f9a760987802acd7ef0b1811e9894/packages/components/modal/src/README.mdx#L62)

maxcheremisin:
> Nice ⚡️ Thanks for improving it!
> 
> Do you think it makes sense to add a sentence in our docs that users can also leverage the native `autoFocus` attr to set the initial focus on interactive elements? Maybe we can add it to the existing [Setting initial focus example](https://github.com/contentful/forma-36/blob/7e46b96fd42f9a760987802acd7ef0b1811e9894/packages/components/modal/src/README.mdx#L62)

Good suggestion. Done!

packages/components/modal/src/Modal.tsx
```diff
@@ -143,19 +143,32 @@ export const Modal = ({
...
+        return;
+      }
+    }
+
+    const initialFocusEl = props.initialFocusRef?.current;
+    if (initialFocusEl) {
+      initialFocusEl.focus?.();
+    } else if (contentEl) {
+      focusFirstWithinNode(contentEl);
     }
```
cdun, line 163: Knowing the following (snippets from MDN), we can simplify a bit:

`activeElement`:
> The [Element](https://developer.mozilla.org/en-US/docs/Web/API/Element) which currently has focus, [<body>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/body) or null if there is no focused element.

`Node.contains`:
> If the otherNode parameter is null, contains() always returns false.

The following should work since `focusFirstWithinNode` is safe to call with a null argument. I removed the check for `contentEl !== activeEl`, though perhaps I'm missing an edge case where that's possible?

```suggestion
    const contentEl = contentRef.current;
    if (contentEl?.contains(document.activeElement)) {
      // A descendant of the modal already has focus.
      return;
    }

    const initialFocusEl = props.initialFocusRef?.current;
    if (initialFocusEl) {
      initialFocusEl.focus?.();
      return;
    }
    
    focusFirstWithinNode(contentEl);
```

packages/components/modal/src/Modal.tsx
```diff
@@ -143,19 +143,32 @@ export const Modal = ({
...
-  }, [props.isShown, props.initialFocusRef]);
+  };
+
+  const onAfterOpen: ModalProps['onAfterOpen'] = (...args) => {
```
cdun, line 166: We should probably wrap in `useCallback` here so that the value passed to the modal is stable between renders

packages/components/modal/src/Modal.tsx
maxcheremisin: > I removed the check for contentEl !== activeEl, though perhaps I'm missing an edge case where that's possible?

the edge case here is that `contentEl` becomes initially focused by `react-modal`

packages/components/modal/src/Modal.tsx
maxcheremisin: i have to disagree, in order to make it _work_, we have to either force users to wrap `onAfterOpen` prop in `useCallback`, or save the original function to a ref object in order to avoid stale closure problem.

but all this doesn't make much sense, because `react-modal` doesn't rely on `onAfterOpen` function identity
https://github.com/reactjs/react-modal/blob/master/src/components/ModalPortal.js#LL229-L253C5

cdun: If it's not a dependency in react-modal, then makes sense to me 👍

cdun (approved):

PR #2451, opened 2023-04-19
[DON'T MERGE] feat: update workbench header styles

massao:
# Purpose of PR

Update workbench header style

### Preview
https://5fd1dda724cc620021ace8c5-zczrvuwgco.chromatic.com/?path=/story/components-workbench--header-overview


fabe:
- The back button seems too large, I think we've used something different for this in the breadcrumb component?
- Border bottom should be gray-200
- The question mark/help icon seems too large, maybe we can use the tiny variant here

denkristoffer:
Maybe we should hide `[data-test-id="cf-ui-icon"]` to remove product icons?

massao:
> Maybe we should hide `[data-test-id="cf-ui-icon"]` to remove product icons?

We could also just not render the icon at all, keep it on the props to not break anything but don't actually render the icon, since we won't be merging this into main anyway

Lelith:
no longer relevant

PR #2453, opened 2023-04-20
feat: add new MissingContent component

Lelith:
# Purpose of PR

This adds a new small utility component which called MissingContent. This will help to display empty cells or other missing content. It accepts a className for overwriting styles, as well as a test-id and an optional title-field, which can be leveraged for further context to what exactly is missing. The goal of the component is to help reduce visual clutter when content is missing.

image


packages/components/utils/src/MissingContent/MissingContent.mdx
```diff
@@ -0,0 +1,74 @@
+---
+title: 'MissingContent'
+type: 'component'
+slug: /utils/missingcontent/
+section: 'utils'
```
denkristoffer, line 5: I don't think we should put components rendered to the user in utils, but if you're thinking there's no other obvious place to put it I would suggest something like an `empty-state` directory that we can expand with other types of empty state components in the future.

denkristoffer (commented):
Don't forget the changeset 😄

packages/components/utils/src/MissingContent/MissingContent.tsx
```diff
@@ -0,0 +1,29 @@
...
+  ...otherProps
+}: MissingContentProps) => {
+  const styles = getStyles();
+
+  return (
+    <Text
+      {...otherProps}
+      data-test-id={testId}
+      className={cx(styles.missingContent, className)}
+      title={title}
```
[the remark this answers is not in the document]
Lelith: In the cases i want to use the component it would be quite helpful to have further information e.g. required content is missing in the default language, description is not yet provided etc. 

So its low-key important for people that are wondering what this placeholder stands for. but not sure if it is crucial information.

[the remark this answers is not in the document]
denkristoffer: We could also enforce an `aria-label` prop.

packages/components/utils/src/MissingContent/MissingContent.mdx
```diff
@@ -0,0 +1,74 @@
...
+      </Table.Cell>
+    </Table.Row>
+    <Table.Row>
+      <Table.Cell>Dog.jpg</Table.Cell>
+      <Table.Cell>
+        <MissingContent title="No Dimensions available" />
+      </Table.Cell>
+    </Table.Row>
+  </Table.Body>
+  <MissingContent />
```
mshaaban0, line 47: Screenshot 2023-04-20 at 16 22 52

```suggestion
```

packages/components/utils/src/MissingContent/MissingContent.tsx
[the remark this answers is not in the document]
mshaaban0: +1 for `aria-label` "forced"

mshaaban0 (commented):
Do you think it makes sense to have all of this JS code to just render `&mdash;`?

Lelith:
@mshaaban0 yes, its a lot of JS for rendering a simple dash, but this makes it easier for developers to follow the design-system. If we do not provide small utilities like this, we end up with 10+ implementations of dashes, where maybe some are the wrong dash, some are the wrong shade of grey etc.

packages/components/empty-state/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "@contentful/f36-utils",
```
massao, line 2:
```suggestion
  "name": "@contentful/f36-empty-state",
```

packages/components/empty-state/src/MissingContent/MissingContent.tsx
```diff
@@ -0,0 +1,29 @@
...
+  testId = 'cf-ui-missing-content',
+  title,
+  ...otherProps
+}: MissingContentProps) => {
+  const styles = getStyles();
+
+  return (
+    <Text
+      {...otherProps}
+      data-test-id={testId}
```
massao, line 22: Text uses the common props so it has `testId`
```suggestion
      testId={testId}
```

packages/components/empty-state/src/MissingContent/MissingContent.mdx
```diff
@@ -0,0 +1,74 @@
...
+  </Heading>
+  <TextLink
+    href="https://f36-storybook.contentful.com/?path=/story/empty-state-missingcontent--basic"
+    target="_blank"
+    alignIcon="end"
+    icon={<ExternalLinkIcon />}
+  >
+    Open in Storybook
+  </TextLink>
+</Stack>
```
massao, line 72: Is this necessary?? shouldn't it be automatically like on the other components?

packages/components/empty-state/src/MissingContent/MissingContent.mdx
```diff
@@ -0,0 +1,74 @@
...
+section: 'utils'
+typescript: ./MissingContent.tsx
+---
+
+`MissingContent` component can be used to render a placeholder for an empty table cell or any inline content that is missing. Its goal is to provide a system information without cluttering the view.
+
+## Import
+
+```jsx static=true
+import { MissingContent } from '@contentful/f36-utils';
```
massao, line 14:
```suggestion
import { MissingContent } from '@contentful/f36-empty-state';
```

packages/components/empty-state/src/MissingContent/MissingContent.mdx
```diff
@@ -0,0 +1,74 @@
+---
+title: 'MissingContent'
+type: 'component'
+slug: /empty-state/missingcontent/
+section: 'utils'
+typescript: ./MissingContent.tsx
+---
```
massao, line 7: should this be on alpha? Or if it's already as stable should it be under f36-components and sync the versions?

packages/components/utils/src/MissingContent/MissingContent.mdx
Lelith: thanks, good catch!

packages/components/empty-state/src/MissingContent/MissingContent.mdx
Lelith: as this is so tiny, i think an alpha is maybe a little overkill?

packages/components/utils/src/MissingContent/MissingContent.tsx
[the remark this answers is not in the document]
Lelith: But aria-label will not appear on hover like title, or? Should we then have both?

packages/components/empty-state/src/MissingContent/MissingContent.mdx
Lelith: copied it by accident, thanks for pointing it out. i somehow have issues running the website locally

packages/components/empty-state/src/MissingContent/MissingContent.mdx
```diff
@@ -0,0 +1,73 @@
...
+title: 'MissingContent'
+type: 'component'
+slug: /empty-state/missingcontent/
+section: 'utils'
```
massao, line 5: We may remove the section, since we don't have other components on the empty state section at the moment

massao, line 5: We may remove the section, since we don't have other components on the empty state section at the moment

packages/components/empty-state/src/MissingContent/MissingContent.mdx
```diff
@@ -0,0 +1,73 @@
...
+      </Table.Cell>
+    </Table.Row>
+    <Table.Row>
+      <Table.Cell>Dog.jpg</Table.Cell>
+      <Table.Cell>
+        <MissingContent label="No Dimensions available" />
+      </Table.Cell>
+    </Table.Row>
+  </Table.Body>
+</Table>
```
massao, line 47: Also on the new components we use separate folder/file examples, take a look into the other components structure

massao:
Also remember to add the package as dependency on the website package.json so the examples on the website work

packages/components/empty-state/src/MissingContent/MissingContent.mdx
massao: It may be due to the env variables

packages/components/utils/src/MissingContent/MissingContent.tsx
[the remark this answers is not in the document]
mshaaban0: Both are good yes 👍🏼

packages/components/empty-state/README.mdx
```diff
@@ -0,0 +1,33 @@
...
+storybook: 'https://f36-storybook.contentful.com/?path=/story/empty-state-missingcontent--basic'
+typescript: ./MissingContent.tsx
+---
+
+`MissingContent` component can be used to render a placeholder for an empty table cell or any inline content that is missing. Its goal is to provide a system information without cluttering the view.
+
+## Import
+
+```jsx static=true
+import { MissingContent } from '@contentful/f36-empty-state';
```
denkristoffer, line 18:
```suggestion
import { MissingContent } from '@contentful/f36-components';
// or
import { MissingContent } from '@contentful/f36-empty-state';
```

denkristoffer (approved):
Looks good, one tiny nit for docs.

PR #2464, opened 2023-05-04
feat: add NavList component in alpha version

massao:
# Purpose of PR

Add NavList component

## PR Checklist

- [ ] Tests are passing


packages/components/navlist/src/NavListItem/NavListItem.styles.ts
```diff
@@ -0,0 +1,77 @@
...
+          '&:focus': {
+            boxShadow: `inset ${tokens.glowPrimary}`,
+          },
+          '&:focus:not(:focus-visible)': {
+            boxShadow: 'unset',
+          },
+          '&:focus-visible': {
+            boxShadow: `inset ${tokens.glowPrimary}`,
+          },
+          '&:active': activeStyle,
```
fabe, line 69: Let's do the same as Menu here, the font-weight (and color) shouldn't change for :active.

packages/components/navlist/src/NavListItem/NavListItem.styles.ts
```diff
@@ -0,0 +1,77 @@
...
+const activeStyle: ObjectInterpolation<undefined> = {
+  backgroundColor: tokens.gray200,
+  color: tokens.gray900,
+  fontWeight: tokens.fontWeightMedium,
+  '&:hover': {
+    backgroundColor: tokens.gray200,
+  },
+};
+
+const disabledStyle: ObjectInterpolation<undefined> = {
```
fabe, line 14: `cursor: not-allowed`?

fabe (approved):

packages/components/navlist/src/NavList/NavList.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { getStyles } from './NavList.styles';
+
+const NAV_LIST_DEFAULT_TAG = 'div';
+
+export interface NavListInternalProps extends CommonProps {
+  children: React.ReactNode;
+  /**
+   * @default div
+   */
+  as?: 'nav' | 'div';
```
mshaaban0, line 18: Should it be `nav` by default?

packages/components/navlist/src/NavList/NavList.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+export type NavListProps<
+  E extends React.ElementType = typeof NAV_LIST_DEFAULT_TAG,
+> = PolymorphicProps<NavListInternalProps, E>;
+
+function _NavList<E extends React.ElementType = typeof NAV_LIST_DEFAULT_TAG>(
+  props: NavListProps<E>,
+  ref: React.Ref<any>,
+) {
+  const styles = getStyles();
```
mshaaban0, line 29: I know we use this pattern a lot, but I don't see a reason to recompute the styles here on every render, wdyt?

packages/components/navlist/src/NavList/NavList.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  const { as, className, children, testId, ...otherProps } = props;
+  const role = as === 'nav' ? 'navigation' : undefined;
```
mshaaban0, line 32: I think should be the other way around, when it's a nav the role is `navigation` out of the box, we only need to set role when it's a `div`

packages/components/navlist/src/NavList/NavList.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  const Element = (as ?? NAV_LIST_DEFAULT_TAG) as React.ElementType;
+
+  return (
+    <Element
+      aria-label="Sidebar"
```
mshaaban0, line 38:
```suggestion
      aria-label="Sidebar Navigation"
```
We probably should expose it as a prop as well. users should be able to provide descriptive text about the menu

mshaaban0 (commented):
Looks really good, just a couple of things to clarify before shipping

packages/components/navlist/src/NavList/NavList.tsx
```diff
@@ -0,0 +1,55 @@
...
+  E extends React.ElementType = typeof NAV_LIST_DEFAULT_TAG,
+> = PolymorphicProps<NavListInternalProps, E>;
+
+function _NavList<E extends React.ElementType = typeof NAV_LIST_DEFAULT_TAG>(
+  props: NavListProps<E>,
+  ref: React.Ref<any>,
+) {
+  const styles = getStyles();
+
+  const { as, className, children, testId, ...otherProps } = props;
```
mshaaban0, line 31:
```suggestion
  const { as = 'nav', className, children, testId, ...otherProps } = props;
```

packages/components/navlist/src/NavList/NavList.tsx
massao: Should we add a custom prop tho?? because the user can already pass an `aria-label` props, that's why I even added it before the `{...otherProps}`.
And adding another prop we would have the custom prop and also the `aria-label`, which should be the priority in this case? 🤔

mshaaban0: Ah yes you're right it'll be overridden, I missed that, thanks 👍🏼

packages/components/navlist/src/NavList/NavList.tsx
massao: This is so the css class is only generated when the component itself is rendered, otherwise just by importing the file it would create the class, bloating the bundle.
And IIRC emotion handles caching this 🤔

packages/components/navlist/src/NavList/NavList.tsx
massao: nice catch, missed that

PR #2470, opened 2023-05-08
feat: create new modal variant fullscreen

massao:
# Purpose of PR

Add new size variant to modal, to use entire screen with 16px margin around.

### DIFF
--- a/packages/components/modal/src/Modal.styles.ts
+++ b/packages/components/modal/src/Modal.styles.ts
@@ -36,6 +36,15 @@ export function getModalStyles(props: {
           width: '100%',
         })
       : null,
+    props.size === 'fullscreen'
+      ? css({
+          maxWidth: `calc(100vw - ${tokens.spacingXl})`,
+          maxHeight: `calc(100vh - ${tokens.spacingXl})`,
+          margin: 0,
+          height: '100vh',
+          width: '100vw',
+        })
+      : null,
     props.className,
   );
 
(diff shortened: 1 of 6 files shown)


packages/components/modal/src/Modal.styles.ts
```diff
@@ -98,10 +112,10 @@ export function getModalStyles(props: {
         props.overlayClassName,
       ),
       afterOpen: css({
-        opacity: 1,
+        opacity: '1 !important',
```
mshaaban0, line 115: What's the reason for important here? is it being overridden by something?

mshaaban0 (commented):
image

image

 
 A couple of design differences I spotted:
 * Modal controls spacing seems off
 * Modal controls border top is missing
 * Modal controls background should have the blur effect
 * Should we pin the modal controls at the bottom of the modal all the time, even if it's not full width?

packages/components/modal/src/Modal.styles.ts
bgutsol: because of the earlier `padding` reset, emotion generates a new class for the fullscreen modal, and it renders below `afterOpen` class, that's why it has bigger specificity. 

CleanShot 2023-05-09 at 13 02 44@2x

I'm wondering if we can somehow reset the padding for the `fullscreen` differently, so we don't need to hack the specificity, wdyt? 🤔

And btw in storybook, it works fine without `!important` 🤷‍♂️

bgutsol (approved):
Looks good, I'd only try to avoid that `!important` if it's possible

damann:
> Should we pin the modal controls at the bottom of the modal all the time, even if it's not full width?
Yes, please.

damann:
> Should we pin the modal controls at the bottom of the modal all the time, even if it's not full width?
Yes, please.
> Also do we want to update the style of the controls on all modals the spacing, border and such??
I'd say yes to the spacing, but the border don't introduce a border to the other modals please. 
@fabe wdyt? Any objections?
Modal_original
Modal_16totheright12tothebottom

mshaaban0 (approved):

PR #2481, opened 2023-05-11
feat: Plain style for `DragHandle` component

bgutsol:
# Purpose of PR

- Introduce plain style for `DragHandle` component

### DIFF
--- a/packages/components/drag-handle/README.mdx
+++ b/packages/components/drag-handle/README.mdx
@@ -49,6 +49,14 @@ The DragHandle can be styled based on the props you pass to it:
 <DragHandle isHovered label="Hovered drag handle" />
 ```
 
+**Plain**:
+
+Removes default background color and border radius.
+
+```jsx
+<DragHandle isPlain label="Plain style drag handle" />
+```
+
 ### Card example
 
 ```jsx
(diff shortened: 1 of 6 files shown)


packages/components/drag-handle/README.mdx
```diff
@@ -49,6 +49,14 @@ The DragHandle can be styled based on the props you pass to it:
 <DragHandle isHovered label="Hovered drag handle" />
 ```
 
+**Plain**:
+
+Removes default background color and border radius.
+
+```jsx
+<DragHandle isPlain label="Plain style drag handle" />
```
denkristoffer, line 57: To keep with convention I'd say this should probably be a `variant`?

bgutsol: That was my initial thought, but do we want to have only 2 variants, like `default` and `plain`? And I'm not sure we gonna introduce any other variant for this component, since `plain` should cover all the cases.

denkristoffer: In my opinion yes, for consistency's sake. I don't think it's a problem that there are only two variants. Users will probably have an easier time understanding what the `plain` variant is if they already know of the prop from other components.

mshaaban0: We have a `transparent` variant on button, maybe we can call it that instead

denkristoffer (approved):

PR #2486, opened 2023-05-24
feat(entity-status-badge): add isScheduled prop

denkristoffer:
# Purpose of PR

Add a prop to `EntityStatusBadge` that lets us mark entities as having a scheduled action. This will make the badge render our `ClockIcon`.

### DIFF
--- a/packages/components/badge/src/EntityStatusBadge/EntityStatusBadge.tsx
+++ b/packages/components/badge/src/EntityStatusBadge/EntityStatusBadge.tsx
@@ -1,8 +1,13 @@
 import React from 'react';
-import type { EntityStatus, ExpandProps } from '@contentful/f36-core';
+import type {
+  EntityStatus,
+  ExpandProps,
+  PickUnion,
+} from '@contentful/f36-core';
+import { ClockIcon } from '@contentful/f36-icons';
 
 import { Badge, type BadgeProps } from '../Badge/Badge';
-import type { BadgeVariant } from '../types';
+import type { BadgeSize, BadgeVariant } from '../types';
 
 const statusMap: { [key in EntityStatus]: BadgeVariant } = {
   published: 'positive',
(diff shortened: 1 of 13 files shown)


massao (approved):
LGTM

packages/components/badge/src/EntityStatusBadge/EntityStatusBadge.tsx
```diff
@@ -13,20 +18,58 @@ const statusMap: { [key in EntityStatus]: BadgeVariant } = {
...
   return (
-    <Badge {...otherProps} size={size} variant={variant} ref={ref}>
+    <Badge
+      {...otherProps}
+      {...(withIcon
+        ? { size, startIcon: <ClockIcon testId="schedule-icon" /> }
+        : { size })}
+      variant={variant}
+      ref={ref}
+    >
```
bgutsol, line 72: nit
```suggestion
    <Badge
      {...otherProps}
      size={size}
      icon={withClockIcon ? <ClockIcon testId="schedule-icon" /> : undefined}
      variant={variant}
      ref={ref}
    >
```

packages/components/badge/src/EntityStatusBadge/EntityStatusBadge.tsx
```diff
@@ -13,20 +18,58 @@ const statusMap: { [key in EntityStatus]: BadgeVariant } = {
...
 ) {
-  const { entityStatus, size = 'default', ...otherProps } = props;
-
+  const {
+    entityStatus,
+    isScheduled = undefined,
+    size = 'default',
+    ...otherProps
+  } = props;
+  const withIcon = isScheduled && size === 'default';
```
bgutsol, line 61:
```suggestion
  const withClockIcon = isScheduled && size === 'default';
```

bgutsol (commented):
1. Let's add example to the website docs as well.
2. There are some changes in the chromatic that looks like wrong ones

packages/components/badge/src/EntityStatusBadge/EntityStatusBadge.tsx
denkristoffer: Typescript doesn't accept it because the startIcon can only be passed when `size === 'small'` 😞

denkristoffer:
>     2. There are some changes in the chromatic that looks like wrong ones
> 
> 
> CleanShot 2023-05-24 at 12 58 49@2x

Thanks for flagging. It's the baseline that's wrong, I had missed setting the size for the last column of badges. The new snapshot is the right one 👍

denkristoffer:
@bgutsol I updated the docs.

Now that I look at it I don't think it makes sense to have content guidelines or accessibility guidelines for this component since neither the content nor the styling can be customised.

bgutsol (approved):

PR #2492, opened 2023-05-31
fix: Table.Body children type

bgutsol:
- fix `Table.Body` children type

### DIFF
--- a/packages/components/table/src/TableBody/TableBody.tsx
+++ b/packages/components/table/src/TableBody/TableBody.tsx
@@ -7,7 +7,7 @@ import {
 } from '@contentful/f36-core';
 
 export type TableBodyInternalProps = CommonProps & {
-  children?: ReactNode;
+  children: ReactNode;
 };
 
 export type TableBodyProps = PropsWithHTMLElement<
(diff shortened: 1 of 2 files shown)


massao (approved):

packages/components/table/src/TableBody/TableBody.tsx
```diff
@@ -7,7 +7,7 @@ import {
 } from '@contentful/f36-core';
 
 export type TableBodyInternalProps = CommonProps & {
-  children: ReactElement | ReactElement[];
+  children?: ReactNode;
```
massao, line 10: should it be nullable tho?

bgutsol: for conditional rendering might be useful?

bgutsol: like this

massao: Do you mean adding a condition as the children?? because even like that if you had `{someCondition && <div />}` it would still be considered as having a children, and it is a ReactNode 🤔
Or is there a case where we would not pass any children to the Table.Body??

bgutsol: I guess there is no such case, that's why I rolled back to required children, thanks for raising it 👍

PR #2493, opened 2023-05-31
feat(pill): add isDraggable prop

denkristoffer:
# Purpose of PR

Add `isDraggable` prop to `Pill` component, which renders the drag handle without passing an `onDrag` prop.

### DIFF
--- a/.changeset/wild-rockets-wink.md
+++ b/.changeset/wild-rockets-wink.md
@@ -1,0 +1,5 @@
+---
+"@contentful/f36-pill": minor
+---
+
+feat(pill): add isDraggable prop
--- a/package-lock.json
+++ b/package-lock.json
@@ -37692,14 +37692,14 @@
     },
     "packages/components/accordion": {
       "name": "@contentful/f36-accordion",
-      "version": "4.40.3",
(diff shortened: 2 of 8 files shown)


massao (approved):
Should we point out that the `dragHandleComponent` is always required when wanting to render the drag handle? And either the `isDraggable` or the `onDrag`, because at the moment we don't have a fallback for the dragHandle component right?? 🤔 

But looks good to me anyway, just an open question

denkristoffer:
@massao There's a built-in fallback although it doesn't use our `DragHandle` component. I was thinking about changing it but I wonder if we initially made it with a `span` for a reason? https://github.com/contentful/forma-36/blob/51a37aa763deeba32b5ca58189e32ca039c51eb4/packages/components/pill/src/Pill.tsx#L84-L86

massao:
> 

Could be because before we didn't have the transparent variant an the DragHandle only had the styling to be used on the cards, if the transparent variant would fit we should use it, right?? It wouldn't add too many dependencies, the DragHandle only depends on the Icon that is already being used here.

denkristoffer:
@massao done.

PR #2494, opened 2023-05-31
docs: add dnd integration docs

denkristoffer:
Replaces a few older examples using other solutions


packages/website/package.json
```diff
@@ -12,24 +12,26 @@
     "@codesandbox/sandpack-react": "^1.17.0",
     "@contentful/f36-components": "^4.41.1",
     "@contentful/f36-docs-utils": "^4.0.2",
+    "@contentful/f36-empty-state": "^4.41.1",
```
bgutsol, line 15: it can be imported from `f36-components`, right?

packages/website/content/integrations/dnd-kit.mdx
```diff
@@ -0,0 +1,15 @@
...
+title: 'Dnd kit'
+slug: '/integrations/dnd-kit/'
+section: 'integrations'
+---
+
+We recommend using the excellent [dnd kit](https://dndkit.com/) library for adding drag and drop functionality to Forma 36 components.
+
+## Examples
+
+### Integrating with `useSortable`
```
bgutsol, line 11: Let's provide one more example with keyboards controls enabled. So with the usage of:
```
useSensors(
    useSensor(PointerSensor),
    useSensor(KeyboardSensor, {
      coordinateGetter: sortableKeyboardCoordinates,
    })
  );
```

And I would also maybe add it to the example with Card component.
WDYT?

bgutsol: And might also be valuable to add an example with different UI components from Forma. For instance `Table`. Basically to show that DnD is working with any UI component.

packages/components/card/stories/EntryCard.stories.tsx
```diff
@@ -27,38 +22,6 @@ export default {
   title: 'Components/Card/EntryCard',
 } as Meta;
 
-const DragHandle = (props: { drag: React.ReactElement }) => {
```
bgutsol, line 30: Should we provide stories with DnD-kit instead?

packages/components/pill/README.mdx
```diff
@@ -32,14 +32,6 @@ import { Pill } from '@contentful/f36-pill';
 
 ```
 
-### Integration example with `react-sortable-hoc`
```
bgutsol, line 35: I guess you will provide a new example when the Pill changes PR is merged, right?

packages/website/content/integrations/dnd-kit.mdx
```diff
@@ -0,0 +1,15 @@
+---
+title: 'Dnd kit'
+slug: '/integrations/dnd-kit/'
+section: 'integrations'
+---
+
+We recommend using the excellent [dnd kit](https://dndkit.com/) library for adding drag and drop functionality to Forma 36 components.
+
```
mshaaban0, line 8:
```suggestion
## Import
```bash
import { DndContext } from '@dnd-kit/core';
import {
  arrayMove,
  horizontalListSortingStrategy,
  SortableContext,
  useSortable,
} from '@dnd-kit/sortable';
import { CSS } from '@dnd-kit/utilities';
```
```

Comment (mshaaban0) [reply packages/website/content/integrations/dnd-kit.mdx:8]: Since imports are not visible in the examples below, would be nice to see what I need to import

**Review by mshaaban0 — APPROVED** [review]

File: packages/website/content/integrations/dnd-kit.mdx (hunk 2 above)
Comment (denkristoffer) [reply packages/website/content/integrations/dnd-kit.mdx:11]: The defaults seem to work with keyboard navigation for me, although I'm guessing the KeyboardSensor makes it nicer somehow? I would like to keep the examples as simple as possible, readers can always dig into dnd kit docs if they care for more.

File: packages/components/card/stories/EntryCard.stories.tsx (hunk 3 above)
Comment (denkristoffer) [reply packages/components/card/stories/EntryCard.stories.tsx:30]: In my opinion this doesn't serve much value in Storybook, and then we have to keep these examples up do date in multiple places.

File: packages/website/package.json (hunk 1 above)
Comment (denkristoffer) [reply packages/website/package.json:15]: Yeah true, this change was `npm` automatically reordering things.

File: packages/components/card/examples/Card/WithCustomDragHandle.tsx [hunk 6]
```diff
@@ -1,76 +1,82 @@
...
+      useSortable({
+        id,
+      });
 
-  const SortableDragHandle = SortableHandle(() => (
-    <DragHandle as="button" className={styles.dragHandle} label="Move card" />
-  ));
+    const style = {
+      transform: CSS.Translate.toString(transform),
+      transition,
```
massao, line 34: Should it also have the z-index logic to show the dragged element always on top?

packages/components/card/stories/EntryCard.stories.tsx
bgutsol: I think it's gonna be valuable to have at least one example of DnD with our card, but not necessarily an EntryCard

packages/components/card/examples/Card/WithCustomDragHandle.tsx
denkristoffer: I added this to the table example.

packages/components/card/stories/EntryCard.stories.tsx
denkristoffer: There's already drag and drop in the Card example with DragHandle: https://forma-36-git-feat-dnd-docs.colorfuldemo.com/components/card#with-a-custom-drag-handle

packages/components/card/examples/Card/WithCustomDragHandle.tsx
```diff
@@ -1,76 +1,88 @@
...
-
-  const SortableList = SortableContainer((props) => (
-    <Flex flexDirection="column">{props.children}</Flex>
-  ));
-
-  const SortableDragHandle = SortableHandle(() => (
-    <DragHandle as="button" className={styles.dragHandle} label="Move card" />
-  ));
+  function SortableCard({ id }) {
+    const { attributes, listeners, setNodeRef, transform, transition } =
```
massao, line 31:
```suggestion
    const { attributes, listeners, setNodeRef, transform, transition, active } =
```

packages/components/card/examples/Card/WithCustomDragHandle.tsx
```diff
@@ -1,76 +1,88 @@
...
+      useSortable({
+        id,
+      });
+    const zIndex = active && active.id === id ? 1 : 0;
```
massao, line 35:
```suggestion
    const zIndex = active?.id === id ? 1 : 0;
```

denkristoffer: We can't use optional chaining in the website code 😢

massao (approved):

PR #2499, opened 2023-06-07
feat: Image component

denkristoffer:
# Purpose of PR

Adds a new `Image` component that will be used as a base for later stuff.

This one might make sense to put in `f36-core` once it's stable instead of having it in its own package. Thoughts?


packages/components/image/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "@contentful/f36-image",
+  "version": "4.0.0",
```
massao, line 3: Should be tagged as alpha?
You can use the script to make it easier, the script also adds it to the exclude file in changeset to avoid publishing it by mistake

massao (approved):

PR #2506, opened 2023-06-21
fix: small Select padding

MayaGillilanContentful:
# Purpose of PR
Small Selects had incorrect padding, which was causing long text to overlap with the chevron
Before:
image

After:
image

### DIFF
--- a/.changeset/large-hotels-wait.md
+++ b/.changeset/large-hotels-wait.md
@@ -1,0 +1,5 @@
+---
+'@contentful/f36-forms': minor
+---
+
+Fix text overlapping for small Selects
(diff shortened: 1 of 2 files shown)


.changeset/large-hotels-wait.md
```diff
@@ -0,0 +1,5 @@
+---
+'@contentful/f36-forms': minor
+---
+
+Fix text overlapping for small Selects
```
denkristoffer, line 5:
```suggestion
- Fix text overlapping for small Selects
```

denkristoffer (approved):

PR #2521, opened 2023-07-10
chore(dependabot): widen versions when possible

denkristoffer:
# Purpose of PR

Set [dependabot versioning strategy](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy) to `widen`. This seems like the right move for libraries.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,3 +11,4 @@ updates:
       - dependencies
     commit-message:
       prefix: chore
+    versioning-strategy: widen


bgutsol:
> Set [dependabot versioning strategy](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy) to `widen`. This seems like the right move for libraries.

@denkristoffer could you briefly describe the difference pls. I couldn't grasp if from the docs you've provided.

Lelith:
just wanted to ask the same question as @bgutsol because i am not sure if i understand the effect of this change.

denkristoffer:
@bgutsol @Lelith As I understand it it will try to widen the accepted range of versions when something new is released. For instance, we point to `react@^17.0.0` and instead of changing that to `react@^18.0.0` dependabot should try `react@^17.0.0 || react@^18.0.0` first. As a library we probably want the widest possible support.

However, the main reason I'm setting this is honestly because I'm trying to force dependabot to generate lockfile changes for #2410. According to Stack Overflow this should be able to do that.

bgutsol (approved):

PR #2529, opened 2023-07-11
chore(deps): bump next-auth

denkristoffer:
# Purpose of PR

Upgrading `next-auth` should get rid of the "version not supported" error we're seeing for node 18.


bgutsol (approved):
👍

PR #2532, opened 2023-07-12
feat(note): add 'icon' prop

cf-remylenoir:
# Purpose of PR

The Note component uses a pre-defined set of icons based on the selected variant and does not allow passing a custom icon. This PR introduces a new prop to allow passing a custom icon, which is inheriting its color/size from the Note variant/title props.

**Examples**

image

image

## PR Checklist

- [x] Tests are passing


packages/components/note/src/Note.tsx
```diff
@@ -53,6 +53,10 @@ export type NoteInternalProps = CommonProps & {
    * @default false
    */
   withCloseButton?: boolean;
+  /**
+   * Expects any of the icon components to override the default Note variant icon
+   */
+  startIcon?: React.ReactElement;
```
cf-remylenoir, line 59: I was hesitant to use `icon` instead of `startIcon`, but to stay consistent with the API of other components I kept it.

Lelith (approved):
Looks good to me!

packages/components/note/src/Note.tsx
```diff
@@ -73,11 +77,19 @@ export const Note = React.forwardRef<HTMLElement, ExpandProps<NoteProps>>(
...
+      startIcon,
       ...otherProps
     } = props;
 
     const styles = getNoteStyles();
 
+    const iconContent = (icon: React.ReactElement) =>
+      React.cloneElement(icon, {
+        size: icon.props.size ?? `${title ? 'medium' : 'small'}`,
+        variant:
```
denkristoffer, line 89: I'm wondering if we should lock the icon variant to the note variant, the same way we do with the button component: https://github.com/contentful/forma-36/blob/fafa03f7c2f2d4425e906e4c9a227a61d3816641/packages/components/button/src/Button/Button.tsx#L54-L81 – my gut feeling says yes, we can always allow more customisation later on if it turns out to be needed.

packages/components/note/src/Note.tsx
```diff
@@ -73,11 +77,19 @@ export const Note = React.forwardRef<HTMLElement, ExpandProps<NoteProps>>(
...
       title,
+      startIcon,
       ...otherProps
     } = props;
 
     const styles = getNoteStyles();
 
+    const iconContent = (icon: React.ReactElement) =>
+      React.cloneElement(icon, {
+        size: icon.props.size ?? `${title ? 'medium' : 'small'}`,
```
denkristoffer, line 88: Do you see a need to open up for size customisation?

packages/components/note/src/Note.tsx
cf-remylenoir: Good call, let's it keep closed for now.

packages/components/note/src/Note.tsx
cf-remylenoir: Not required for our first use case of using a custom component, let's keep the defaults for now.

.changeset/silly-bears-lie.md
```diff
@@ -0,0 +1,5 @@
+---
+"@contentful/f36-note": minor
+---
+
+feat(note): add 'icon' prop
```
denkristoffer, line 5:
```suggestion
- feat(note): add 'icon' prop
```

denkristoffer (approved):
Thanks, looks good to me but I just remembered the changeset format we're using on the website has changed to bullets.

PR #2535, opened 2023-07-17
fix: apply style refresh to multiselect  [HOMER-2159]

Lelith:
# Purpose of PR
fixes smaller spacing and highlight issues on the multiselect and applies the style refresher from the menu component-

before:
image

image

after:
image

image


fabe:
Looks great!

1. Could we make the search input fixed/sticky when scrolling the item list?
2. The vertical padding of the container here should be the same as the horizontal one:
  image
  image

3. Between each item, there should be 2px spacing vertically.

bgutsol (approved):
Looks great ✨

It might be not relevant to these changes, but `With Title` example has a weird spacing in storybook.

CleanShot 2023-07-17 at 17 53 13@2x

fabe:
1. Could we vertically center the search? I think that broke in the last commit.
  image

2. What do we think about autofocusing the search when opening? Or an accessibility no-no?

Lelith:
> Looks great ✨
> 
> It might be not relevant to these changes, but `With Title` example has a weird spacing in storybook.
> 
> CleanShot 2023-07-17 at 17 53 13@2x

that storybook example has a weird spacing because its wrapped in a unstyled ul-tag

fabe (approved):

PR #2542, opened 2023-07-17
refactor: leverage react-day-picker functionality instead of cusomising

bgutsol:
# Purpose of PR

With the recently added [CaptionLayout](https://react-day-picker.js.org/api/types/CaptionLayout) prop to `react-day-picker`  we can now remove our customisation that has been reproducing the same behaviour.


Lelith (approved):
Looks good to me!

PR #2548, opened 2023-07-21
chore: [F36-932] update sidebar links

cf-remylenoir:
# Purpose of PR

We want to move the list of Deprecated V3 Components down the list, as well as rename `Utils` to `Utilities`.

**Before**

https://github.com/contentful/forma-36/assets/103024358/632d344d-bace-406d-9333-a9ffed336619

**After**

https://github.com/contentful/forma-36/assets/103024358/73f4720e-531a-4ac6-825d-d1d10ece975f

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/website/components/SidebarSection.tsx
+++ b/packages/website/components/SidebarSection.tsx
@@ -98,6 +98,16 @@ export function SidebarSection({
   const { currentPage } = useCurrentLocation();
   const { data: session } = useSession();
 
+  // move the deprecated link section to the end of the components list
+  const deprecatedLinkSectionIndex = links.findIndex((link) =>
+    link.title.includes('Deprecated V3 Components'),
+  );
+
+  if (deprecatedLinkSectionIndex !== -1) {
+    const removedElement = links.splice(deprecatedLinkSectionIndex, 1);
+    links.push(removedElement[0]);
+  }
+
   // don't list auth protected sections in the sidebar if the user is not logged in.
   if (isAuthProtected && !session) {
     return null;
--- a/packages/website/components/SidebarSection.tsx
+++ b/packages/website/components/SidebarSection.tsx
@@ -98,16 +98,6 @@ export function SidebarSection({
   const { currentPage } = useCurrentLocation();
   const { data: session } = useSession();
 
-  // move the deprecated link section to the end of the components list
-  const deprecatedLinkSectionIndex = links.findIndex((link) =>
-    link.title.includes('Deprecated V3 Components'),
-  );
-
-  if (deprecatedLinkSectionIndex !== -1) {
-    const removedElement = links.splice(deprecatedLinkSectionIndex, 1);
-    links.push(removedElement[0]);
-  }
-
   // don't list auth protected sections in the sidebar if the user is not logged in.
   if (isAuthProtected && !session) {
     return null;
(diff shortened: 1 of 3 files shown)


packages/website/components/SidebarSection.tsx
```diff
@@ -98,6 +98,16 @@ export function SidebarSection({
   const { currentPage } = useCurrentLocation();
   const { data: session } = useSession();
 
+  // move the deprecated link section to the end of the components list
+  const deprecatedLinkSectionIndex = links.findIndex((link) =>
```
giotiskl, line 102: I am a bit ignorant about how the repo is setup but this filtering on the links feels like business logic that the `SidebarSection` should be agnostic about.

Where are the `links` being passed down as props? Can we do the filtering there and just pass down the links properly filtered? Also can we extract this logic into a function with a fitting name (e.g. `sortLinks`?) so that it's easier to read.

cf-remylenoir:
I could not find this information but, it makes sense to NOT push a changeset for such changes, oder?

packages/website/components/SidebarSection.tsx
cf-remylenoir: Well, you actually raise a good point, I could actually manually place it in the code below, no need to manipulate the links.

https://github.com/contentful/forma-36/blob/af819f3fe8be9f4ea012544b82feb8cbd99c9731/packages/website/pages/%5B...slug%5D.tsx#L166-L171

The only difference will be that it will not be a collapsable section, but its own expanded section, which seems fine to me.

Screenshot 2023-07-21 at 17 18 51

giotiskl: No strong opinions here as I'm lacking context, the others can plus 1 it, but it sounds fine to me.

mshaaban0:
> I could not find this information but, it makes sense to NOT push a changeset for such changes, oder?

You don't need to generate a changeset if it doesn't affect the packages. Website will be deployed both ways

cf-remylenoir:
> > I could not find this information but, it makes sense to NOT push a changeset for such changes, oder?
> 
> You don't need to generate a changeset if it doesn't affect the packages. Website will be deployed both ways

Makes sense, I removed the changeset 😉

packages/website/components/SidebarSection.tsx
```diff
@@ -98,6 +98,16 @@ export function SidebarSection({
...
+    link.title.includes('Deprecated V3 Components'),
+  );
+
+  if (deprecatedLinkSectionIndex !== -1) {
+    const removedElement = links.splice(deprecatedLinkSectionIndex, 1);
+    links.push(removedElement[0]);
+  }
```
mshaaban0, line 109: 💭 wouldn't the side bar look much better if the sections are collapsed but the one the current active component is in, expanded?

mshaaban0 (approved):

packages/website/components/SidebarSection.tsx
cf-remylenoir: I think the sidebar UX could be improved, the mix of collapsable/static elements, font weights, and indentation levels (subsection links) makes it hard to scan it IMO.

packages/website/components/SidebarSection.tsx
cf-remylenoir: @mshaaban0 what do you reckon on this one ☝️ ?

PR #2559, opened 2023-08-15
chore: [F36-889] full height button group divider

maxcheremisin:
# Purpose of PR

ButtonGroup: add full height divider to Primary and Positive: 1px #FFFFFF at 20% opacity;
Button: Remove Primary/Positive button border color

https://github.com/contentful/forma-36/assets/22265863/fb1e4ddf-a8c7-46d6-8077-a7a2dd7ad3ac

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/.changeset/grumpy-bags-complain.md
+++ b/.changeset/grumpy-bags-complain.md
@@ -1,0 +1,5 @@
+---
+'@contentful/f36-button': minor
+---
+
+chore: full height button group divider; remove positive/primary button border color
--- a/packages/components/button/src/Button/Button.styles.ts
+++ b/packages/components/button/src/Button/Button.styles.ts
@@ -7,11 +7,11 @@ import { hexToRGBA } from '@contentful/f36-utils';
 const variantActiveStyles = (variant: ButtonVariant): CSSObject => {
   switch (variant) {
     case 'primary':
-      return { backgroundColor: tokens.blue700, borderColor: tokens.blue700 };
+      return { backgroundColor: tokens.blue700 };
     case 'secondary':
       return { backgroundColor: tokens.gray200 };
     case 'positive':
-      return { backgroundColor: tokens.green700, borderColor: tokens.green700 };
+      return { backgroundColor: tokens.green700 };
     case 'negative':
       return { backgroundColor: tokens.gray200 };
     case 'transparent':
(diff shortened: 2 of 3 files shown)


packages/components/button/src/ButtonGroup/ButtonGroup.tsx
```diff
@@ -39,17 +39,24 @@ function _ButtonGroup(
...
-          key,
-          className: cx(
-            styles.groupContent,
-            (child as React.ReactElement).props.className,
-          ),
-        });
+        return (
+          <>
+            {withDivider && index !== 0 ? (
+              <div className={styles.divider} />
```
Lelith, line 49: ❗️To be honest I am not a fan of this approach. As it adds multiple html-elements only to provide a little line. 

I am happy to hear what the others think, but I would preferably use a pseudo element instead of adding new divs.

packages/components/button/src/ButtonGroup/ButtonGroup.tsx
```diff
@@ -39,17 +39,24 @@ function _ButtonGroup(
...
+            ) : null}
```
Lelith, line 50: 🧹 I suggest using the `&&` notion instead of a null return, but thats just a nitpick as i think rendering an additional html tag for every element is the bigger issue here.

Lelith (changes requested):
I don't think adding additional html tags for a little line is a good pattern. 
If you want to get rid of the property in the styles, you could go for class combination and a pseudo element e.g. `::before`

packages/components/button/src/ButtonGroup/ButtonGroup.tsx
maxcheremisin: i tried pseudo elements first, the problem is that button has `overflow: hidden` styles and the border, which doesnt allow to make the divider full height.
happy to try other approaches if you have any suggestions

maxcheremisin: image


i also tried overriding border-right/left, but as you can see button has border-top/bottom as well, what make the divider cut in corners


image

and i tried to remove the border completely, and the result is visually what we want, but it makes button 2px shorter in width then it used to be

maxcheremisin: hm, i think i can remove top/bottom border only, so it will keep the same with

maxcheremisin: updated (including PR description)

cf-remylenoir (changes requested):
The `negative` variant does not look right when passing the `withDivider` prop. Could you please have a look at it?

It would be worth adding it to the `ButtonGroup` "Overview" story as well, to make sure we automatically capture any regression in the future with Chromatic.

buttongroup-withdivider-negative

buttongroup-withdivider-negative--focus

maxcheremisin:
> The `negative` variant does not look right when passing the `withDivider` prop. Could you please have a look at it?
> 
> It would be worth adding it to the `ButtonGroup` "Overview" story as well, to make sure we automatically capture any regression in the future with Chromatic.
> 
> buttongroup-withdivider-negative
> 
> buttongroup-withdivider-negative--focus

thats the problem with API, `withDivider` is supposed to be used with primary/positive buttons.
we have a ticket for this https://contentful.atlassian.net/browse/F36-903

cf-remylenoir (approved):
Based on the Button component from our Figma library, it seems only the `secondary` and `negative` variants receive a border. I am happy with these changes, let's get the review from the designers.

Lelith (approved):
Thank you for putting up the fight to avoid the extra html tags. Looks good!

.changeset/grumpy-bags-complain.md
```diff
@@ -0,0 +1,5 @@
+---
+'@contentful/f36-button': minor
+---
+
+chore: [F36-889] full height button group divider; remove positive/primary button border color
```
cf-remylenoir, line 5:
```suggestion
chore: full height button group divider; remove positive/primary button border color
```

PR #2562, opened 2023-08-17
docs: update select examples

Lelith:
# Purpose of PR

Updates the Documentation for Select component, adding examples for different sizes as well as updating the the correct FormControl usage

### DIFF
--- a/packages/components/forms/examples/SelectWithValidationExample.tsx
+++ b/packages/components/forms/examples/SelectWithValidationExample.tsx
@@ -3,8 +3,8 @@ import { Select, FormControl } from '@contentful/f36-components';
 
 export default function SelectWithValidationExample() {
   return (
-    <FormControl as="fieldset">
-      <FormControl.Label as="legend">Label for select</FormControl.Label>
+    <FormControl>
+      <FormControl.Label>Label for select</FormControl.Label>
       <Select id="optionSelect" name="optionSelect">
         <Select.Option value="optionOne">Option 1</Select.Option>
         <Select.Option value="optionTwo">Long Option 2</Select.Option>
--- a/packages/components/forms/examples/SelectWithValidationExample.tsx
+++ b/packages/components/forms/examples/SelectWithValidationExample.tsx
@@ -3,8 +3,8 @@ import { Select, FormControl } from '@contentful/f36-components';
 
 export default function SelectWithValidationExample() {
   return (
-    <FormControl>
-      <FormControl.Label>Label for select</FormControl.Label>
+    <FormControl as="fieldset">
+      <FormControl.Label as="legend">Label for select</FormControl.Label>
       <Select id="optionSelect" name="optionSelect">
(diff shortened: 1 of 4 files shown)


packages/components/forms/examples/SelectWithValidationExample.tsx
```diff
@@ -3,8 +3,8 @@ import { Select, FormControl } from '@contentful/f36-components';
 
 export default function SelectWithValidationExample() {
   return (
-    <FormControl>
-      <FormControl.Label>Label for select</FormControl.Label>
+    <FormControl as="fieldset">
```
denkristoffer, line 6: Why do we want to use fieldsets and legends? This breaks the label behaviour.

cf-remylenoir: Hey Kristoffer, what is breaking exactly?

Lelith: i think the description for accessibility where all wrong in the first place 🫠 because if you add the things that where recommendet initially (using id, name, form control) they stop working / are no longer connected with each other.

denkristoffer: You can't click to focus the element.

cf-remylenoir: Interestingly it seems to be supported, see [this example](https://codesandbox.io/s/billowing-waterfall-pyjwkv?file=/index.html).

cf-remylenoir: Sorry, the context of my example wasn't correct. I get it now, looking around to material around fieldset/legend with select 👀

Lelith: So just because i am fully confused at this moment. Do understand it correclty: Id and name are only necessary for input elements when we do *not* use `FormControl` because it has it baked in.

What exactly was the intend behind Fieldset and Legend? Only for radio and checkbox groups?

denkristoffer: > What exactly was the intend behind Fieldset and Legend? Only for radio and checkbox groups?

@Lelith The way I see it those elements are used for grouping, but not to replace labels. I think the MDN example is good as it shows how to use `<legend>` and `<label>` together: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/legend

cf-remylenoir: The intent behind `<fieldset />` and `<legend />` is to group fields and offer a visual representation of that group, it also allows disabling all fields of the group if the attribute is passed. 

So `<select />` can also be included in a `<fieldset />`, but It might make more sense to use `<fieldset />` together with `<select />` if there are multiple fields, not when it is alone. In any case, a `<label />` should always be set with its field.

cf-remylenoir (approved):

PR #2565, opened 2023-08-24
fix(asset): set full width

denkristoffer:
# Purpose of PR

Sets `width: 100%` on Asset root. This is a bug we somehow never got fixed in F36 itself, I think no one ever made a ticket.

Tested in the web app.

### DIFF
--- a/.changeset/khaki-suits-perform.md
+++ b/.changeset/khaki-suits-perform.md
@@ -1,0 +1,5 @@
+---
+"@contentful/f36-asset": patch
+---
+
+fix(asset): set full width on Asset root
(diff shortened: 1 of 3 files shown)


massao (approved):
LGTM
Should we consider having a `fullWidth` prop on the Box as we have in Flex?? 🤔 
Or actually why only Flex has the `fullWidth` prop, and Box and Grid don't??

denkristoffer:
> LGTM Should we consider having a `fullWidth` prop on the Box as we have in Flex?? 🤔 Or actually why only Flex has the `fullWidth` prop, and Box and Grid don't??

Yeah that could make sense to me. I think consistency of our base components is not our strong point 😅

PR #2567, opened 2023-08-25
feat(Autocomplete): add controllable isOpen, inputValue props; extend text behaviour on item select

maxcheremisin:
# Purpose of PR

What's changed:

- ❌ `clearAfterSelect` (`boolean`) replaced with 🆕 `textOnAfterSelect` (`'clear' | 'preserve' | 'replace'`) to be able to preserve the text input value when item selected
- 🆕 `inputValue` prop added to be able to clear the text input programmatically (e.g. on menu close)
- 🆕 `isOpen` prop added (and corresponding `onOpen`, `onClose` handlers) to be able to control menu state and subscribe to its change

https://github.com/contentful/forma-36/assets/22265863/65807b83-0400-422c-9a54-33ce439be045

### DIFF
--- a/packages/components/autocomplete/src/Autocomplete.test.tsx
+++ b/packages/components/autocomplete/src/Autocomplete.test.tsx
@@ -97,6 +97,29 @@ describe('Autocomplete', () => {
       expect(mockOnSelectItem).toHaveBeenCalledWith('Apple 🍎');
     });
 
+    it('clears the input after item is selected when "clearAfterSelect" is true', async () => {
+      const user = userEvent.setup();
+
+      renderComponent({ clearAfterSelect: true });
+
+      // Type one letter in the input to open the list
+      await user.type(screen.getByRole('textbox'), 'a');
+
+      // checks if the list is visible
+      expect(screen.getByRole('listbox')).toBeVisible();
+
+      // go to the list first item
+      await user.keyboard('[ArrowDown]');
+
+      // press Enter to select the item
+      await user.keyboard('[Enter]');
+
+      // checks if the list got closed and the value of the input is an empty string
+      expect(screen.getByRole('listbox', { hidden: true })).not.toBeVisible();
+      expect(screen.getByRole('textbox')).toHaveValue('');
+      expect(mockOnSelectItem).toHaveBeenCalledWith('Apple 🍎');
+    });
+
     it('clears the input after item is selected when "textOnAfterSelect" is "clear"', async () => {
       const user = userEvent.setup();
 
--- a/packages/components/autocomplete/src/Autocomplete.test.tsx
+++ b/packages/components/autocomplete/src/Autocomplete.test.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
-import { render, screen, fireEvent, waitFor } from '@testing-library/react';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import { getStringMatch } from '@contentful/f36-utils';
 import { Autocomplete, AutocompleteProps } from './Autocomplete';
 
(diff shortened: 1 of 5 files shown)


packages/components/autocomplete/src/Autocomplete.tsx
```diff
@@ -296,11 +363,12 @@ function _Autocomplete<ItemType>(
...
               onFocus={(e) => {
                 onFocus?.(e as React.FocusEvent<HTMLInputElement>);
-                if (!isOpen) {
-                  toggleMenu();
-                }
+                openMenu();
+              }}
+              onBlur={(e) => {
+                onBlur?.(e as React.FocusEvent<HTMLInputElement>);
+                inputProps.onBlur(e);
```
maxcheremisin, line 370: was missing and didn't fire onBlur properly

packages/components/autocomplete/src/Autocomplete.tsx
```diff
@@ -249,12 +322,6 @@ function _Autocomplete<ItemType>(
           if (selectedItem) {
             onSelectItem(selectedItem);
           }
-          if (clearAfterSelect) {
```
maxcheremisin, line 252: moved to `onInputValueChange`

packages/components/autocomplete/src/Autocomplete.tsx
```diff
@@ -249,12 +322,6 @@ function _Autocomplete<ItemType>(
...
-            handleInputValueChange('');
-          }
```
maxcheremisin, line 254: moved to `onInputValueChange`

packages/components/autocomplete/src/Autocomplete.test.tsx
```diff
@@ -140,6 +140,40 @@ describe('Autocomplete', () => {
...
+
+      // go to the list first item
+      fireEvent.keyDown(input, {
+        key: 'ArrowDown',
+      });
+
+      // press Enter to select the item
+      fireEvent.keyDown(input, {
+        key: 'Enter',
+      });
```
Chaoste, line 169: [nitpick]

Considering that we try to replace usage of fireEvent with userEvent for more realistic input testing, as well as some other recommended testing styles (e.g. avoid test IDs but use realistic queries), I would rewrite it like this:

```suggestion
      await userEvent.type(screen.getByRole('textbox'), 'a');
      expect(await screen.findByRole('menuitem', { name: /apple/ }).toBeInTheDocument();
      await userEvent.type(screen.getByRole('textbox'), '{arrowdown}{enter}');
```

All of these changes are also further explained in [this blog post](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library) which I can highly recommend to understand why we shouldn't use test IDs, etc.

Chaoste (commented):
Looks good to me - just a nitpick on the testing style. I didn't approve as this should be down by the owning team.

packages/components/autocomplete/src/Autocomplete.test.tsx
maxcheremisin: i just copied one of the test cases and adjusted it for the new prop. but i refactored the test file now using user event and getByRole

Chaoste: Thank you! I know that many of our repositories are still in this old state. Just want to avoid that reproduce it - maybe we can get rid of these things by slowly refactoring it bit by bit.

denkristoffer (changes requested):
Prop changes like this are a breaking change. Since the `Autocomplete` is considered stable we will have to deprecate instead and remove in v5.

maxcheremisin:
> Prop changes like this are a breaking change. Since the `Autocomplete` is considered stable we will have to deprecate instead and remove in v5.

thanks! done

cf-remylenoir (commented):
It looks good, could you please add an example of using the new props in the documentation and storybook as per in the video? I was not able to reproduce it myself 🤔

maxcheremisin:
> It looks good, could you please add an example of using the new props in the documentation and storybook as per in the video? I was not able to reproduce it myself 🤔

curious what you think - there is an example for `textOnAfterSelect`, but i didn't add more examples of `inputValue` and `isOpen` to the docs because the use case is pretty niche and it might overload the docs with too much information.

but i defo can add more stories to the storybook

maxcheremisin:
> > It looks good, could you please add an example of using the new props in the documentation and storybook as per in the video? I was not able to reproduce it myself 🤔
> 
> curious what you think - there is an example for `textOnAfterSelect`, but i didn't add more examples of `inputValue` and `isOpen` to the docs because the use case is pretty niche and it might overload the docs with too much information.
> 
> but i defo can add more stories to the storybook

extended `ControlledFromOutside` story with more props

cf-remylenoir (approved):
Looking good 👍 I [noted](https://contentful.atlassian.net/wiki/spaces/PROD/pages/4295426148/Forma+36+-+v5+delivery) down the deprecation for the next major release.

denkristoffer (approved):

PR #2568, opened 2023-08-29
docs: show deprecated prop badge

maxcheremisin:
# Purpose of PR

Add a clear way to mark props as deprecated in the docs.

Example: https://forma-36-git-feat-docsdeprecated.colorfuldemo.com/components/autocomplete#props-api-reference:~:text=the%20root%20element-,clearAfterSelect

Before:

image

After:

image

image

### DIFF
--- a/packages/f36-docs-utils/src/PropsTable/PropsTable.tsx
+++ b/packages/f36-docs-utils/src/PropsTable/PropsTable.tsx
@@ -5,6 +5,7 @@ import {
   Subheading,
   Text,
   Table,
+  Tooltip,
 } from '@contentful/f36-components';
 
 import { getPropsTableStyles } from './PropsTable.styles';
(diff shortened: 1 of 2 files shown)


Lelith (approved):

packages/f36-docs-utils/src/PropsTable/PropsTable.tsx
cf-remylenoir, line 1: This looks great! 

The way I understand it, if a prop has the `@deprecated` decorator without content, the badge won't be shown at all, is that correct?

maxcheremisin: right, currently it works that way. but we can also show the badge with no tooltip if you prefer

https://github.com/contentful/forma-36/assets/22265863/ad51a5f3-9dc1-45af-b3bc-96177bd65b11

cf-remylenoir: Unless we set a linting rule to prevent contributors from not adding a deprecation guideline, I'd say so 🙂

maxcheremisin: done

cf-remylenoir (approved):
Thank you for offering this improvement 👌

PR #2576, opened 2023-09-11
feat(IconButton): Adjust IconButton size. Deprecate `size="large"` and `children`

maxcheremisin:
# Purpose of PR

- adjust IconButton size (Small: 32 x 32 px, Medium: 40 x 40 px)
- deprecated implicit inherited props from base Button component (children and size="large")

Before:
image

After:
image


damann:
Screenshot 2023-09-11 at 13 56 57
@maxcheremisin In the preview the two sizes are the same. Otherwise this is good to go from my point of view.

maxcheremisin:
> Screenshot 2023-09-11 at 13 56 57
> @maxcheremisin In the preview the two sizes are the same. Otherwise this is good to go from my point of view.

I didn't touch this part. The Icon is smaller on the second row. But I changed this example to preview the small IconButton (not only the Icon itself)

image

Lelith:
Can i still produce large buttons with icons? or is this considered a breaking change?

massao:
> Can i still produce large buttons with icons? or is this considered a breaking change?

Since it only picks the `size` from the Base Button type, I thing it is possible to pass `large` as size, and the style returns an empty {}, so it won't override anything from the button, and it pass the size down to the button too, so it should get the `large` size.

Yeah, just tried it on the storybook of this branch: https://5fd1dda724cc620021ace8c5-jktwsdgcgl.chromatic.com/?path=/story/components-button-iconbutton--basic&args=variant:secondary;size:large

maxcheremisin:
> Can i still produce large buttons with icons? or is this considered a breaking change?

Gonna be a breaking change in the next major release. For now it works the same way as @massao mentioned.

cf-remylenoir (approved):
Looking good 👍

PR #2577, opened 2023-09-13
feat: new field type icons

denkristoffer:
New icons have two different size variants with different SVG paths. This means we cannot easily manipulate the SVG with CSS, but will instead have to serve different code. Here's a proposal for how we can tackle this.


packages/components/icon/src/Icon.tsx
```diff
@@ -32,59 +33,39 @@ export type IconVariant =
...
-  variant?: IconVariant;
-  /**
-   * Custom SVG viewBox attribute to use
-   */
-  viewBox?: SVGAttributes<SVGSVGElement>['viewBox'];
-};
+type NewOrOldIconProps =
+  | {
+      // Render the icon with new sizes and styling. Not for general use.
+      isNewIcon: true;
```
bgutsol, line 39: let's not use the `new` as a version differentiator, since it's too ambiguous. Let's use something more specific like `V2` for clarity and future-proofing the codebase for subsequent updates.

denkristoffer: Fair point but as a counter I will say that these icons are not v2 or similar, for now they will live side by side with the others. However, the prop is meant to be completely removed at some point, and it should not be used by the public.

denkristoffer: I'm very open to suggestions for different names though! 😄

mshaaban0:
The following packages has been published manually as under beta tag:

```json
"@contentful/f36-icons": "4.28.0-beta.0"
"@contentful/f36-icon": "4.51.0-beta.0"
```

cf-remylenoir:
Hey @denkristoffer, thank you for creating this proposal. I would be happy to discuss it further with you and about the new icons delivery strategy.

denkristoffer:
@cf-remylenoir I believe we're dropping the two-variant approach so we can recreate the new icons with the existing way of doing things.

PR #2578, opened 2023-09-15
feat(Tooltip): support ReactElement as content

miguelcrespo:
# Purpose of PR

Support `ReactElement` as Content for `Tooltips`

## PR Checklist

- [X] Tests are passing

### DIFF
--- a/packages/components/avatar/src/Avatar/Avatar.tsx
+++ b/packages/components/avatar/src/Avatar/Avatar.tsx
@@ -3,7 +3,12 @@ import { cx } from 'emotion';
 
 import { type CommonProps } from '@contentful/f36-core';
 import { Image, type ImageProps } from '@contentful/f36-image';
-import { Tooltip, type TooltipProps } from '@contentful/f36-tooltip';
+import {
+  Tooltip,
+  type TooltipProps,
+  type TooltipInternalProps,
+  type WithEnhancedContent,
+} from '@contentful/f36-tooltip';
 
 import { convertSizeToPixels, getAvatarStyles } from './Avatar.styles';
 import type { ColorVariant } from './utils';
@@ -20,7 +25,9 @@ export interface AvatarProps extends CommonProps {
   /**
    * A tooltipProps attribute used to conditionally render the tooltip around root element
    */
-  tooltipProps?: Omit<TooltipProps, 'children'>;
+  tooltipProps?: CommonProps &
+    WithEnhancedContent &
+    Omit<TooltipInternalProps, 'children'>;
   variant?: Variant;
   colorVariant?: ColorVariant;
   icon?: React.ReactElement;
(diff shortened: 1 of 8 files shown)


packages/components/tooltip/stories/Tooltip.stories.tsx
```diff
@@ -212,3 +213,28 @@ export const WithDelays = () => {
...
+
+export const WithReactElement = () => {
+  return (
+    <>
+      <SectionHeading as="h3" marginBottom="spacingS">
+        With React Elements as children
+      </SectionHeading>
+
+      <Flex marginBottom="spacingS">
+        <Tooltip
```
Lelith, line 225: 💡 I think the example would be better without the delay and isVisible properties

packages/components/tooltip/stories/Tooltip.stories.tsx
```diff
@@ -212,3 +213,28 @@ export const WithDelays = () => {
...
+          isVisible
+          hideDelay={60000}
+          label="Render a paragrah with a React Element"
+          content={
+            <Flex>
+              <p className={css({ marginBottom: 0 })}>I am a React Element</p>
```
Lelith, line 231: I think it would be better if we change the example a bit to feature like a headline with a paragraph, to already set a good example when it can be used.
```suggestion
              <div><SectionHeading>Example Heading</SectionHeading><p>I am a React Element</p></div>
```

packages/components/tooltip/README.mdx
```diff
@@ -50,3 +50,5 @@ import { Tooltip } from '@contentful/f36-tooltip';
   <TextLink>Hover me</TextLink>
 </Tooltip>
 ```
+
+- When using `ReactElement` as the content, it's recommended to use the `ScreenReaderOnly` component when displaying critical information.
```
Lelith, line 54: This paragraph should be part of the "Content Guidlines" area on the element and extended
```suggestion
The Tooltip component allows you to pass React elements as content. However, this should be used with care. React.Element as content is best suited for text formatting purposes. It should not be used for interactive elements like links, buttons, or form elements. Additionally, extensive images should be avoided, except for text-decorator icons that give semantic meaning to the text shown in the tooltip (e.g., warning signs or other relevant icons).

```

Also extend the Accessibility part with something like:

Screen readers cannot access the content of a tooltip, so crucial information should not be placed exclusively within a tooltip. To improve accessibility use the <ScreenreaderOnly> component that contains the same text as the tooltip.

Lelith (changes requested):
We need to change the documentation a bit more. Also it would be good if you could directly add an Example in the Avatar Group with the react.element as content.
Also add an additional example to the tooltip,example files

packages/components/avatar/src/Avatar/Avatar.tsx
```diff
@@ -20,7 +25,9 @@ export interface AvatarProps extends CommonProps {
   /**
    * A tooltipProps attribute used to conditionally render the tooltip around root element
    */
-  tooltipProps?: Omit<TooltipProps, 'children'>;
+  tooltipProps?: CommonProps &
+    WithEnhancedContent &
+    Omit<TooltipInternalProps, 'children'>;
```
Lelith, line 29: @denkristoffer 🥷 thank you sensei for teaching me advanced TS

Lelith (approved):

denkristoffer:
@miguelcrespo Don't forget the changeset.

PR #2584, opened 2023-09-28
fix(TextLink): vertical alignment when used within flexbox container

maxcheremisin:
# Purpose of PR

Fix TextLink vertical alignment when used within flexbox container.

## Problem

When used within a flexbox container TextLink stretches to full height. It used to have `style="align-items: center"` so when it's stretched it still would look aligned. But with recent changes in  https://github.com/contentful/forma-36/pull/2574 it was changed to `style="align-items: baseline"` to align it when used within inline text elements.

With this PR I'd like to return previous behaviour back (when TextLink is stretched to full height, and its elements are centrally aligned), but at the same time keep text and icon aligned to baseline when used within other text.

🚨 But to make it work I had to add an extra DOM element.

image

<details>
  <summary>Bug screenshots</summary>

 
  image
 
  image
 
  image
  
</details>

### DIFF
--- a/.changeset/heavy-geese-lick.md
+++ b/.changeset/heavy-geese-lick.md
@@ -1,0 +1,5 @@
+---
+'@contentful/f36-text-link': patch
+---
+
+fix(TextLink): vertical alignment when used within flexbox container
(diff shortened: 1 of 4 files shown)


maxcheremisin:
Worth mentioning that I had another idea to add `style="align-self: center"` to TextLink container, it would fix the central alignment, but it would make impossible to align the TextLink element to start or end.

ghepting (approved):
lgtm! will hopefully avoid regressions in consumers, thanks for catching :)

massao (approved):

PR #2589, opened 2023-10-13
chore(Button): update stories

maxcheremisin:
# Purpose of PR

Update Button components stories.


damann (changes requested):
**Button**
- Button size variants: Would you mind removing the "(default)" on the medium sizes?
- Please break out the button with really long content (truncated) into its own section, because it is not a size variant.
**Button group**
- Variant of merged button group without divider is missing from the overview
- "Collapsed" property should be removed
Screenshot 2023-10-13 at 13 12 37

cf-remylenoir (approved):
Besides Daniel's feedback, it looks good to me 👍

maxcheremisin:
@damann, sorry for the long delay 😬

> Button size variants: Would you mind removing the "(default)" on the medium sizes?

✅ done

> Please break out the button with really long content (truncated) into its own section, because it is not a size variant.
Button group

✅ done

> Variant of merged button group without divider is missing from the overview

❓ doesn't it conflict with this [ticket](https://contentful.atlassian.net/browse/F36-903) ("Buttons in merged group should always have dividers.")?
i think we shouldn't show this variant if we gonna remove it in future, so consumers dont rely on this behaviour now

> "Collapsed" property should be removed

⚠️ I removed it from the select dropdown, but i can't remove it from typings cause we still support it (it will be gone automatically with the next major release)

damann (approved):

PR #2600, opened 2023-10-27
fix: simplify ButtonGroup child type

bgutsol:
# Purpose of PR

- simplify `ButtonGroup` child prop type and `BaseCard` badge prop type

### DIFF
--- a/.changeset/chilly-baboons-sip.md
+++ b/.changeset/chilly-baboons-sip.md
@@ -1,0 +1,5 @@
+---
+'@contentful/f36-button': minor
+---
+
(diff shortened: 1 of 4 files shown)


bgutsol:
> Good spot! It seems we have a similar case for the [BaseCard](https://github.com/contentful/forma-36/blob/fa7a33e5c15f7dd45ab4bb4aa636fbc08a59f376/packages/components/card/src/BaseCard/BaseCard.types.ts#L43) component.

Good one, I'm gonna change it there as well 👍

PR #2605, opened 2023-11-08
fix: IconButton size type

bgutsol:
- Fix `IconButton` size prop marked as deprecated. Leave it only as comment for size `large`.

CleanShot 2023-11-08 at 15 51 40@2x

Related PR: 
https://github.com/contentful/forma-36/pull/2576

### DIFF
--- a/.changeset/khaki-cows-divide.md
+++ b/.changeset/khaki-cows-divide.md
@@ -1,0 +1,5 @@
+---
+'@contentful/f36-button': patch
+---
+
+Fix IconButton size prop type.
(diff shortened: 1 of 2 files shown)


maxcheremisin (approved):

cf-remylenoir (approved):
Nice catch 👍

maxcheremisin:
there is a similar comment for entityType prop, wonder why it works there
image

cf-remylenoir:
> there is a similar comment for entityType prop, wonder why it works there image

My hunch is that because there is some more text after the `@deprecated` decorator it is not assigning it to the whole property

PR #2610, opened 2023-11-14
fix(story): fix bug in Pill story where user cannot change draggable …

lilbitner:
# Purpose of PR

This PR fixes a bug found in the Pill `basic` and `In small container `story that prevents a user from changing the component from its draggable state to its default state and vise versa. 

Essentially the prop, `isDraggable`, was not being passed effectively to the story component. Worth noting that if the props are spread, the `onDrag` prop is always true as an `actionHandler`, so it will always render the story with drag features. 

Originally this was noticed within the `basic` story, but I also fixed the logic within the `In small container` story  as well. This might not be necessary, but I figured the user might want to view the small container example with or without the drag handler - I can certainly remove that adjustment 👍 
chrome-capture-2023-10-15-pill-fix

## PR Checklist

- [X] Tests are passing

### DIFF
--- a/packages/components/pill/stories/Pill.stories.tsx
+++ b/packages/components/pill/stories/Pill.stories.tsx
@@ -215,7 +215,7 @@ export const InSmallContainer: Story<PillInternalProps> = (args) => {
         <Flex flexDirection="row" marginBottom="spacingM">
           <Box marginRight="spacingXs">
             <Pill
-              draggable={args.isDraggable}
+              isDraggable={args.isDraggable}
               label={args.label}
               className={styles.pill}
               onDrag={onDrag}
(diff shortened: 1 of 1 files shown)


packages/components/pill/stories/Pill.stories.tsx
```diff
@@ -205,6 +205,7 @@ export const InSmallContainer: Story<PillInternalProps> = (args) => {
       maxWidth: 200,
     }),
   };
+  const onDrag = args.isDraggable ? args.onDrag : undefined;
```
lilbitner, line 208: Only pass the `onDrag` callback if the `isDraggable` argument has been adjusted within the story controls. (otherwise the `onDrag` is always truthy)

damann (approved):

damann:
@lilbitner, I've given my approval, but we'll still always need an engineer's sign-off since I can't evaluate any potential code issues. Thanks!

Lelith (approved):
Looks good to me!

PR #2619, opened 2023-11-28
refactor: add gap instead of hardcoded margin right [CFCX-835]

massao:
# Purpose of PR

Replace hardcoded margin-right with gap, so when used without a children it won't have a spacing to the right of the input


bgutsol (approved):
This might require manually checking WebApp and exp-packages if we use custom labels there and add that missing gap while bumping forma.

massao:
> This might require manually checking WebApp and exp-packages if we use custom labels there and add that missing gap while bumping forma.

yeah, I'm taking a look to see if we have any places, and will open PRs if we have

PR #2618, opened 2023-11-27
feat(DensityContainer): add package

cf-remylenoir:
# Purpose of PR

## PR Checklist

- [ ] I have read the relevant `readme.md` file(s)
- [ ] All commits follow our [Git commit message convention](https://github.com/contentful/forma-36/tree/main/packages/forma-36-react-components#commits)
- [ ] Tests are added/updated/not required
- [ ] Tests are passing
- [ ] Storybook stories are added/updated/not required
- [ ] Usage notes are added/updated/not required
- [ ] Has been tested based on [Contentful's browser support](https://www.contentful.com/faq/about-contentful/#which-browsers-does-contentful-support)
- [ ] Doesn't contain any sensitive information

PR #2620, opened 2023-11-30
feat: add density context []

massao:
# Purpose of PR

- Add density context provider to `f36-utils` package
- Consume density context on `f36-buttons` package

This should unblock: https://github.com/contentful/forma-36/pull/2618


cf-remylenoir (approved):
Happy to get feedback from other folks

obama-medal

cf-remylenoir:
I realized I forgot to add the documentation for the `useDensity` hook/util, will add it in a follow-up.

PR #2636, opened 2023-12-12
style(formcontrol): adjust spacing and placement of validation message

lilbitner:
## Purpose

The purpose of this PR is to fix the alignment, spacing, and order of the Validation messages with respect to their accompanying input component within a `FormControl` context. 

Originally I implemented a solution that would render the `ValidationMessage` at the level of the `BaseCheckbox` to align the validation message automatically with the Checkbox label, as opposed to it's alignment now existing with the great container. Based on PR comments and further Slack discussion with @massao, there are inconsistencies with this approach, given the already established pattern of rendering `FormControl.ValidationMessage` at the level of the input. This established pattern gives the user great flexibility with where they place the validation message i.e at the bottom of several inputs in a `CheckBox` group or after a help text.  It also makes it a bit complicated for how we can make this alignment automated. It seems that the fact that the `ValidationMessage` is not aligned comes with the territory of this flexibility, and the user is still able to accomplish this alignment via a `marginLeft` prop passed to `ValidationMessage`. 

Example of validation message being awkwardly aligned if we were to programmatically ensure this alignment with Checkbox group inputs: 
image

[Example](https://f36.contentful.com/playground?code=N4Igxg9gJgpiBcICWBbADhATgFwAQCUYBDMPAM0whVwHJNjSaBuAHQDtUMddhcAxLCgDCENtkoAbADS4hACxhgA1gCMIAD1wBfXBSq0AApDEwxZAK4SA9GQDMANgC0kdKNPYAzs3bsY6rniwZESW5OZspEiiuACi6kToEjAAFACUPOy4uPTY5phsuMmZWbgAPAKYwqLiEBK4RB4AvCwgZEgwElAeMNgtAHzFJWUVVWKSAHQAMkQqHfVNLUkA5qZQ-fKKqhqlViMiY7VTMx0DBUMlpRvKauqD51kAbkQS5jDNIBBo2FEFAIwtd3uSCg7zACmuGkcn2+okc-xAgJKp3uQwA8l8frh-md7jsrlt1MiUeVBPsahJxgAJDpoAAqfjwKCImCWSDYkxgZGw7w8aBIbKWkxauCZLLZtM+7zYbn61IkaFw2AZuiwuGhmN+Oz21QmcrpDKJuO1BwpADVnsCiDC2ABZGAeDxEFYi5ms9mc7ktXn8tiC4Wit0StBSmUgPoxTCUTAi+2OlZa0k6w7miSW612h1OmCGsq7RMmompVhsLTsEBaIA) that the user is still able to accomplish this alignment via a `marginLeft` prompt. 

Therefore I have implemented this `marginLeft` prop as discussed ^. The current `FormControl` story looks as so: 

Screenshot 2023-12-21 at 1 43 55 PM

Another part of this PR is to ensure there is consistency between `FormControl` components when it comes to rendering the help text and validation message, at least in our code examples. The help text should appear above the validation message, as per [these](https://www.figma.com/file/Pd9pnDy181xBiZRbXYK4hr/F36-1039-%E2%80%93-Form-components?node-id=1%3A774&mode=dev) designs. I only found this inconsistency within the below documentation example. I also adjusted this spacing to resemble the designs. 

### Before:
Screenshot 2023-12-12 at 3 20 02 PM

### After:
Screenshot 2023-12-12 at 3 26 14 PM

### Follow-up work:

- The validation icon seems like a pixel or two off with alignment, but I believe there is already a similar [ticket](https://contentful.atlassian.net/browse/F36-919) to adjust this within the backlog.

- Additionally, there is a [ticket](https://contentful.atlassian.net/browse/F36-1083) to implement the new error/warning icon itself, as the designs differ from the result here.

## PR Checklist

- [X] Tests are passing

### DIFF
--- a/.changeset/bright-brooms-join.md
+++ b/.changeset/bright-brooms-join.md
@@ -1,0 +1,5 @@
+---
+'@contentful/f36-forms': minor
+---
+
+Reduce top margin spacing of ValidationMessage component
--- a/packages/components/forms/examples/FormControlInvalidExample.tsx
+++ b/packages/components/forms/examples/FormControlInvalidExample.tsx
@@ -6,8 +6,8 @@ export default function FormControlInvalidExample() {
     <FormControl isInvalid>
       <FormControl.Label>Name</FormControl.Label>
       <TextInput />
-      <FormControl.ValidationMessage>Error</FormControl.ValidationMessage>
       <FormControl.HelpText>Please enter your first name</FormControl.HelpText>
+      <FormControl.ValidationMessage>Error</FormControl.ValidationMessage>
     </FormControl>
   );
 }
--- a/packages/components/forms/src/BaseCheckbox/BaseCheckbox.styles.ts
+++ b/packages/components/forms/src/BaseCheckbox/BaseCheckbox.styles.ts
@@ -16,11 +16,7 @@ const sizeToStyle = (size) => {
   };
 };
 
-const getSupplementalTextStyle = ({
-  size,
-  type,
-  isValidationMessage = false,
-}) => {
+const getSupplementalTextStyle = ({ size, type }) => {
   let inputWidth = tokens.spacingM;
   if (type === 'switch') {
     inputWidth = sizeToStyle(size).width;
@@ -28,7 +24,7 @@ const getSupplementalTextStyle = ({
 
   return {
     marginLeft: `calc(${inputWidth} + ${tokens.spacingXs})`,
-    marginTop: isValidationMessage ? tokens.spacing2Xs : 0,
+    marginTop: 0,
   };
 };
 
@@ -72,9 +68,6 @@ const getStyles = ({
     type === 'switch' && sizeToStyle(size),
   ]),
   helpText: css(getSupplementalTextStyle({ size, type })),
-  inlineValidationMessage: css(
-    getSupplementalTextStyle({ size, type, isValidationMessage: true }),
-  ),
 });
 
 export default getStyles;
--- a/packages/components/forms/src/BaseCheckbox/BaseCheckbox.styles.ts
+++ b/packages/components/forms/src/BaseCheckbox/BaseCheckbox.styles.ts
@@ -16,7 +16,7 @@ const sizeToStyle = (size) => {
   };
 };
 
-const getSupplementalTextStyle = ({ size, type }) => {
+const getHelpTextStyle = ({ size, type }) => {
   let inputWidth = tokens.spacingM;
   if (type === 'switch') {
     inputWidth = sizeToStyle(size).width;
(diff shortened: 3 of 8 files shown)


ghepting (approved):

ghepting:
I approved, but I'm not the design authority! Definitely recommend getting a forma-36 core contributor's eyes/input too. :)

denkristoffer:
I'm not sure I'm keen on this API where we can now pass validation errors in multiple ways. Is it safe to say that the message should always be indented for checkboxes and radios? Maybe we can handle it automatically in these cases? 🤔

massao:
I would rather have the inline prop as a boolean, and render the correct validation message styled based on that, rather than having two props accepting a validation message

lilbitner:
> I'm not sure I'm keen on this API where we can now pass validation errors in multiple ways. Is it safe to say that the message should always be indented for checkboxes and radios? Maybe we can handle it automatically in these cases? 🤔

Sure, I was wavering between just using the original `isInvalid` boolean to render this inline message, but opted to disrupt as little as possible. Happy to remove the additional prop and just render the inline message per the validation state.

lilbitner:
> I would rather have the inline prop as a boolean, and render the correct validation message styled based on that, rather than having two props accepting a validation message

Ah gotcha, maybe I am missing a prop that you might be referring to, but it seems like we would need an explicit prop that accepts the string validation message? @massao

lilbitner:
> I'm not sure I'm keen on this API where we can now pass validation errors in multiple ways. Is it safe to say that the message should always be indented for checkboxes and radios? Maybe we can handle it automatically in these cases? 🤔

Yeah I was definitely trying to disrupt as little as possible here, but I do think it makes sense if we just always render the validation message inline for this component, if the state `isInvalid`. I think the user will still need to pass an explicit validation message through the `inlineValidationMessage` prop (can rename to `validationMessage`)? unless I am missing another prop that already exists here for that @denkristoffer

massao:
> > I would rather have the inline prop as a boolean, and render the correct validation message styled based on that, rather than having two props accepting a validation message
> 
> Ah gotcha, maybe I am missing a prop that you might be referring to, but it seems like we would need an explicit prop that accepts the string validation message? @massao

yeah, I didn't explain it well at all, sorry about it.
What I meant is that it would be better to have a prop that just marks that the validation message to be rendered with correct style based on that prop, but keep using the FormControl one. wouldn't be work on just one component though. Probably would need to add the value on the context and check it on the ValidationMessage component itself 🤔 

But since this is just for styling purposes it would be a better way to do it, for consistency rather than introducing another place for a validation message, also what would happen if there was a `FormControl.ValidationMessage`, and the some other string on this new prop?? 
Also this could be confusing to users as, why is is that only this form element has a prop for validation message and such

lilbitner:
Have discussed with @massao via slack, and as per his suggestions, I believe it makes sense to revert the adjustments I made, and rely on the user to implement the alignment when using the `ValidationMessage` component via the `marginLeft` prop. See PR description for more details.

massao (approved):

massao:
Looks good to me, just make sure to add a new changeset to ensure that the new ValidationMessage version gets released: https://github.com/contentful/forma-36/blob/main/RELEASES.md#adding-changesets

denkristoffer (approved):

PR #2637, opened 2023-12-14
feat: icons API

denkristoffer:
Draft PR to facilitate discussion around how we want to build icons going forward.

Discuss here: https://contentful.atlassian.net/wiki/spaces/PROD/pages/4418830338/Forma+36+v5+icons+proposal

### DIFF
--- a/packages/components/icons/src/v5/ArrowCounterClockwise.tsx
+++ b/packages/components/icons/src/v5/ArrowCounterClockwise.tsx
@@ -1,0 +1,5 @@
+import { ArrowCounterClockwise as PhosphorIcon } from '@phosphor-icons/react';
+
+import { generateForma36Icon } from '../../../icon/src/generateIcon';
+
+export const ArrowCounterClockwiseIcon = generateForma36Icon(PhosphorIcon);
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ You will find each package’s instructions in their README files, check the [Pa
 
 ### Storybook for f36-components
 
-We use storybook with our react component library to develop components. You can start it from the root of the repo, just run `npm run-script storybook`
+We use storybook with our react component library to develop components. You can start it from the root of the repo, just run `npm run-script storybook` after building.
 
 ## Commits & releases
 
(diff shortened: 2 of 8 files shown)


packages/components/icons/src/v5/ArrowCounterClockwise.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+const ArrowCounterClockwise = wrapPhosphorIcon(PhosphorIcon);
+
+export const ArrowCounterClockwiseIcon = generateForma36Icon({
+  variants: {
+    [IconVariant.Active]: (props) => (
+      <ArrowCounterClockwise {...props} weight="fill" />
+    ),
+    [IconVariant.Default]: ArrowCounterClockwise,
+  },
```
cf-remylenoir, line 18: It would be great if we could abstract the variant logic within the generator to make our contributors' lives easier and less prone to errors.

Following the [design](https://www.figma.com/file/ILoICAP1fxyvE7xmpWiJBi/Forma-36-Assets?type=design&node-id=1699-22&mode=design&t=1LzNdfhx1togJCsI-0), we would always have the same set of icons (3 sizes, 2 variants (default/active)).

If it turns out to be too tricky, we could create a script to generate the icon automatically given the Phosphor name, wdyt?

denkristoffer: Yeah I hear you and will give it some thought. To be able to solve multiple use cases where we want to render different SVGs based on a component prop, I'm thinking we might need to expose a function like this, where variants are set explicitly, and another level of abstraction that achieves what you are suggesting.

denkristoffer: I added another level of abstraction.

cf-remylenoir: Thank you, this is much cleaner! ✨ 

I can see the addition of a script to automatically generate the icon files/changes based on the icon(s) contributors want to add. 

I will review it thoroughly when I am back, but feel free to go ahead if other folks review & approve 👍

Lelith (approved):
I really do like the approach, the code looks clean and straight forward for me. It also looks like it is easy to extend the variations, esp after the abstraction layer you added.

to me it looks like a good approach to move forward.

denkristoffer:
Closing for #2647 – please go there 😄

PR #2639, opened 2023-12-20
chore: update registry to github packages [CFCX-857]

massao:
# Purpose of PR

Migrate our publish config to github packages registry.

Follow up for: https://github.com/contentful/forma-36/pull/2640

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,8 +16,8 @@ jobs:
       - vault/get-secrets:
           template-preset: packages-read
       - run: |
-        echo "//npm.pkg.github.com/:_authToken=${GITHUB_PACKAGES_READ_TOKEN}" > ~/.npmrc
-        echo "@contentful:registry=https://npm.pkg.github.com" >> ~/.npmrc
+          echo "//npm.pkg.github.com/:_authToken=${GITHUB_PACKAGES_READ_TOKEN}" > ~/.npmrc
+          echo "@contentful:registry=https://npm.pkg.github.com" >> ~/.npmrc
       - attach_workspace:
           at: ~/circleci-f36-build
       - node/install-packages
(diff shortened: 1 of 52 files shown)


scripts/prerelease.mjs
andipaetzold, line 1: This is only reformatting?

.circleci/config.yml
```diff
@@ -13,6 +13,11 @@ jobs:
     working_directory: ~/circleci-f36-build
     steps:
       - checkout
+      - vault/get-secrets:
+          template-preset: packages-read
+      - run: |
+          echo "//npm.pkg.github.com/:_authToken=${GITHUB_PACKAGES_READ_TOKEN}" > ~/.npmrc
+          echo "@contentful:registry=https://npm.pkg.github.com" >> ~/.npmrc
```
andipaetzold, line 20: Do we need this? We never read from our private registry.

scripts/prerelease.mjs
massao: Mostly reformatting, but there also added the `dryRun` to be able to just test without doing changes

.circleci/config.yml
massao: Without this, it was failing to install the browserlist:
https://app.circleci.com/pipelines/github/contentful/forma-36/11276/workflows/94cfd9cb-ea24-4742-a631-3d8f116eb414/jobs/31390

Probably because all our packages on github are marked as internal visibility

scripts/prerelease.mjs
```diff
@@ -1,86 +1,144 @@
...
+          '@contentful/',
+          '',
+        )}`
+      : [registry, name].join('/');
 
-  // Avoid chageset publishing it, by adding the package to the ignore list
-  await ignorePackage(name)
+  if (!dryRun) {
+    await updatePackageJson(path.resolve(dir, 'package.json'), newVersion);
+    // Avoid chageset publishing it, by adding the package to the ignore list
```
andipaetzold, line 102:
```suggestion
    // Avoid changeset publishing it, by adding the package to the ignore list
```

andipaetzold (approved):

PR #2644, opened 2024-01-08
feat(baseCheckbox): adds density support for Checkbox, Radio components.

stephanLeece:
# Purpose of PR

This PR adds density support to allow for decreased font-size and line height when the component is wrapped in a density container. Adds examples for the Checkbox and Radio, to the DensityContainer storybook stories

Ticket: https://contentful.atlassian.net/browse/CFCX-956

## PR Checklist

- [x] Tests are passing


packages/components/forms/src/BaseCheckbox/types.ts
```diff
@@ -48,4 +48,5 @@ export interface BaseCheckboxInternalProps
    */
   helpText?: string;
   onChange?: ChangeEventHandler<HTMLInputElement>;
+  density?: 'low' | 'high';
```
cf-remylenoir, line 51: You should be able to replace this with the Density type.

https://github.com/contentful/forma-36/blob/408cd9aadf665d06f5b06856b6cea7ffa6d4c4e8/packages/components/utils/src/useDensity/useDensity.ts?plain=1#L3

packages/components/forms/src/BaseCheckbox/BaseCheckbox.styles.ts
```diff
@@ -25,14 +25,20 @@ const getHelpTextStyle = ({ size, type }) => {
   return {
     marginLeft: `calc(${inputWidth} + ${tokens.spacingXs})`,
     marginTop: 0,
+    fontSize: density === 'high' && tokens.fontSizeS,
+    lineHeight: density === 'high' && tokens.lineHeightS,
```
cf-remylenoir, line 29: We could define the default (low) value here too, based on the design spec, as you did for the `BaseCheckbox` Text component.

stephanLeece: good shout, couldn't hurt eh?

stephanLeece: done :)

cf-remylenoir:
@damann happy to hear your feedback on the addition of the default values (based on your [design spec](https://www.figma.com/file/IZfaouOHnhiX8OVbVAAPzx/F36-911-High-density?type=design&node-id=116-1560&mode=design&t=zmvNxxorR1XlYSCD-0)). 

The Chromatic changes show how much our baselines changed from this update. 
I am happy to approve them, but I would love to get your approval too.

Previously, no default values were assigned to the elements and they were inherited from the global values (i.e. line-height-default': '1.5').

cf-remylenoir (commented):
Looks good to me, could you please add a changeset too ([see](https://github.com/contentful/forma-36/pull/2644#issuecomment-1881582344)), since the components' API was affected?

Happy to approve after :)

stephanLeece:
> Looks good to me, could you please add a changeset too ([see](https://github.com/contentful/forma-36/pull/2644#issuecomment-1881582344)), since the components' API was affected?
> 
> Happy to approve after :)

Thanks! Added as a patch, based on example here https://github.com/contentful/forma-36/pull/2620/files as they are of similar scope

.changeset/wild-wombats-change.md
```diff
@@ -0,0 +1,5 @@
+---
+'@contentful/f36-forms': patch
+---
+
+Consume density context on Checkbox, Radio components
```
cf-remylenoir, line 5: I'd love to make the message more explicit, what about: 

```suggestion
Add high-density support to Checkbox and Radio components.
```

cf-remylenoir (approved):
Thanks! Once the Chromatic baseline snapshots are accepted by @damann / myself, we can proceed with the merge 👍

.changeset/wild-wombats-change.md
cf-remylenoir: Ah maybe the revert put the previous message back 👀

stephanLeece: Good shout, updated!

damann (changes requested):
It appears like the distance between checkboxes/radio buttons in the group is smaller than in the design specs.
Screenshot 2024-01-10 at 15 25 37

stephanLeece:
> It appears like the distance between checkboxes/radio buttons in the group is smaller than in the design specs. Screenshot 2024-01-10 at 15 25 37

Is that something we should resolve with design tweaks to the stack component? 

As i've only made changes to the font-size and line-height, i'm not sure how we could fix this with changes to the individual checkbox/radio components...

stephanLeece:
@damann based on storybook, it seems like they are set to have a 0.5rem vertical gap?

Screenshot 2024-01-10 at 15 34 53

stephanLeece:
Okidoki, i've narrowed down the spacing / line height issue. 

Screenshot 2024-01-10 at 15 58 19

`lineHeightCondensed` corresponds to 1.25, which differs from the current default line height. 

`lineHeightM` corresponds to 1.25rem, which should match the current line height. 

I'll go ahead and update in my code.

packages/components/forms/package.json
```diff
@@ -7,9 +7,11 @@
   },
   "dependencies": {
     "@contentful/f36-core": "^4.58.0",
+    "@contentful/f36-density-container": "^4.0.0-alpha.3",
```
cf-remylenoir, line 10: I was too late to recommend installing this as dev-dependency instead since it is not required for the package to be functional?

PR #2650, opened 2024-01-10
feat(formControl): adds density support for Form control elements

stephanLeece:
# Purpose of PR

This PR adds density support to the FormControl Element.

Fort size, line height, icon size and spacing will be reduced when the component is wrapped in a density container. 

Adds example FormControl story

Updates Stories that import `DensityContainer` to use relative import father than from package, to fix changeset generation errors 

Ticket: https://contentful.atlassian.net/browse/CFCX-957

## PR Checklist

- [x] Tests are passing


stephanLeece:
@damann UI changes are here: https://www.chromatic.com/review?appId=5fd1dda724cc620021ace8c5&number=2650&type=linked

Note that the Horizontal margin for the Validation message has increased to 8px for the low-density version, and 4px for the high. Previously it was at 4px as the default.

package-lock.json
```diff
@@ -11,6 +11,9 @@
         "packages/*",
         "packages/components/*"
       ],
+      "dependencies": {
+        "@contentful/f36-density-container": "^4.0.0-alpha.3"
+      },
```
cf-remylenoir, line 16: Is this intentional?

cf-remylenoir (commented):
One comment regarding the new dependency in the lock file, besides that it is all good 👍  I will let @damann review the snapshots.

package-lock.json
stephanLeece: Ah good shout, I'll need to regenerate the lock file. Will do :)

stephanLeece: updated, refefences to `"@contentful/f36-density-container"` which were added in previous PR, are now removed, since we are using relative import now

cf-remylenoir (approved):
Nice one! ✨

damann (approved):

PR #2653, opened 2024-01-15
feat(forms): add high-density support to Select

cf-remylenoir:
# Purpose of PR

Enables high-density support for the `Select` component (medium size only).

Ticket: https://contentful.atlassian.net/browse/CFCX-958

Docs: https://forma-36-git-feat-select-high-density-support.colorfuldemo.com/components/select#density-support
Storybook: https://5fd1dda724cc620021ace8c5-znbenjvdei.chromatic.com/?path=/story/form-elements-select--with-density-support

with-ellipsis

## PR Checklist

- [x] Tests are passing


cf-remylenoir:
@damann the text gets slightly cut with the current `height/padding` values (see the bottom of the `p`).

To keep the `height` in control (32px), I suggest going one more level down with the `padding`.

| Value | Outcome |
|--------|--------|
| Top/Bottom: spacingXs (8px) | xs |
| Top/Bottom: spacing2Xs (2px) | 2xs |

packages/components/forms/src/Select/Select.styles.ts
```diff
@@ -1,16 +1,29 @@
...
           height: '32px',
+          fontSize: tokens.fontSizeM,
+          lineHeight: tokens.lineHeightM,
+          borderRadius: tokens.borderRadiusMedium,
         }
       : {
-          padding: `10px ${tokens.spacingL} 10px ${tokens.spacingS}`,
-          height: '40px',
+          padding: isHighDensity
+            ? `${tokens.spacingXs} ${tokens.spacingL} ${tokens.spacingXs} ${tokens.spacingXs}`
```
Lelith, line 17: I am a little concerned about the amount of inline ifs for this one variable. Would it not be easier to read and maintain to switch the whole class or write a "getStyles" sort of function? So we would have one if and avoid touching each and every single property

packages/components/forms/src/Select/Select.styles.ts
```diff
@@ -28,6 +43,7 @@ export function getSelectStyles({ isInvalid, isDisabled, size }) {
     outline: 'none',
     border: `1px solid ${tokens.gray300}`,
     cursor: 'pointer',
+    textOverflow: 'ellipsis',
```
cf-remylenoir, line 46: The `text-overflow` handling was missing, resulting in cut text when too long

| Before | After |
|--------|--------|
|  no-ellipsis | with-ellipsis |

packages/components/forms/src/Select/Select.styles.ts
cf-remylenoir: It makes sense, adjusted 👍

damann (approved):

cf-remylenoir:
@Lelith please let me know if you are satisfied with the changes 🙂

PR #2654, opened 2024-01-17
feat: update multiselect [HOMER-2649]

Lelith:
# Purpose of PR
- extends the multiselect component with a new reference prop for controlling the clear search field button.
- adds examples for how to use the references
- fixes smaller design issues e.g. text-color, text-alignment

image

image


packages/components/multiselect/src/Multiselect.tsx
```diff
@@ -72,6 +72,11 @@ export interface MultiselectProps extends CommonProps {
    */
   toggleRef?: React.Ref<HTMLButtonElement>;
 
+  /**
+   * Use this prop to get a ref to the reset search button of the component
+   */
+  resetSearchRef?: React.Ref<HTMLButtonElement>;
```
jfctfl, line 78: Why are you using a ref here instead of passing a callback `onSearchReset` that is called (when defined) in the onClick handler `resetSearchInput ` of the `IconButton`?

packages/components/multiselect/examples/MultiselectReferenceExample.tsx
```diff
@@ -0,0 +1,76 @@
...
+    );
+    setFilteredItems(newFilteredItems);
+  };
+
+  const handleSelectItem = (event) => {
+    const { checked, value } = event.target;
+    if (checked) {
+      setSelectedItems((prevState) => [...prevState, value]);
+    } else {
+      const newSelectedFruits = selectedItems.filter(
```
mdmjg, line 33: [nit] why are we calling them fruits? There's nothing about fruits in this file

mdmjg (changes requested):

packages/components/multiselect/src/Multiselect.tsx
Lelith: Resetting the Search Field is not super trivial (its a react 16+ issue, because you can no longer access the native value of a text input https://stackoverflow.com/a/46012210/17269164) , this component already has this function properly implemented and so it would be a waste of time to write the exact same code to reset that field every time you want to have the search input field
https://github.com/contentful/forma-36/blob/main/packages/components/multiselect/src/Multiselect.tsx#L185

So just offering a ref that calls this function is the proper way to go. We also do this already for the toggle menu.

packages/components/multiselect/examples/MultiselectReferenceExample.tsx
Lelith: ah, sorry, copy paste issue 😅

packages/components/multiselect/src/Multiselect.tsx
Lelith: Will try to see if i can refactor it to use an imperative ref to only expose the clear functin, thanks :)

Lelith: I do not think this works, because this component also exposes other refs for other parts of it (toggling the list e.g.) and imperativeHandle only works when the component is a forwardRef component. It does not really fit our use-case here unfortunately :/

jfctfl: But the MultiSelect is wrapped in a `forwardRef` or what are you referring to?

Lelith: That forwardRef is already used as a reference to the outer popover component. So i can not setup a second forwardRef for that clearSearch Button.

Lelith: What i could do is refactor all refs to be one imperative forwardRef, which would be a breaking change at this point in time.

mdmjg (approved):

packages/components/multiselect/src/Multiselect.tsx
Lelith: @denkristoffer @mshaaban0 @cf-remylenoir what do you think about the usage of impertative Refs for big components like this one?

So far the component is not that widely used, so its a good idea to do any breaking changes while its still considered alpha?

cf-remylenoir: Considering this is an alpha component I reckon it is okay to change its API if necessary 👍  

Ensuring we notify the [consumers](https://github.com/search?q=org%3Acontentful+%3CMultiselect+-repo%3Acontentful%2Fforma-36++NOT+is%3Aarchived&type=code) and follow-up to update the usage accordingly.

Lelith: I spoke with Kristoffer and we both think its better to take a change to the way we approach refs into a bigger picture so that forma components approach refs always the same instead of using an imperative ref approach this one time.

What i will do is to group the props better.

denkristoffer:
If you move these props into the new object that's also a breaking change, right?

miguelcrespo (approved):

Lelith:
@denkristoffer i refactored the component and separated the whole search part from the component in its own sub component. I added a new sub-component prop `searchProps` to the component and marked the existing offered props as deprecated. In this way we can introduce the change in a non breaking way.

denkristoffer (approved):
Thank you 🙇

PR #2659, opened 2024-01-24
feat: add high density for textinput [CFCX-958]

massao:
Add support for high density for text inputs

### DIFF
--- a/packages/components/forms/src/BaseInput/BaseInput.styles.ts
+++ b/packages/components/forms/src/BaseInput/BaseInput.styles.ts
@@ -80,8 +80,8 @@ const getStyles = ({
     },
     high: {
       borderRadius: tokens.borderRadiusSmall,
-      lineHeight: tokens.lineHeightS,
-      fontSize: tokens.fontSizeS,
+      lineHeight: tokens.lineHeightMHigh,
+      fontSize: tokens.fontSizeMHigh,
     },
   };
 
--- a/packages/components/forms/src/BaseInput/BaseInput.styles.ts
+++ b/packages/components/forms/src/BaseInput/BaseInput.styles.ts
@@ -1,19 +1,27 @@
 import { css } from 'emotion';
 import tokens from '@contentful/f36-tokens';
 import type { CSSObject } from '@emotion/serialize';
+import type { Density } from '@contentful/f36-utils';
+import { BaseInputInternalProps } from './types';
 
-const getSizeStyles = ({ size }): CSSObject => {
+type getSizeStylesProps = Pick<BaseInputInternalProps, 'size'> & {
+  density?: Density;
+};
+
+const getSizeStyles = ({ size, density }: getSizeStylesProps): CSSObject => {
+  const isHighDensity = density === 'high';
   if (size === 'small') {
     return {
-      padding: `${tokens.spacing2Xs} ${tokens.spacingXs}`,
-      height: '32px',
-      maxHeight: '32px',
+      padding: tokens.spacingXs,
+      minHeight: isHighDensity ? '24px' : '32px',
+      maxHeight: isHighDensity ? '24px' : '32px',
     };
   }
 
   return {
-    height: '40px',
-    maxHeight: '40px',
+    padding: isHighDensity ? tokens.spacingXs : `10px ${tokens.spacingS}`,
+    minHeight: isHighDensity ? '32px' : '40px',
+    maxHeight: isHighDensity ? '32px' : '40px',
   };
 };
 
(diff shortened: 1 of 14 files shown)


packages/components/forms/src/BaseInput/BaseInput.styles.ts
```diff
@@ -1,100 +1,138 @@
...
+
+      '&::placeholder': {
+        color: tokens.gray500,
+      },
 
-  inputWithIcon: css({
-    paddingLeft: size === 'small' ? tokens.spacingXl : '38px',
-  }),
+      '&:active, &:active:hover': {
+        borderColor: isInvalid
```
Lelith, line 95: nitpick: i think with this nested inline ifs, the styles here become harder to read and understand. Maybe it would be good to use a helper method for returning invalid and disabled styles?

packages/components/forms/src/TextInput/input-group/InputGroup.tsx
```diff
@@ -22,7 +23,9 @@ const _InputGroup = (
   ref: React.Ref<HTMLDivElement>,
 ) => {
   const { children, className, spacing = 'none', ...otherProps } = props;
-  const styles = getStyles({ spacing });
+  const density = useDensity();
+  console.log({ density });
```
cf-remylenoir, line 27:
```suggestion
```

packages/components/forms/stories/InputGroup.stories.tsx
cf-remylenoir, line 1: I need to have a deeper look at it but it seems the high-density button styling differs from what it is supposed to be (padding).

Screenshot 2024-01-26 at 15 01 38

Screenshot of when the Button high-density story was still there (it used to be in the DensityContainer story but the component has been removed, I was sure to have moved into its own Button story 🙈)
Screenshot 2024-01-26 at 15 01 48

cf-remylenoir: Oh, could it be the flex wrapping behavior?

massao: uhmm I think this is related to what I commented about the components having some conflicting styles, removing the minHeight from some components seems to solve that, will try it out and see if that doesn't break anything somewhere else 🤔

massao: Last commit should have fixed it, it was an issue with the buttons having wrong minHeight, and some components were depending on the minHeight of the buttons, like the copyButton, which caused them to work weirdly when in the Input Group, and trying to fix them with a override on the InputGroup only caused more problems 🤷 
Downside is that the user needs to specify correctly the size of all elements in the InputGroup

packages/components/forms/src/BaseInput/BaseInput.styles.ts
```diff
@@ -1,100 +1,140 @@
...
-        ? tokens.gray300
-        : tokens.blue600,
-      boxShadow: isInvalid
-        ? tokens.glowNegative
-        : isDisabled
-        ? 'none'
-        : tokens.glowPrimary,
+    high: {
+      borderRadius: tokens.borderRadiusSmall,
+      lineHeight: tokens.lineHeightS,
```
cf-remylenoir, line 83: We now have typography [tokens](https://forma-36-git-feat-high-density-typography-scale.colorfuldemo.com/tokens/typography) for high densities:

```suggestion
      lineHeight: tokens.lineHeightMHigh,
      fontSize: tokens.fontSizeMHigh
```

packages/components/forms/src/BaseInput/BaseInput.styles.ts
```diff
@@ -1,100 +1,140 @@
...
+      fontSize: tokens.fontSizeS,
```
cf-remylenoir, line 84: We now have typography [tokens](https://forma-36-git-feat-high-density-typography-scale.colorfuldemo.com/tokens/typography) for high densities:

```suggestion
      lineHeight: tokens.lineHeightMHigh,
      fontSize: tokens.fontSizeMHigh
```

packages/components/forms/stories/InputGroup.stories.tsx
cf-remylenoir: Thanks!

> Downside is that the user needs to specify correctly the size of all elements in the InputGroup 

I am not sure what to follow here, based on your changes I don't see why that would be necessary.

packages/components/forms/src/BaseInput/BaseInput.styles.ts
```diff
@@ -1,100 +1,140 @@
...
+const getSizeStyles = ({ size, density }: getSizeStylesProps): CSSObject => {
+  const isHighDensity = density === 'high';
   if (size === 'small') {
     return {
-      padding: `${tokens.spacing2Xs} ${tokens.spacingXs}`,
-      height: '32px',
-      maxHeight: '32px',
+      padding: tokens.spacingXs,
+      minHeight: isHighDensity ? tokens.spacingL : tokens.spacingXl,
+      maxHeight: isHighDensity ? tokens.spacingL : tokens.spacingXl,
```
cf-remylenoir, line 17: I can see in Storybook the computed height of the `small` TextInput is `24px` whereas in the design spec it is `28px`. But here, with a height of `24px` (`spacingL`) + a vertical padding of `8px` (`spacingXs`), it should also equal `28px`.

packages/components/forms/stories/InputGroup.stories.tsx
massao: > I am not sure what to follow here, based on your changes I don't see why that would be necessary.

I meant as if the user wants to they could have a button small with a input as medium, or vice-versa, since we don't 
force the height from the InputGroup or anything like that.

packages/components/forms/src/BaseInput/BaseInput.styles.ts
massao: Good point I didn't see that the rendered on the design for `small` was `28px`, since all the others had the same height as the minHeight I assumed this was the case, or maybe this is a mistake on the design @damann ?? 🤔

cf-remylenoir: Good point regarding the `box-sizing` 😉

damann (approved):

cf-remylenoir (approved):
Thank you ✨

PR #2661, opened 2024-01-30
fix(components): fix rendering glitch on Safari 17.x

z0al:
# Purpose of PR

Fix a glitch on Safari 17.x where the view gets corrupted on scroll.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/.changeset/eighty-lobsters-cheat.md
+++ b/.changeset/eighty-lobsters-cheat.md
@@ -1,0 +1,5 @@
+---
+'@contentful/f36-workbench': patch
+---
+
+- fix rendering glitch on Safari 17.x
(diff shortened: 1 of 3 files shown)


mshaaban0 (approved):

z0al:
Looks like this caused a visual regression in some of the stories e.g. 

Screenshot 2024-01-30 at 16 55 07

I'm unsure how the change is relevant to those stories.

cf-remylenoir:
> Looks like ~this caused a visual regression in some of the stories~ actually these regressions are _already in [main](https://f36-storybook.contentful.com/?path=/story/components-autocomplete--controlled-from-outside)_ e.g.
> 
> Screenshot 2024-01-30 at 16 55 07
> Since the UI Test step is required, is it Ok if I skip it in this case @forma-36 ?

We will follow up on the snapshots, they aren't related to your changes and this issue is impacting our users ability to work with Contentful. Let's merge this.

z0al:
Sounds, good. could you please merge it @cf-remylenoir ? I don't have admin rights to force-merge.

PR #2664, opened 2024-02-01
fix: remove maxHeight from buttons []

massao:
# Purpose of PR

Fixes issue when using buttons with inputs and button being kept in a smaller size


cf-remylenoir:
Let's get this one over the finish line 🏁

cf-remylenoir (approved):

PR #2667, opened 2024-02-06
chore: fix secret name on changelog job []

massao:
# Purpose of PR

The `semantic-release` template of the vault doesn't have the `READ` secret, only the `WRITE`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -149,7 +149,7 @@ jobs:
       - vault/get-secrets:
           template-preset: semantic-release
       - run: |
-          echo "//npm.pkg.github.com/:_authToken=${GITHUB_PACKAGES_READ_TOKEN}" > ~/.npmrc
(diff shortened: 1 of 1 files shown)


cf-remylenoir (approved):

.circleci/config.yml
```diff
@@ -149,7 +149,7 @@ jobs:
       - vault/get-secrets:
           template-preset: semantic-release
```
andipaetzold, line 150: Wouldn't it be better to change this to `packages-read` and keep the env name?
If we don't need to write, we should use the token with fewer privileges

giotiskl (approved):

.circleci/config.yml
massao: I think this would not work in this case, because we update the Changelog files and we push it back to github, so we need some tokens that don't exist on the `packages-read`, the `GIT_AUTHOR_EMAIL` and `GIT_AUTHOR_NAME`

andipaetzold: ah - i thought we only read. my bad

andipaetzold (approved):

PR #2676, opened 2024-02-15
feat: extend SelectAllOption props to enable setting individual itemIds

Lelith:
# Purpose of PR
 This will allow for the MultiSelect SubComponent SelectAllOption to handover an individual itemId to further customize this components experience.


cf-remylenoir (approved):
It looks good to me. I was curious to see if the API documentation was reflecting how to use this and realized [we don't have it rendered yet](https://forma-36-git-feat-multiselectselectallitemidasprop.colorfuldemo.com/components/multiselect#props-api-reference). We should keep this in mind for when the component exits the `alpha`.

Lelith:
@cf-remylenoir what do you mean with we do not have it rendered? there is an example here: 
https://forma-36-git-feat-multiselectselectallitemidasprop.colorfuldemo.com/components/multiselect#selectall-option 
and one in the storybook: https://f36-storybook.contentful.com/?path=/story/components-multiselect--with-select-all

.changeset/swift-shrimps-sneeze.md
```diff
@@ -0,0 +1,5 @@
+---
+'@contentful/f36-multiselect': patch
+---
+
+Allows now setting an individual itemId for the SelectAll sellect option
```
Lelith, line 5:
```suggestion
Allows now setting an individual itemId for the SelectAllOption component which is a divert from MultiselectOption
```

cf-remylenoir:
> @cf-remylenoir what do you mean with we do not have it rendered? there is an example here: https://forma-36-git-feat-multiselectselectallitemidasprop.colorfuldemo.com/components/multiselect#selectall-option and one in the storybook: https://f36-storybook.contentful.com/?path=/story/components-multiselect--with-select-all

Thanks @Lelith, I meant the actual API prop table 🕵️

Lelith:
@cf-remylenoir ah i think the prop table might have issues with sub-components as SelectAll is only a compound Component and its not a prop on the Multiselect Component itself.

PR #2677, opened 2024-02-16
feat(multi-select): forward popover onClose

jfctfl:
MultiSelect: Forwards the `onClose` Popover-prop inside the MultiSelect component ensuring it's called.


bgutsol (approved):
lgtm

PR #2680, opened 2024-02-23
fix: drop transform to UTC when creting machine readable datetime [SPA-1849]

Chaoste:
# Purpose of PR

In the UI, we sometimes see mismatching times being rendered if you're not in the UTC timezone. That is because some datetime utils transform the time to the UTC timezone (`formatMachineReadableDateTime`) while others don't (`formatDateAndTime`). When we use both in the UI (e.g. for rendering scheduled actions), we get mismatching times from two different timezones (current vs. UTC).

I started this PR by first re-creating our use case to proof that this function is the root cause. Afterward, I adjusted the function and confirmed that the tests were successful.

## PR Checklist

- [x] Tests are passing


Lelith (approved):
Looks good to me!

cf-remylenoir:
I forgot to tell you to add the changeset 🙈

PR #2686, opened 2024-02-28
feat(modal): updates Modal.Header to accept child component

stephanLeece:
# Purpose of PR
We have a current opportunity that requires displaying a stepper component within a modal header:

Screenshot 2024-02-28 at 16 02 53

To enable this, we wanted to update the Forma36 component to take a new prop, `children`.

## PR Checklist

- [x] Tests are passing


packages/components/modal/src/ModalHeader/ModalHeader.tsx
```diff
@@ -48,6 +50,7 @@ export const ModalHeader = ({
           </Text>
         )}
       </Subheading>
+      {children}
```
cf-remylenoir, line 53: Do we need to account for any styling/alignment? What should be the default behavior (vertically centered with the rest)?

stephanLeece: The parent has justify-content: space between, so it should have equal spacing between the title/subtitle & close button. That won't necessarily be centred if the title is particularly long. 

I considered applying a max width to other elements or refactoring to a grid perhaps, but figured that would have potentially breaking changes for existing usage. I reckon this is a reasonable trade-off, WDYT?

cf-remylenoir (approved):

PR #2694, opened 2024-04-03
feat(entry card): entry card supports custom action button

msieroslawska:
# Purpose of PR

This PR enables consumers of Forma36 to provide their own React component as a replacement for the default `...` action menu.
If the custom component is provided, it takes precedence over default action menu. Otherwise the card renders as before with support for actions provided as Entry Card props.

**Default action menu:**
entry-card-action-menu

**Custom component:**
entry-card-custom-button

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/components/card/src/BaseCard/DefaultCardHeader.tsx
+++ b/packages/components/card/src/BaseCard/DefaultCardHeader.tsx
@@ -48,15 +48,52 @@ export const stopEvents = (e: React.MouseEvent<HTMLElement>) => {
 export const DefaultCardHeader = (
   props: Pick<
     BaseCardInternalProps,
-    'type' | 'icon' | 'badge' | 'actions' | 'actionsButtonProps'
+    | 'type'
+    | 'icon'
+    | 'badge'
+    | 'actions'
+    | 'actionsButtonProps'
+    | 'customActionButton'
   >,
 ) => {
-  const { icon, type, actions, actionsButtonProps, badge } = props;
+  const { icon, type, actions, actionsButtonProps, badge, customActionButton } =
+    props;
   const styles = getHeaderStyles();
+
+  const renderActionButton = () => {
+    if (customActionButton) {
+      return (
+        <Flex
+          // don't propagate click event, so onClick handler on the card is not triggered
+          onClick={stopEvents}
+          alignItems="center"
+        >
+          {customActionButton}
+        </Flex>
+      );
+    }
+    if (actions && actions.length > 0) {
+      return (
+        <Flex
+          // don't propagate click event, so onClick handler on the card is not triggered
+          onClick={stopEvents}
+          alignItems="center"
+        >
+          <CardActions buttonProps={actionsButtonProps}>{actions}</CardActions>
+        </Flex>
+      );
+    }
+
+    return null;
+  };
+
   return (
     <Flex
       flexWrap="wrap"
-      className={cx(styles.header, actions && styles.headerWithActions)}
+      className={cx(
+        styles.header,
+        (actions || customActionButton) && styles.headerWithActions,
+      )}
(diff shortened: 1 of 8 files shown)


cf-remylenoir:
Thank you for your contribution, Marta!

packages/components/card/src/BaseCard/DefaultCardHeader.tsx
cf-remylenoir, line 1: _Adding a comment here for easier thread discussion._

 Changes look good to me, the only things I noticed and would want @damann 's feedback on are the following:
 
 **Color**
- The default actions icon color is: #111B2B
- Custom actions icon color is: #5A657C

**Padding/size**
- We already had the entry card header badge/actions button misalignment/varying in sizes
- Having a custom actions icon button exacerbates it, depending on the icon size/trim

Should we address this to improve spacing harmony/consistency? 

It doesn't have to be in the PR, since it would change the scope of this initial work.

| Default | Custom |
|--------|--------|
| default-actions-icon | custom-actions-icon |

denkristoffer (commented):
Shouldn't forwarding `actionsButtonProps` be enough here? Is there a reason you cannot use it for this purpose?

msieroslawska:
> Shouldn't forwarding `actionsButtonProps` be enough here? Is there a reason you cannot use it for this purpose?

Depends how we look at this - do we want to allow the consumer of Forma to define their own button component (how I approached this) or do we provide a scaffolding where button is defined and we only ask for the icon and the click handler.

However, I cannot simply start to rely on existence of `actionsButtonProps` (if provided: render custom button; if not, render default `...` action button). Current implementation uses this prop to pass down extra button props for the action menu in the default `...` case, changing this logic could possibly break any customers' code that is providing this prop at the moment.
And I cannot reuse the current implementation because we do not want the action menu at all (in our case, we want to surface the option to delete the card immediately with one click).

packages/components/card/src/BaseCard/DefaultCardHeader.tsx
msieroslawska: Is the comment about color crossed out on purpose?

Re: padding
a) I do not see the size difference between Default and Custom 🙈 
b) If the Forma consumer provides a completely custom button component with whatever icon they use inside of it, would we even have control over size and padding of the icon on our side? Technically they can pass _any_ component and it's out of our control if it has an icon or what styling is used inside.
c) While I am all in into solving consistency issues, I would strongly advocate here to have this as a separate follow up to unblock my main sprint work.

msieroslawska:
> My understanding here is that you can use the button props to change both the icon and the click handler, which should let you avoid rendering the actions menu. Would that not work?

Sadly, no. Passing `icon` as part of `actionsButtonProps` would override the icon, but everything else is [hardcoded in the component](https://github.com/contentful/forma-36/blob/c4402360922f452d48ec7ff3059db1a913c1fae7/packages/components/card/src/BaseCard/CardActions.tsx) to be rendered as a menu.

denkristoffer:
I've been playing around with it and you're right, it's not easily doable at the moment. Nevertheless this approach seems the least opinionated to me so it's one I would say we should work towards in the future, perhaps with the menu trigger not being there by default.

I think what you've done in this PR is good for non-breaking change 👍

denkristoffer (approved):

packages/components/card/src/BaseCard/DefaultCardHeader.tsx
cf-remylenoir: Yes, I will revisit these comments with @damann when he is back, you can proceed with the merge 👍

cf-remylenoir (approved):

cf-remylenoir:
@all-contributors please add @msieroslawska for code

PR #2706, opened 2024-04-09
feat: deprecate Icon Trimmed Variants

Lelith:
# Purpose of PR

Deprecate the trimmed Icon variant, they will be removed with the next Iteration over Icons.

image

### DIFF
--- a/packages/components/icon/examples/AllIconsExample.tsx
+++ b/packages/components/icon/examples/AllIconsExample.tsx
@@ -11,8 +11,6 @@ import {
 } from '@contentful/f36-components';
 
 export default function AllIconsExample() {
-  console.log(f36icons);
-
   const [deprecatedIcons, availableIcons] = Object.entries(f36icons).reduce(
     (result, [name, icon]) => {
       result[name.includes('Trimmed') ? 0 : 1].push([name, icon]);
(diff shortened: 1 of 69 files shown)


packages/components/icon/examples/AllIconsExample.tsx
```diff
@@ -1,29 +1,72 @@
...
+  Box,
+} from '@contentful/f36-components';
 
 export default function AllIconsExample() {
-  return (
-    <Grid columns="repeat(2, 1fr)">
-      {Object.entries(f36icons).map(([name, icon]) => {
-        const Component = icon;
+  console.log(f36icons);
+
```
denkristoffer, line 15:
```suggestion
```

packages/components/icon/src/Icon.tsx
```diff
@@ -74,6 +74,7 @@ export type IconInternalProps = CommonProps & {
   size?: IconSize;
   /**
    * Whether or not to trim the icon width, i.e. set `width` to `auto`
+   * @deprecated IconsV5 will no longer support the trimmed option.
```
denkristoffer, line 77:
```suggestion
   * @deprecated Trimmed icons will be removed in a future major release
```

I think we should be less specific here, haha 😅

denkristoffer (commented):
Looks good to me, I wonder if we should discuss this console warning a bit more before merging it. It would bring a lot of noise to our own products, but I'm not against having it per se. It might be too early to bring in such a warning though?

Lelith:
@denkristoffer yeah, i think the warning could become something like a "phase 2" and probaply only a short time period after we did our first migration shot but when other projects might still be using the old ones.

damann (approved):

denkristoffer (approved):

PR #2708, opened 2024-04-11
feat: add onClick prop to breadcrumb [CFCX-1433]

massao:
# Purpose of PR

We want to be able to pass custom onClick to breadcrumb to be able to use tracking and similars.


denkristoffer (commented):
Don't forget the changeset.

massao:
> Don't forget the changeset.

This is still in alpha, adding changeset ignores alpha and beta currently, something that we need to take a look into as well

denkristoffer (approved):
Well in that case!

miguelcrespo (approved):
Would be cool to update the documentation to show this new possibility:
https://f36.contentful.com/components/header 
https://github.com/contentful/forma-36/blob/40700191b96113a2dc1b7b57197daf5632afbef6/packages/components/header/examples/WithBreadcrumbs.tsx#L18

Lelith (approved):
Looks good to me!

PR #2709, opened 2024-04-11
feat(autocomplete): add prop to control clear button visibility

veu:
# Purpose of PR

In some cases the button to clear the filter input may not be needed and, because it shares its position with the "open list" button, it can hide the "open list" button when needed. Especially when the `isReadOnly` prop is set to true there is little reason to reset the filter input and doing so hides the selected item even when `selectedItem` is passed.

To avoid breaking existing behavior this PR introduces a new property called `showClearButton` which allows to control whether the button is shown from the outside. It can be used to hide it completely or to fine-tune its behavior which is especially useful in combination with `selectedItem` and `inputValue` props.

### showClearButton unset

image
image

### showClearButton = false

image
image

### showClearButton = true

image
image

## PR Checklist

- [x] Tests are passing


Lelith (approved):
Looks good to me! Don't forget to add the changeset  via `npx changeset` before merging. :)

cf-remylenoir (approved):

PR #2647, opened 2024-01-10
feat: v5 icons

denkristoffer:
# Purpose of PR

## PR Checklist

- [ ] I have read the relevant `readme.md` file(s)
- [ ] All commits follow our [Git commit message convention](https://github.com/contentful/forma-36/tree/main/packages/forma-36-react-components#commits)
- [ ] Tests are added/updated/not required
- [ ] Tests are passing
- [ ] Storybook stories are added/updated/not required
- [ ] Usage notes are added/updated/not required
- [ ] Has been tested based on [Contentful's browser support](https://www.contentful.com/faq/about-contentful/#which-browsers-does-contentful-support)
- [ ] Doesn't contain any sensitive information

PR #2716, opened 2024-04-17
feat:(codemod) add v5-icons codemod

massao:
Add options to run v5-alpha codemods.

Also remove v4 codemods


denkristoffer:
Initial look seems good to me, I'll try to play around with it in a bit. I wonder if it's worth trying to support v4 as well though, should we just remove it? Less code would make it easier to maintain, maybe we could even move to `tsc` 🤔

massao:
> Initial look seems good to me, I'll try to play around with it in a bit. I wonder if it's worth trying to support v4 as well though, should we just remove it? Less code would make it easier to maintain, maybe we could even move to `tsc` 🤔

Yeah, was thinking on that too, but since we discussed of having tsc in a separate PR, was thinking in doing all the removing and such in a separate branch

packages/forma-36-codemod/transforms/v5/icons.js
```diff
@@ -0,0 +1,147 @@
...
+const iconsMap = {
+  ArrowBackward: 'ArrowLeft',
+  ArrowDown: 'CaretDownFill',
+  ArrowDownward: 'ArrowDown',
+  ArrowForward: 'ArrowRight',
+  ArrowUp: 'CaretUpFill',
+  ArrowUpward: 'ArrowUp',
+  Audio: 'FileAudio',
+  Calendar: 'CalendarBlank',
+  ChatBubble: 'ChatFill',
```
cf-remylenoir, line 18: We might need to double-check that list when we are ready, for instance, this v5 icon is not part of https://github.com/contentful/forma-36/pull/2647 yet (we use `Chat` 🐱 ).

denkristoffer:
Great work! There are a few behaviours I'm wondering if we can fix. First one is that I'm seeing a lot of added parentheses here, is this something we can avoid touching? I think these would get removed again when we ran prettier, so I'd rather avoid them in the first place.

Second one is how the codemods insist on creating new blocks of imports. Can we replace the old import in place instead?

Screenshot 2024-04-18 at 16 19 04

packages/forma-36-codemod/transforms/v5/icons.js
denkristoffer: I imagine we will be updating this mapping a few times no matter what. We need to press design for a final list though.

massao:
> Great work! There are a few behaviours I'm wondering if we can fix. First one is that I'm seeing a lot of added parentheses here, is this something we can avoid touching? I think these would get removed again when we ran prettier, so I'd rather avoid them in the first place.
> 
> Second one is how the codemods insist on creating new blocks of imports. Can we replace the old import in place instead?
> 
> Screenshot 2024-04-18 at 16 19 04

1. Currently looking into this one, it shouldn't be adding those
2. We are using `jscodeshift-add-imports` to add the import and that is how it handles adding new import, we could write our own code to try and solve that tho.

denkristoffer (approved):

PR #2717, opened 2024-04-18
fix(breadcrumb): remove max-width

denkristoffer:
# Purpose of PR

Removes `max-width` from breadcrumb buttons.

### DIFF
--- a/packages/components/header/package.json
+++ b/packages/components/header/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@contentful/f36-header",
-  "version": "4.0.0-alpha.5",
+  "version": "4.0.0-alpha.6",
   "description": "Forma 36: Header component",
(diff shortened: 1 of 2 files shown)


massao (approved):

denkristoffer:
@massao Do I need to bump version manually here? I can't remember for alpha versions 🤔

massao:
> @massao Do I need to bump version manually here? I can't remember for alpha versions 🤔

You can run `build`, and after that run `npm run prerelease`, it starts an interactive prompt to choose package and version, and if it should be published

denkristoffer:
@massao But I should do that for this PR before merging 😄

massao:
> @massao But I should do that for this PR before merging 😄

Yes, since the script also change the version on the package.json 😄

PR #2721, opened 2024-04-23
feat(badge): add prop to skip CSS text transforms

MayaGillilan:
# Purpose of PR

This adds a prop to the Badge component so as to not have to overwrite styles when using it with text that should remain consistently capitalized and not adhere to the Badge capitalization restrictions

## PR Checklist

- [x] Tests are added/updated/**not required**
- [x] Tests are passing
- [x] Storybook stories are added/updated/**not required**
- [x] Usage notes are added/updated/**not required**

### DIFF
--- a/packages/components/badge/src/Badge/Badge.styles.ts
+++ b/packages/components/badge/src/Badge/Badge.styles.ts
@@ -64,7 +64,7 @@ const sizeToStyles = ({ size }: { size: BadgeSize }): CSSObject => {
 };
 
 export const getBadgeStyles = (
-  textTransform: CSSProperties['textTransform'],
+  textTransform: CSSProperties['textTransform'] = 'capitalize',
 ) => ({
   badge: ({ variant, size }: BadgeStylesProps) =>
     css({
--- a/packages/components/badge/src/Badge/Badge.styles.ts
+++ b/packages/components/badge/src/Badge/Badge.styles.ts
@@ -2,6 +2,7 @@ import tokens from '@contentful/f36-tokens';
 import { css } from 'emotion';
 import type { BadgeVariant, BadgeSize, BadgeStylesProps } from '../types';
 import type { CSSObject } from '@emotion/serialize';
+import { CSSProperties } from 'react';
 
 const variantToStyles = ({ variant }: { variant: BadgeVariant }): CSSObject => {
   switch (variant) {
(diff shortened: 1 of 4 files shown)


packages/components/badge/src/Badge/Badge.styles.ts
```diff
@@ -83,9 +83,9 @@ export const getBadgeStyles = () => ({
   badgeText: css({
     color: 'currentcolor',
     lineHeight: 'inherit',
-    textTransform: 'lowercase',
+    textTransform: noTextTransforms ? 'none' : 'lowercase',
     '::first-letter': {
-      textTransform: 'uppercase',
+      textTransform: noTextTransforms ? 'none' : 'uppercase',
```
z0al, line 88: Was the existing transforms trying to simulate `text-transform: capitalize` ? 

If yes then maybe it's better to replace two with one transform and change `noTextTransforms` to `textTransform: 'none' | 'capitalize` | ...`? Just something to think about.

packages/components/badge/src/Badge/Badge.tsx
cf-remylenoir, line 1: What do you think about having a more scalable API?

```
textTransform?: CSSProperties['textTransform'];
```

MayaGillilan: Sure - I wanted to do minimal changes since I don't want to intrude too much on f36 behavior, but if you think it would be best to do it that way I'm up for it :)

packages/components/badge/src/Badge/Badge.tsx
```diff
@@ -32,16 +32,16 @@ export type BadgeInternalProps = CommonProps & {
...
-  noTextTransforms?: boolean;
+  textTransform?: CSSProperties['textTransform'];
 };
 
 export type BadgeProps = PropsWithHTMLElement<BadgeInternalProps, 'div'>;
 
 export const Badge = React.forwardRef<HTMLDivElement, ExpandProps<BadgeProps>>(
   (props, ref) => {
-    const styles = getBadgeStyles(props.noTextTransforms);
+    const styles = getBadgeStyles(props.textTransform ?? 'capitalized');
```
z0al: Nit: I'd change the parameter in `getBadgeStyles` to be have a default value instead to make sure we always fall back to the same value

```typescript
export const getBadgeStyles = (
  textTransform: CSSProperties['textTransform'] = 'capitalize',
) => ({
```

z0al (approved):

PR #2723, opened 2024-04-23
fix(navlist.item): Add `disabled` attribute to disabled NavList.Item component

colomolo:
# Purpose of PR
Adding actual `disabled` attribute to HTML element of NavList.Item when `isDisabled` is passed.

## PR Checklist

- [x] Tests are added/updated/**not required**
- [x] Tests are passing
- [x] Storybook stories are added/updated/**not required**
- [x] Usage notes are added/updated/**not required**


denkristoffer (commented):
We need to do some validation before passing this prop, `disabled` is not supported by `<a>`, which is the default for NavList. Are you intending to use this for button elements?

colomolo:
> We need to do some validation before passing this prop, `disabled` is not supported by `<a>`, which is the default for NavList. Are you intending to use this for button elements?

Yeah, somehow I missed that. Will check.

colomolo:
> We need to do some validation before passing this prop, `disabled` is not supported by `<a>`, which is the default for NavList. Are you intending to use this for button elements?

However:
https://github.com/contentful/forma-36/blob/main/packages/components/button/src/Button/Button.tsx#L124

I think it should be fine with disabled. Wdyt?

denkristoffer:
> > We need to do some validation before passing this prop, `disabled` is not supported by `<a>`, which is the default for NavList. Are you intending to use this for button elements?
> 
> However: [`main`/packages/components/button/src/Button/Button.tsx#L124](https://github.com/contentful/forma-36/blob/main/packages/components/button/src/Button/Button.tsx?rgh-link-date=2024-04-26T12%3A00%3A06Z#L124)
> 
> I think it should be fine with `disabled` attribute. Wdyt?

@colomolo I don't think I follow. The [default tag for the NavListItem is `<a>`](https://github.com/contentful/forma-36/blob/main/packages/components/navlist/src/NavListItem/NavListItem.tsx#L13), which doesn't accept `disabled`. If you [pass `disabled` here](https://github.com/contentful/forma-36/pull/2723/files#diff-136d85777bb47a18ed195eb239621b8b5e75701dbbcfd24cadf0b58d25280f87R69) it will complain for anyone not rendering the `<MenuListItem as="button" />`

colomolo:
> @colomolo I don't think I follow, could you explain what you read from that?

I mean that in Button component `disabled` is [explicitly assigned](https://github.com/contentful/forma-36/blob/main/packages/components/button/src/Button/Button.tsx?rgh-link-date=2024-04-26T12%3A00%3A06Z#L124) to `a` element:
```
<a {...otherProps} {...commonProps} disabled={isDisabled}>
```

I just tried it and can't see any errors:
image

denkristoffer:
> > @colomolo I don't think I follow, could you explain what you read from that?
> 
> I mean that in Button component `disabled` is [explicitly assigned](https://github.com/contentful/forma-36/blob/main/packages/components/button/src/Button/Button.tsx?rgh-link-date=2024-04-29T10%3A35%3A59Z#L124) to `a` element:
> 
> ```
> <a {...otherProps} {...commonProps} disabled={isDisabled}>
> ```

Right, I see. So it's not going to throw a warning but just do nothing instead. In that case I think we can land this as it is for now, although ideal case would still be handling per element type.

denkristoffer (approved):

PR #2728, opened 2024-04-25
fix: update mapping

denkristoffer:
# Purpose of PR

Update codemod mapping.


packages/forma-36-codemod/transforms/v5/icons.js
```diff
@@ -45,14 +47,18 @@ const iconsMap = {
   ExternalLink: 'ArrowSquareOut',
   FaceHappy: 'Smiley',
   Filter: 'FunnelSimple',
+  Folder: 'FolderSimple,
```
massao, line 50:
```suggestion
  Folder: 'FolderSimple',
```

massao (approved):
just one small type but LGTM

PR #2730, opened 2024-04-25
[DO NOT MERGE] feat: navigation light mode 

Lelith:
# Purpose of PR
Change the color schema of the Navbar component from dark to light. This is not a new variation but a complete switch of style. The overall structure and layout of the component is unchanged. 
But some of the classNames are changed in order to be more meaningful.


denkristoffer:
As far as I can tell from the design we don't want the background coloured border to separate the items:
Screenshot 2024-04-29 at 14 47 08

denkristoffer:
Are you also seeing this weird behaviour with the background cut in half on the hover?
Screenshot 2024-04-29 at 16 54 19

packages/components/navbar/src/NavbarSwitcher/NavbarSwitcher.tsx
```diff
@@ -41,14 +41,15 @@ function _NavbarSwitcher(
...
-      className={cx(styles.root, className)}
+      className={cx(styles.navbarSwitcher, className)}
       testId={testId}
       alignItems="center"
       fullHeight
       gap="spacingXs"
     >
       {logo || <ContentfulLogoIcon />}
-      <MenuIcon className={styles.menuIcon} size="small" variant="white" />
+      <MenuIcon size="medium" variant="secondary" />
```
denkristoffer, line 51: I think this should be `small`.

denkristoffer (commented):
I am noticing a few minor differences from the design, I don't know how much we want to nitpick them though. Any thoughts @damann?

Screenshot 2024-04-30 at 10 29 47

- There's a difference in the green colour used for the environments.
- The selected state is currently "inside" the hover state background, in the design it lives below it.
- The burger menu icon seems smaller in the design.

damann:
Thanks for your attention to detail @denkristoffer!
Please still do the following changes @Lelith:

- Environment colors: green600 and orange500
- Help icon color: grey700
- Avatar/initials: background grey300, on hover grey400, Initials grey700
- Burger icon: size <small>, grey700
- Trial button outline and font color: purple600

Thanks! ♥️♥️♥️

packages/components/navbar/src/NavbarSwitcher/NavbarSwitcher.styles.ts
```diff
@@ -33,23 +32,22 @@ export const getNavbarSwitcherStyles = () => ({
         display: 'none',
       },
       '&:hover li': {
-        backgroundColor: hexToRGBA(tokens.colorWhite, 0.15),
+        backgroundColor: '#f2f4f6', // calculated hex color to avoid opaque color blending
```
denkristoffer, line 35: I'm trying to understand what was happening here, can you explain? Were two layers positioned on top of each other or something, which would cause issues with the opacity? Can we explain how this colour was calculated, so we can easily do it again when the colours change?

packages/components/navbar/stories/Navbar.stories.tsx
```diff
@@ -37,7 +37,7 @@ const Account = ({
 }: Partial<NavbarAccountProps>) => (
   <Navbar.Account
     username="username"
-    avatar="https://images.ctfassets.net/iq4lnigp6fgt/2EEEk92Kiz6KxREsjBLPAN/810d5a21650d91abad12e95da4cd3beb/2021-06_Everyone_is_Welcome_here_1_.png?fit=fill&f=top_left&w=48&h=48"
+    avatar="https://www.gravatar.com/avatar/f6d4b637888d9c661921216832a9126e?s=50&d=https%3A%2F%2Favatars.contentful.com%2Finitials%2Fv1%2FJD%2F50"
```
denkristoffer, line 40:
```suggestion
    avatar="https://avatars.contentful.com/initials/v1/JD/50"
```

I would suggest we don't use third-party assets since they go down once in a while and it leads to Chromatic snapshots breaking. This is not really a big concern for gravatar, but still good practice.

denkristoffer (approved):

cf-remylenoir:
Nit: how's the focus state supposed to look like @damann?

Screenshot 2024-04-30 at 15 27 20

packages/components/navbar/stories/Navbar.stories.tsx
Lelith: Thank's i was not aware of that URL, just copied the one from my avatar in the app 😅

cf-remylenoir: TIL too!

packages/components/navbar/src/NavbarSwitcher/NavbarSwitcher.styles.ts
Lelith: The individual li elements are overlapping as we do not longer have the gap between them. I also could not put the color on the wrapping element as it would show a fully wrong width (because of the overlapping child elements). I tried to adjust the clipped path that was on them before that created the white halo. But i was not able to get it really working without looking wonky (is there a tool that calculates the right clip path for you?)

Anyhow, this is why I decided to not use the opacity. I simply did a color picker from the desired gray tone rendered.

Lelith:
> Nit: What should the focus state look like for the different parts @damann?
> 
> Some of them have an extra border color (space/env switcher + avatar) and others only the focus blur.
> 

@cf-remylenoir i believe the blue glow should be correct and not the black borders.

cf-remylenoir:
Something we can follow up on in your absence unless you tackle it before is the loading state.

It seems there are some slight differences in the sizes and alignment of the skeleton elements (i.e. navbar items aren't vertically centered)

**Current navigation bar**

https://github.com/contentful/forma-36/assets/103024358/807e58a6-095b-4ce7-a2b0-3806c9eb69e3

**New navigation bar skeleton** (switch between Storybook Basic and Loading Skeleton stories, it might not be accurate ?)

https://github.com/contentful/forma-36/assets/103024358/c9920385-cddd-4ba3-a7cf-6c75f9825100

cf-remylenoir:
I will take care of adjusting the Skeleton alignment/sizing + I also realized the navigation height is greater than in the design (`92px` instead of `81px`).

cf-remylenoir (dismissed):
Marking as changes requested to avoid potential merge by mistake

cf-remylenoir:
> I think there's still a bit of vertical shift between the skeleton and the navbar items but it's negligible and probably not worth tackling until we change the active state.

This should be now fixed, I increased the offset.

PR #2731, opened 2024-04-25
fix(badge): restore default text transform

denkristoffer:
# Purpose of PR

Recent Badge change included a default CSS `text-transform: capitalize`, but this is not the same as the default we had, where only the first word was capitalized. This fixes that, but still allows users to avoid the transform overall. Also aligns with our existing styling APIs where we pass arguments to the individual style keys rather than the overall function.


stephanLeece (approved):
nice

cf-remylenoir (approved):

PR #2732, opened 2024-04-26
feat: V5 icons codemod

massao:
# Purpose of PR

Icons v5 have updates props, we remove not supported props and update the values of the size prop.

### DIFF
--- a/packages/forma-36-codemod/package.json
+++ b/packages/forma-36-codemod/package.json
@@ -21,8 +21,7 @@
   "devDependencies": {
     "@types/semver": "^7.3.12",
     "eslint": "^8.43.0",
-    "jest": "^29.3.1",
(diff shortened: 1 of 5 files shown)


packages/forma-36-codemod/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@contentful/f36-codemod",
   "description": "Forma 36 Codemod",
-  "version": "4.5.0",
+  "version": "5.0.0-alpha.0",
```
denkristoffer, line 4: We need to update the deps as well here, we're still pointing to icons v4.

massao: good point, actually I think we don't need the icons dependency here 🤔

denkristoffer (approved):

PR #2733, opened 2024-04-26
feat: add custom omit utility type

cf-remylenoir:
# Purpose of PR

We need this custom Omit utility type to support upcoming library updates where some extended-type properties get lost when using Omit.

## Reading 🍿 
- https://github.com/microsoft/TypeScript/issues/36981
- https://github.com/microsoft/TypeScript/issues/54451

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -76,7 +76,7 @@ export type PickUnion<UnionType, Keys> = Exclude<
  * Also known as mapped type or homomorphic.
  * Use this custom Omit utility type to preserve extended Record properties.
  *
- * Why? Using Omit on extended Records results won't preserve the structure.
+ * Why? Using Omit on extended Records does not preserve the structure.
  * - https://github.com/microsoft/TypeScript/issues/36981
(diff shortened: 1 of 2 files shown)


packages/core/src/types.ts
```diff
@@ -71,3 +71,15 @@ export type PickUnion<UnionType, Keys> = Exclude<
   UnionType,
   Exclude<UnionType, Keys>
 >;
+
+/**
+ * Also known as mapped type or homomorphic.
```
cf-remylenoir, line 76:
```suggestion
 * Also known as mapped omit or homomorphic.
```

denkristoffer (commented):
You don't want to keep the name `MappedOmit`?

cf-remylenoir:
> You don't want to keep the name `MappedOmit`?

I don't have a strong opinion on this one, I renamed it.

denkristoffer (approved):

cf-remylenoir:
Ugh, I forgot to export the actual type from the barrel file 😄

PR #2746, opened 2024-05-07
build: switch to dual package emitting

cf-remylenoir:
# Purpose of PR

> [!NOTE]
> This is only for v5 alpha/beta pre-releases and won't be merged into `main`

Adds setup for emitting 2 package.json files for ESM and CJS, making use of [tsconfig-to-dual-package](https://github.com/azu/tsconfig-to-dual-package).

This approach has the benefit of clearly declaring what type should be used since it relies on 2 package.json and prevents Node/TypeScript configuration confusion or change for the consumers, and mostly issues with our direct dependency for the icons (Phosphor)

## Motivation and readings 🍿 
- https://github.com/microsoft/TypeScript/issues/49462
- https://github.com/microsoft/TypeScript/issues/54593


.changeset/config.json
cf-remylenoir, line 1: Remnant of publishing the wrong package during some testing. To be removed and published package to be unpublished.

packages/components/icon/package.json
```diff
@@ -1,29 +1,35 @@
...
     },
     "./package.json": "./package.json"
   },
   "source": "src/index.ts",
-  "types": "dist/types/index.d.ts",
   "sideEffects": false,
   "type": "module",
+  "main": "./dist/cjs/index.js",
+  "module": "./dist/esm/index.js",
+  "types": "./dist/esm/index.d.ts",
```
cf-remylenoir, line 25: For backward compability

.changeset/config.json
```diff
@@ -7,6 +7,7 @@
     }
   ],
   "ignore": [
+    "@contentful/f36-icon-alpha",
```
cf-remylenoir, line 10: Remnant of publishing the wrong package during some testing. To be removed and [published package](https://www.npmjs.com/package/@contentful/f36-icon-alpha) to be unpublished.

packages/components/icon/package.json
```diff
@@ -1,29 +1,34 @@
...
+      },
+      "default": "./dist/cjs/index.js"
     },
     "./package.json": "./package.json"
   },
-  "source": "src/index.ts",
-  "types": "dist/types/index.d.ts",
   "sideEffects": false,
   "type": "module",
+  "main": "./dist/cjs/index.js",
```
andipaetzold, line 22: `type: module` & `main` being CJS seems wrong

packages/components/icon/package.json
```diff
@@ -49,5 +53,8 @@
   "publishConfig": {
     "registry": "https://npm.pkg.github.com/",
     "access": "public"
+  },
+  "engines": {
+    "node": ">=16"
```
andipaetzold, line 58: Why is this required? It's a frontend package.

cf-remylenoir: It isn't, to be removed.

PR #2747, opened 2024-05-07
feat: [CFISO-1472] Implement new Pagecut style to Layout component

stephanLeece:
# Purpose of PR
Updates the Layout alpha component to the new `Modern and Scalable U.I` design.

Tickets: 
https://contentful.atlassian.net/browse/CFISO-1472
https://contentful.atlassian.net/browse/CFISO-1520

- New Narrow variant with max Layout.body width of 720px
- narrow, wide variants have top border radius, box shadow, 16px horizontal margin and 12px top margin
- Layout default variant is now `wide` as it is more ...widely used.

Full screen:
Screenshot 2024-05-14 at 17 22 28

Wide: 
Screenshot 2024-05-14 at 17 21 51

Narrow:
Screenshot 2024-05-14 at 17 22 51

## PR Checklist

- [x] Tests are passing


packages/components/layout/src/Layout.tsx
```diff
@@ -73,9 +74,15 @@ const _Layout = (props: LayoutProps, ref: Ref<HTMLDivElement>) => {
           flexGrow={1}
           testId={contentTestId}
         >
-          {leftSidebar}
+          {React.isValidElement(leftSidebar) &&
+            React.cloneElement(leftSidebar, {
```
stephanLeece, line 78: maybe there's an easier way to pass a prop to a component that is itself passed as a prop?

packages/components/layout/src/Layout.tsx
```diff
@@ -73,9 +74,15 @@ const _Layout = (props: LayoutProps, ref: Ref<HTMLDivElement>) => {
...
+              sidebarSide: 'left',
+            } as Partial<unknown> & Attributes)}
           {children}
-          {rightSidebar}
+          {React.isValidElement(rightSidebar) &&
+            React.cloneElement(rightSidebar, {
+              sidebarSide: 'right',
+            } as Partial<unknown> & Attributes)}
         </Flex>
```
stephanLeece, line 86: maybe there's an easier way to pass a prop to a component that is itself passed as a prop?

packages/components/layout/src/Layout.tsx
```diff
@@ -73,9 +74,15 @@ const _Layout = (props: LayoutProps, ref: Ref<HTMLDivElement>) => {
...
+          {React.isValidElement(leftSidebar) &&
+            React.cloneElement(leftSidebar, {
+              sidebarSide: 'left',
+            } as Partial<unknown> & Attributes)}
           {children}
-          {rightSidebar}
+          {React.isValidElement(rightSidebar) &&
+            React.cloneElement(rightSidebar, {
+              sidebarSide: 'right',
+            } as Partial<unknown> & Attributes)}
```
stephanLeece, line 85: maybe there's an easier way to pass a prop to a component that is itself passed as a prop?

damann:
@stephanLeece I noticed that the last item in the breadcrumb lacks a margin. It should be the same as for the breadcrumb links.

Screenshot 2024-05-08 at 10 20 25


Two tiny details in the documentation: 

1. the example content (form) has a padding. Ideally that would be removed. 
2. The background there is white. It should be gray/100

Screenshot 2024-05-08 at 10 19 26
-36/assets/67584870/e0423f04-ad7b-4bf4-a21c-383a59ee9000">

Generally speaking, could you please also update the documentation to show the different variants: left, right, both sidebars, and full, wide, narrow?

Let me know if you have questions.

stephanLeece:
> @stephanLeece I noticed that the last item in the breadcrumb lacks a margin. It should be the same as for the breadcrumb links.
> 
> Screenshot 2024-05-08 at 10 20 25
> Two tiny details in the documentation:
> 
> 1. the example content (form) has a padding. Ideally that would be removed.
> 2. The background there is white. It should be gray/100
> 
> Screenshot 2024-05-08 at 10 19 26 -36/assets/67584870/e0423f04-ad7b-4bf4-a21c-383a59ee9000">
> Generally speaking, could you please also update the documentation to show the different variants: left, right, both sidebars, and full, wide, narrow?
> 
> Let me know if you have questions.

One thing regarding the background - i will add it for the example, but it's not part of the layout component itself. So we will need to add that background to the parent component when using it in the web app.

packages/components/layout/src/Layout.styles.ts
```diff
@@ -12,16 +12,19 @@ export const getLayoutMaxWidthStyles = (variant: LayoutProps['variant']) => {
...
   });
 };
 
 export const getLayoutStyles = (variant: LayoutProps['variant']) => ({
-  root: css({
-    width: '100%',
+  root: css(getLayoutMaxWidthStyles(variant), {
+    backgroundColor: tokens.colorWhite,
+    boxShadow:
+      '0px 6px 16px -2px rgba(25, 37, 50, 0.1), 0px 3px 6px -3px rgba(25, 37, 50, 0.15)',
```
denkristoffer, line 23: @damann Is this a new token we want to reuse?

packages/components/layout/src/Layout.tsx
denkristoffer: I think using `:first-child` and `:last-child` styling on the Sidebar should do it.

denkristoffer (commented):
I think we are missing the spacing distribution / handling of the different elements mentioned in the ticket for the `narrow` variant.

packages/components/layout/src/Layout.tsx
stephanLeece: Will give it a shot, cheers!

packages/components/layout/src/Layout.styles.ts
stephanLeece: The border-radius also does not match an existing token.

stephanLeece:
> I think we are missing the spacing distribution / handling of the different elements mentioned in the ticket for the `narrow` variant.

I have the fixed 720px width set on the layout body for narrow variant. Spacing should be handled automatically. Is there something else missing? 

Screenshot 2024-05-10 at 11 36 50

packages/components/layout/src/Layout.tsx
stephanLeece: yeah much simpler, cheers

packages/components/layout/src/LayoutHeader.styles.ts
```diff
@@ -8,6 +8,7 @@ export const getLayoutHeaderStyles = (
 ) => ({
   root: css({
     borderBottom: `1px solid ${tokens.gray200}`,
+    padding: `0 ${tokens.spacingL}`,
```
denkristoffer, line 11: As discussed we cannot have padding here, it will need to be handled by the child component to avoid breaking existing headers. Sidebar and body should behave the same to be consistent.

denkristoffer (changes requested):

packages/components/layout/src/Layout.styles.ts
denkristoffer, line 23: We will have to do the box shadow inside the padding of the outer layout root if it is to work with fullscreen pages. If it helps, think of the page cut as something that should take up space on the page, not just a shadow.

stephanLeece: have removed it for the fullscreen variant

packages/components/layout/src/LayoutHeader.styles.ts
stephanLeece: removed

packages/components/layout/src/Layout.styles.ts
```diff
@@ -12,16 +12,20 @@ export const getLayoutMaxWidthStyles = (variant: LayoutProps['variant']) => {
 
   return css({
     width: '100%',
-    maxWidth: '1550px',
-    padding: `0 ${tokens.spacingXs}`,
+    maxWidth: '1524px',
+    boxShadow:
+      '0px 6px 16px -2px rgba(25, 37, 50, 0.1), 0px 3px 6px -3px rgba(25, 37, 50, 0.15)',
+    borderRadius: `10px 10px 0 0`,
+    margin: '12px 16px 0',
```
denkristoffer, line 19: I think we need to hide overflow on this element to ensure child elements get cut by the border radius.

stephanLeece:
```css
  '& > *': {
      borderRadius: `10px 10px 0 0`,
    },
```

?

stephanLeece: I don't think we'll have a reliable way to do this beyond ensuring the child component is given adequate spacing so that it doesn't clip the rounded corners.

denkristoffer (commented):
Code looks good to me, I think, but let me play around with it a bit as well 🙂

packages/components/layout/src/Layout.tsx
```diff
@@ -56,26 +56,28 @@ const _Layout = (props: LayoutProps, ref: Ref<HTMLDivElement>) => {
...
-        {header}
-
+      <Flex className={cx(styles.root, className)} justifyContent="center">
         <Flex
-          className={cx(styles.mainContainer, contentClassName)}
-          flexGrow={1}
-          testId={contentTestId}
+          {...otherProps}
+          as="section"
+          ref={ref}
```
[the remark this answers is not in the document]
stephanLeece: Thanks for the spot, done!

I had a look at usage and couldn't see anything being passed that would fall under otherProps, that isn't already accounted for. I see `CommonProps` includes `style`, but would assume we'd just use `className` for that..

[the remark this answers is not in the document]
denkristoffer: Yes that sounds right. I don't know if people can pass other props we don't account for, data attributes or something maybe. In general I think it's good API design to spread props on some element in a consistent way, but I don't think it's a battle we need to fight right here, right now.

packages/components/layout/src/Layout.tsx
```diff
@@ -37,7 +37,7 @@ const _Layout = (props: LayoutProps, ref: Ref<HTMLDivElement>) => {
     header,
     leftSidebar,
     rightSidebar,
-    variant = 'fullscreen',
+    variant = 'wide',
     className,
     testId = 'cf-layout',
```
denkristoffer, line 42:
```suggestion
    testId = 'cf-ui-layout',
```

We missed the namespace here.

denkristoffer:
I'm closing this for the `next` branch: https://github.com/contentful/forma-36/tree/next, let's keep future work there.

PR #2757, opened 2024-05-13
feat: update Avatar and Navbar.Account to display initials if avatar is not available [AHOY-3140] [AHOY-3146]

budimirbudimir:
# Purpose of PR

- Updating Avatar component and updating its dependent Navbar.Account component to contain the initials fallback option, for rendering initials if avatar URL is absent.
- If neither - avatar URL not initials - are present, display fallback option of grey background.

## PR Checklist

- [x] Tests are passing


packages/components/avatar/src/Avatar/Avatar.tsx
```diff
@@ -65,7 +67,9 @@ function _Avatar(
       {...otherProps}
     >
       {isFallback ? (
-        <div className={styles.fallback} data-test-id={`${testId}-fallback`} />
+        <div className={styles.fallback} data-test-id={`${testId}-fallback`}>
+          {initials}
```
Annmary12, line 71: What happens in a case where there are no initials or image URLs?

packages/components/navbar/src/NavbarAccount/NavbarAccount.tsx
```diff
@@ -51,11 +54,12 @@ function _NavbarAccount(
           className={cx(styles.root, className)}
           testId={testId}
         >
-          <img
+          {/* <img
```
Annmary12, line 57: If this block of code isn't required, shall we remove it altogether? Alternatively, is there a specific reason for keeping it commented out?

packages/components/avatar/src/Avatar/Avatar.tsx
budimirbudimir: We will display the same grey background like until now.
You can check it in the deployed version of Storybook [here](https://5fd1dda724cc620021ace8c5-pzscbqwoqw.chromatic.com/?path=/story/components-avatar--overview).

packages/components/navbar/src/NavbarAccount/NavbarAccount.tsx
budimirbudimir: Good catch, thanks!

packages/components/navbar/src/NavbarAccount/NavbarAccount.tsx
```diff
@@ -8,11 +8,13 @@ import {
   type ExpandProps,
 } from '@contentful/f36-core';
 import { NavbarMenu } from '../NavbarMenu/NavbarMenu';
+import { Avatar } from '../../../avatar/src/Avatar';
```
denkristoffer, line 11: This should point to the package, not the relative path.

denkristoffer (commented):
Code looks good to me but I wonder if we want to do some sort of handling of strings that are too long? Slice them after 2 or 3 characters maybe?

Annmary12:
> Code looks good to me but I wonder if we want to do some sort of handling of strings that are too long? Slice them after 2 or 3 characters maybe?

Hi @denkristoffer, I don't believe this will be necessary since the strings consist of two-character initials.

packages/components/navbar/src/NavbarAccount/NavbarAccount.tsx
Annmary12: done

denkristoffer:
> > Code looks good to me but I wonder if we want to do some sort of handling of strings that are too long? Slice them after 2 or 3 characters maybe?
> 
> Hi @denkristoffer, I don't believe this will be necessary since the strings consist of two-character initials.
> 
> _Side note: I'm taking over from Budimir as he is on vacation._

The thing we have to remember is that the design system is also used outside of our own product, so I'd prefer to handle this case so customers can't break the component. I'd either slice the string or change the type to only take two characters.

Annmary12:
> > > Code looks good to me but I wonder if we want to do some sort of handling of strings that are too long? Slice them after 2 or 3 characters maybe?
> > 
> > 
> > Hi @denkristoffer, I don't believe this will be necessary since the strings consist of two-character initials.
> > _Side note: I'm taking over from Budimir as he is on vacation._
> 
> The thing we have to remember is that the design system is also used outside of our own product, so I'd prefer to handle this case so customers can't break the component. I'd either slice the string or change the type to only take two characters.

I see your point, that makes sense. I will add that now

packages/components/avatar/src/Avatar/Avatar.tsx
```diff
@@ -65,7 +67,9 @@ function _Avatar(
       {...otherProps}
     >
       {isFallback ? (
-        <div className={styles.fallback} data-test-id={`${testId}-fallback`} />
+        <div className={styles.fallback} data-test-id={`${testId}-fallback`}>
+          {initials?.substring(0, 2).toUpperCase()}
```
Annmary12, line 71: @denkristoffer, I have implemented the feedback. Please review again

denkristoffer (approved):
Thank you, looks good to me 🙇

PR #2760, opened 2024-05-16
feat: introduce size variants for light navbar

Lelith:
This enables controlling the size of the navbar content via a dedicated prop instead the need to handover specific width sizes.

image

### DIFF
--- a/packages/components/navbar/src/Navbar.tsx
+++ b/packages/components/navbar/src/Navbar.tsx
@@ -6,7 +6,6 @@ import {
 } from '@contentful/f36-core';
 import React from 'react';
 import { getNavbarStyles } from './Navbar.styles';
-import { cx } from 'emotion';
 type NavbarOwnProps = CommonProps & {
   children?: React.ReactNode;
   account?: React.ReactNode;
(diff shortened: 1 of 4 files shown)


packages/components/navbar/src/Navbar.tsx
```diff
@@ -48,12 +51,14 @@ function _Navbar(props: ExpandProps<NavbarProps>, ref: React.Ref<HTMLElement>) {
     topRightItems,
     contentMaxWidth = '100%',
     testId = 'cf-ui-navbar',
+    className,
+    variant = 'wide',
```
denkristoffer, line 55: Although `wide` should be our default, we will have a much easier time migrating to the new Layout if we go with `fullscreen` for now. I'm a bit torn, but since it's an alpha I'm thinking we can always switch it to `wide` later on. What do you think?

Lelith: I would rather do the necessary changes now than executing a code-mod later and would go directly with the wide variant as default. 
At least this way the fullscreen is a conscious decision from the beginning and we can challenge if its even needed, when it appears to us.
While if we change it later via a code mod we will not think twice over the usage of the fullscreen variant. 

But maybe @damann can weight in if fullscreen should be the exception or the norm?

denkristoffer: We need `fullscreen` everywhere we don't have the page cut. So everywhere but 4 screens.

denkristoffer (approved):

PR #2761, opened 2024-05-17
fix: properly pass testId as data-test-id to components []

massao:
# Purpose of PR

We were passing the `testId` prop without renaming it, what would make tests that rely on the `data-test-id` to fail

### DIFF
--- a/packages/components/icon/src/utils/wrapPhosphorIcon.tsx
+++ b/packages/components/icon/src/utils/wrapPhosphorIcon.tsx
@@ -15,12 +15,16 @@ export function wrapPhosphorIcon(PhosphorIcon: PhosphorIcon) {
     isActive = false,
     color = isActive ? tokens.blue500 : tokens.gray900,
     size = 'medium',
+    testId,
     ...props
   }: GeneratedIconProps & { weight?: IconWeight }) => {
+    const commonProps = { ...props, 'data-test-id': testId };
     return (
       <Icon
-        {...props}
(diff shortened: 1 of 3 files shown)


packages/components/icon/src/utils/wrapPhosphorIcon.tsx
```diff
@@ -15,12 +15,16 @@ export function wrapPhosphorIcon(PhosphorIcon: PhosphorIcon) {
...
     size = 'medium',
+    testId,
     ...props
   }: GeneratedIconProps & { weight?: IconWeight }) => {
+    const commonProps = { ...props, 'data-test-id': testId };
     return (
       <Icon
-        {...props}
-        as={() => <PhosphorIcon {...props} color={color} size={sizes[size]} />}
+        {...commonProps}
```
denkristoffer, line 24: Shouldn't `Icon` take the `testId` prop?

massao: I had tried that, but didn't work, but looking deeper into it, seems that using a function on the `as` prop, breaks how the Icon and underling Box works.
It expects a React Element and would pass the props of the Icon and Box to it, but we are passing an arrow function that don't accept props.

So whatever we pass here to the Icon is being left out

massao: I'm taking a look into how to fix that

PR #2762, opened 2024-05-17
fix: properly pass the props from Icon to underling icon []

massao:
# Purpose of PR

We were not passing correctly the props to the icons.


andipaetzold (approved):
Why are you updating the package.json?

denkristoffer (approved):

PR #2766, opened 2024-05-27
refactor: pass color prop to icon from Button [CFISO-1471]

massao:
# Purpose of PR

To enable usage of the Icons V5 alpha we need to pass the color prop from the Button to the Icons

### DIFF
--- a/packages/components/button/src/Button/Button.tsx
+++ b/packages/components/button/src/Button/Button.tsx
@@ -76,7 +76,10 @@ function _Button<E extends React.ElementType = typeof BUTTON_DEFAULT_TAG>(
             // We need to pass the color to the icons to enable the usaged of the V5 icons
             // it may change in the future
             color:
-              (variant === 'transparent' && icon.props.variant === undefined && icon.props.color) || 'currentColor',
+              (variant === 'transparent' &&
+                icon.props.variant === undefined &&
+                icon.props.color) ||
+              'currentColor',
             // we want to allow variants for icons for transparent buttons
             variant:
               (variant === 'transparent' && icon.props.variant) ||
--- a/packages/components/button/src/Button/Button.tsx
+++ b/packages/components/button/src/Button/Button.tsx
@@ -7,6 +7,7 @@ import {
   type PolymorphicComponent,
   type ExpandProps,
 } from '@contentful/f36-core';
+import tokens from '@contentful/f36-tokens';
 import { useDensity } from '@contentful/f36-utils';
 import { Spinner } from '@contentful/f36-spinner';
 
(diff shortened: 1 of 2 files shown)


cf-remylenoir (approved):

packages/components/button/src/Button/Button.tsx
```diff
@@ -73,6 +73,10 @@ function _Button<E extends React.ElementType = typeof BUTTON_DEFAULT_TAG>(
         >
           {React.cloneElement(icon, {
             size: icon.props.size ?? `${size === 'large' ? 'medium' : 'small'}`,
+            // We need to pass the color to the icons to enable the usaged of the V5 icons
+            // it may change in the future
+            color:
+              (variant === 'transparent' && icon.props.color) || 'currentColor',
```
denkristoffer, line 79:
```suggestion
              variant === 'transparent' && icon.props.variant === undefined ? icon.props.color : 'currentColor',
```

Would this work?

Lelith: what is interesting to me is, that this issue only appears in user_interface not when i try to use any of the new icons in storybook or anywhere else. (which is why i can not tell you if either of these variations are working)

denkristoffer: @Lelith what do you mean by "this issue", exactly? What's not working? 🤔

Lelith: the wrong color of the icon. on storybook the styles do not overwrite the color set from the button.

Lelith: image
this class is not added in storybook examples
image

massao: @Lelith This is because in Storybook the IconButton isn't using the V5 Icons, and still uses the V4. But since Phosphor icons works with the color prop, we need to pass it down, which causes the `Box` component that is used by the `Icon` to add it to the CSS

Lelith: @massao as i was trying to replicate the behaviour, i actually installed them in my local working branch and my example is using a v5 icon.

massao: @denkristoffer We would need something like this, so we only keep the icon.props.color if it exists, otherwiser it would be reset to the default values on the V5 icons 🤔 

```suggestion
              (variant === 'transparent' && icon.props.variant === undefined && icon.props.color) || 'currentColor',
```

Lelith: Anyways, it's just an observation, i think the fix should still work and cover our bases. Let's merge and check it out in the user_interface branches

Lelith (approved):

PR #2771, opened 2024-05-29
fix(header): remove margin on noWrap

denkristoffer:
# Purpose of PR

@massao the Header title has a left margin at the moment that we only want on elements that aren't the first child. Can we safely remove it from the `noWrap` class?


denkristoffer:
Nope, checked the UI tests, it doesn't work. We want the margin when there's a separator, but not when the title is on its own.

PR #2785, opened 2024-06-10
feat: single line navigation

Lelith:
# Purpose of PR
reduce the size and structure of the Navbar to a single-line.
It contains the following breaking changes:

- removal of Components `NavbarSearch` and `NavbarHelp`
- Prop change of `NavbarItem`: 
  - `title` is no longer required
  -  accepts new property `label` which will be set as `aria-label` for Items without a title 
- Prop changes of `Navbar`
  - removes the following props:   
    - `badge`
    - `bottomRightItems`
    - `children`
    - `help`
    - `search`
    - `topRightItems`
  - introduces new props
    - `mainNav` as replacement for `children`
    -  `secondaryNav` as a replacement for the other removed properties   
    
Screenshots:
     
image

image


packages/components/navbar/src/Navbar.tsx
```diff
@@ -9,33 +9,33 @@ import { getNavbarStyles } from './Navbar.styles';
...
    */
-  bottomRightItems?: React.ReactNode;
-  children?: React.ReactNode;
+  logo?: React.ReactNode;
+
+  /** Environment Switcher component */
+  switcher?: React.ReactNode;
+
+  /** Main Navigation Elements */
+  mainNav?: React.ReactNode;
```
denkristoffer, line 22: Can you write out these names?

package-lock.json
cf-remylenoir, line 1: Is it intentional? There aren't changes in the `package.json`.

Lelith: hm, strange. i did a rebase and needed to npm install again i did not apply something like that on purpose, will revert.

packages/components/navbar/src/NavbarItemIcon/NavbarItemIcon.styles.ts
```diff
@@ -6,6 +6,9 @@ export const getNavbarItemIconStyles = () => ({
     '&:last-child': {
       marginLeft: tokens.spacing2Xs,
     },
+    '&:only-child': {
```
cf-remylenoir, line 9: TIL about the `only-child` pseudo-class 😄

packages/components/navbar/src/CompoundNavbar.ts
cf-remylenoir, line 1: I wonder what is the boundary between `Navbar.TopbarItem` and `Navbar.Item` now 🤔  Should we consolidate everything into `Navbar.Item`?

cf-remylenoir:
Not related to your changes but I noticed we have an a11y issue with the account button.

Screenshot 2024-06-10 at 14 21 34

damann:
Does that mean the Avatar component has the same issues? Or is it only in the navigation?
> Not related to your changes but I noticed we have an a11y issue with the account button.

cf-remylenoir:
> Does that mean the Avatar component has the same issues? Or is it only in the navigation?
> 
> > Not related to your changes but I noticed we have an a11y issue with the account button.



> Does that mean the Avatar component has the same issues? Or is it only in the navigation?
> 
> > Not related to your changes but I noticed we have an a11y issue with the account button.

Only in the navigation where we wrap the avatar with a button to display the account settings.

packages/components/navbar/src/CompoundNavbar.ts
Lelith: I think this component can be deleted like the Help or Search Helper Components.

denkristoffer (commented):
Can we make the root element `as="header"` for better accessibility?

packages/components/navbar/src/Navbar.tsx
```diff
@@ -9,33 +9,33 @@ import { getNavbarStyles } from './Navbar.styles';
...
+  secondaryNavigation?: React.ReactNode;
+
+  /** User Account Component  */
+  account?: React.ReactNode;
+
   /**
    * Defines the max-width of the content inside the navbar.
    * @default '100%'
    */
   contentMaxWidth?: string;
```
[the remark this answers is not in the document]
Lelith: So we would only have fullscreen and the wide variant with the fixed width? 
i think we can talk about this again on the responsive task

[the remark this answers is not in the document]
denkristoffer: Yeah basically controlling the width with the variant. But it makes sense to deal with later.

cf-remylenoir (approved):
Looking great! ✨

PR #2789, opened 2024-06-10
feat(utils): add `truncate` utility function

andipaetzold:
# Purpose of PR

Introduce a new utility function `truncate`.

> `truncate` is a utility function that truncates a given string from the left, right or the middle. The removed characters are replaced with a customizable string.

## PR Checklist

- [x] Tests are passing


packages/components/utils/src/truncateMiddle/truncateMiddle.mdx
```diff
@@ -0,0 +1,29 @@
...
+
+### Basic usage
+
+```ts static=true
+import { truncateMiddle } from '@contentful/f36-utils';
+
+truncateMiddle('long string', 5);
+// lo…ng
+
+truncateMiddle('long string', 7, '***);
```
denkristoffer, line 27:
```suggestion
truncateMiddle('long string', 7, '***');
```

denkristoffer (commented):
Sorry if the wording of the ticket was imprecise, but we want to always keep 6 characters on each side of the truncation. Maybe this should be configurable as well to be more generally useful?

andipaetzold:
> Sorry if the wording of the ticket was imprecise, but we want to always keep 6 characters on each side of the truncation.

I built the method in a way that matches the expectations that I assume the user of this function has. Imo, when a user truncates a string, they want to truncate that string to a specific length: The number of characters they have available. They don't care that much about the number of characters are before and after the ellipsis. Their goal is to get the string down to a max length. By ignoring the size of the replacement, the user has to do math themselves to get the string to a specific length.

Would you envision the implementation to look like this?

```
function truncateMiddle(
  str: string,
  maxLength: number,
  replacement = '…',
) {
  if (str.length <= 2 * maxLength) {
    return str;
  }

  return `${str.slice(0, maxLength)}${replacement}${str.slice(-maxLength)}`;
}

```

> Maybe this should be configurable as well to be more generally useful?

What do you want to configure?

denkristoffer:
@andipaetzold
> What do you want to configure?

The number of characters shown before and after the cut.

I do agree with you about what the expectation for a function such as this would be from a user's perspective. At the same time we currently need to always show 6 characters on each side of the cut. I will get back to you once I have confirmed the final behaviour with Daniel.

packages/components/utils/src/truncate/truncate.ts
```diff
@@ -0,0 +1,39 @@
...
+   * Number of characters to keep at the start of the string
+   */
+  start?: number;
+  /**
+   * Number of characters to keep at the end of the string
+   */
+  end?: number;
+  /**
+   * String to replace the truncated part
+   */
```
massao, line 12: maybe add the default value?? same for the other props

andipaetzold: You mean as jsdoc?

massao: just so it shows up on the editor as well

massao (approved):

damann:
@stephanLeece @massao First of all thanks for picking up on this idea! Are the numbers you can set minimum or absolute characters? I imagined it to be the later.

denkristoffer (approved):

andipaetzold:
@damann 

> First of all thanks for picking up on this idea! Are the numbers you can set minimum or absolute characters? I imagined it to be the later.

The numbers passed in via the options are the number of characters of the original string that should remain in the truncated string.
If the original string is 5 characters, but you want to keep 3 at the front and 3 at the back, no truncation is done and the original string is returned.

damann:
> The numbers passed in via the options are the number of characters of the original string that should remain in the truncated string. If the original string is 5 characters, but you want to keep 3 at the front and 3 at the back, no truncation is done and the original string is returned.

So, the truncation would stop at the specified number of characters, but it would also allow for more characters depending on the available space?

denkristoffer:
> > The numbers passed in via the options are the number of characters of the original string that should remain in the truncated string. If the original string is 5 characters, but you want to keep 3 at the front and 3 at the back, no truncation is done and the original string is returned.
> 
> So, the truncation would stop at the specified number of characters, but it would also allow for more characters depending on the available space?

@damann No, the handling of hiding and showing text will have to happen in the web app itself. This function is a helper to do that.

PR #2795, opened 2024-06-12
refactor: pass pointerEvents none to children of button

massao:
# Purpose of PR

In some cases when using SVGs and other elements inside a button the event handler would consider them as being the target of the event, and not the button.
By adding the `pointer-events: none` to the children, we make it so it consistently consider event target being the button.

Also adding an onClick action on the storybook of button to be easier to check.

**Before**

https://github.com/contentful/forma-36/assets/1071799/d66e1e7c-bdf9-44fa-9009-88d3f8158915

**After**

https://github.com/contentful/forma-36/assets/1071799/416652cd-4bc1-4a1b-8de7-db6207c30bff

### DIFF
--- a/.changeset/modern-lions-poke.md
+++ b/.changeset/modern-lions-poke.md
@@ -1,0 +1,5 @@
+---
+'@contentful/f36-button': patch
+---
+
+fix issue with event having wrong target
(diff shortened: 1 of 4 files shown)


cf-remylenoir (approved):

massao:
> This is a breaking change depending on how F36 users use the `IconButton` today. Are we sure we want to release this as a `patch`? Also, in theory, all events should bubble up, so this CSS shouldn't be necessary.

We could make the selector SVG specific, what is the main issue, being that SVG is a separate HTML object, what makes the event no bubble out from the SVG to the button, that belongs to another HTML object.

andipaetzold (changes requested):
As discussed, please close this PR

PR #2803, opened 2024-06-21
feat: adjust switcher position and style

denkristoffer:
# Purpose of PR

## PR Checklist

- [ ] Tests are passing


packages/components/navbar/examples/BasicNavbarExample.tsx
```diff
@@ -5,17 +5,7 @@ import { EntryIcon, ImageSquareIcon } from '@contentful/f36-icons';
...
-          <Navbar.SwitcherItem isCircle>
-            <span role="img" aria-label="org logo">
-              ⚡️
-            </span>
-          </Navbar.SwitcherItem>
-          <Navbar.SwitcherItem>Space</Navbar.SwitcherItem>
-          <Navbar.SwitcherItem envVariant="master">master</Navbar.SwitcherItem>
-        </Navbar.Switcher>
-      }
+      switcher={<Navbar.Switcher envVariant="master">Space</Navbar.Switcher>}
```
massao, line 8: it's import Navbar from the '@contentful/f36-components' stable version, the envVariant prop doesn't exist there.
I think we may need to import directly from the navbar package

massao (commented):
Codewise looks good, but seems to have an issue with types and versions.

I did test locally, and the `@contentful/f36-navbar` is getting the version 4.67.0 installed inside the `forma-36-react-components/node_modules` what causes tsc to complain about the envVariant on the Switcher.

packages/components/navbar/examples/BasicNavbarExample.tsx
denkristoffer: I don't think we can do that in examples, I will revert examples.

massao (approved):

PR #2804, opened 2024-06-24
fix(navbar): centred env icon

denkristoffer:
# Purpose of PR

Centred environment icon

### DIFF
--- a/packages/components/navbar/src/NavbarSwitcher/NavbarEnvVariant.tsx
+++ b/packages/components/navbar/src/NavbarSwitcher/NavbarEnvVariant.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import { NavbarSwitcherProps } from '../NavbarSwitcher/NavbarSwitcher';
 import { EnvironmentAliasIcon, EnvironmentIcon } from '@contentful/f36-icons';
-import { getNavbarEnvVariantStyles } from './NavbarSwitcher.styles';
+import tokens from '@contentful/f36-tokens';
 
 export type NavbarEnvVariantProps = Pick<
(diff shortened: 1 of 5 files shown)


Lelith (approved):

packages/components/navbar/src/NavbarSwitcher/NavbarEnvVariant.tsx
```diff
@@ -12,19 +12,11 @@ export function NavbarEnvVariant({
...
-      }
-    >
-      {isAlias ? (
-        <EnvironmentAliasIcon size="tiny" />
-      ) : (
-        <EnvironmentIcon size="tiny" />
-      )}
-    </span>
+  return isAlias ? (
+    <EnvironmentAliasIcon color={color} size="tiny" />
```
Lelith, line 18: are these not still old icons?

denkristoffer: No they are from the alpha: https://github.com/contentful/forma-36/blob/afc9f97d979e6e516c63b3f43d44348d70b629db/packages/components/navbar/package.json#L14

packages/components/navbar/src/NavbarSwitcher/NavbarSwitcher.tsx
cf-remylenoir, line 1: I can't seem to be able to use `ENTER` to activate the button

Lelith: i think in storybook there is no on-click action behind the switcher

cf-remylenoir (approved):

PR #2805, opened 2024-06-24
feat: increase hit area for navigation buttons []

massao:
# Purpose of PR

- Improve hitareas for navigation elemenents
- Improve responsiveness of navigation


packages/components/navbar/src/utils.styles.ts
```diff
@@ -19,3 +22,17 @@ export const getGlowOnFocusStyles = (shadow: string = tokens.glowPrimary) => ({
     boxShadow: shadow,
   },
 });
+
+export const increaseHitArea = (minSize = '44px'): CSSObject => ({
+  position: 'relative',
+  '&:after': {
```
denkristoffer, line 28: I think this should be `::before` to render the element "below" the item.

massao: I was using `::before` before, but it conflicts with avatar that already has a `::before`, and functionally wise, both seems to work, or did you find an issue with the after??

denkristoffer: Hmm that does make sense. For the Avatar I had to use `::before` exactly because it would otherwise visually render the wrong way. I figured it would be the same here. I haven't actually tested the code yet 😄

cf-remylenoir:
Could you please add a story for the mobile view or see if we can fix the Storybook viewport feature? I tried using the Storybook viewport feature to get the responsiveness but it does not seem to work 🤔 

Screenshot 2024-06-25 at 15 52 41

massao:
> Could you please add a story for the mobile view or see if we can fix the Storybook viewport feature? I tried using the Storybook viewport feature to get the responsiveness but it does not seem to work 🤔

Added a `WIth Responsiveness` Story to make it easier, and not have to change all the others

damann:
Screenshot 2024-07-01 at 17 12 14

Once the secondary navigation icons are hidden:
- Change icon sizes to medium
- Max out the space available to the Space switcher to show as much of the label as possible, with a minimum distance of 16px to the Menu button

Not specified in the design yet:
- We should also increase the tap target of the menu to make them touch-friendly
- Add a divider line between General settings and Space settings

packages/components/navbar/src/NavbarSwitcher/NavbarSwitcher.styles.ts
```diff
@@ -26,6 +26,16 @@ export const getNavbarSwitcherStyles = () => ({
     '&:has(> span:last-child:nth-child(3))': {
       minWidth: '12ch',
     },
+    maxWidth: '15vw',
+    [mqs.xsmall]: {
```
Lelith, line 30: unfortunately we can not use `@container` queries as our emotion version is too old. therefore i needed to use 4 different breakpoints to make the environment switcher fit for the different variations

denkristoffer:
>     * minimum gap is added but as we can not use container based queries (because of our outdated emotion version) it's super hard to find the sweet spot of width for the switcher.

@Lelith Just guessing but if something is not supported in emotion 10 it would probably be due to the version of `stylis`. Maybe we can force it to the latest version and container queries could work. Do you want to try it?

cf-remylenoir:
> > ```
> > * minimum gap is added but as we can not use container based queries (because of our outdated emotion version) it's super hard to find the sweet spot of width for the switcher.
> > ```
> 
> @Lelith Just guessing but if something is not supported in emotion 10 it would probably be due to the version of `stylis`. Maybe we can force it to the latest version and container queries could work. Do you want to try it?
> 
> Looks like we need to use stylis v4.1.3 or newer [thysultan/stylis#304](https://github.com/thysultan/stylis/pull/304)

Actually yes, that could work

Lelith:
@denkristoffer @cf-remylenoir we can try it for sure, i am just at a loss where all the configuration and packages are defined for it, they are not noted in any of the package.json files for forma?

denkristoffer:
> @denkristoffer @cf-remylenoir we can try it for sure, i am just at a loss where all the configuration and packages are defined for it, they are not noted in any of the package.json files for forma?

We have to add an `"overrides"` key in `package.json`. Docs: https://docs.npmjs.com/cli/v10/configuring-npm/package-json#overrides

Lelith:
@cf-remylenoir @denkristoffer synced with Daniel and we decided to increase all icon sizes for the mobile views. had to do some overwrites for that as we do not have media queries in JS and the icon sizes are not a token, i had to handwrite them. Should we add these sizes also as tokens?

damann:
> @cf-remylenoir @denkristoffer synced with Daniel and we decided to increase all icon sizes for the mobile views. had to do some overwrites for that as we do not have media queries in JS and the icon sizes are not a token, i had to handwrite them. Should we add these sizes also as tokens?

Does that mean they don't use `Icon size="medium"`?

Lelith:
@damann the icon size properties are not setup responsive, they have a single fixed size. If we want to support more responsive designs in the future, we should establish a pattern for Icon Sizes soon

cf-remylenoir (approved):
QA'ed with success, codewise looks good to me. I am looking forward to the v5 with the emotion bump to v11 to use container queries 🙏

PR #2806, opened 2024-06-24
chore: exclude react-day-picker.js.org from link check

denkristoffer:
# Purpose of PR

Exclude https://react-day-picker.js.org/ from the link checker job.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -89,7 +89,7 @@ jobs:
           command: npx wait-on http://localhost:3000
       - run:
           name: Run links checker
-          command: npx blc http://localhost:3000 -ro --exclude "https://medium.com/contentful-design" --exclude "https://github.com/contentful/forma-36" --exclude "https://www.figma.com/@contentful"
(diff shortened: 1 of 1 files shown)


stephanLeece:
I'm happy to go with this approach and just exclude the links, but you may wanna run this by @andipaetzold  first as it looks like [his changes ](https://github.com/contentful/forma-36/pull/2748/files)just updated the broken links rather than excluding them from the checks. So perhaps there's some benefit of keeping the checks in that I'm unaware of...

cf-remylenoir (approved):

denkristoffer:
> I'm happy to go with this approach and just exclude the links, but you may wanna run this by @andipaetzold first as it looks like [his changes ](https://github.com/contentful/forma-36/pull/2748/files)just updated the broken links rather than excluding them from the checks. So perhaps there's some benefit of keeping the checks in that I'm unaware of...

Good point, seems like the website actually changed. It redirects to https://daypicker.dev/ now. Closed for #2807

PR #2809, opened 2024-06-25
fix: navbar switcher colour

denkristoffer:
# Purpose of PR

Small visual tweaks to the switcher button.


massao (approved):
LGTM, why is chromatic showing so many changes though? 🤔

damann (approved):

denkristoffer:
> LGTM, why is chromatic showing so many changes though? 🤔

@massao I'm not certain but I think Chromatic is comparing against `main`? I changed the Storybook render wrapper a bit on the light navigation branch to be able to change viewport width all the way down, we had a set min-width before.

PR #2812, opened 2024-06-26
refactor: not animate Collapse on first render

massao:
# Purpose of PR

- Add check to disable animation on mount.
- Add Story with expanded style.

### DIFF
--- a/packages/components/collapse/src/Collapse.tsx
+++ b/packages/components/collapse/src/Collapse.tsx
@@ -1,10 +1,10 @@
-import React, { useLayoutEffect, useRef } from 'react';
+import React, { useEffect, useLayoutEffect, useRef } from 'react';
 import {
   Box,
   type CommonProps,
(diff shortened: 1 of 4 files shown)


packages/components/collapse/src/Collapse.tsx
```diff
@@ -61,6 +62,20 @@ export const Collapse = ({
     };
 
     if (current) {
+      // Don't animate on first render
+      if (isMounted.current) {
```
denkristoffer, line 66: Shouldn't this be the other way around?

massao: No, there is a transition already on the `.styles` file, that we override with the 0s transition, that we remove here.

mdmjg (approved):

cf-remylenoir (approved):

andipaetzold (approved):
LGTM

Don't forget to add a changeset

PR #2835, opened 2024-07-26
feat: update styles for section heading

massao:
# Purpose of PR

We want to replace the all caps section heading and update the font color to make the contrast more accessible

## Todo
- [x] Update version and release as alpha before merge


Lelith (approved):
Looks good to me!

PR #2839, opened 2024-07-30
fix(<entitylistitem>): differentiate click events for entire entity c…

ethan-ozelius-contentful:
…ard, and menu

because of javascript event propagation, it was previously impossible to differentiate when a user was clicking the menu trigger/menu and when they were clicking anywhere else in the entity card.

re LUMOS-119

# Purpose of PR

## PR Checklist

- [ ] Tests are passing


jsdalton (approved):

Lelith (changes requested):
I have some concerns accessibility wise. 
Yes, this approach finally fixes the "button inside button" issue, but this actually just shifts the problem, because when we render it as a div with a onClick handler, it is no longer keyboard accessible. One thing i noticed is that we do not even show the proper mouse-cursor that indicates that the whole Item is clickable.

In my opinion the proper solution would be to no longer have the menu as a child of the EntityListItem-Button but wrap them in one div container and then target each button with a proper onClick handler. We can even go that far and absolutely position the menu-trigger on top of the other button, so the right-corner of the box would still be clickable with the main-interaction.

WDYT?

package.json
```diff
@@ -19,6 +19,7 @@
     "storybook": "start-storybook -p 9001 -c ./scripts/.storybook",
     "storybook:build": "build-storybook -c ./scripts/.storybook -o dist-storybook --webpack-stats-json",
     "test": "jest",
```
ethan-ozelius-contentful, line 21: Not strictly related to this PR, but I find myself using watch mode a lot when developing test for forma.

packages/components/entity-list/src/EntityListItem/EntityListItem.tsx
```diff
@@ -186,107 +188,109 @@ export const EntityListItem = ({
           </Skeleton.Container>
         </article>
       ) : (
```
ethan-ozelius-contentful, line 190: This looks like a much bigger change than it is.  All I'm doing is adding a `<>` and moving the `<Element ...` and  `{actions && (...` elements alongside each other.

packages/components/entity-list/stories/EntityList.stories.tsx
```diff
@@ -12,30 +13,75 @@ export default {
...
+      console.log('[ <MenuItem> ] Delete');
+    }}
+  >
+    Delete
+  </MenuItem>,
+];
+
 export const Basic: Story<EntityListProps> = () => (
   <EntityList>
     <EntityList.Item
```
ethan-ozelius-contentful, line 45: I could take or leave these storybook changes.  I found them very useful while developing, mostly the changes are debugging in nature, like updating the title to reflect the eventual type of element that will be rendered `button` or `a`, and adding onClick console messages to ensure that callbacks are firing correctly.

jsdalton (approved):

packages/components/entity-list/src/EntityListItem/EntityListItem.styles.ts
```diff
@@ -21,17 +21,23 @@ export const getEntityListItemStyles = () => ({
         backgroundColor: tokens.gray100,
       },
     }),
```
ethan-ozelius-contentful, line 23: Made this a function to set `cursor` to pointer if element is clickable.

ethan-ozelius-contentful:
> I have some concerns accessibility wise.
> Yes, this approach finally fixes the "button inside button" issue, but this actually just shifts the problem, because when we render it as a div with a onClick handler, it is no longer keyboard accessible. One thing i noticed is that we do not even show the proper mouse-cursor that indicates that the whole Item is clickable.
> In my opinion the proper solution would be to no longer have the menu as a child of the EntityListItem-Button but wrap them in one div container and then target each button with a proper onClick handler. We can even go that far and absolutely position the menu-trigger on top of the other button, so the right-corner of the box would still be clickable with the main-interaction.


@Lelith, thanks for the feedback, the accessibility call out is a really good point.

I've separated the action menu from the rest of the components in `<EntityCardListItem>`, so that now click events don't clash with each other.  Separating the action menu had the added benefit of removing the necessity for the `onMenuClick` prop that was going to be introduced by this PR  🎉.

As far as accessibility, 
1. I've updated the cursor of`<a>` and `<button>` <EntityCardListItems>`s to be a pointer, while `<article>` <EntityCardListItems>`s will inherit their cursor.
2. I've improved the box-shadow of `<EntityCardListItems>` see "Keyboard accessible differentiated events" gif.

Lastly, I've removed the `as` prop that this PR was originally going to introduce.  I agree with your accessibility concerns that it would introduce too many potential issues if we allow consumers to specify any type of element.

packages/components/entity-list/stories/EntityList.stories.tsx
Lelith: i think it's good to have these examples, let's keep them!

packages/components/entity-list/stories/EntityList.stories.tsx
```diff
@@ -12,30 +13,68 @@ export default {
...
-      title="Entry 1"
+      title="{ ONE } onClick & onMenuClick, href=undefined - <button>"
```
Lelith, line 39:
```suggestion
      title="{ ONE } onClick & onMenuClick defined, href undefined, button"
```

packages/components/entity-list/stories/EntityList.stories.tsx
```diff
@@ -12,30 +13,68 @@ export default {
...
+      onClick={() => {
+        console.log('{ ONE } onClick()');
+      }}
       description="Description"
       contentType="My content type"
       status="published"
     />
     <EntityList.Item
-      title="Entry 2"
+      title="{ TWO } href, onClick=undefined - <a>"
```
Lelith, line 49:
```suggestion
      title="{ TWO } hre defined, onClick undefined, a-tag"
```

packages/components/entity-list/stories/EntityList.stories.tsx
```diff
@@ -12,30 +13,68 @@ export default {
...
+        console.log('{ FOUR } onClick()');
+      }}
       description="Description"
       contentType="My content type"
       status="archived"
       entityType="Release"
     />
     <EntityList.Item
-      title="Entry 4"
+      title="{ FIVE } Presentational entity card - <article>"
```
Lelith, line 77:
```suggestion
      title="{ FIVE } Presentational entity card - article"
```

packages/components/entity-list/stories/EntityList.stories.tsx
```diff
@@ -12,30 +13,68 @@ export default {
...
       contentType="My content type"
       status="draft"
       entityType="Page"
+      onClick={() => {
+        console.log('{ THREE } onClick()');
+      }}
     />
     <EntityList.Item
-      title="Entry 3"
+      title="{ FOUR } onClick() defined, href=undefined - <button>"
```
Lelith, line 67:
```suggestion
      title="{ FOUR } onClick defined, href undefined - button"
```

packages/components/entity-list/stories/EntityList.stories.tsx
```diff
@@ -12,30 +13,68 @@ export default {
...
     <EntityList.Item
-      title="Entry 2"
+      title="{ TWO } href, onClick=undefined - <a>"
+      href={'https://contentful.com'}
+      description="Description"
+      contentType="My content type"
+      status="published"
+    />
+    <EntityList.Item
+      title="{ THREE } onClick defined & href defined - <a> "
```
Lelith, line 56:
```suggestion
      title="{ THREE } onClick defined & href defined - a-tag "
```

Lelith (approved):
Thank you very much for putting in the time and  solving the root issues of this component.

cf-remylenoir:
FYI please avoid using tags in the commit message or changelog because they break the website build once the changelog description gets injected into `changelog.mdx` (after the merge).

It happened in the past on a few occasions, I am looking into adding a lint check for Markdown files to prevent this from happening.

PR #2840, opened 2024-08-05
feat: update section heading style

massao:
# Purpose of PR

Update the style of the SectionHeading component so the color contrast is accessible and the aesthetic appeal is more modern.

### DIFF
--- a/packages/components/typography/src/SectionHeading/SectionHeading.tsx
+++ b/packages/components/typography/src/SectionHeading/SectionHeading.tsx
@@ -43,14 +43,17 @@ function _SectionHeading<
       as={SECTION_HEADING_DEFAULT_TAG}
       testId={testId}
       marginBottom="spacingL"
-      fontWeight="fontWeightDemiBold"
-      fontColor="gray900"
+      fontWeight="fontWeightMedium"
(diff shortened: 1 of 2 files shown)


Lelith (approved):

packages/components/typography/src/SectionHeading/SectionHeading.tsx
```diff
@@ -43,14 +43,17 @@ function _SectionHeading<
...
       lineHeight={density === 'high' ? 'lineHeightSHigh' : 'lineHeightS'}
       className={cx(
         css({
-          letterSpacing: tokens.letterSpacingWide,
-          textTransform: 'uppercase',
+          letterSpacing: tokens.letterSpacingDefault,
+          textTransform: 'lowercase',
+          '&:first-letter': {
+            textTransform: 'uppercase',
+          },
```
denkristoffer, line 56: Do we need this at all?

massao: This is to make sure that the styling will be consistent, since because of the previous `uppercase` for everything, there may be places where the string is all lowercase or something like that

PR #2850, opened 2024-08-23
feat: enable built-in tooltip for icon buttons

Lelith:
# Purpose of PR
This enables a built in tooltip for icon-buttons to enhance their user-experience. The IconButton will now accept `tooltipProps`, once it received this property, it will render a tooltip with a default showDelay of `600ms`

It is recommended to use the same content for the tooltip as the aria-label. 

image

image

### DIFF
--- a/packages/components/button/src/IconButton/IconButton.tsx
+++ b/packages/components/button/src/IconButton/IconButton.tsx
@@ -119,7 +119,7 @@ function _IconButton<
 _IconButton.displayName = 'IconButton';
 
 export const IconButton: PolymorphicComponent<
-  ExpandProps<IconButtonInternalProps>,
+  ExpandProps<ExtendedIconButtonProps>,
   typeof ICON_BUTTON_DEFAULT_TAG,
   'disabled'
 > = React.forwardRef(_IconButton);
--- a/packages/components/button/src/IconButton/IconButton.tsx
+++ b/packages/components/button/src/IconButton/IconButton.tsx
@@ -21,7 +21,7 @@ type WithTooltipOrNot =
       /**
        * Triggers, wheter or not to render the tooltip
        */
-      withTooltip?: true;
+      withTooltip?: boolean;
 
       /**
        * A tooltipProps attribute used to conditionally render the tooltip around root element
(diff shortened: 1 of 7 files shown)


denkristoffer (commented):
Does it make sense to have a `withTooltip` prop (continuing naming pattern from other components) where the `content` defaults to the same value as `aria-label`? So you don't necessarily have to define it twice.

cf-remylenoir:
Could we keep the tooltip aligned with the center of the target?

Screenshot 2024-08-26 at 11 06 38

Lelith:
@cf-remylenoir i think that was just an issue of the surrounding divs
image

cf-remylenoir (approved):
Nice addition, @damann are we ready on the design side to ensure code & design parity?

damann:
> Nice addition, @damann are we ready on the design side to ensure code & design parity?

Not yet, but I can update it quickly.

packages/components/button/stories/IconButton.stories.tsx
```diff
@@ -46,6 +46,38 @@ basic.args = {
   variant: 'transparent',
 };
 
+export const withTooltip = ({ icon, iconProps, ...props }) => (
```
denkristoffer, line 49: Can we have this story—or an additional one—default to tooltip open so we can pick it up with Chromatic?

packages/components/button/src/IconButton/IconButton.tsx
```diff
@@ -49,24 +67,44 @@ function _IconButton<
...
+      aria-label={ariaLabel}
+      {...otherProps}
     />
   );
+
+  if (withTooltip) {
+    const {
+      showDelay = 600,
+      content = ariaLabel,
+      ...restTooltipProps
```
massao, line 97: not a biggie or anything, but in other places we usually use other rather than rest

packages/components/button/src/IconButton/IconButton.tsx
```diff
@@ -32,6 +38,18 @@ interface IconButtonInternalProps
...
+
+  /**
+   * Triggers, wether or not to render the tooltip
+   */
+  withTooltip?: boolean;
+
+  /**
+   * A tooltipProps attribute used to conditionally render the tooltip around root element
+   */
+  tooltipProps?: CommonProps &
```
massao, line 50: Should we have a type check like the one in [Header](https://github.com/contentful/forma-36/blob/a3c52a6d5b4c8a43895ab64324c31427ff39c26e/packages/components/header/src/Header.tsx#L24-L35)?? 
So the editor would complain about a tooltipProps without the withTooltip??

PR #2851, opened 2024-08-26
feat: add tooltip for icon only nav items cfiso 1677

Lelith:
# Purpose of PR

Shows a delayed tooltip for icon-only menu items with the item title

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -41530,7 +41530,7 @@
     },
     "packages/components/navbar": {
       "name": "@contentful/f36-navbar",
-      "version": "5.0.0-alpha.30",
+      "version": "5.0.0-alpha.31",
       "license": "MIT",
       "dependencies": {
         "@contentful/f36-avatar": "4.68.1",
--- a/packages/components/navbar/package.json
+++ b/packages/components/navbar/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@contentful/f36-navbar",
-  "version": "5.0.0-alpha.30",
+  "version": "5.0.0-alpha.31",
   "description": "Forma 36: Navbar component",
   "scripts": {
     "build": "tsup"


packages/components/navbar/src/NavbarItem/NavbarItem.tsx
```diff
@@ -83,6 +84,16 @@ function _NavbarItem(
     </Comp>
   );
 
+  if (!title) {
+    item = (
+      <div>
+        <Tooltip content={label} placement="bottom" showDelay={600} usePortal>
+          {item}
```
cf-remylenoir, line 91: Is this correct? We are re-assigning the `item` value on line 88 and then use it again from within here 🤔

denkristoffer: Yes that's possible, like `i = i + 1`.

cf-remylenoir: I see 👍

andipaetzold, line 91: This probably has been discussed before: Why do we make the tooltip part of the navbar item? The same behavior can be already achieved today by combining `Tooltip` and `NavbarItem` outside of f36.

denkristoffer: If we want to always have tooltips for these items I think it makes sense to compose them here.

damann (approved):

packages/components/navbar/src/NavbarAccount/NavbarAccount.styles.ts
```diff
@@ -16,38 +15,20 @@ const notificationVarianColorMap: Record<
...
     },
     getGlowOnFocusStyles(),
     increaseHitArea(),
   ),
-  avatar: css({
-    borderRadius: '50%',
-    display: 'block',
-    height: '24px',
-    width: '24px',
-  }),
```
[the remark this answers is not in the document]
Lelith: that class was not even used, i think it was other lines i deleted.

packages/components/navbar/src/NavbarItem/NavbarItem.styles.ts
```diff
@@ -44,6 +44,7 @@ export const getNavbarItemStyles = ({ title }) => ({
       boxSizing: 'border-box',
       transition: `color ${tokens.transitionDurationShort} ${tokens.transitionEasingCubicBezier}`,
       borderRadius: tokens.borderRadiusMedium,
+      height: '30px',
```
cf-remylenoir, line 47: Is it the fix to avoid the loading skeleton & complete state jump?

packages/components/navbar/src/NavbarAccount/NavbarAccount.styles.ts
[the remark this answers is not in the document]
cf-remylenoir: Ah maybe then

packages/components/navbar/src/NavbarItem/NavbarItem.styles.ts
Lelith: yes, but i had a chat with Andi and we have found a different approach, will be fixed with a different pr, but for time beeing, this is necessary so the buttons with icon only and a tooltip don't become too small

packages/components/navbar/src/NavbarAccount/NavbarAccount.styles.ts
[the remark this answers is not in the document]
Lelith: it was the padding :) fixed it

cf-remylenoir (approved):
Nice ✨

PR #2857, opened 2024-09-03
feat: add environment to navbar switcher

massao:
# Purpose of PR

Add environment name to the navbar switcher

### Preview
image

### DIFF
--- a/packages/components/navbar/src/NavbarSwitcher/NavbarSwitcher.tsx
+++ b/packages/components/navbar/src/NavbarSwitcher/NavbarSwitcher.tsx
@@ -1,4 +1,4 @@
-import React, { type ReactNode } from 'react';
+import React from 'react';
 import { getNavbarSwitcherStyles } from './NavbarSwitcher.styles';
 import { Button } from '@contentful/f36-button';
 import {
@@ -9,6 +9,10 @@ import {
 } from '@contentful/f36-core';
 import { cx } from 'emotion';
 import { NavbarEnvVariant } from './NavbarEnvVariant';
+import { NavbarSwitcherSkeleton } from './NavbarSwitcherSkeleton';
+import { CaretRightIcon } from '@contentful/f36-icons';
+import { Text } from '@contentful/f36-typography';
+import tokens from '@contentful/f36-tokens';
 
 type TruncateOptions = {
   /**
@@ -23,7 +27,7 @@ type TruncateOptions = {
   end?: number;
 };
 
-function splitSpaceName(
+function splitString(
   string: string,
   {
     start: startLength = 5,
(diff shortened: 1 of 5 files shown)


packages/components/navbar/src/NavbarSwitcher/NavbarSwitcher.tsx
```diff
@@ -41,35 +45,90 @@ function splitSpaceName(
...
+      <span>{start}</span>
+      {middle && (
+        <span className={styles.switcherSpaceNameTruncation}>{middle}</span>
+      )}
+      {end && <span>{end}</span>}
+    </Text>
+  ) : (
+    <Text fontColor={isEnv ? envColor : 'gray600'}>{value}</Text>
+  );
+};
```
cf-remylenoir, line 114: Could we consolidate colors like so?

```suggestion
const SwitcherLabel = ({
  value,
  styles,
  envVariant,
  type,
}: SwitcherLabelProps) => {
  const [start, middle, end] =
    typeof value === 'string' ? splitString(value) : [];
  const isEnv = type === 'environment';
  const isMaster = envVariant === 'master';
  const fontColor = isEnv && isMaster ? 'green600' : 'gray600';
  
  return start !== undefined ? (
    <Text fontColor={fontColor}>
      <span>{start}</span>
      {middle && (
        <span className={styles.switcherSpaceNameTruncation}>{middle}</span>
      )}
      {end && <span>{end}</span>}
    </Text>
  ) : (
    <Text fontColor={fontColor}>{value}</Text>
  );
};
```

cf-remylenoir: Or rather

```typescript
switch(true) {
  case isEnv && isMaster:
    return 'green600';
  case isEnv && !isMaster || isEnv:
  default:
    return 'gray600';
}
```

packages/components/navbar/src/NavbarSwitcher/NavbarSwitcher.tsx
```diff
@@ -41,35 +45,98 @@ function splitSpaceName(
...
-  children: ReactNode;
-  isCircle?: boolean;
-  envVariant?: 'master' | 'non-master';
-  isAlias?: boolean;
-  ariaLabel?: string;
-};
+type NavbarLoadingProps =
+  | {
+      isLoading?: true;
+      environemnt?: never;
```
denkristoffer, line 51:
```suggestion
      environment?: never;
```

packages/components/navbar/src/NavbarSwitcher/NavbarSwitcher.tsx
```diff
@@ -41,35 +45,98 @@ function splitSpaceName(
...
+      space?: never;
+    }
+  | {
+      isLoading?: false;
+      environemnt?: string;
```
denkristoffer, line 56:
```suggestion
      environment?: string;
```

packages/components/navbar/src/NavbarSwitcher/NavbarSwitcher.tsx
```diff
@@ -41,35 +45,98 @@ function splitSpaceName(
...
+    typeof value === 'string' ? splitString(value) : [];
+  const envColor = useMemo(() => {
+    const isEnv = type === 'environment';
+    const isMaster = envVariant === 'master';
+    switch (true) {
+      case isEnv && isMaster:
+        return 'green600';
+      default:
+        return 'gray600';
+    }
```
denkristoffer, line 108: Can we do if statements and not try to be overly clever?

packages/components/navbar/src/NavbarSwitcher/NavbarSwitcher.tsx
```diff
@@ -41,35 +45,98 @@ function splitSpaceName(
...
+      space?: string;
+    };
+
+type NavbarSwitcherOwnProps = CommonProps &
+  NavbarLoadingProps & {
+    isCircle?: boolean;
+    envVariant?: 'master' | 'non-master';
+    isAlias?: boolean;
+    ariaLabel?: string;
+    environment?: React.ReactNode;
```
denkristoffer, line 66: Does this type still make sense with the new `NavbarLoadingProps`?

packages/components/navbar/src/NavbarSwitcher/NavbarSwitcher.tsx
cf-remylenoir: Sorry @massao 😄

cf-remylenoir (approved):
It looks good to me, my only remark would be related to the poor contrast of the caret icon color with its background. @damann what do you think, does it satisfy an accessible contrast ratio?

damann:
> It looks good to me, my only remark would be related to the poor contrast of the caret icon color with its background. @damann what do you think, does it satisfy an accessible contrast ratio?

Let's bump it up to grey500 (contrast ration 4.57:1)

damann (approved):

PR #2859, opened 2024-09-05
fix: avoid necessity for overwriting height and scrolling behavior  [CFISO-1861]

Lelith:
# Purpose of PR

This is a refactor / rewrite of the layout component structure as well as it's styling. It forces the main container, which gives the white background always to be the height of the current view height - offset from the navigation.

The offset will be adjusted for the content height based on wether or not the component contains a header.

The scrolling behavior changes from beeing on the outer wrapper to the individual content elements as soon as at least one sidebar is available. To make all three different areas individually scrollable.

This also extended the storybook example for this component to show different variations with, without header and with different amounts of content.

variant narrow, header and right sidebar
image

variant wide header, no sidebars
image

variant wide, header and both sidebars
image

variant fullscreen, header and both sidebars
image


damann:
Thanks a bunch for this, Kathrin! 
Please consider that sidebars should always be tugged to the side of the page cut. Only the content becomes narrow. At least that was the intent.

Lelith:
> Thanks a bunch for this, Kathrin! Please consider that sidebars should always be tugged to the side of the page cut. Only the content becomes narrow. At least that was the intent.

ah, i decided to change that, because we got the feedback that it looked super weird when the content became scrollable and it looked like the gap is unintentional.

denkristoffer:
> > Thanks a bunch for this, Kathrin! Please consider that sidebars should always be tugged to the side of the page cut. Only the content becomes narrow. At least that was the intent.
> 
> @damann ah, i thought we might want to change that, because we got the feedback that it looked super weird when the content became scrollable and it looked like the gap is unintentional.

The gap should still be scrollable, meaning you can hover it and scroll the content from there, so it's "part of the content but not really". It's one of the lovely challenges we face due to the way we handle overflow in the web app 😅

damann:
Not quite sure what the scrolling has to do with the gap. Can we discuss this in power hour, please?

packages/components/layout/src/Layout.styles.ts
```diff
@@ -1,6 +1,13 @@
 import tokens from '@contentful/f36-tokens';
 import { css } from 'emotion';
 import type { LayoutProps } from './Layout';
+import type { LayoutContextType } from './LayoutContext';
+
+export const NAVBAR_HEIGHT = 60;
```
cf-remylenoir, line 6: I prefer to avoid calculating sizes/positions based on external components we don't depend on in the package, but I understand we utilized all of our jokers' cards. If we go down that road, should we import this constant from the Navigation package to represent the relationship between the two?

Lelith: I agree with you, let me double check one last time if the usage of grid could solve this now. Otherwise I agree with you and we should set it up as a cross-package constant.

packages/components/layout/src/Layout.styles.ts
```diff
@@ -20,28 +30,107 @@ export const getLayoutMaxWidthStyles = ({
...
+    width: `calc(100% - ${tokens.spacingM} * 2)`,
+  });
+};
+
+const getContentContainerGridTemplateColumns = ({
+  variant,
+  withLeftSidebar,
+  withRightSidebar,
+}: {
+  variant: LayoutContextType['variant'];
```
cf-remylenoir, line 44: Could we have used the already imported LayoutProps instead?

```suggestion
  variant: LayoutProps['variant'];
```

packages/components/layout/src/Layout.styles.ts
```diff
@@ -20,28 +30,107 @@ export const getLayoutMaxWidthStyles = ({
...
+    },
+    '&:last-child': {
+      borderLeft: `1px solid ${tokens.gray200}`,
+    },
   }),
-  contentContainer: css({
+});
+
+export const getLayoutHeaderStyles = (
+  variant: LayoutContextType['variant'],
```
cf-remylenoir, line 127: And same here?

```suggestion
  variant: LayoutProps['variant'],
```

packages/components/layout/src/Layout.styles.ts
```diff
@@ -1,6 +1,15 @@
 import tokens from '@contentful/f36-tokens';
 import { css } from 'emotion';
 import type { LayoutProps } from './Layout';
+import type { LayoutContextType } from './LayoutContext';
```
cf-remylenoir, line 4:
```suggestion
```

packages/components/layout/src/Layout.tsx
```diff
@@ -42,48 +42,48 @@ const _Layout = (props: LayoutProps, ref: Ref<HTMLDivElement>) => {
...
   return (
     <LayoutContextProvider value={contextValue}>
       <Flex
         ref={ref}
         testId={testId}
-        className={cx(styles.root, className)}
         as="section"
-        justifyContent="center"
+        {...otherProps}
+        className={styles.layoutMainContainer}
```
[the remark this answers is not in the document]
Lelith: the main container was not stylable in the previous version, but you are right. As this is now the most outer layer we should forward the class prop to this container. Will update thanks, for calling out :)

packages/components/layout/src/Layout.styles.ts
```diff
@@ -20,28 +29,107 @@ export const getLayoutMaxWidthStyles = ({
     boxShadow: withBoxShadow
       ? '0px 6px 16px -2px rgba(25, 37, 50, 0.1), 0px 3px 6px -3px rgba(25, 37, 50, 0.15)'
       : 'unset',
-    borderRadius: `10px 10px 0 0`,
-    margin: `0 ${tokens.spacingM}`,
+    borderRadius: '10px 10px 0 0',
+    margin: `0 auto`,
+    width: `calc(100% - ${tokens.spacingM} * 2)`,
```
cf-remylenoir, line 34: Could you please comment on the rationale behind this calculation?

Lelith: I removed the outer wrapper that was there to center the main container via flexbox, because it was messing with the container heights. Intead I am using now our good old friend `margin: 0 auto` but we want that even on  smaller screens, there is a 1rem gap between the white container and the rest of it, i can not set the width to 100%. It needs to be `100% - applied margin`

happy to add a comment to the code.

cf-remylenoir: Thank you for explaining, I think it is worth storing `${tokens.spacingM} * 2` in a variable to make this clear + add a comment, yes 💯

cf-remylenoir (approved):
It looks good to me, thank you for doing this work! 

Something not introduced by your changes but that we should fix (can be in a follow-up): we violate a11y landmarks because we are nesting the `<header>` tag.

See the [permitted parents](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/header#technical_summary).

Screenshot 2024-09-16 at 11 48 20

denkristoffer:
The changes don't work with the visual modeler: https://github.com/contentful/visual-modeler/pull/459

Now, I'm pretty sure this works fine once the VM goes into the web app, but it seems like there's an expectation now that the Layout is at a minimum always rendered with a navbar?

Lelith:
@denkristoffer yeah, the layout is now only to be able to be used as the outer wrapper for content in the user_interface with a navbar. otherwise its impossible to pair the sidebars, scrolling behavior AND stretching it always to the bottom of the page. Its no longer a generic "3 columns and a header" component. 

to make it more flexible we could offer a offset prop so that the default offset is the navbar or you can overwrite it.

denkristoffer:
How does this work at the moment on pages with the fullscreen behaviour? Are they fine?

cf-remylenoir:
> How does this work at the moment on pages with the fullscreen behavior and no nav (Like live preview)? Are they fine?

It looks fine, but other views lack padding at the bottom and their large vertical content is cut. 
I am looking at whether it comes from the changes here or the styling override in the web app.

massao (approved):
LGTM

cf-remylenoir:
> LGTM

Thanks for the review, after discussing with Daniel about the sidebar widths, I am working on updating the API to have various sizes (`280px` and `340px`). The default is `280px`. I will push the changes soon.

PR #2861, opened 2024-09-09
fix: add and export height constant

Lelith:
this PR is linked to https://github.com/contentful/forma-36/pull/2859
it sets a fixed height to the header component and exports the variable so other components can reuse it.


packages/components/header/src/Header.styles.ts
cf-remylenoir, line 1: Nit: would it make sense to have it in a `constants.ts` file instead of within the styles?

cf-remylenoir (approved):

PR #2862, opened 2024-09-09
fix: add fixed height to navbar

Lelith:
this pr is related to https://github.com/contentful/forma-36/pull/2859 it adds a fixed height to the light navbar and exports it, so it is useable in other components


andipaetzold:
Just to confirm: This doesn't change how things look, but only adds a height to ensure the height will stay at 60px.

massao (approved):

packages/components/navbar/src/Navbar.styles.ts
cf-remylenoir, line 1: Nit: would it make sense to have it in a `constants.ts` file instead of within the styles?

PR #2863, opened 2024-09-09
refactor: enable children on navbar switcher

massao:
# Purpose of PR

In some use cases we don't have space and environment, in those cases we pass a string that will appear on the switcher

### Preview
image

### DIFF
--- a/packages/components/navbar/src/NavbarSwitcher/NavbarSwitcher.tsx
+++ b/packages/components/navbar/src/NavbarSwitcher/NavbarSwitcher.tsx
@@ -48,11 +48,13 @@ function splitString(
 type NavbarLoadingProps =
   | {
       isLoading?: true;
+      children?: React.ReactNode;
       environment?: never;
       space?: never;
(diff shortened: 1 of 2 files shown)


Lelith (approved):
lgtm!

damann:
Hey Massao, I strongly recommend to remove that behavior. It is misleading. The label "account settings" should be displayed in the page header. I'm working on a proposal that eradicates the need to change the main navigation for different settings areas (org, space, account). For now, I suggest we use the breadcrumb in the header to show the account settings label and remove it from the main navigation.

damann:
> @damann the string is not hardcoded in there, it only allows us to also hand over strings.

Thanks. That’s fine, but there shouldn’t be any text in that place that resembles the environment switcher. I hadn’t specified this earlier as I only came across it recently, and it got buried under other priorities 🙇. As mentioned, I’m working on a better alternative, but for now, the interim solution should be to remove or hide it.

PR #2871, opened 2024-09-12
feat: add CDN

andipaetzold:
Creating a CDN for Forma using Fastly Compute to host the Geist font: https://vercel.com/font.

Adds a GH action that auto-deploys the CDN when merging to `main`.

The files will be deployed as `https://cdn.f36.contentful.com/*`. You can try an example deploy with an older font version: https://cdn.f36.contentful.com/fonts/geist-sans-1.3.0.woff2


packages/cdn/fastly.toml
```diff
@@ -0,0 +1,11 @@
+# This file describes a Fastly Compute package. To learn more visit:
+# https://developer.fastly.com/reference/fastly-toml/
+
+description = "Forma 36 CDN"
+language = "javascript"
+manifest_version = 2
+name = "@contentful/f36-cdn"
```
lewiscowper, line 7:
```suggestion
name = "f36-cdn"
```

I think with the `@` and `/` it might fail to successfully deploy.

Lelith (approved):

.github/workflows/deploy-cdn.yml
```diff
@@ -0,0 +1,21 @@
...
+    branches:
+      - main
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install dependencies
+        run: npm ci --workspace=@contentful/f36-cdn
```
denkristoffer, line 14: Today I learned.

denkristoffer (approved):

andipaetzold:
> Just a general question but shouldn't a CDN generally be served from a different domain to avoid wasting bytes on cookies from the web app?

Yes, to save every single byte, we would need to use a totally separate domain. For the Web App itself, the HTML is loaded from app.contentful.com and the static files from static.contentful.com to reduce the number of cookies. Not all cookies have strict same site policy, so they are also sent to static.contentful.com.

Fonts are only loaded once per page load (+within apps) and are not render blocking. Having a separate domain would have an insignificant performance improvement. By using a `contentful.com` subdomain, customers inspecting the network traffic know that the traffic goes to Contentful and not a third party service.

PR #2874, opened 2024-09-12
chore: add pre releases config and job

massao:
# Purpose of PR

We want to make it possible to automatically release new alpha versions on the next branch.


.changeset/pre.json
```diff
@@ -0,0 +1,54 @@
...
+    "@contentful/f36-icon": "5.0.0-alpha.20",
+    "@contentful/f36-icons": "5.0.0-alpha.31",
+    "@contentful/f36-image": "4.68.1",
+    "@contentful/f36-layout": "5.0.0-alpha.15",
+    "@contentful/f36-list": "4.68.1",
+    "@contentful/f36-menu": "4.68.1",
+    "@contentful/f36-modal": "4.68.1",
+    "@contentful/f36-multiselect": "4.24.0",
+    "@contentful/f36-navbar": "5.0.0-alpha.35",
+    "@contentful/f36-navlist": "4.1.0-alpha.1",
```
massao, line 30: This and the progress-stepper won't be bumped to the v5 when leaving the prerelease mode

.circleci/config.yml
```diff
@@ -139,6 +139,33 @@ jobs:
           command: |
             npm run-script changeset:publish
 
+  release_next:
```
massao, line 142: We need to also add these changes to the main branch, right?

RELEASES.md
```diff
@@ -86,3 +86,79 @@ For that we need to take some precautions:
...
+
+### Differences in the CI
+
+The main difference in the release jobs on the CI, is that in the `next` branch we don't generate the `.changelogrc` file and don't generate the github releases.
+
+Other than that the proccess is the same:
+
+1. Build the packages
+2. Update the versioning
+3. commit the updated package.json file
```
massao, line 164: Let me know if something is not clear, or if you want to change something

scripts/prerelease.mjs
```diff
@@ -40,6 +40,24 @@ async function main() {
...
+    )} branch, you don't need to run this script.\n\n`,
+  );
+  const { skip } = await inquirer.prompt([
+    {
+      name: 'skip',
+      message: 'Do you want to stop the script?',
+      type: 'confirm',
+    },
+  ]);
+  if (skip) return;
```
massao, line 59: I added this to discourage the usage of the script, or should we entirely remove it??

.circleci/config.yml
denkristoffer: I don't think so, isn't it only Vault that only works on `main`?

.changeset/config.json
```diff
@@ -1,70 +1,13 @@
...
-      "@contentful/f36-tooltip",
-      "@contentful/f36-typography",
-      "@contentful/f36-components",
-      "@contentful/f36-image",
-      "@contentful/f36-avatar"
-    ]
-  ],
   "access": "public",
-  "baseBranch": "main",
+  "baseBranch": "next",
```
Lelith, line 7: ❓ these changes are only for the time while its a next branch or? So we will not merge it back to the main branch once we have the V5 finalized?

RELEASES.md
```diff
@@ -86,3 +86,79 @@ For that we need to take some precautions:
...
 You can check the script [here](https://github.com/contentful/forma-36/blob/c6b10071959a085b21e49f5411a5ebff2f8a70d6/scripts/prerelease.mjs)
+
+## Next branch
+
+The next branch is used for the new version of Forma 36, which will probably have breaking changes. To help the development of the new packages, we have setup an automatic publishing of the packages on that branch to the `alpha` tag.
+These packages aren't shown on the website as they can have breaking changes often.
+
+### How to automatic publish
+
+To make changeset publish a new version of the packages it's the same as in the `main` branch, it just requires having a changeset generated and specifing the packages.
```
Lelith, line 97: ❓ will this also be the `npx changeset `command? Maybe would be good to add the exact command then also here in this seciton

.changeset/config.json
massao: This is because sadly we can't have mixed prereleases with releases, so we can only auto publish prereleases from the next branch.
And when merging it back to the main branch from the next, it would mean moving it to stable, for the main branch to auto publish it as stable, that I described on the RELEASES.md.

For alpha packages on the main branch we would still need to manually release sadly, or have two PRs, one pointing to the next branch to have the package published and one to the main branch to add the documentation.

RELEASES.md
massao: Yeah, that is the same command, I'll add the command on the documentation to make it clear

cf-remylenoir (commented):
Let's have a quick tour of these changes together, that would help me review better.

RELEASES.md
```diff
@@ -86,3 +86,122 @@ For that we need to take some precautions:
...
+# To test how it would update the changelogs and how it would behave as in main branch, replace the .changeset/config.json file with the one from the main branch, before running changeset version
+
+npx changeset version;
+# This will remove the pre.json file, and update all package.json files that need to be updated to the version without the -alpha tag.
+
+```
+
+### Changeset configuration
+
+To make it possible we take advantage of changeset feature of prereleasing, to make it possible we have the [.changeset/pre.json](./.changeset/pre.json) which specifies which tag to use and the initial version of the packages, which is used to calculate the alpha version, for example:
```
cf-remylenoir, line 145:
```suggestion
We use the pre-releasing changeset feature and the [.changeset/pre.json](./.changeset/pre.json), which specifies which tag to use and the packages' initial versions. It allows to control the alpha version bump, for example:
```

RELEASES.md
```diff
@@ -86,3 +86,122 @@ For that we need to take some precautions:
...
+  "fixed": [] // We don't want to have fixed or linked versions of alpha releases
+  "baseBranch": "next", // The branch which changesets uses to check versions
+  "bumpVersionsWithWorkspaceProtocolOnly": true, // Avoid bumping dependencies on other packages
+}
+```
+
+Notes:
+
+- The changelog is not updated because when we exit the prerelease state the changeset will compile all the changesets and apply onto the new version. eg. 5.0.0 will have all the changesets of the 5.0.0-alpha.\*
+- The `bumpVersionsWithWorkspaceProtocolOnly` makes it possible to avoid bumping packages on the packages that are dependent on the pacakge being updated, this is because we want alpha packages to depend on stable packages, rather than have all of them depending on alpha packages. ie. we would make all packages bump to an alpha core in case we update it.
```
cf-remylenoir, line 177:
```suggestion
> [!NOTE]
> 
> - The changelog is not updated because after exiting the prerelease state, all the changesets will be compiled and applied to the new version. For example, `5.0.0` will have all the changesets of the `5.0.0-alpha.\*`
> - The `bumpVersionsWithWorkspaceProtocolOnly` prevents bumping packages' versions after a package they depend on is updated. This is because we want alpha packages to depend on stable packages. We would bump all packages to an alpha core if we allow it.
```

RELEASES.md
```diff
@@ -86,3 +86,122 @@ For that we need to take some precautions:
...
+# Always select major change
+```
+
+Differently from the `main` branch, neither the packages changelog or the **What's new page** will be updated, and the changeset files will not be removed.
+
+### Testing how the version would change locally
+
+It's possible to test how the version would change locally, make sure that your last changes are commited, and you can run the following commands depending on what needs to be tested.
+
+Remember to revert the changes made through the `npx changeset version` and not commit them.
```
cf-remylenoir, line 110:
```suggestion
After committing your work, test the version's change by running these commands.

> [!WARNING]
> Remember to revert the changes made through the `npx changeset version` and not commit them.
```

cf-remylenoir (commented):
Thank you for expanding the documentation about releases, I commented on some editorial nits ✍️

PR #2880, opened 2024-09-18
fix(Avatar): avatar size and loading skeleton

cf-remylenoir:
# Purpose of PR

Fixes the Avatar's loading skeleton rendering.

We extend the size property to receive custom sizes. This should prevent people from using custom styling that does not deeply apply to the Avatar > Image > Skeleton loader component, see below:

**Current rendering with custom styling**

https://github.com/user-attachments/assets/2cfe3e24-d050-4745-a06d-739236e8bb25

## Loading skeleton

> [!TIP]
> One can observe the loading skeleton before/after in the deployed Storybook by blocking the request to the avatar image:
> - In the Avatar stories
> - In the Navbar stories

**Before**

https://github.com/user-attachments/assets/1a4c6edc-870b-45c6-a7ff-a65fa9bdd64a

**After**

https://github.com/user-attachments/assets/fde616f5-b3e9-488c-b890-f0772b043c3d

## Storybook update

- Add loading state to all `Loading` section
- Includes custom size via props examples

https://github.com/user-attachments/assets/477eba63-f937-4aff-9fa2-5116d4fb6296

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/components/avatar/src/Avatar/Avatar.tsx
+++ b/packages/components/avatar/src/Avatar/Avatar.tsx
@@ -10,7 +10,12 @@ import {
 } from '@contentful/f36-tooltip';
 
 import { getAvatarStyles } from './Avatar.styles';
-import { getSizeInPixels, type Size, type ColorVariant } from './utils';
+import {
+  getSizeInPixels,
+  type ColorVariant,
+  type Size,
+  type SizeInPixel,
+} from './utils';
 
 export type Variant = 'app' | 'user';
 
@@ -21,10 +26,10 @@ export interface AvatarProps extends CommonProps {
    */
   isLoading?: boolean;
   /**
-   * Use the available sizes or a numerical custom one, e.g. '52' or '52px'
+   * Use the available sizes or a numerical custom one, e.g. '52px'
    * @default 'medium'
    */
-  size?: Size | string;
+  size?: Size | SizeInPixel;
   initials?: string;
   src?: ImageProps['src'];
   /**
--- a/packages/components/avatar/src/Avatar/Avatar.tsx
+++ b/packages/components/avatar/src/Avatar/Avatar.tsx
@@ -9,10 +9,8 @@ import {
   type WithEnhancedContent,
 } from '@contentful/f36-tooltip';
 
-import { convertSizeToPixels, getAvatarStyles } from './Avatar.styles';
-import type { ColorVariant } from './utils';
-
-export type Size = 'tiny' | 'small' | 'medium' | 'large';
+import { getAvatarStyles } from './Avatar.styles';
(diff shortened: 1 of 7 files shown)


damann (approved):

packages/components/avatar/src/Avatar/Avatar.tsx
```diff
@@ -23,9 +21,10 @@ export interface AvatarProps extends CommonProps {
    */
   isLoading?: boolean;
   /**
+   * Use the available sizes or a numerical custom one, e.g. '52' or '52px'
    * @default 'medium'
    */
-  size?: Size;
+  size?: Size | string;
```
denkristoffer, line 27: I would rather adhere to the sizes we have in our product than introduce this to break the rules 🤷

damann: Without having a deep understanding of it, I’m not seeing the value in allowing overrides for images and skeletons. Could you provide an example of when these overrides would be necessary?

cf-remylenoir: The account setting page uses the Avatar component, with a custom size via CSS override, which results in this 
https://github.com/user-attachments/assets/2cfe3e24-d050-4745-a06d-739236e8bb25

With the proposed changes, the size can be specified via the Avatar props (e.g. `52px`), does not require CSS override, and is propagated to the Image + skeleton loader components.

andipaetzold: > I would rather adhere to the sizes we have in our product than introduce this to break the rules 🤷

I generally agree with that. Most other components don't allow that kind of override, but we limit users to specific values.

In this case, CSS overrides are not straight forward. The size impacts multiple CSS properties and relies on knowing the internal structure.

With the proposed type, we loose all autocompletion for the property. I would therefore go for `Size | `${number}px` or similar. That way, the IDE still suggests the official sizes.

Alternatively, we could allow passing a custom size but hide that from the public API.

cf-remylenoir: Good point for the typing, template literal is a better approach.

packages/components/avatar/src/Avatar/utils.ts
```diff
@@ -33,3 +39,42 @@ export function applyMuted(color: string): string {
...
+
+/**
+ * Utility function to convert the given size variant/custom size to pixels
+ *
+ * @param size
+ * @returns The variant or custom size in pixels, e.g. '32px'
+ */
+export function getSizeInPixels(
+  size: AvatarProps['size'],
+): AvatarProps['size'] {
```
andipaetzold, line 78:
```suggestion
): SizeInPixel {
```

PR #2898, opened 2024-10-04
fix(Avatar): don't overlap image with color border

andipaetzold:
# Purpose of PR

In the current version of the avatar component, the colored border overlaps with the image. The colored border is about 2 or 3px wide, so a total of 6px horizontal and vertical could be covered. With this PR, the size of the image is reduced to reduce overlaps.

The existing implementation was mixing size numbers with size strings. To shrink the image correctly, its size must be calculated. I, therefore, changed methods to always return numbers and to only add the `px` suffix at the very last moment. Imo, it's now way easier to understand when a variable is a number or a string.

Additionally, the overlay icon is round again.

## Screenshot

image

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/components/avatar/src/Avatar/Avatar.styles.ts
+++ b/packages/components/avatar/src/Avatar/Avatar.styles.ts
@@ -37,6 +37,9 @@ export const getAvatarStyles = ({
   colorVariant: ColorVariant;
 }) => {
   const borderRadius = variant === 'app' ? APP_BORDER_RADIUS : '100%';
+
+  // The inner border radius is smaller than the outer border radius
+  // See https://github.com/webuild-community/advent-of-sharing/blob/main/2022/day-06.md
   const innerBorderRadius =
     variant === 'app'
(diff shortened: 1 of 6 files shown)


packages/components/avatar/src/Avatar/Avatar.styles.ts
```diff
@@ -48,23 +59,29 @@ export const getAvatarStyles = ({
...
-      width: finalSize,
-      svg: {
-        borderRadius,
-      },
+      padding: getTotalBorderWidth(colorVariant),
+
+      // color variant border
       '&::after': {
-        borderRadius,
+        borderRadius: borderRadius,
```
cf-remylenoir, line 84: Nit (or left out for differentiation with `innerBorderRadius`?)

```suggestion
        borderRadius,
```

cf-remylenoir (approved):
Looking good. Remember to accept the new Chromatic snapshots.

PR #2906, opened 2024-10-09
fix: security vulnerability 130

Lelith:
This removes an unused dev-dependency that introduces a possible security vulnerability. https://github.com/contentful/forma-36/security/dependabot/130

### DIFF
--- a/packages/website/package.json
+++ b/packages/website/package.json
@@ -65,7 +65,6 @@
     "@emotion/babel-preset-css-prop": "^10.2.1",
     "@next/eslint-plugin-next": "^13.1.6",
     "babel-plugin-emotion": "10.2.2",
-    "broken-link-checker": "^0.7.8",
     "dotenv": "^16.0.3",
     "prettier": "^2.8.8",
     "ts-node": "^10.9.1",
--- a/package-lock.json
+++ b/package-lock.json
@@ -15351,50 +15351,6 @@
         "node": ">=4"
       }
     },
-    "node_modules/bhttp": {
-      "version": "1.2.8",
-      "dev": true,
-      "license": "WTFPL OR CC0-1.0",
-      "dependencies": {
-        "bluebird": "^2.8.2",
-        "concat-stream": "^1.4.7",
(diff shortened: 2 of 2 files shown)


packages/website/package.json
cf-remylenoir, line 1: Aren't we using this package on the CI for the `check-links` step? https://github.com/contentful/forma-36/blob/6e0cf8593d91b185383422fcc05626688c5fe7be/.circleci/config.yml#L90-L92

The `blc` command is coming from this package https://www.npmjs.com/package/broken-link-checker

cf-remylenoir: But the step passes on the CI here, even without the package. I wonder if it runs another package called `blc` (https://www.npmjs.com/package/blc) but I assume it would fail. 🤔

andipaetzold: The output in CI is very different from current `main`. So, I'd also assume a different package is installed and used here.
To avoid that issue, we should not use `npx` in CI, but npm run ...` instead.

denkristoffer: Wouldn't `npx` ask for confirmation before running something uninstalled though? Maybe it's different in CI environments.

Lelith: interestingly, this CI steps breaks on the next PR 
https://app.circleci.com/pipelines/github/contentful/forma-36/13264/workflows/df132e92-3fb6-4f89-993b-fe5048bd067b/jobs/37315?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary

denkristoffer: I think our npm cache on Circle is set up wrong then, it shouldn't be sharing the cache between branches 🤔

Lelith: i am going to close these two prs for now, as my main goal was to address the high security vulnerabilities. will follow up on the rest at another time.

andipaetzold: @denkristoffer With `CI=true` `npx` doesn't require a confirmation. To prevent that, you can use `--no-install`

PR #2907, opened 2024-10-09
chore: remove unneeded dependency pin for request

Lelith:
Removes version overwrite for "request" package dependency, which will fix https://github.com/contentful/forma-36/security/dependabot/126


andipaetzold (approved):

Lelith:
closing as there might be a bigger underlying issue on our CI setup. we will follow up at a later point

PR #2916, opened 2024-10-18
chore: less jest-axe configuration

denkristoffer:
# Purpose of PR

Clean up some jest-axe configuration which makes it possible to get rid of our `axeHelper` and do `import { axe } from 'jest-axe';` directly in tests.


cf-remylenoir (approved):
Nice, thanks!

PR #2927, opened 2024-11-01
fix(get-string-match): trim strings

denkristoffer:
# Purpose of PR

There's a case where `getStringMatch` returns no matches


denkristoffer:
Handled in https://github.com/contentful/forma-36/pull/2928/

PR #2928, opened 2024-11-04
feat(multiselect): handle items without match & trim search value

andipaetzold:
# Purpose of PR

Handles two edge cases:

1. When the search value doesn't match the option text

E.g. when an option with the label `Apple` is rendered while the search value is `Lemon`. Currently, the multiselect item is rendered without any label - only the checkbox. With this PR, the option option label is rendered without any highlighting.

2. The search value has leading or trailing white space

Currently,  when an option has the label `Apple` and the search value is ` pp ` (not the white space) only the checkbox is rendered without any text. With the fix mentioned above, `Apple` is rendered without any highlighting. By internally trimming the search value, `Apple` is rendered with `pp` highlighted.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/.changeset/neat-lizards-remember.md
+++ b/.changeset/neat-lizards-remember.md
@@ -2,4 +2,4 @@
 '@contentful/f36-multiselect': minor
 ---
 
-improve option text highlighting by trimming the search value
+improve option label highlighting by trimming the search value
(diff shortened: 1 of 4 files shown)


andipaetzold:
> > E.g. when an option with the label `Apple` is rendered while the search value is `Lemon`. Currently, the multiselect item is rendered without any label - only the checkbox. With this PR, the option option label is rendered without any highlighting.
> 
> Wouldn't it render the no matches message?

That message is only shown when no item matches at all.

massao (approved):

PR #2933, opened 2024-11-12
fix: implement type safety fallback

Lelith:
With some package update, typescript errors got interpreted as more strict. This implements the slug handling in a typesafe way with fallbacks.

### DIFF
--- a/packages/website/pages/[...slug].tsx
+++ b/packages/website/pages/[...slug].tsx
@@ -295,9 +295,12 @@ export const getStaticPaths: GetStaticPaths = async () => {
           ?.sectionCollection?.items[0];
     }
 
+    const sectionSlug = section?.slug ?? 'default-section';
+    const itemSlug = item?.slug ?? 'default-item';
     const slug = item.authProtected
-      ? [section.slug, 'protected', item.slug]
-      : [section.slug, item.slug];
+      ? [sectionSlug, 'protected', itemSlug]
+      : [sectionSlug, itemSlug];
+
     return {
       params: {
         slug,


packages/website/pages/[...slug].tsx
```diff
@@ -295,9 +295,12 @@ export const getStaticPaths: GetStaticPaths = async () => {
           ?.sectionCollection?.items[0];
     }
 
+    const sectionSlug = section?.slug ?? 'default-section';
+    const itemSlug = item?.slug ?? 'default-item';
```
Comment (denkristoffer) [inline packages/website/pages/[...slug].tsx:299]: Do these exist?

Comment (Lelith) [reply packages/website/pages/[...slug].tsx:299]: they usually do, but the type script error is now blocking the build, so i implemented a quick fallback to avoid the error.
the issue is that neither item nor section are properly typed and so this is the point where it falls trough. i can also go ahead and do a wider refactoring, fixing the initial types.

Comment (denkristoffer) [reply packages/website/pages/[...slug].tsx:299]: I mean do these default sections and items exist or will something break if they actually got used?

Comment (Lelith) [reply packages/website/pages/[...slug].tsx:299]: Ahh, no, just the slug name and item would be displayed like that.

cf-remylenoir:
Are we able to pinpoint what package bump caused this regression? 
Looking at the latest merged pull requests, I see that they were successful.

Lelith:
@denkristoffer @cf-remylenoir i am also not sure, i was trying it out with the latest master where it failed immediately. and also versions above, so maybe its not an update in this package but one on circle ci? i mean its "just" a linter issue, so it seems like it could be based on some node or npm version or something like that.

PR #2934, opened 2024-11-18
feat: export media queries

Lelith:
# Purpose of PR

This exports the media queries used in the navbar in order to also make them available in elements used inside the navbar.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -42050,7 +42050,7 @@
     },
     "packages/components/icons": {
       "name": "@contentful/f36-icons",
-      "version": "5.0.0-alpha.33",
+      "version": "5.0.0-alpha.34",
       "license": "MIT",
       "dependencies": {
         "@contentful/f36-core": "^4.67.3",
@@ -42382,7 +42382,7 @@
     },
     "packages/components/navbar": {
       "name": "@contentful/f36-navbar",
-      "version": "5.0.0-alpha.37",
+      "version": "5.0.0-alpha.39",
       "license": "MIT",
       "dependencies": {
         "@contentful/f36-avatar": "^4.74.1",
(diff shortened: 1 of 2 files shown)


veu (approved):

veu (approved):

packages/components/navbar/src/NavbarSwitcher/NavbarSwitcher.styles.ts
```diff
@@ -14,6 +14,19 @@ export const getNavbarSwitcherStyles = ({
   text: css({
     color: 'currentcolor',
     fontWeight: 'inherit',
+    lineHeight: 'unset',
+    'span:first-child': {
```
cf-remylenoir, line 18: It appears we are really only targeting the start `<span>` here.

Looking at the usage of `styles.text` from the condition @ L-88-97 - https://github.com/contentful/forma-36/pull/2934/files#diff-78e584a28d5563d6442791d3f7b1f32d1c9eae70a296d7f83293ca73c1810d8fR88-R97

If we don't have a `start`, then the `value` equals a string without `<span>` because we pass the `space` string to the `<SwitcherLabel value={space} />`.

If confirmed, let's move it to a new key for better understandability.

cf-remylenoir (dismissed):

packages/components/navbar/src/Navbar.tsx
```diff
@@ -32,6 +32,9 @@ type NavbarOwnProps = CommonProps & {
   /** Navigation displayed on mobile versions */
   mobileNavigation?: React.ReactNode;
 
+  /** breakpoint to determine when to show the mobile navigation */
+  mobileNavigationBp?: 'small' | 'medium';
```
cf-remylenoir, line 36: What would be the deciding factor to select one or the other? The number of nav items (i.e. the organization settings)?

Lelith: yes, for the organization settings navigation, there is already not enough space for the switcher with the main navigation.

cf-remylenoir: We can revisit this logic once we support container queries, we should be able to make it fully dynamic based on the surrounding elements 🤞

cf-remylenoir (approved):
QA looking good 👍

PR #2935, opened 2024-11-18
fix: hide only endIcon when Button is in loading state

bgutsol:
# Purpose of PR

- we don't need to hide `startIcon` when Button is in loading state.


Lelith (approved):
Looks good to me!

Issue #2339, opened 2022-11-24
🐛 Bug - Text component should not take "color" prop


denkristoffer:
# Forma 36 bug report

## Summary

The `Text` component should not take the prop `color` but instead throw a type error. I keep using `color` instead of `fontColor` because I'm silly and I never get an error, but I don't get the result I expect either 🙁 The prop type comes from the fact that we're extending the default HTML attributes. We should probably `Omit` `color`. 

Example: https://f36.contentful.com/playground?code=N4Igxg9gJgpiBcICWBbADhATgFwATFwDEAbGADwBpcAVcvAX1wDNMIVcByAAUgDtsY-JgFdiAeiYBmAGwBaSOgi9B2AM4cA3AB1eO8hhy5YTAIai8I3mGxIluAIJo0ACgCU+HblyYY2YZl5cAB4ScmZSMgARJB9rW14AXi1wCGJhFF0QAD5PL2DaMjxIYiwkkB8oAFYABmrkrIBJXFUACwhRKFxsFtYAd1wTLoBPNBhcGExWTABCILEC7BzAvKCF5iVsAGFU0uSKmrrsptb24k6AIzGK2fm6Ja850LIl+hB6IA

Wake1st:
@denkristoffer, could you provide an example of how you're passing in `color` - is it via `style` prop?

denkristoffer:
Thanks for requesting that @Wake1st, I've added an example to the issue.

PR #2947, opened 2024-12-09
fix: omit color prop to avoid using it over fontColor

Lelith:
Text components accepted the "color" default prop, but it was actually never used as we favor fontColor as property. This change reflects this behavior now in the Type and helps avoid this mistake.

### DIFF
--- a/packages/components/typography/src/Text/Text.tsx
+++ b/packages/components/typography/src/Text/Text.tsx
@@ -47,10 +47,13 @@ function wordBreakStyle() {
 }
 
 export type TextProps<E extends React.ElementType = typeof TEXT_DEFAULT_TAG> =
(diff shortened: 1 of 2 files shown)


cf-remylenoir:
Fixes https://github.com/contentful/forma-36/issues/2339 :)

packages/components/typography/src/Text/Text.tsx
cf-remylenoir, line 1: There is a type mismatch we need to prevent, it happens when spreading the props of some of the typography components to the Text component, I could experience it for:
- the Caption
- the DisplayText
- the Heading
- the SectionHeading
- the Subheading

Lelith:
closing in favor of: https://github.com/contentful/forma-36/pull/2952

PR #2952, opened 2024-12-10
feat: extend props and fix spread operator for typography components

Lelith:
The components
- Caption
- DisplayText
- Heading
- Paragraph
- SectionHeading
- Subheading

Where all having an type issue, when using the spread operator. This fixes this issue, stemming from the "as" property. It also allows now to overwrite `fontColor` and `marginBottom` to allow for easier variations of these components.

### DIFF
--- a/.changeset/olive-bikes-pump.md
+++ b/.changeset/olive-bikes-pump.md
@@ -1,0 +1,5 @@
+---
+'@contentful/f36-typography': minor
+---
+
+Enables overwriting of fontColor and marginBottom props on Typography components. Also fixes a Type Issue with the as property.
--- a/.changeset/olive-bikes-pump.md
+++ b/.changeset/olive-bikes-pump.md
@@ -2,4 +2,6 @@
 '@contentful/f36-typography': minor
 ---
 
-Enables overwriting of fontColor and marginBottom props on Typography components. Also fixes a Type Issue with the as property.
+Enables overwriting of fontColor and marginBottom props on Typography components: Caption, DisplayText, Heading, Paragraph, SectionHeading, Subheading
+Omits 'color' default property from Text component, in favor of 'fontColor' property.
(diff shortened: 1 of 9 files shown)


packages/components/typography/src/Caption/Caption.tsx
```diff
@@ -1,18 +1,16 @@
...
 import { FontWeightTokens } from '@contentful/f36-tokens';
 import type {
-  CommonProps,
-  MarginProps,
   PolymorphicComponent,
   PolymorphicProps,
   ExpandProps,
 } from '@contentful/f36-core';
-import { Text } from '../Text';
+import { Text, TextProps } from '../Text';
```
cf-remylenoir, line 8: Could you please add the type prefix to all `TextProps` imports you've made?

```suggestion
import { Text, type TextProps } from '../Text';
```

.changeset/rich-geckos-doubt.md
cf-remylenoir, line 1: Do you want to close #2947, since your changes are part of this PR too? Or do you prefer to keep both PRs?

Lelith: closed the other

.changeset/olive-bikes-pump.md
cf-remylenoir, line 1: I am unsure if both updates should be released under a single patch changelog. 

@massao from experience is it gonna trigger 2 consecutive version bumps and release or only one?

Lelith: I can remove the first changelog entry and we do one minor changelog with both changes? i can change the description to be more explicit that one is about typescript and one is about the new property overwrites.

cf-remylenoir: Sounds good

cf-remylenoir (approved):

PR #2955, opened 2024-12-12
feat: update Notification to support React 18

tkers:
# Purpose of PR

The Notification package is currently broken when upgrading to React 19 (and somewhat breaks the use of React 18) because it uses the deprecated `render` function instead of the newer `createRoot()` method of mounting a component. 

I figured keeping support for previous React versions is desired so I added a conditional import to use this new method when running React 18 or higher.

## PR Checklist

- [ ] Tests are passing


Lelith:
Hey @tkers,
thank you so much for your contribution. We are currently in preparation for a new major release and will include a full upgrade to react18 in that version. 
This is why i am closing this PR for now. 

Best
Kathrin

PR #2968, opened 2025-01-07
fix(menu): don’t focus menu if menu has just been removed

veu:
# Purpose of PR

Fix edge case where a menu is opened and removed from the DOM at the same time before we can focus it in the next tick.

## PR Checklist

- [x] Tests are passing


Lelith (approved):
Looks good to me!

PR #2973, opened 2025-01-13
feat: Add onClick optional property to ProgressStepper

leeeandroo:
feat: Add onClick optional property to `ProgressStepper`

# Purpose of PR

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/components/progress-stepper/src/Step/Step.tsx
+++ b/packages/components/progress-stepper/src/Step/Step.tsx
@@ -97,6 +97,7 @@ function _Step(props: StepProps, ref: React.Ref<HTMLLIElement>) {
 
     return isClickable ? (
       <button
+        type="button"
         className={classNames}
         onClick={(e) => handleStepClick(e, stepNumber)}
       >
--- a/packages/components/progress-stepper/src/Step/Step.tsx
+++ b/packages/components/progress-stepper/src/Step/Step.tsx
@@ -75,6 +75,18 @@ function _Step(props: StepProps, ref: React.Ref<HTMLLIElement>) {
     }
   };
 
+  const getStepText = () => {
+    if (labelText) {
+      return labelText;
+    }
+
+    if (stepStyle === 'number') {
+      return `Step ${stepNumberToDisplay}`;
+    }
+
+    return state.charAt(0).toUpperCase() + state.slice(1);
+  };
+
   const handleStepClick = (
     e: React.MouseEvent<HTMLElement>,
     stepNumber: number,
@@ -100,6 +112,7 @@ function _Step(props: StepProps, ref: React.Ref<HTMLLIElement>) {
         type="button"
         className={classNames}
         onClick={(e) => handleStepClick(e, stepNumber)}
+        aria-label={getStepText()}
       >
         {content}
       </button>
--- a/packages/components/progress-stepper/src/Step/Step.tsx
+++ b/packages/components/progress-stepper/src/Step/Step.tsx
@@ -96,13 +96,12 @@ function _Step(props: StepProps, ref: React.Ref<HTMLLIElement>) {
     const isClickable = onClick && state !== 'disabled';
 
     return isClickable ? (
-      <a
-        href="#"
+      <button
         className={classNames}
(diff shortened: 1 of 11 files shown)


packages/components/progress-stepper/src/Step/Step.tsx
```diff
@@ -68,6 +75,39 @@ function _Step(props: StepProps, ref: React.Ref<HTMLLIElement>) {
...
+      [styles.complete]: state === 'complete',
+      [styles.error]: state === 'error',
+      [styles.warning]: state === 'warning',
+    });
+
+    const content = renderStep();
+    const isClickable = onClick && state !== 'disabled';
+
+    return isClickable ? (
+      <button
```
whitelisab, line 99: When this is changed to a button element, it is now affecting the spacing of the numbers/icons within the step circle. When the `stepStyle` is number, the change is ever so slight, but the number is no longer centered:

| Not clickable (span element)   | Clickable (button element) |
| -------- | ------- |
| Screenshot 2025-01-13 at 11 00 32 PM | Screenshot 2025-01-13 at 11 00 26 PM  |

Also, when the component uses the icon `stepStyle`, the impact is more noticable:

| Not clickable (span element)   | Clickable (button element) |
| -------- | ------- |
| Screenshot 2025-01-13 at 11 01 21 PM | Screenshot 2025-01-13 at 11 01 11 PM |

whitelisab (commented):
Overall this approach looks great! ⭐ I added a comment about the impact of changing the step from a span to a button on the step styles (numbers and icons are no longer centered). I also think it would be great to add another example to the [component documentation](https://f36.contentful.com/components/progress-stepper) showing how to use this new `onClick` prop.

packages/components/progress-stepper/src/Step/Step.tsx
leeeandroo: Thanks for checking @whitelisab !
I could not simulate the misalignment on that level you show in the screenshots. But I did some changes, could you please check it again? (And let me know how you saw it in this way, so next time I can check it properly)

leeeandroo:
> also think it would be great to add another example to the [component documentation](https://f36.contentful.com/components/progress-stepper) showing how to use this new `onClick` prop.

Added the example page. Let me know if something is missing.

leeeandroo:
> @leeeandroo I tried the chromatic generated storybook, I think the labels below the step number should also be clickable. What do you think?

That's a good point @john-e , thanks for the suggestion. Added, please have a look again 🙏

packages/components/progress-stepper/src/Step/Step.tsx
cf-remylenoir: Introducing the button requires updating the focus style when navigating with the keyboard. 

https://github.com/user-attachments/assets/3f52d290-9a89-495e-a6df-7b8d8fcb2983

As you can see in the video, we don't see the step number focused state on steps with a dark blue border. All focusable elements across the Design System components use the `tokens.glowPrimary` style for their focus ring.

**With that said, here are my suggestions:**
- for the step number, use the `tokens.glowPrimary` instead of the current dark blue color
- for the label that becomes a button, use the Forma 36 Button component instead. You will inherit the right focus glow, it also ensures consistency with the use of actionable components across the app. It could be used with the `small` size and `transparent` variant, with additional custom styling to reduce its width/height, if necessary.

@damann do we also need to update the Figma component to introduce the onClick feature?

leeeandroo: > for the step number, use the tokens.glowPrimary instead of the current dark blue color

Done!

> for the label that becomes a button, use the Forma 36 Button component instead. You will inherit the right focus glow, it also ensures consistency with the use of actionable components across the app. It could be used with the small size and transparent variant, with additional custom styling to reduce its width/height, if necessary.

I have changed the button to the f36 Button component, but I have also added the `tabIndex` as `-1`, so user will only navigate to the step number/icon. Let me know if you think it is OK

cf-remylenoir (approved):
Looking good on my end 👍  Thank you for your contribution!

PR #2976, opened 2025-01-15
feat: Allow overriding onClick of ProgressStepper on child Steps

leeeandroo:
# Purpose of PR

To allow client to use the `onClick` functionality with a bit more of flexibility, I am forwarding now all props from child component to the cloned Step. 
In this way, now it is possible to instead of adding a global `onClick` to have it on `Step` level, allowing us to control for example that **only previous steps are clickable**.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/components/progress-stepper/src/ProgressStepper.tsx
+++ b/packages/components/progress-stepper/src/ProgressStepper.tsx
@@ -60,6 +60,7 @@ function _ProgressStepper(
         key: `steps-rendered-${index}`,
         stepNumber: index,
         onClick,
+        ...(child as React.ReactElement).props,
       });
       return stepChild;
     });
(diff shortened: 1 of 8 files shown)


packages/components/progress-stepper/src/ProgressStepper.tsx
```diff
@@ -60,6 +60,7 @@ function _ProgressStepper(
         key: `steps-rendered-${index}`,
         stepNumber: index,
         onClick,
+        ...(child as React.ReactElement).props,
```
leeeandroo, line 63: Forwarding `props` from child element to the actual cloned `Step`.

Lelith:
FYI: don't forget to add the changeset :) `npx changeset`

PR #2996, opened 2025-02-14
chore: filter out Dependabot from CDN deploy workflow

cf-remylenoir:
# Purpose of PR

Because Dependabot can't access the `FASTLY_API_TOKEN` secret, the workflow fails on all Dependabot merges to `main` and creates a lot of noise in our ci alerting channel.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/.github/workflows/deploy-cdn.yml
+++ b/.github/workflows/deploy-cdn.yml
@@ -7,6 +7,7 @@ on:
 
 jobs:
   deploy:
+    if: ${{ github.actor != 'dependabot[bot]' }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4


.github/workflows/deploy-cdn.yml
```diff
@@ -7,6 +7,7 @@ on:
 
 jobs:
   deploy:
+    if: ${{ github.actor != 'dependabot[bot]' }}
```
andipaetzold, line 10: Did you test this? Is the wrapping in `${{` and `}}` required?

cf-remylenoir: 1. I haven't tested it, as it is triggered when pushing to `main`. 
Do you recommend a testing tool? I came across [act](https://github.com/nektos/act).

2. I followed the documentation, which uses curly braces. It seems we use both syntaxes across our projects, with and without.

andipaetzold: Sounds good. Adding testing is overkill. I was just not sure about the syntax.

andipaetzold (approved):

veu (approved):

PR #3006, opened 2025-02-28
feat(multiselect): allow full control over option label rendering

denkristoffer:
# Purpose of PR

This PR adds support for custom label rendering in the `Multiselect.Option` component. This is useful in cases where you want to render more than just a string inside the label, however it also moves the responsibility of highlighting the values onto yourself. For that purpose I've also exported the `HighlightedItem` component.

It looks like this—notice the icon which even has a tooltip on hover.

Screenshot 2025-02-28 at 18 05 41


Lelith:
Looks all good so far to me, could you add a example of how to use it with a react child to the examples and in the storybook stories? so the usage is documented.

damann:
We'll also have to update the Figma library accordingly to include the `HighlightedItem` component

packages/components/multiselect/src/MultiselectOption.tsx
```diff
@@ -43,14 +58,18 @@ export const MultiselectOption = ({
           className={styles.optionText}
           data-test-id={`cf-multiselect-list-item-${itemId}`}
         >
-          <HighlightedItem item={label} inputValue={searchValue} />
+          {typeof label === 'string' ? (
+            <HighlightedItem item={label} inputValue={searchValue} />
+          ) : (
+            children
+          )}
```
cf-remylenoir, line 65: Using a React Node involves adding custom CSS to align the item label with whatever needs to be rendered,  This feels wrong, I would feel better with a Flex wrapper and default `center` alignment. What do you think?

image

denkristoffer: Yeah makes sense, I will add it. I hope it won't affect anything it shouldn't.

cf-remylenoir: Since the `label` padding already induces visual centering, this must be the preferred method for most of the use cases introduced by the custom React node prop.

cf-remylenoir:
For me it looks good, @damann let's coordinate the release with the Figma library. Wdyt?

cf-remylenoir:
Could you please update with the latest `main` before merging? It will remove one of the UI Tests unrelated to your changes, and some packages used for the changelog/release pipeline were updated.

cf-remylenoir (approved):
Code-wise looking good, waiting for @damann's approval

Lelith (approved):
thanks for the example updates

PR #3023, opened 2025-03-19
fix: auto reposition submenu when out of viewport [SPA-2627]

Chaoste:
# Purpose of PR

When rendering a `<Submenu>` close to the border of the window, it might render out of bounds as it's currently forced to not auto align itself.

## PR Checklist

- [x] Tests are passing

## Before

https://github.com/user-attachments/assets/644133cd-abcc-4578-9441-212f05a72336

## After

https://github.com/user-attachments/assets/b711559c-2868-48c3-b73f-ed6b51f7c8b8

### DIFF
--- a/.changeset/itchy-rice-kneel.md
+++ b/.changeset/itchy-rice-kneel.md
@@ -1,0 +1,5 @@
+---
+'@contentful/f36-menu': minor
+---
+
+Allow overwriting placement and isAutoalignmentEnabled for Submenu component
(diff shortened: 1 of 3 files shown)


Lelith:
@Chaoste thank you very much for taking the time to propose a solution. I think the best way forward to not accidentally break implementations that depend on the autoAlignment set to false, we should instead stop omitting the property and set false as the default behavior.

Also to be more flexible with the submenu we should do the same with the placement property. So allowing it as prop and set the `"right-start"` as default

cf-remylenoir (approved):

PR #3041, opened 2025-04-24
feat:remove hardcoded copy from navbar [UFO-1308]

Lelith:
# Purpose of PR
this enables full translatability of the Contentful platform and apps that are making use of the navbar component. All new props are optional and use the copy from before as their default values. 

There are no visual differences expected

### DIFF
--- a/packages/components/navbar/src/Navbar.tsx
+++ b/packages/components/navbar/src/Navbar.tsx
@@ -54,9 +54,9 @@ type NavbarOwnProps = CommonProps & {
   variant?: 'wide' | 'fullscreen';
 
   /**
-   * aria labels for the navbar
+   * aria labels for different areas of the navigation bar
    */
-  aria: {
+  aria?: {
     labelMainNavigation?: string;
     labelSecondaryNavigation?: string;
     labelPromotions?: string;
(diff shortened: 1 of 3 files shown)


packages/components/navbar/src/Navbar.tsx
```diff
@@ -35,6 +35,12 @@ type NavbarOwnProps = CommonProps & {
   /** breakpoint to determine when to show the mobile navigation */
   mobileNavigationBp?: 'small' | 'medium';
 
+  /**
+   * label of the mobile navigation trigger button
+   * @default 'Menu'
+   */
+  mobileNavigationLabel?: string;
```
Lelith, line 42: I'm actually thinking if it would be better to also group the properties for the navbar mobile navigation

cf-remylenoir: I don't have a strong opinion on this one, both approaches work for me.

cf-remylenoir (approved):

packages/components/navbar/src/Navbar.tsx
Lelith: i choose now to go for the nested properties to have them grouped together.

PR #3045, opened 2025-05-06
feat: enable localisation of pagination [UFO-1310]

Lelith:
# Purpose of PR

Enables the localisation of the Pagination component. Following new, optional Properties are introduced:

```
/**
   * Label for the View per page selector
   * @default "View"
   */
viewPerPageLabel?: string;

 
navigationButtonProps?: {
/**
   * Label for the next button
   * @default "Next"
   */
  nextLabel?: string;
  
/**
   * Label for the previous button
   * @default "Previous"
   */
  previousLabel?: string;
  
/**
   * Aria label for the next button
   * @default "To next page"
   **/
  nextAriaLabel?: string;

  /**
   * Aria label for the previous button
   * @default "To previous page"
   */
  previousAriaLabel?: string;
}


Lelith:
@cf-remylenoir i've slept now over the naming of these properties and I am no longer sure if these are the right names. 
For the close buttons, i have started using `closeButtonAriaLabel` 
so if i would follow that pattern, first describing the component, then describing the property, i think i need to switch around the names. 

so `navigationButtonProps.nextLabel` and `navigationButtonProps.nextAriaLabel`

that would also fit better in with `viewPerPageLabel`

But I am now doubting myself and would like your opinion as well.

cf-remylenoir:
> @cf-remylenoir i've slept now over the naming of these properties and I am no longer sure if these are the right names. For the close buttons, i have started using `closeButtonAriaLabel` so if i would follow that pattern, first describing the component, then describing the property, i think i need to switch around the names.
> 
> so `navigationButtonsProps.nextLabel` and `navigationButtonsProps.nextAriaLabel` instead of `navigationButtonsProps.labelNext` and `navigationButtonsProps.ariaLabelNext`
> 
> that would also fit better in with `viewPerPageLabel`
> 
> But I am now doubting myself and would like your opinion as well.

I think it is better this way and feels more natural in this order.

PR #3048, opened 2025-05-07
feat: enable i18n for Multiselect Component [UFO-1750]

Lelith:
# Purpose of PR

Enables localisation of the Multiselect Component by offering new optional properties.

For Multiselect Component
```
clearButtonProps = {
  /**
   * Aria label for the clear button
   * @default 'Clear selection'
   */
  ariaLabel?: string;

  /**
   * Tooltip for the clear button
   * @default 'Clear selection'
   */
  tooltip?: string;
};

/**
   * Aria label for toggle button that opens the list
   */
  toggleButtonAriaLabel?: string;
```

For Multiselect.SelectAllOption
```
selectAllOptionLabel? = {
  /**
   * Label for the select all option when it is checked
   * @default 'Deselect all'
   */
  checked: string;

  /**
   * Label for the select all option when it is unchecked
   * @default 'Select all'
   */
  unchecked: string;
};
```
No visual differences

### DIFF
--- a/packages/components/multiselect/src/Multiselect.tsx
+++ b/packages/components/multiselect/src/Multiselect.tsx
@@ -62,7 +62,8 @@ export interface MultiselectProps extends CommonProps {
   toggleRef?: React.Ref<HTMLButtonElement>;
 
   /**
-   * Aria label for toggle button that opens the list
+   * Aria label for the toggle button that opens the list
+   * @default 'Toggle Multiselect'
    */
   toggleButtonAriaLabel?: string;
 
--- a/packages/components/multiselect/src/Multiselect.tsx
+++ b/packages/components/multiselect/src/Multiselect.tsx
@@ -17,6 +17,20 @@ import FocusLock from 'react-focus-lock';
 import type { MultiselectSearchProps as SearchProps } from './MultiselectSearch';
 import { MultiselectSearch } from './MultiselectSearch';
 
+type ClearButtonProps = {
+  /**
+   * Aria label for the clear button
+   * @default 'Clear selection'
+   */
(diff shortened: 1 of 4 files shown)


packages/components/multiselect/src/Multiselect.tsx
```diff
@@ -17,6 +17,20 @@ import FocusLock from 'react-focus-lock';
 import type { MultiselectSearchProps as SearchProps } from './MultiselectSearch';
 import { MultiselectSearch } from './MultiselectSearch';
 
+type ClearButtonProps = {
+  /**
+   * Aria label for the clear button
+   * @default 'Clear selection'
+   */
+  ariaLabel?: string;
```
veu, line 25: Both properties are labeled as optional but there is no fallback for either of them, only for the whole object. If you overwrite only one, the other will be empty. Is that intended?

packages/components/multiselect/src/Multiselect.tsx
```diff
@@ -47,6 +61,11 @@ export interface MultiselectProps extends CommonProps {
    */
   toggleRef?: React.Ref<HTMLButtonElement>;
 
+  /**
+   * Aria label for toggle button that opens the list
+   */
+  toggleButtonAriaLabel?: string;
+
```
cf-remylenoir, line 68:
```suggestion
  /**
   * Aria label for the toggle button that opens the list
   * @default 'Toggle Multiselect'
   */
  toggleButtonAriaLabel?: string;

```

cf-remylenoir (approved):

cf-remylenoir:
The lock difference comes from #3052

PR #3051, opened 2025-05-14
feat: add i18n-utils package

veu:
Add utilities for formatting commonly used data types

# Purpose of PR

## PR Checklist

- [ ] Tests are passing


packages/f36-i18n-utils/src/formatters.ts
```diff
@@ -0,0 +1,66 @@
+import dayjs from 'dayjs';
+import 'dayjs/locale/de';
+import 'dayjs/locale/fr';
+import localizedFormat from 'dayjs/plugin/localizedFormat';
+
+dayjs.extend(localizedFormat);
+
+export function formatDate(locale: string, date: string): string {
+  return dayjs(date)
+    .locale(mapToDayjsLocale(locale))
```
cf-remylenoir, line 10: Following up from my [previous comment](https://github.com/contentful/forma-36/pull/3051/files#r2088927994), from reading the docs, I understand this is to activate the received locale? Still curious if `de` and `fr` would be dynamically imported.

veu: I’ve changed the imports to happen dynamically so nothing gets imported until the function is called. With the right setup it’s easy to split this into separate chunks (and even split up the locales separately if that ever becomes necessary).

packages/f36-i18n-utils/src/formatters.ts
```diff
@@ -0,0 +1,80 @@
+export async function formatDate(
+  locale: string,
+  date: string,
+): Promise<string> {
+  const dayjsLocale = mapToDayjsLocale(locale);
+  return Promise.all([
+    import('dayjs'),
+    importDayjsLocalePackage(dayjsLocale),
```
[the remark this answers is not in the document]
veu: Theoretically but I’m not sure there’s a bundler that could analyze that. We want to know the imported packages.

scripts/changesets/changelog-generate.js
```diff
@@ -15,6 +15,7 @@ const ignorePkgs = [
   '@contentful/f36-cdn',
   '@contentful/f36-website',
   '@contentful/f36-docs-utils',
+  '@contentful/f36-i18n-utils',
```
cf-remylenoir, line 18: We want this package to have changelogs, right?

PR #3108, opened 2025-06-17
fix: set SVG viewport to 24 24 to render correctly []

Chaoste:
# Purpose of PR

Three of the Icon SVGs use the viewport `0 0 13 18` while the icon actually covers 24x24 pixels.

Screenshot 2025-06-17 at 14 39 17
Screenshot 2025-06-17 at 14 39 29
Screenshot 2025-06-17 at 14 39 37

## PR Checklist

- [x] Tests are passing


Lelith:
@Chaoste while i think fixing this is needed, the icons are no longer supported for the next version. As alternative you should use the TextSubscriptIcon from v5 (looks like this) https://phosphoricons.com/?q=text-subscript

If we require this as custom svg icon, can you please add it instead to the next branch with the fixed svg box. So we already are future proofing the usage of this Icon.

Chaoste:
> @Chaoste while i think fixing this is needed, the icons are no longer supported for the next version. As alternative you should use the TextSubscriptIcon from v5 (looks like this) https://phosphoricons.com/?q=text-subscript
> 
> If we require this as custom svg icon, can you please add it instead to the next branch with the fixed svg box. So we already are future proofing the usage of this Icon.

Thanks @Lelith for calling out. We don't need those in our team. It's just something I noticed years ago and since then it jumped into my eye everytime I opened that page 😄 If we're gonna get rid of those anyways, no need to fix it.

Lelith (approved):

PR #3111, opened 2025-06-18
chore: [UFO-1809] Update all icons from v4 to v5

missating:
# Purpose of PR

## PR Checklist

- [ ] Tests are passing


packages/f36-docs-utils/package.json
```diff
@@ -21,7 +21,7 @@
   "dependencies": {
     "@contentful/f36-components": "^4.21.4",
     "@contentful/f36-tokens": "^4.0.0",
-    "@contentful/f36-icons": "^4.21.4",
+    "@contentful/f36-icons": "^5.0.0-alpha.44",
```
cf-remylenoir, line 24: After double-checking, the icons aren't used in this package.

package.json
```diff
@@ -134,5 +134,9 @@
   "volta": {
     "node": "18.20.1",
     "npm": "8.3.0"
+  },
+  "dependencies": {
+    "@contentful/f36-icons": "^5.0.0-alpha.44",
+    "@radix-ui/react-tabs": "^1.1.12"
```
cf-remylenoir: What is the reason we have these dependencies installed at the root?

missating: It was throwing an error locally after building that it needed those

Issue #3092, opened 2025-06-05
💡 Proposal -


0math:
# Forma 36 contribution proposal

## The problem 

**Total Number usage card is implemented differently on different pages**
Each Usage tab (Organization settings -> Usage) has its own implementation and styling of usage cards. This reveals the following issues:
- lack of unified approach to structure/style usage components, i.e. code base is difficult to understand, usage cards' UI is different in different places
- usage cards are not reusable (e.g. by other teams that need to add "Total Number" card)

These are current "Total Number" cards on different pages:

![Image](https://github.com/user-attachments/assets/fcda0023-2fe0-409e-8ea0-b98ce0594f80)
![Image](https://github.com/user-attachments/assets/eff1ee36-bdd2-4e04-9339-892fba0d820e)
![Image](https://github.com/user-attachments/assets/aa4549f3-dfd0-47ca-be8a-8a7381876a39)

## The proposed solution

Currently "Total Number" cards are structured and styles off of several Forma 36 components.
In order to address the both problems, "Total Number" can be a separate Forma36 component.

The component should be generic and should address all variations of usage from the examples above, i.e. used space in GB, words counts, consumption units etc.
The following cards (taken from AI Actions usage page) are suggested to be the best examples of how the new component should look like:

![Image](https://github.com/user-attachments/assets/3e3a64c6-10a0-462f-8ec9-f3c8e2dd146c)

 

## Breaking changes

No breaking changes.

Lelith:
@0math I talked with the team today about the approaches and we agreed that the Compund Component approach gives us more scalability for the future. So The different parts of the Card should not be handed over as props but as child-components to the UsageCard.

Overall this is a thumbs up from our Team to start implementing this new Component. Let me know if you need any support from me for implementation details. We can also do a pair programming session to get started :)

0math:
Hey @Lelith , thank you for your feedback. Compound Component approach makes sense to me.
The order of elements matters but we can configure correct order on the Generic Usage Dashboard template level.
I will start implementing the component and will reach out for check ins 👍

PR #3122, opened 2025-07-01
Add UsageCard component

0math:
# Purpose of PR

This adds new component, `UsageCard`. Related proposal is https://github.com/contentful/forma-36/issues/3092.

Screenshot 2025-07-08 at 23 50 09

With `UsageCount` as a child:

Screenshot 2025-07-10 at 18 17 24
Screenshot 2025-07-10 at 18 17 33

## PR Checklist

- [ ] Tests are passing


packages/components/usagecard/src/UsageCard.styles.ts
```diff
@@ -0,0 +1,33 @@
+import { css } from 'emotion';
+import tokens from '@contentful/f36-tokens';
+import { UsageCardProps } from './UsageCard';
+
+export const getUsageCardStyles = (variant: UsageCardProps['variant']) => {
+  if (variant === 'info') {
+    return {
+      usageCard: css({
+        borderRadius: '6px',
```
Lelith: 💅 You should optimize the code here more to avoid duplications. You could write two functions, one getUsageCardStyles and one which contains the additional stylings for the info variant. So you can avoid having most of the basic stylings duplicated,

For example in the Badge it was done similar: https://github.com/contentful/forma-36/blob/next/packages/components/badge/src/Badge/Badge.styles.ts#L73

packages/components/usagecard/src/UsageCard.styles.ts
```diff
@@ -0,0 +1,33 @@
...
+        backgroundColor: tokens.gray100,
+        border: `2px solid ${tokens.gray100}`,
+        padding: '1.3rem',
```
Lelith: 🤔 this spacing for padding seems a bit off to me. In forma we usually only use to the spacings available as tokens. This helps us have consistency. We do not have a spacing for 1.3rem, can you go and double check with your designer if either `spacingM: 1rem` or `spacingL: 1.5rem` would work in that case? 
These are also default spacings of the two size variations of the Card. So you would not need to overwrite it at all.

https://github.com/contentful/forma-36/blob/next/packages/components/card/src/Card/Card.styles.ts#L5

packages/components/usagecard/src/UsageCard.styles.ts
```diff
@@ -0,0 +1,33 @@
...
+        flexDirection: 'column',
+        justifyContent: 'space-between',
+        '@media (max-width: 1024px)': {
+          marginBottom: tokens.spacingM,
+        },
```
Lelith: ⚠️ we normally do not set outside spacing for reusable components in forma. Because the outside spacing should be set in the moment the component is used. Therefore the Card component accepts a margin variable as a property.  Not having outside spacing set on components make them more predictable. Therefore i advise to remove it from the style definitions.

packages/components/usagecard/src/UsageCard.styles.ts
Lelith: 🧹 the border radius of 6px is already the default of the Card component. So it does not need to be added here: https://github.com/contentful/forma-36/blob/next/packages/components/card/src/BaseCard/BaseCard.styles.ts#L27

packages/components/usagecard/src/UsageCard.test.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import { UsageCard } from './UsageCard';
+
+describe('UsageCard', function () {
+  it('renders', () => {
+    const tree = render(<UsageCard>hello world</UsageCard>);
+
+    expect(tree).toBeTruthy();
+  });
```
Lelith: Could you add one test case for a11y issues please? Also best to do one of reach variant, so we know the color contrast for the variants are working.


````

  it('has no a11y issues', async () => {
    const { container } = render(<UsageCard>Card</UsageCard>);
    const results = await axe(container);

    expect(results).toHaveNoViolations();
  });
````

Maybe also adding another one with all possile children and props.

File: packages/components/usagecard/src/UsageCard.tsx [hunk 5]
```diff
@@ -0,0 +1,70 @@
...
+
+  return (
+    <UsageCardContextProvider value={contextValue}>
+      <Card
+        {...otherProps}
+        className={styles.usageCard}
+        ref={ref}
+        data-test-id={testId}
+      >
+        <Flex flexDirection="column" style={{ height: '100%' }}>
```
Lelith: ❓Is this additonal Flex neccessary when we have the column flex style already on the card?

packages/components/usagecard/src/UsageCard.tsx
```diff
@@ -0,0 +1,70 @@
...
+      variant,
+    }),
+    [variant],
+  );
+
+  return (
+    <UsageCardContextProvider value={contextValue}>
+      <Card
+        {...otherProps}
+        className={styles.usageCard}
```
Lelith: 💅 you should allow adding `className` as a property and combine them, to allow additional styling when using. Because it will not combine them with using the `...otherProps` spread
```suggestion
        className={cx(styles.usageCard, className)}
```

packages/components/usagecard/src/UsageCard.tsx
```diff
@@ -0,0 +1,70 @@
...
+        ref={ref}
+        data-test-id={testId}
+      >
+        <Flex flexDirection="column" style={{ height: '100%' }}>
+          <Box style={{ flex: 1 }}>
+            {header}
+            {children}
+          </Box>
+          {description && (
+            <Box style={{ marginTop: 'auto' }}>{description}</Box>
```
Lelith: ❓will this margin setting result in the same space between these elements always? Should we maybe use the flex property `gap` instead?

packages/components/usagecard/src/UsageCardDescription/UsageCardDescription.styles.ts
```diff
@@ -0,0 +1,10 @@
+import { css } from 'emotion';
+import tokens from '@contentful/f36-tokens';
+
+export const getUsageCardDescriptionStyles = () => {
```
Lelith: As this style is so minimal, we could also put it into the UsageCard styles file

packages/components/usagecard/src/UsageCard.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  const contextValue: UsageCardContextType = useMemo(
+    () => ({
+      variant,
+    }),
+    [variant],
+  );
+
+  return (
+    <UsageCardContextProvider value={contextValue}>
```
Lelith: ❓what exactly are we using the context for? I don't see it reused in any of the compound components?

packages/components/usagecard/src/UsageCardDescription/UsageCardDescription.tsx
```diff
@@ -0,0 +1,47 @@
...
+      {...otherProps}
+      ref={ref}
+      className={cx(styles.usageCardDescription, className)}
+      testId={testId}
+    >
+      {children}
+      {text}{' '}
+      <TextLink target="_blank" rel="noopener noreferrer" href={link}>
+        {linkTitle}
+      </TextLink>
```
Lelith: 🤔 I'm not sure if its such a good idea to have a hard position of the link at the end of the text. Depending on the languages we are going to use, we never know if the link needs to be positioned within the sentence or at the beginning instead of the end.
I personally would not receive these as a property at all and it can just be handed over as part of the `children` property.

packages/components/usagecard/src/UsageCardHeader/UsageCardHeader.styles.ts
```diff
@@ -0,0 +1,25 @@
+import tokens from '@contentful/f36-tokens';
+import { css } from 'emotion';
+
+export const getUsageCardHeaderStyles = () => {
+  return {
+    usageCardHeader: css({
+      width: '100%',
+      maxWidth: '1550px',
```
Lelith: ❓where is this max-Width coming from?

packages/components/usagecard/src/UsageCardHeader/UsageCardHeader.tsx
```diff
@@ -0,0 +1,53 @@
...
+  } = props;
+  const styles = getUsageCardHeaderStyles();
+
+  return (
+    <Subheading
+      {...otherProps}
+      ref={ref}
+      className={cx(
+        styles.usageCardHeader,
+        styles.subheadingWithIcon,
```
Lelith: ❓should the subheadingWithIcon class not only be applied when there is a tooltip  property?

packages/components/usagecard/src/UsageCard.styles.ts
0math: Thank you, I missed that Card has a `padding` prop.

packages/components/usagecard/src/UsageCard.tsx
0math: I initially wanted to use it with UsageCard.Description but abandoned this idea. I removed the context, thank you.

packages/components/usagecard/src/UsageCard.tsx
0math: No, we can get rid of it.

packages/components/usagecard/src/UsageCard.tsx
0math: I think it stopped serving its purpose, it doesn't do anything now, I removed it.

packages/components/usagecard/src/UsageCardDescription/UsageCardDescription.tsx
0math: Make sense to me. I kept children only.

packages/components/usagecard/src/UsageCardHeader/UsageCardHeader.styles.ts
0math: We used in our old dashboard. It doesn't make sense now, I removed it.

0math:
> Thanks for the code-changes, it looks good. I'm just thinking if we should merge the UsageCount first and then we are able to sow the Card combined with the UsageCount for people to better understand how these two interact with each other.

Yes, I'm thinking the same. We can merge this one first: https://github.com/contentful/forma-36/pull/3129 and then I can add examples of with UsageCount inside UsageCard.

0math:
@Lelith @cf-remylenoir I added `UsageCount` as a child to examples and stories.

packages/components/usage-card/examples/UsageCardInfoExample.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import { UsageCard } from '@contentful/f36-components';
+import { TextLink } from '@contentful/f36-text-link';
+import { UsageCount } from '@contentful/f36-components';
```
Lelith:
```suggestion
import { TextLink } from '@contentful/f36-text-link';
import { UsageCard, UsageCount } from '@contentful/f36-components';
```

packages/components/usage-card/examples/UsageCardInfoExample.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+export default function UsageCardInfoExample() {
+  return (
+    <UsageCard
+      variant={'info'}
```
Lelith:
```suggestion
      variant="info"
```

packages/components/usage-card/examples/UsageCardUsageExample.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import { UsageCard } from '@contentful/f36-components';
+import { TextLink } from '@contentful/f36-text-link';
+import { UsageCount } from '@contentful/f36-components';
```
Lelith:
```suggestion
import { TextLink } from '@contentful/f36-text-link';
import {UsageCard,  UsageCount } from '@contentful/f36-components';
```

packages/components/usage-card/examples/UsageCardUsageExample.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+export default function UsageCardUsageExample() {
+  return (
+    <UsageCard
+      variant={'usage'}
```
Lelith:
```suggestion
      variant="usage"
```

Lelith (approved):
just some tiny nitpicks, then we are ready for release 🚀

PR #3130, opened 2025-07-08
feat: update header component and prepare for v5 release

Lelith:
# Purpose of PR

Iteration over the Header component to align it with latest Scalable UI designs and improve the prop API

* bring back segmentation dividing lines between back button, breadcrumbs and title
* change the need to force the variant via prop to title, when it should be default behavior
* adding tests
* extend examples in storybook

ToDo
[] adding migration guidelines


packages/components/header/src/Header.styles.ts
```diff
@@ -36,11 +35,4 @@ export const getHeaderStyles = ({
     // Reduce vertical padding when there's a filter in the header
     padding: hasFilters ? `${tokens.spacingXs} 0` : `${tokens.spacingS} 0`,
   }),
-  separator: css({
```
Lelith, line 39: this class was added in the original version but never used.

packages/components/header/src/Header.tsx
```diff
@@ -82,32 +79,34 @@ function _Header<E extends ElementType = typeof HEADER_DEFAULT_TAG>(
...
       {...otherProps}
     >
       <div className={styles.context}>
         <Flex alignItems="center" gap="spacingXs">
           {variant === 'title' ? (
-            <HeaderTitle title={title} variant="title" />
+            <HeaderTitle title={title} variant={variant} />
           ) : (
-            <>
+            <Segmentation>
```
Lelith, line 110: going to clarify with Daniel wether or not this is now right as it was purpusfully removed between this version and v4

packages/components/header/src/Header.tsx
```diff
@@ -82,32 +79,34 @@ function _Header<E extends ElementType = typeof HEADER_DEFAULT_TAG>(
     metadata,
     title,
     withBackButton,
-    variant = 'breadcrumb',
```
Lelith, line 85: i removed the variant prop as it was not obvious that variant had to be set to "title" specifically when there where no breadcrumbs or back-button.
I believe this logic is better handled inside the component. But will calrify with Daniel wether or not the size of the title should be forcable from outside

packages/components/header/src/Header.styles.ts
```diff
@@ -1,14 +1,13 @@
 import { css } from 'emotion';
```
Lelith, line 1: need to clarify about the border-bottom, if it should be dependent on the breadcrumbs or not.

Lelith: the bottom border should be determined by the layout component instead, based on the presence of sidebars. no sidebars no bottom border.

packages/components/header/src/Header.tsx
Lelith: segmentation should happen, but only between the breadcrumb and the title, not after the button

packages/components/header/MIGRATION.md
```diff
@@ -0,0 +1,51 @@
+# Migration Guideline Header from 4.X.X to 5.X.X
+
+The header component is to be used mostly within the Layout component. It has undergone significant design and property changes. The most siginificant is the change of the background color from `gray100` to `white`
+
+## Header
+
+The Header component is meant to include the title of a page and therefore on default renders tie title as `h1`. It can receive multiple sub-components via the props `actions` `filters` `metadata`. For navigiation inside a feature, it can receive `breadcrumbs` in an array
```
missating, line 7: Typo: `renders the title`

missating (commented):
Code LGTM

packages/components/header/MIGRATION.md
cf-remylenoir, line 7:
```suggestion
The Header component is meant to include the title of a page and, therefore, by default renders the title as `h1`. It can receive multiple sub-components via the props `actions` `filters` `metadata`. For navigation inside a feature, it can receive `breadcrumbs` in an array.
```

packages/components/header/README.mdx
```diff
@@ -45,6 +45,23 @@ import { Header } from '@contentful/f36-header';
...
+| metadata    | Optional additional information to display in the header, will appear after the title.                                        | Text, Badge, Tag             |
+| filters     | Optional element displayed in the center of the header, typically used to render refinement/search UI.                        | TextInput, Select, Search    |
+| actions     | Optional elements to display as complementary actions (e.g. buttons) related to the current page/route.                       | Button, IconButton, Dropdown |
+
+## Accessibility
+
+- The Header component renders a `<header>` landmark and a heading (`<h1>` by default). Use the title prop for a meaningful heading, and adjust the heading level with `titleProps.as` as needed for your page structure.
+- If using multiple headers, use the `as` prop to avoid duplicate landmarks and keep the hierarchy of the content integer.
+- Breadcrumbs and back buttons improve navigation for all users. Ensure they have descriptive, accessible labels.
+- All interactive elements in actions and filters must be keyboard accessible and clearly labeled.
```
cf-remylenoir, line 63:
```suggestion
- The Header component renders a `<header>` landmark and a heading (`<h1>` by default). Use the `title` prop for a meaningful heading, and adjust the heading level with `titleProps.as` as needed for your page structure.
- If using multiple headers, use the `as` prop to avoid duplicate landmarks and keep the hierarchy of the content integer.
- Breadcrumbs and back buttons improve navigation for all users. Ensure they have descriptive and accessible labels.
- All interactive elements in actions and filters must be keyboard accessible and labeled following our content guidelines.
```

cf-remylenoir (commented):
A few comments ✍️ 

I like the info note approach above the components in the stories to provide more information 👍 
I will look at the code changes with fresher eyes & brain tomorrow.

cf-remylenoir (approved):

packages/components/header/MIGRATION.md
cf-remylenoir, line 1: My only nit would be to have property keys, types, or values wrapped in inline-code styling, e.g.:

`testId` | `string` | Test id for the root element. (default: `cf-ui-header`) | New
-- | -- | -- | --

PR #3132, opened 2025-07-16
feat: Add custom placeholder prop to Datepicker component

colomolo:
# Purpose of PR
Currently Datepicker component's default placeholder shows today date. Adding a custom placeholder prop to override this behavior if other placeholder is needed.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/components/datepicker/README.mdx
+++ b/packages/components/datepicker/README.mdx
@@ -82,6 +82,14 @@ If you need to show more than a month at once, use the `numberOfMonths` property
 
 ```
 
+### Custom placeholder
+
+To replace default placeholder with your custom one, use the `placeholder` property.
+
+```jsx file=./examples/Datepicker/Placeholder.tsx
+
+```
+
 ### Open by default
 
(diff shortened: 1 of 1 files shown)


packages/components/datepicker/src/Datepicker.tsx
```diff
@@ -36,6 +36,11 @@ export type DatepickerProps = CommonProps & {
    */
   defaultIsOpen?: boolean;
 
+  /**
+   * Custom placeholder for the input field
+   */
+  customPlaceholder?: string;
```
cf-remylenoir: I wonder if it is necessary to have `custom` in here, `placeholder` could be sufficient 🤔  Wdyt @Lelith @veu ?

veu: I’d drop it. Every argument passed it implicitly custom.

colomolo: Good point, yes.

PR #3134, opened 2025-07-28
fix(core): adjust Grid, Grid.Item and Flex components alignment properties

cf-remylenoir:
# Purpose of PR

Adjust the Flex and Grid components alignment properties

**Grid**

Adds the following properties

- `justifyItems`
- `alignItems`
- `placeItems`
- `placeContent`

**Grid Item**

Adds the following properties

- `justifySelf`
- `alignSelf`
- `placeSelf`

**Flex**

Deprecates the `justifyItems` property, which is ignored on flexbox layouts.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/core/src/Flex/Flex.tsx
+++ b/packages/core/src/Flex/Flex.tsx
@@ -53,15 +53,15 @@ export interface FlexInternalProps
    * Defines a gap between flexbox items. */
   gap?: Spacing | string;
   /**
-   * Defines how flexbox/grid items are aligned according to the main axis, within a flexbox/grid container. */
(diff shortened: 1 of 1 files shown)


packages/core/stories/Grid.stories.tsx
```diff
@@ -19,6 +19,14 @@ export default {
   parameters: {
     propTypes: [Grid['__docgenInfo'], Grid.Item['__docgenInfo']],
   },
+  argTypes: {
```
cf-remylenoir, line 22: For better controls with type inference, we can replace these Storybook props with a `select` once we upgrade the Storybook to LTS, in the Forma v6 work.

PR #3139, opened 2025-08-08
feat: add format truncated string list to i18n utils

pypas:
# Purpose of PR
Adds `formatTruncatedStringList` in order to replace the [joinAndTruncate](https://github.com/contentful/user_interface/blob/e0b547580791fd90d42360af73f5c384ad2e97a0/src/javascripts/utils/StringUtils.ts#L116) function from `user_interface`.

This function joins a string list with commas, adding "and X more" to the end of the sentence (localized). 

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/packages/f36-i18n-utils/package.json
+++ b/packages/f36-i18n-utils/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@contentful/f36-i18n-utils",
   "description": "Forma 36 i18n utilities",
-  "version": "1.0.0-alpha.1",
+  "version": "1.0.0-alpha.2",
   "private": true,
   "license": "MIT",
   "files": [
--- a/packages/f36-i18n-utils/src/formatters.ts
+++ b/packages/f36-i18n-utils/src/formatters.ts
@@ -15,6 +15,34 @@ export function formatStringList(locale: string, items: string[]): string {
   }`;
 }
 
+/**
+ * Joins the strings with commas and a final 'and X more' (determined by maxLength).
+ * The sentence is localized based on the provided locale.
+ *
+ * formatTruncatedStringList('en-US', ['a', 'b', 'c', 'd'], 2)
+ * // => 'a, b and 2 more'
+ */
+export function formatTruncatedStringList(
+  locale: string,
+  list: string[],
+  maxLength: number,
+) {
+  if (list.length <= maxLength) {
+    return formatStringList(locale, list);
+  }
+
+  // This prevents finishing the sentence with 'and 1 more'
+  if (list.length === maxLength + 1) {
+    if (maxLength === 1) return formatStringList(locale, list);
+    maxLength = maxLength - 1;
+  }
+
+  const restLength = list.length - maxLength;
+  const initialList = list.slice(0, maxLength);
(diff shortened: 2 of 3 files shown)


packages/f36-i18n-utils/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@contentful/f36-i18n-utils",
   "description": "Forma 36 i18n utilities",
-  "version": "1.0.0-alpha.1",
+  "version": "1.0.0-alpha.2",
```
cf-remylenoir: You can revert this change.

For alpha packages, the release is done locally via the `npm run prerelease` command, which automatically bumps the version. I can support you and do it for you once the PR is approved 👍 .

```suggestion
  "version": "1.0.0-alpha.1",
```

pypas: Oops, I missed that, thanks a lot for letting me know @cf-remylenoir

packages/f36-i18n-utils/src/formatters.ts
```diff
@@ -15,6 +15,34 @@ export function formatStringList(locale: string, items: string[]): string {
...
+  list: string[],
+  maxLength: number,
+) {
+  if (list.length <= maxLength) {
+    return formatStringList(locale, list);
+  }
+
+  // This prevents finishing the sentence with 'and 1 more'
+  if (list.length === maxLength + 1) {
+    if (maxLength === 1) return formatStringList(locale, list);
```
cf-remylenoir: Nit: to stay consistent with the rest

```suggestion
    if (maxLength === 1) {
      return formatStringList(locale, list);
    }
```

packages/f36-i18n-utils/src/formatters.ts
```diff
@@ -15,6 +15,34 @@ export function formatStringList(locale: string, items: string[]): string {
   }`;
 }
 
+/**
+ * Joins the strings with commas and a final 'and X more' (determined by maxLength).
+ * The sentence is localized based on the provided locale.
+ *
+ * formatTruncatedStringList('en-US', ['a', 'b', 'c', 'd'], 2)
+ * // => 'a, b and 2 more'
```
cf-remylenoir, line 18: Nit: when more than 2 items are connected, a comma is added before `and`

- a, b, and 2 more
- a and b

cf-remylenoir: BUT, while it is true in English, I have no idea how it applies to other locales, so I'm happy to leave it for later.

cf-remylenoir: I came across this https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat

cf-remylenoir (approved):
Looking great, thank you for your contribution ❤️ 

~~I will prerelease the package locally and push the package version bump.
Then, you can go ahead with the merge.~~

EDIT: I forgot this package was private (for now) and only accessible via our GitHub organization registry. 
Hence, we don't prerelease on NPM yet. Merging the PR should trigger the release and version changelog update 👍

PR #3148, opened 2025-08-20
chore: update codemod script

missating:
# Purpose of PR

## PR Checklist

- [ ] Tests are passing

### DIFF
--- a/packages/forma-36-codemod/transforms/v5/icons.js
+++ b/packages/forma-36-codemod/transforms/v5/icons.js
@@ -362,8 +362,8 @@ const updateToV5Icons = function (file, api) {
         from: importName,
         to:
           v4IconName === 'Icon'
-            ? '@contentful/f36-icon-alpha'
-            : '@contentful/f36-icons-alpha',
+            ? '@contentful/f36-icon'
+            : '@contentful/f36-icons',
         outputComponentName: newComponentName,
       });
     }
@@ -393,7 +393,7 @@ const updateToV5Icons = function (file, api) {
     source = changeImport(j, source, {
       componentName: iconProps.localName,
       from: iconProps.importName,
-      to: '@contentful/f36-icons-alpha',
+      to: '@contentful/f36-icons',
       outputComponentName: iconProps.localName,
     });
   }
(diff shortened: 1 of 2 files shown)


Lelith (approved):
Looks good to me!

packages/forma-36-codemod/bin/packages-list.mjs
```diff
@@ -1,5 +1,5 @@
 export const PACKAGES_UPGRADE = {
-  '@contentful/f36-icons-alpha': 'npm:@contentful/f36-icons@alpha',
+  '@contentful/f36-icons': '^5.0.0-alpha.49',
```
cf-remylenoir, line 2: Once we exit alpha and release, the alpha tag will be removed. We must point to this version instead.

```suggestion
  '@contentful/f36-icons': '^5.0.0',
```

packages/forma-36-codemod/transforms/v5/icons.js
```diff
@@ -110,7 +111,7 @@ const iconsMap = {
   Spreadsheet: 'Table',
   Star: 'Star',
   Subscript: 'TextSubscript',
-  Superscript: 'TextSuperscript',
+  Superscript: 'TextSuperscriptIcon',
```
cf-remylenoir: Why `TextSuperscriptIcon` when all the others do not have the `Icon` prefix?

packages/forma-36-codemod/bin/packages-list.mjs
Lelith: so we update already to this? then we can not use the script anymore until we have the v5 fully released

cf-remylenoir: I assume this PR prepares the Codemod for the release and public usage. Do we still need to use it internally?

Lelith: there are still the custom product-icons missing, which also need replacement. So I personally would like to have the script available for cleaning these instances up.

cf-remylenoir: Fair enough, if there are significant occurrences to replace

cf-remylenoir (approved):

PR #3149, opened 2025-08-20
chore: remove deprecated apis + update component statuses

cf-remylenoir:
# Purpose of PR

- Removes previously deprecated APIs & components + pre-releases the components
- Adjusts the components' status metadata

## PR Checklist

- [x] Tests are passing


packages/components/multiselect/src/Multiselect.tsx
```diff
@@ -219,18 +192,16 @@ function _Multiselect(props: MultiselectProps, ref: React.Ref<HTMLDivElement>) {
...
 
   const handoverSearchProps =
     Object.keys(searchProps).length > 0
       ? searchProps
       : {
-          onSearchValueChange: props.onSearchValueChange,
-          searchPlaceholder: props.searchPlaceholder,
-          searchInputName: props.searchInputName,
-          searchInputRef: props.searchInputRef,
+          onSearchValueChange: searchProps.onSearchValueChange,
```
Lelith: 🧹 I think we can get rid of the if / else block all together when now only searchProps exist

packages/components/multiselect/src/Multiselect.tsx
```diff
@@ -77,39 +77,12 @@ export interface MultiselectProps extends CommonProps {
    */
   triggerButtonProps?: ButtonProps;
 
-  /**
```
Lelith, line 80: As this change was not part of any major release before, we should add it to the migration guideline for V5. Do you want to do this in this PR or should we do a follow-up story?

cf-remylenoir: Let's do that as a follow-up 👍

Lelith (approved):
nice work fixing the icons overview for the documentation website! 
Looks good to me!

PR #3150, opened 2025-08-25
feat: make list formatters more flexible

MayaGillilan:
# Purpose of PR

Simply making the list formatters more flexible by adding an optional type param to allow the other two types of formatting for better i18n use

### DIFF
--- a/packages/f36-i18n-utils/src/formatters.ts
+++ b/packages/f36-i18n-utils/src/formatters.ts
@@ -1,11 +1,17 @@
+type ListFormatType = 'conjunction' | 'disjunction' | 'unit';
+
 export function formatNumber(locale: string, value: number): string {
   return Intl.NumberFormat(locale).format(value);
 }
 
-export function formatStringList(locale: string, items: string[]): string {
+export function formatStringList(
+  locale: string,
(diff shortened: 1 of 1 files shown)


cf-remylenoir:
> For some reason I can't seem to select this package for a changeset as it doesn't show in the list 👀 can anyone help with that? This should trigger a minor bump

Hey, yes, this package is in `alpha` mode; there isn't a changeset process available until it is stable. We can support with the pre-release process 👍

packages/f36-i18n-utils/src/formatters.ts
cf-remylenoir, line 1: Could you please add a few assertions of the other types of usage?

MayaGillilan: Of course, sorry I forgot that

cf-remylenoir: No worries 🙂

cf-remylenoir (approved):
@missating could you please support with the pre-release of this new alpha version? Then, once the package is published and the version bump committed, we must merge against the `next` branch (as we locked the `main` branch until the v5 release).

PR #3170, opened 2025-09-04
fix(icons): pin phosphor to version with new names and working ESM build

denkristoffer:
# Purpose of PR

Pin the version of Phosphor react package to the first version of the package that implements the new export names with the `Icon` suffix. This version is also the latest version that has a working CJS and ESM build so rather than allowing versions above `2.1.8` I pinned it.

### DIFF
--- a/packages/components/icon/package.json
+++ b/packages/components/icon/package.json
@@ -33,7 +33,7 @@
   "dependencies": {
     "@contentful/f36-core": "^5.1.0",
     "@contentful/f36-tokens": "^5.1.0",
-    "@phosphor-icons/react": "^2.1.5",
+    "@phosphor-icons/react": "2.1.8",
     "emotion": "^10.0.17"
   },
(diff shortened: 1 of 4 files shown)


packages/components/icon/package.json
```diff
@@ -33,7 +33,7 @@
   "dependencies": {
     "@contentful/f36-core": "^5.1.0",
     "@contentful/f36-tokens": "^5.1.0",
-    "@phosphor-icons/react": "^2.1.5",
+    "@phosphor-icons/react": "^2.1.8",
```
missating:
```suggestion
    "@phosphor-icons/react": "2.1.8",
```

denkristoffer: If I can get npm to share versions properly it shouldn't be necessary to pin it everywhere. I guess we can, though 🤔

missating (approved):

PR #3171, opened 2025-09-05
Breaking: React 19 Update

Lelith:


packages/components/icon/package.json
cf-remylenoir, line 57:
```suggestion
    "node": ">=22"
```

packages/components/icons/package.json
cf-remylenoir, line 58:
```suggestion
    "node": ">=22"
```

package.json
```diff
@@ -6,6 +6,19 @@
     "packages/*",
     "packages/components/*"
   ],
+  "description": "<h1><img src=\"./forma-icon.svg\" height=\"24\"> Forma 36 - The Contentful Design System</h1>",
+  "main": ".eslintrc.js",
```
cf-remylenoir: Two things:
- this file is removed in this PR; it can't be reached.
- this field should point to the entry point of the output, typically the `index.js` file; it shouldn't be the ESLint configuration. In our case, there isn't an entry point at the monorepo root; all packages have their own.

https://docs.npmjs.com/cli/v11/configuring-npm/package-json#main

package.json
```diff
@@ -6,6 +6,19 @@
     "packages/*",
     "packages/components/*"
   ],
+  "description": "<h1><img src=\"./forma-icon.svg\" height=\"24\"> Forma 36 - The Contentful Design System</h1>",
```
cf-remylenoir: This should be a string, not an HTML or MDX expression - https://docs.npmjs.com/cli/v11/configuring-npm/package-json#description-1.

```suggestion
  "description": "Forma 36 - The Contentful Design System",
```

package.json
```diff
@@ -16,123 +29,94 @@
...
-    "webpack": "5"
-  },
-  "simple-git-hooks": {
-    "pre-commit": "npm run-script pretty:quick",
-    "commit-msg": "npx commitlint -e \"$@\""
-  },
-  "engines": {
-    "node": ">=18.0.0",
-    "npm": ">=8.3.0",
-    "yarn": "Please use npm"
```
cf-remylenoir, line 132: We might want to keep the engines object property to enforce the minimum compatible version when installing packages

package.json
```diff
@@ -16,123 +29,94 @@
...
-    "tsup": "^7.2.0",
-    "turbo": "^1.10.13",
-    "typescript": "4.7.3",
-    "wait-on": "^8.0.3",
-    "webpack": "5"
-  },
-  "simple-git-hooks": {
-    "pre-commit": "npm run-script pretty:quick",
-    "commit-msg": "npx commitlint -e \"$@\""
-  },
```
cf-remylenoir, line 128: Is this taken care of elsewhere?

package.json
Lelith: good catch, will fix

package.json
Lelith: great callout, will remove the line

package.json
Lelith: the knive was too sharp on this one, will re-add the githooks

package.json
Lelith: will re-add

veu (approved):

PR #3172, opened 2025-09-05
feat:[MOI-5862] Make "included" word switcheable in UsageCount

0math:
This introduces to changes:
- Adds "included" word in UsageCount component as a prop (`includedLabel`)(see screenshot)
- Fixes `tooltip` type in UsageCard component

Screenshot 2025-09-10 at 18 09 00

# Purpose of PR

## PR Checklist

- [ ] Tests are passing


.changeset/petite-boats-carry.md
```diff
@@ -0,0 +1,6 @@
+---
+'@contentful/f36-usage-count': minor
+'@contentful/f36-usage-card': minor
```
cf-remylenoir, line 3: Nit: according to the PR title, it is a fix. 

If it is truly a fix, then the changeset should be a `patch` and not a `minor` change.

If it is indeed a `minor` change, then update the PR title to `feat`.

https://semver.org

0math: Thank you @cf-remylenoir . Updated the title. Do you have an idea what's blocking the merge at the moment? I understand that the failed jobs are expected, trying to figure out why merging is blocked.

cf-remylenoir: Update your branch to get the latest changes; it will resolve the `f36-link-checker` job. As for the Vercel build, it is due to Vercel dropping support for Node 18. We are working on updating Node to LTS in the next major version. It does not block the merge, but the documentation website will not get updates until then.

packages/components/usage-count/src/UsageCount.tsx
cf-remylenoir, line 1: Should it rather be a `boolean` to display the suffix rather than a `string`? I wonder what other suffixes besides `included` could be used?

0math: @cf-remylenoir I initially implemented it as boolean: https://github.com/contentful/forma-36/pull/3172/commits/2e201b5e46cc2b29a232d2f0cfc62609f2be24e5. Later @Lelith suggested to make it string to avoid translating it in the component.
`included` is assumed in most cases, but now that I think about it, it's good option to allow other suitable labels.

PR #3180, opened 2025-09-18
feat: upgrade eslint and fix issues :six: [UFO-2005]

Lelith:
# 🚀 Purpose of PR  
Upgrade **`eslint`** to the latest version and resolve all newly introduced issues and warnings.  

---

## 🔧 Main Fixes  

- **Flat Config Migration**  
  Migrated the ESLint configuration to the new **flat config format**.  

- **React Hooks Rules Compliance**  
  - Per [React Hooks rules](https://reactjs.org/docs/hooks-rules.html), hooks may only be used in React components or React functions.  
  - Previously, many components were named with a leading underscore (e.g. `_Component`) and then exported with `forwardRef` under their proper name.  
  - Since React requires component names to start with an **uppercase letter**, this violated linting rules.  
  - **Change:** Naming convention is now `ComponentNameBase` → then wrapped in `forwardRef`.  
    - ✅ Avoids invalid React components  
    - ✅ Makes intention clearer  

- **Polymorphic Components + `forwardRef`**  
  - Using `forwardRef` stripped away polymorphic types, causing type errors downstream.  
  - **Fix:** Exported components are now typecast with `as` to preserve polymorphic typing.  

- **Storybook Stories Typing**  
  - Storybook story typing changed to also respect hooks rules.  
  - Old pattern (using a `render` function inside a Story) violated linting.  
  - **Fix:** Stories are now written as explicit React components.  

- **Display Names**  
  - Even though React 18/19 no longer require `displayName`, all components now explicitly set it.  
  - ✅ Improves DX (developer experience) and debugging consistency.  

---

## ⚠️ Rules That Could Be Further Improved  

- **Polymorphic Refs**  
  - Handling refs as polymorphic props caused complex type issues.  
  - Current solution: `ref` is typed as `any` in most polymorphic components, with linting rules suppressed where necessary.  

- **Looser Typings in Stories & Tests**  
  - To satisfy ESLint + hooks rules, typings in Storybook stories and test cases are intentionally more flexible.  

---

## 📦 Additional Changes  

- Improved overall **test coverage**.  
- **Breaking Change (Skeleton component):**  
  - Removed two exported types that were identical to the parent component’s type.  
  - Consumers should switch to the parent type.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -53,13 +53,6 @@
         "eslint-plugin-react": "^7.37.5",
         "eslint-plugin-react-hooks": "^5.2.0",
         "eslint-plugin-storybook": "^9.1.6",
-        "eslint-config-prettier": "^10.1.8",
-        "eslint-plugin-import": "^2.32.0",
-        "eslint-plugin-jest": "^29.0.1",
-        "eslint-plugin-jsx-a11y": "^6.10.2",
-        "eslint-plugin-react": "^7.37.5",
-        "eslint-plugin-react-hooks": "^5.2.0",
-        "eslint-plugin-storybook": "^9.1.6",
         "globals": "^16.4.0",
         "globby": "^15.0.0",
         "inquirer": "^12.9.6",
@@ -3015,12 +3008,24 @@
       "resolved": "packages/components/asset",
       "link": true
     },
+    "node_modules/@contentful/f36-autocomplete": {
+      "resolved": "packages/components/autocomplete",
+      "link": true
+    },
+    "node_modules/@contentful/f36-avatar": {
+      "resolved": "packages/components/avatar",
+      "link": true
+    },
     "node_modules/@contentful/f36-badge": {
       "resolved": "packages/components/badge",
       "link": true
     },
     "node_modules/@contentful/f36-button": {
       "resolved": "packages/components/button",
+      "link": true
+    },
+    "node_modules/@contentful/f36-card": {
+      "resolved": "packages/components/card",
       "link": true
     },
     "node_modules/@contentful/f36-cdn": {
@@ -3047,6 +3052,10 @@
       "resolved": "packages/core",
       "link": true
     },
+    "node_modules/@contentful/f36-datepicker": {
+      "resolved": "packages/components/datepicker",
+      "link": true
+    },
     "node_modules/@contentful/f36-datetime": {
       "resolved": "packages/components/datetime",
       "link": true
@@ -3061,6 +3070,10 @@
     },
     "node_modules/@contentful/f36-empty-state": {
       "resolved": "packages/components/empty-state",
+      "link": true
+    },
+    "node_modules/@contentful/f36-entity-list": {
+      "resolved": "packages/components/entity-list",
       "link": true
     },
     "node_modules/@contentful/f36-forms": {
@@ -3095,6 +3108,22 @@
       "resolved": "packages/components/list",
       "link": true
     },
+    "node_modules/@contentful/f36-menu": {
+      "resolved": "packages/components/menu",
+      "link": true
+    },
+    "node_modules/@contentful/f36-modal": {
+      "resolved": "packages/components/modal",
+      "link": true
+    },
+    "node_modules/@contentful/f36-multiselect": {
+      "resolved": "packages/components/multiselect",
+      "link": true
+    },
+    "node_modules/@contentful/f36-navbar": {
+      "resolved": "packages/components/navbar",
+      "link": true
+    },
     "node_modules/@contentful/f36-navlist": {
       "resolved": "packages/components/navlist",
       "link": true
@@ -3103,12 +3132,20 @@
       "resolved": "packages/components/note",
       "link": true
     },
+    "node_modules/@contentful/f36-notification": {
+      "resolved": "packages/components/notification",
+      "link": true
+    },
     "node_modules/@contentful/f36-pagination": {
       "resolved": "packages/components/pagination",
       "link": true
     },
     "node_modules/@contentful/f36-pill": {
       "resolved": "packages/components/pill",
+      "link": true
+    },
+    "node_modules/@contentful/f36-popover": {
+      "resolved": "packages/components/popover",
       "link": true
     },
     "node_modules/@contentful/f36-progress-stepper": {
@@ -3137,6 +3174,10 @@
     },
     "node_modules/@contentful/f36-tokens": {
       "resolved": "packages/forma-36-tokens",
+      "link": true
+    },
+    "node_modules/@contentful/f36-tooltip": {
+      "resolved": "packages/components/tooltip",
       "link": true
     },
     "node_modules/@contentful/f36-typography": {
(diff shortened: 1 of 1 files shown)


packages/components/copybutton/src/CopyButton.tsx
```diff
@@ -134,6 +139,8 @@ function _CopyButton(
       ref={ref}
     />
   );
+
+  // ToDo reimplement tooltip
   // return (
```
missating: Can we add the files with these Todo's to the tooltip ticket? so we don't forget any.

cf-remylenoir (commented):
Let's get this over the finish line 👀

packages/components/button/src/IconButton/__snapshots__/IconButton.test.tsx.snap
```diff
@@ -0,0 +1,31 @@
+// Jest Snapshot v1, https://jestjs.io/docs/snapshot-testing
```
cf-remylenoir, line 1: Chromatic already covers Visual Regression Testing. Is there another reason why it was implemented here?

packages/components/button/stories/ButtonComponent.stories.tsx
```diff
@@ -12,7 +12,7 @@ export default {
   title: 'Components/Button components',
   component: Button,
   parameters: {
-    propTypes: Button['__docgenInfo'],
+    propTypes: (Button as any)?.__docgenInfo,
```
cf-remylenoir: Isn't it breaking the Storybook controls? I'm unable to verify since the Storybook build is currently broken.

packages/components/copybutton/src/CopyButton.tsx
```diff
@@ -94,6 +97,8 @@ function _CopyButton(
 
       // @ts-expect-error -- The return type of `execCommand` can also be string
       if (result === 'unsuccessful') {
+        // eslint-disable-next-line no-console
```
cf-remylenoir, line 97: Is this to surface errors on the browser's console?

Edit: this comment is about line `101`, hidden from this comment 😄

packages/components/header/src/Segmentation.tsx
```diff
@@ -37,14 +37,15 @@ export type SegmentationProps<
...
+function SegmentationBase<
+  E extends ElementType = typeof SEGMENTATION_DEFAULT_TAG,
+>(
   {
     children,
     className,
     segments,
     separator: SeparatorComponent,
     testId = 'cf-ui-segmentation',
-    ...otherProps
```
cf-remylenoir, line 47: What was the issue with these props? This would qualify as a breaking change; let's make sure it is captured in the changeset.

packages/components/navlist/src/Navlist.test.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
```
cf-remylenoir, line 1: Is it intentional?

packages/f36-docs-utils/src/PropsTable/PropertyType/PropertyType.tsx
```diff
@@ -25,7 +25,9 @@ interface PropertyTypeProps {
 export function PropertyType({ type, name }: PropertyTypeProps) {
   if (name === 'as') {
     return (
-      <PropertyValue value="HTML Tag or React Component (e.g. div, span, etc)" />
+      <PropertyValue
+        value={'HTML Tag or React Component (e.g. div, span, etc)'}
```
cf-remylenoir, line 29: Nit, the curly braces aren't needed. Is the change to single quotes coming from prettier, or can we keep double quotes? 

```suggestion
        value='HTML Tag or React Component (e.g. div, span, etc)'
```

```suggestion
        value="HTML Tag or React Component (e.g. div, span, etc)"
```

packages/core/src/utils/getEntityStatusStyles.ts
cf-remylenoir, line 1: This might conflict with @missating's latest work or variant, specifically the color mapping. Additionally, these `color` base tokens appear to be an attempt to be used as semantic tokens. Could it be true, @damann?

packages/core/src/GlobalStyles/GlobalStyles.tsx
```diff
@@ -21,7 +21,7 @@ const cssReset = css`
 
   /* Remove list styles on ul, ol elements with a list role, which suggests default styling will be removed */
   ul[role='list'],
-  ol[role='list'] {
+  ol[role='list'] {bn
```
cf-remylenoir: Typo?
```suggestion
  ol[role='list'] {
```

packages/components/utils/src/Portal/Portal.test.tsx
cf-remylenoir, line 1: Interesting choice of child rendering with emoji in the first place 😄

packages/components/skeleton/MIGRATION.md
```diff
@@ -0,0 +1,104 @@
+# Migration of Skeleton from v5.x to v6.x
+
+## Breaking change: Consolidated prop types
+
+The duplicated component prop types `SkeletonDisplayTextProps` and `SkeletonBodyTextProps` have been removed. Use the single unified `SkeletonTextProps` instead.
+
+### Why was this changed?
```
cf-remylenoir, line 7:
```suggestion
### Why did it change?
```

packages/components/button/src/IconButton/__snapshots__/IconButton.test.tsx.snap
Lelith: To save time, I let Copilot write the tests. will remove the snapshot test!

packages/components/button/stories/ButtonComponent.stories.tsx
Lelith: Yeah, hm, they are still there but no longer radio buttons or dropdowns. So, I have to double-check how that properly works.

packages/components/copybutton/src/CopyButton.tsx
Lelith: yes, i think that was the intention behind this console output

packages/components/accordion/src/AccordionContext.ts
```diff
@@ -1,8 +1,7 @@
 import React from 'react';
-import { Accordion } from './Accordion';
 
 export type AccordionContextType = {
-  align?: 'start' | 'end';
+  align: 'start' | 'end';
```
cf-remylenoir, line 4: This change should be part of the changelog ✍️

vitest.shims.d.ts
cf-remylenoir, line 1: Could you educate me on this declaration file usage?

packages/components/button/src/IconButton/__snapshots__/IconButton.test.tsx.snap
cf-remylenoir: This still needs to be removed as well as the snapshot setup

packages/components/accordion/src/AccordionContext.ts
veu: It’s a change in the contex and not in the user facing type so changelog won’t be needed.

PR #3185, opened 2025-09-29
feat: update website dependencies to use react 19 :six: [UFO-1944]

missating:
# Purpose of PR

This PR updates the `packages/website` to be compatible with `nextjs 15`, `next/mdx 15` and `react 19`

**Here's a summary of everything that changed:**

1. Brought back the build script for the website, with a change from `Turborepo v2.5` switching from automatic package scoping to explicitly using` --filter` argument in the run script

https://turborepo.com/docs/crafting-your-repository/running-tasks#using-filters

```
   "docs:next": "turbo run start --filter=@contentful/f36-website --parallel --no-cache",
    "docs:next:build": "turbo run build:prod --filter=@contentful/f36-website --parallel --force",
    "docs:next:start-prod": "turbo run start:prod --filter=@contentful/f36-website --parallel"
```
Also updated `turbo.json` file to include `env vars` as Turbo v2 isolates tasks for caching and parallelism, local builds inherit shell env by default but CI build was failing because of this. Adding the env tells Turbo to inject the needed env var into the isolated task and make them available during build

2. Updated `emotion to v11` and changed all packages to their new names: https://emotion.sh/docs/emotion-11 Also the imports where updated to reflect this change, along with `babel.config.js` 

3. Got rid of  `experimental: { esmExternals: 'loose' }` in the `next.config.js` that was added to resolve import issues between ESM/CJS when building with next on the website (https://github.com/phosphor-icons/react/issues/135) We no longer need to transpile the phosphor icons as the recommended pinned version works `phosphor-icons/react@v2.1.8`

4. Updated `next-mdx-remote to v5` which introduced breaking changes. We had some data passed to the source props used to render custom components but with the update those were removed. A custom plugin was added to manually map and make this data available, see here: https://github.com/contentful/forma-36/pull/3185/commits/34ae0a643475791814f3402811a4e94710fb03b0
then we adjusted the custom components to be able to render `inline code blocks`, `fenced code blocks` and `live code blocks with file imports` correctly, see here: https://github.com/contentful/forma-36/pull/3185/commits/c920ebd9e1b91fc1776543c359fb3cdfa38ff0d9

Also, extended the components needed in `mdxrenderer` to include the missing ones needed for the `Typography Page` so it renders correctly.

The tokens import on the mdx components pages were updated to match the new v6 api, from this: https://github.com/contentful/forma-36/blob/main/packages/website/mdx-components/BorderRadiusTable.tsx to https://github.com/contentful/forma-36/blob/a44e472820291ae4736ae50a2b5bf52c00770846/packages/website/mdx-components/BorderRadiusTable.tsx

5. Updated [react-copy-to-clipbaord](https://github.com/nkbt/react-copy-to-clipboard) to [copy-to-clipbaord](https://github.com/sudodoki/copy-to-clipboard) since the former doesn't support react 19

6. In the `DatePicker` component we currently use `v8` of the `react-day-picker` but the peer dependencies of this package doesn't resolve `react 19` so I added it to the overrides in package.json to bypass this issue for now (its possible to upgrade to v9 of this package as it works with react 19 but it introduces a lot of breaking changes that is out of scope for this PR)

### DIFF
--- a/packages/website/next-env.d.ts
+++ b/packages/website/next-env.d.ts
@@ -1,5 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
+// eslint-disable-next-line @typescript-eslint/triple-slash-reference
 /// <reference path="./.next/types/routes.d.ts" />
 
 // NOTE: This file should not be edited


package-lock.json
veu, line 1: Why is this change so big when there is just one tiny change in the package.json?

packages/website/package.json
```diff
@@ -29,28 +28,29 @@
...
+    "next": "^15.1.0",
     "next-auth": "^4.24.11",
-    "next-mdx-remote": "^3.0.8",
+    "next-mdx-remote": "^5.0.0",
     "next-transpile-modules": "^10.0.1",
     "prism-react-renderer": "1.3.5",
     "qs": "^6.10.3",
-    "react": "17.0.2",
-    "react-copy-to-clipboard": "^5.1.0",
+    "react": ">=19.1.0",
```
veu: Let’s limit this to minor version updates as all major versions will need some manual work.

packages/website/pages/[...slug].tsx
```diff
@@ -0,0 +1,325 @@
...
+    const matches = content.match(shortIntroRegex);
+
+    if (matches !== null) {
+      shortIntroText = matches[0];
+      mainContentText = content.replace(matches[0], '');
+    }
+
+    const shortIntro = await serialize({
+      value: shortIntroText,
+      path: mdxSource.filepath,
```
Comment (veu) [inline packages/website/pages/[...slug].tsx]: For my understanding: why do we need to pass a path?

packages/website/public/docs-images/typeface.png
veu, line 1: Looks like it’s no longer used.

packages/website/public/oh-snap.png
veu, line 1: Looks like it’s no longer used.

package-lock.json
missating: I did a clean install & it still didn't change but this also contains packages for the website too

veu: You’re right. Still feels like a lot seeing that a third of the file has changed. Is the website that heavy in dependencies?

packages/website/pages/[...slug].tsx
Comment (missating) [reply packages/website/pages/[...slug].tsx]: ah, good catch, `shortIntro` is pulled from `frontmatter` inside the mdx file so we don't need to add a path for it since its just a text. i would remove, thanks.

PR #3186, opened 2025-10-01
feat: Upgrade components :six: [UFO-1987] [UFO-1988] [UFO-1989] [UFO-1992]

Lelith:
##  Summary
This PR replaces the **deprecated `react-popper`** package with its actively maintained successor, **[`@floating-ui/react`](https://floating-ui.com/)**.  
`floating-ui` provides improved flexibility, performance, and more robust positioning behavior for UI elements.

Also contains react 19 upgrades for the following other components
* Modal
* Notification
* Tabs

---

## Affected Components
The following components previously relied directly on `react-popper` and have been refactored to use `floating-ui`:
- **Popover**
- **Menu**
- **Tooltip**

Additionally, several components that indirectly depend on these were updated or tested to ensure compatibility:
- Autocomplete
- Avatar
- IconButton
- Card
- CopyButton
- Datepicker
- EntityList
- Multiselect
- Navbar
- Pill
- UsageCard

---

## Changes
- Removed `react-popper` dependency.
- Added and configured `@floating-ui/react`.
- Refactored Popover, Menu, and Tooltip to use `floating-ui`.
- Updated positioning logic and event handling for better consistency and performance.
- Adjusted tests and stories where necessary to reflect the new implementation.

### **Menu**
- `Menu.Submenu` has been **deprecated**.  
  - Submenus can now be created using the standard `Menu` component.  
  - See migration guidelines for updated usage examples.  

### **Placement Property**
- The `placement` prop for **Popover**, **Menu**, and **Tooltip** has changed to match the `floating-ui` API.  
  - Old placement values (e.g. `"auto-start"`, `"auto-end"`) need to be updated.  
  - Refer to the migration guide for the full list of valid placement options.

### DIFF
--- a/packages/components/forms/stories/InputGroup.stories.tsx
+++ b/packages/components/forms/stories/InputGroup.stories.tsx
@@ -182,6 +182,12 @@ export const Overview = () => {
             variant="secondary"
             icon={<LockSimpleIcon />}
             aria-label="Lock"
+          />
+
+          <IconButton
+            variant="secondary"
+            icon={<LockSimpleIcon />}
+            aria-label="Lock"
             withTooltip
             tooltipProps={{ content: 'Tooltip text' }}
           />


packages/components/menu/Migration.md
```diff
@@ -0,0 +1,217 @@
+# Migration of Menu from v5.x to 6.x
+
+## Breaking change: Changed prop types
+
+The Men component introduces two changes:
```
veu:
```suggestion
The Menu component introduces two changes:
```

packages/components/menu/Migration.md
```diff
@@ -0,0 +1,217 @@
...
+| `id`                     | `string`                                                                                                                                                                                       | `string`                                                                                                                                                                      | undefined → undefined                 | Unchanged                                                                                                                                     |
+| `offset`                 | `[number, number]`                                                                                                                                                                             | `number` OR { mainAxis?: number; crossAxis?: number; alignmentAxis?: number }`                                                                                                | [1,4] → 0                             | Changed Requires migration (see mapping)                                                                                                      |
```
veu, line 29: Sanity check: does that look different?

packages/components/menu/Migration.md
```diff
@@ -0,0 +1,217 @@
...
+
+```tsx
+<Menu>
+  <Menu.Trigger>
+    <Button>Actions</Button>
+  </Menu.Trigger>
+  <Menu.List>
+    <Menu.Item>Item 1</Menu.Item>
+    <Menu>
+      <Menu.SubmenuTrigger>
```
veu: Mismatch in opening/closing tag. Also, do we want to deprecate `SubmenuTrigger` as well?

packages/components/autocomplete/src/Autocomplete.test.tsx
```diff
@@ -78,15 +78,15 @@ describe('Autocomplete', () => {
...
+      expect(screen.queryByRole('listbox')).toBeVisible();
 
       // go to the list first item
       await user.keyboard('[ArrowDown]');
 
       // checks if the first item of the list gets selected
       const listFirstItem = screen.getAllByRole('option')[0];
       expect(listFirstItem.getAttribute('aria-selected')).toBe('true');
-      expect(listFirstItem.getAttribute('class')).toContain('highlighted');
+      expect(listFirstItem.getAttribute('data-highlighted')).toBeTruthy();
```
veu, line 89: Is the `data-highlighted` attribute used for anything but the test? Looks like we are checking the same thing twice in the test.

packages/components/autocomplete/src/AutocompleteItems.tsx
```diff
@@ -47,8 +47,10 @@ export const AutocompleteItems = <ItemType,>(
               styles.item({}),
               highlightedIndex === itemIndex && styles.highlighted,
             ])}
+            data-highlighted={highlightedIndex === itemIndex}
             data-test-id={`cf-autocomplete-list-item-${itemIndex}`}
           >
+            {/* @ts-expect-error render item type issues*/}
```
veu: TypeScript is correct about complaining here. You want to limit the type to make sure that the result is always a `ReactNode` but currently you’re allowing an `ItemType` that can’t be rendered.
The easiest way to fix this is to reduce the line below to `{renderItem(item, inputValue)}` and make `renderItem` required in `Autocomplete`.

packages/components/autocomplete/stories/Autocomplete.stories.tsx
```diff
@@ -92,7 +91,7 @@ export const Basic = (args: AutocompleteProps<string>) => {
     >
       {/* It’s not necessary to pass "Fruit" (type of one item)  */}
       <Autocomplete<string>
-        {...args}
+        placeholder="select your favorit fruit"
```
veu:
```suggestion
        placeholder="select your favorite fruit"
```

packages/components/autocomplete/stories/Autocomplete.stories.tsx
```diff
@@ -178,7 +171,7 @@ export const ControlledFromOutside = (args: AutocompleteProps<Produce>) => {
       alignItems="start"
     >
       <Autocomplete<Produce>
-        {...args}
+        placeholder="select your favorit fruit"
```
veu:
```suggestion
        placeholder="select your favorite fruit"
```

packages/components/autocomplete/stories/Autocomplete.stories.tsx
```diff
@@ -236,7 +226,7 @@ export const UsingGroupedItems = (args: AutocompleteProps<Produce>) => {
     >
       {/* It’s not necessary to pass "Fruit" (type of one item)  */}
       <Autocomplete<Produce>
-        {...args}
+        placeholder="select your favorit fruit"
```
veu:
```suggestion
        placeholder="select your favorite fruit"
```

packages/components/autocomplete/stories/Autocomplete.stories.tsx
```diff
@@ -281,7 +268,7 @@ export const MultipleSelection = (args: AutocompleteProps<Produce>) => {
...
```
veu:
```suggestion
        placeholder="select your favorite fruit"
```

packages/components/avatar/src/AvatarGroup/AvatarGroup.tsx
```diff
@@ -48,14 +48,20 @@ function _AvatarGroup(
       })}
     >
       {childrenToRender.map((child, index) => {
+        if (!React.isValidElement(child)) {
```
veu, line 51: Should we swallow those errors silently or print a warning at least in dev mode?

packages/components/card/examples/AssetCard/AssetCardActionsExample.tsx
```diff
@@ -8,14 +9,7 @@ export default function AssetCardActionsExample() {
...
       src="https://bit.ly/31yL3Ps"
-      actions={[
-        <MenuItem key="copy" onClick={() => alert('copy')}>
-          Copy
-        </MenuItem>,
-        <MenuItem key="delete" onClick={() => alert('delete')}>
-          Delete
-        </MenuItem>,
-      ]}
+      actions={[<li key="copy">Copy</li>, <li key="delete">Delete</li>]}
```
veu: Is this change temporary because we can’t use the `MenuItem` or are we changing how we’re defining actions?

packages/components/forms/src/Checkbox/Checkbox.tsx
```diff
@@ -66,5 +66,5 @@ const CheckboxBase = (
   );
 };
 
-CheckboxBase.displayName = 'Checkbox';
+CheckboxBase.displayName = 'CheckboxBase';
```
veu:
```suggestion
CheckboxBase.displayName = 'Checkbox';
```

packages/components/icon/stories/Icon.stories.tsx
```diff
@@ -31,7 +31,9 @@ export const Sizes: StoryFn = () => {
 
       <Stack>
         <Flex alignItems="center" gap="spacingS">
-          <CalendarBlankIcon size="tiny" /> <Text>Tiny</Text>
+          <Icon as={CalendarBlankIcon} testId="my-other-test-id" size="tiny" />
```
veu, line 34: What functionality do the `testId` properties have outside of a test?

packages/components/modal/src/Modal.tsx
```diff
@@ -187,7 +187,9 @@ export const Modal = ({
           />
         )}
         <ModalContent {...otherProps.modalContentProps}>
-          {otherProps.children}
+          {typeof otherProps.children === 'function'
```
veu, line 190: Shouldn’t we execute the function in this case instead of skipping it?

packages/components/multiselect/src/Multiselect.tsx
```diff
@@ -1,21 +1,24 @@
...
-import { cx } from 'emotion';
+import { cx } from '@emotion/css';
 
 import { mergeRefs, type CommonProps, Flex } from '@contentful/f36-core';
 import { Button, IconButton, type ButtonProps } from '@contentful/f36-button';
 import { CaretDownIcon, XIcon } from '@contentful/f36-icons';
 
 import { SkeletonContainer, SkeletonBodyText } from '@contentful/f36-skeleton';
-import { Popover, type PopoverProps } from '@contentful/f36-popover';
+import { Popover, type PopoverProps } from '../../popover';
```
veu: Does this mean we have a duplicated `Popover` component when importing both `@contentful/f36-popover` and `@contentful/f36-multiselect`?

packages/components/navbar/src/NavbarItem/NavbarItem.tsx
```diff
@@ -120,7 +121,10 @@ const isNavbarItemHasMenu = <E extends React.ElementType>(
 ): props is PolymorphicProps<NavbarItemAsMenuProps, E> =>
   Boolean(props.children);
 
-export const NavbarItem: PolymorphicComponent<
+NavbarItemBase.displayName = 'NavbarItemBase';
```
veu:
```suggestion
NavbarItemBase.displayName = 'NavbarItem';
```

packages/components/tooltip/stories/Tooltip.stories.tsx
```diff
@@ -81,7 +81,7 @@ AutoPlacement.args = {
   ),
 };
 
-const AutoPlacementSourceCode = `<Tooltip place="auto" content={<>I will reposition automatically<br/>when you scroll</>}>
+const AutoPlacementSourceCode = `<Tooltip placement="auto" content={<>I will reposition automatically<br/>when you scroll</>}>
```
veu, line 89: This looks off

Image

packages/components/multiselect/src/Multiselect.tsx
Lelith: that's a leftover from when i was developing in a state where popover was not able to build yet. will revert that change! good catch. it should import popover from the popover packae and not the relative path!

packages/components/icon/stories/Icon.stories.tsx
Lelith: In the current implementation we have an issue that the specific property `testId` gets handed to the underlying svg tag, which causes a warning message in the console about an unsupported property. I fixed the implementation of this forwarding, which is why I wanted to test / showcase that it works also with custom testIds and not only the default testId we already set.

packages/components/menu/Migration.md
Lelith: I am not sure what you are refering to. The offset value currently is expected to come in an array, but floating-ui wants it to be a specific shaped object.
The axes themselves are then also dependent on the placement. Like the main axis is always the axis of the placement (bottom / top or left/right)

packages/components/menu/Migration.md
Lelith: SubmenuTrigger is a component we need to keep as this Trigger has to have a specific look and other than the Menu.Trigger as it should look like one of the Menu.Items with a caret on the end.

packages/components/modal/src/Modal.tsx
Lelith: we have to double check the functionality here. All tests and examples did run as expected, which is why i thought this is just a type issue to be solved, because if children are a function, they can not be rendered directly like `{children}`

packages/components/menu/Migration.md
veu: Mainly that the resulting UI appears to be different because the default offset is gone.

packages/components/tooltip/src/useTooltip.tsx
```diff
@@ -0,0 +1,95 @@
...
+  if (placement === 'auto') {
+    middleware.push(autoPlacement({ padding: 5 }));
+  } else {
+    sanitizedPlacement = placement;
+    middleware.push(shift(), flip());
+  }
+
+  if (
+    sanitizedPlacement.startsWith('right') ||
+    sanitizedPlacement.toString().startsWith('left') ||
```
veu:
```suggestion
    sanitizedPlacement.startsWith('left') ||
```

packages/components/tooltip/src/useTooltip.tsx
```diff
@@ -0,0 +1,95 @@
...
+  placement = 'auto',
+  showDelay,
+  hideDelay,
+  usePortal,
+}: TooltipOptions = {}) {
+  const arrowRef = React.useRef(null);
+  const [isOpen, setIsOpen] = React.useState(isVisible);
+
+  let sanitizedPlacement: Placement = 'top';
+  const middleware = [];
```
veu: This should be wrapped in `useMemo` to prevent constant rerendering due to the newly created array.

packages/components/tooltip/src/useTooltip.tsx
```diff
@@ -0,0 +1,95 @@
...
+    open: isOpen,
+    onOpenChange: setIsOpen,
+    whileElementsMounted: autoUpdate,
+    middleware,
+  });
+
+  const context = data.context;
+
+  const hover = useHover(context, {
+    delay: { open: showDelay, close: hideDelay },
```
veu, line 79: Let’s make a lot of friends and use the [safe polygon](https://floating-ui.com/docs/useHover#safepolygon) while we’re at it.
```
handleClose: safePolygon({ requireIntent: false }),
```
Note: `requireIntent` should be `false` to not punish people who are (sometimes) slow with the mouse.

packages/components/tooltip/src/TooltipTrigger.tsx
```diff
@@ -0,0 +1,53 @@
...
+>(function TooltipTrigger({ children, tooltipId, ...otherProps }, propRef) {
+  const context = useTooltipContext();
+
+  const childRef: React.Ref<HTMLElement> | undefined = (
+    children as unknown as { ref?: React.Ref<HTMLElement> }
+  ).ref;
+  const baseRef = useMergeRefs([propRef, context.refs.setReference]);
+  const ref = useMergeRefs([childRef, baseRef]);
+
+  if (React.isValidElement(children)) {
```
veu: Should we log an error at least in dev mode rather than silently hiding the element?

packages/components/tooltip/src/Tooltip.test.tsx
```diff
@@ -41,55 +42,39 @@ describe('Tooltip', () => {
...
-
-    await user.hover(screen.getByText('Hover me'));
-
-    await waitFor(() =>
-      expect(
-        screen.getByRole('tooltip').getAttribute('data-popper-placement'),
-      ).toBe('left'),
-    );
-  });
+  // Enable again once emotion/jest is compatible again
```
veu: Let’s mark this as an explicit TODO to make sure we find it again.

packages/components/card/examples/EntryCard/EntryCardActionsExample.tsx
```diff
@@ -9,12 +10,14 @@ export default function EntryCardActionsExample() {
...
       actions={[
-        <MenuItem key="copy" onClick={() => alert('copy')}>
-          Copy
-        </MenuItem>,
-        <MenuItem key="delete" onClick={() => alert('delete')}>
-          Delete
-        </MenuItem>,
+        <li>Copy</li>,
+        <li>Delete</li>,
+        // <MenuItem key="copy" onClick={() => alert('copy')}>
```
veu: Let’s add an explicit TODO to fix this.

packages/components/card/examples/InlineEntryCard/BasicInlineEntryCardExample.tsx
```diff
@@ -1,64 +1,47 @@
...
 
 export default function BasicInlineEntryCardExample() {
   return (
     <>
       <InlineEntryCard
-        actions={[
-          <MenuItem key="copy">Copy</MenuItem>,
-          <MenuItem key="delete">Delete</MenuItem>,
-        ]}
+        actions={[<li key="copy">Copy</li>, <li key="delete">Delete</li>]}
```
veu: TODO

packages/components/card/examples/InlineEntryCard/InlineEntryCardInTextExample.tsx
```diff
@@ -1,15 +1,13 @@
...
 export default function InlineEntryCardInTextExample() {
   return (
     <Text>
       Macbeth (/məkˈbɛθ/, full title The Tragedie of Macbeth) is a tragedy by{' '}
       <InlineEntryCard
-        actions={[
-          <MenuItem key="copy">Copy</MenuItem>,
-          <MenuItem key="delete">Delete</MenuItem>,
-        ]}
+        actions={[<li key="copy">Copy</li>, <li key="delete">Delete</li>]}
```
veu: TODO

packages/components/card/examples/InlineEntryCard/InlineEntryCardWithActionsExample.tsx
```diff
@@ -10,18 +14,18 @@ export default function InlineEntryCardWithActionsExample() {
   return (
     <InlineEntryCard
       actions={[
-        <MenuItem key="edit" onClick={handleEditAction}>
+        <li key="edit" onClick={handleEditAction}>
```
veu: TODO

packages/components/multiselect/src/Multiselect.tsx
```diff
@@ -331,47 +293,60 @@ function _Multiselect(props: MultiselectProps, ref: React.Ref<HTMLDivElement>) {
...
-                searchValue={searchValue}
-                focusList={focusList}
-              />
+              </div>
             )}
-            {isLoading && <ListItemLoadingState />}
-
-            {!isLoading && optionsLength > 0 && (
-              <ul className={styles.list} data-test-id="cf-multiselect-items">
-                {hasSearch ? enrichOptions(children) : children}
```
veu, line 362: Why are we removing this logic? Is it implemented somewhere else or are we changing the behavior?

packages/components/popover/src/PopoverTrigger/PopoverTrigger.tsx
```diff
@@ -1,19 +1,50 @@
...
-    ...getTriggerProps(child.props, child.ref),
-    'aria-haspopup': child.props['aria-haspopup'] ?? 'dialog',
+  // Existing ref on the child (if any) so we can merge it.
+  const childRef: React.Ref<HTMLElement> | undefined = (
+    children as unknown as { ref?: React.Ref<HTMLElement> }
+  ).ref;
+
+  const ref = useMergeRefs([context?.refs.setReference, propRef, childRef]);
+
+  if (!React.isValidElement(children)) return null;
```
veu: Should we log this as a warning at least in dev mode rather than silently hiding whatever is passed?

packages/components/popover/src/usePopover.ts
```diff
@@ -0,0 +1,134 @@
...
+  isOpen = false,
+  offset: offsetOption = 0,
+  onClose,
+  renderOnlyWhenOpen = true,
+  usePortal = true,
+  closeOnEsc = true,
+  closeOnBlur = true,
+  autoFocus = true,
+}: PopoverOptions = {}) {
+  const [labelId, setLabelId] = React.useState<string | undefined>();
```
veu, line 47: `labelId` and `descriptionId` are read but I cannot find a way to set them.

packages/components/tooltip/src/useTooltip.tsx
cf-remylenoir: The middleware should not be memoized, every middleware mutation (offset) gets accumulated and the tooltip displayed a bit further every time otherwise.

packages/components/tooltip/src/TooltipContent.tsx
```diff
@@ -0,0 +1,46 @@
...
+  const wrappedTooltipContents = content ? (
+    <Box
+      {...state.getFloatingProps()}
+      {...otherProps}
+      className={cx(styles.tooltip, className)}
+      style={{ ...style, ...state.floatingStyles }}
+      ref={ref}
+      id={id}
+    >
+      <span aria-label={label}>{content}</span>
```
cf-remylenoir: Aria labels can't be used on static elements

```suggestion
      <span>{content}</span>
```

packages/components/card/examples/InlineEntryCard/BasicInlineEntryCardExample.tsx
cf-remylenoir: What is todo?

packages/components/tooltip/src/TooltipContent.tsx
```diff
@@ -0,0 +1,46 @@
...
+>(({ content, label, style, className, id, ...otherProps }, propRef) => {
+  const state = useTooltipContext();
+  const ref = useMergeRefs([state.refs.setFloating, propRef]);
+  const styles = getTooltipContentStyles(state.isOpen);
+  if (!state.isOpen) {
+    return null;
+  }
+
+  const wrappedTooltipContents = content ? (
+    <Box
```
cf-remylenoir, line 29: We need to keep the `as="span"`, otherwise we break the `InlineEntryCard` behavior.

packages/components/tooltip/src/Tooltip.tsx
```diff
@@ -205,79 +159,38 @@ export const Tooltip = ({
...
-      className={cx(styles.tooltip, className)}
-      testId={testId}
-      onMouseEnter={() => {
-        clearTimeouts();
-        setShow(true);
-      }}
-      onMouseLeave={() => {
-        clearTimeouts();
-        setShow(false);
-      }}
```
cf-remylenoir, line 224: Why don't we trigger the display anymore? The InlineEntryCard tooltip is not triggered when hovering now.

packages/components/popover/src/PopoverContent/PopoverContent.tsx
```diff
@@ -18,27 +23,49 @@ export type PopoverContentProps = PropsWithHTMLElement<
...
+  }
+
   const content = (
     <div
       {...otherProps}
-      {...getPopoverProps(otherProps, ref)}
       className={cx(styles.container, className)}
+      style={{ ...context.floatingStyles }}
+      {...getFloatingProps()}
+      aria-labelledby={context.labelId}
```
cf-remylenoir: Aria labels can't be used on static elements
```suggestion
```

package.json
```diff
@@ -86,13 +86,6 @@
     "eslint-plugin-react": "^7.37.5",
     "eslint-plugin-react-hooks": "^5.2.0",
     "eslint-plugin-storybook": "^9.1.6",
-    "eslint-config-prettier": "^10.1.8",
-    "eslint-plugin-import": "^2.32.0",
-    "eslint-plugin-jest": "^29.0.1",
-    "eslint-plugin-jsx-a11y": "^6.10.2",
-    "eslint-plugin-react": "^7.37.5",
-    "eslint-plugin-react-hooks": "^5.2.0",
-    "eslint-plugin-storybook": "^9.1.6",
```
cf-remylenoir: Info, these were duplicated, I removed them.

packages/components/tooltip/src/TooltipTrigger.tsx
```diff
@@ -0,0 +1,57 @@
...
+    console.error(
+      'Only valid React elements are supported - https://react.dev/reference/react/isValidElement',
+    );
+    return null;
+  }
+
+  const childType = typeof children.type;
+  const childProps = children.props as Record<string, unknown>;
+  const ariaHasPopup = (childProps['aria-haspopup'] ??
+    'dialog') as React.AriaAttributes['aria-haspopup'];
```
cf-remylenoir, line 28: How does it relate to https://floating-ui.com/docs/userole#native-roles?

packages/components/modal/src/ModalLauncher/ModalLauncher.tsx
```diff
@@ -58,18 +58,29 @@ async function closeAll(): Promise<void> {
...
+    // Reuse the existing root if present
+    const modalData = openModalsIds.get(rootElId);
+    if (modalData) {
+      root = modalData.root;
+    } else {
+      root = ReactDOM.createRoot(rootDom);
+    }
+  } else {
+    root = ReactDOM.createRoot(rootDom);
+  }
```
missating: This was updated because of the `console error` when using the modal launcher. it complains about an exiting root. This just prevents that.

veu: We’ve added the `finally` statement above that removes the ID from `openModalsIds` even when unmounting the root fails. This code assumes that the root is still under the ID in `openModalsIds` if something goes wrong. Shouldn’t these be mutually exclusive or is there another case that I’m overlooking?

missating: The `getRoot` function is designed to reuse DOM containers but the `open` function was always calling `ReactDOM.createRoot()` on the returned DOM element. So if we called `ModalLauncher.open()` twice with the same `modalid` it first creates DOM container + React root then second call reuses DOM container but tries to create another React root which didn't account for React 18's restriction that each DOM container can only have one React root which was the console error I was talking about.

You are correct about the `finally` statement but it was doing an unconditional cleanup without checking if the entry existed and this didn’t account for the fact that another cleanup path could run concurrently which I’ve now updated to run without the side effects.

veu: Thanks for the explanation. 🙂

packages/components/popover/src/usePopover.ts
missating: I'm thinking this is supposed to be like this, so they can be managed by the consuming component (eg: setting meaningful Ids that relates to the specific use case) rather than auto generated by the hook?

packages/components/accordion/stories/Accordion.stories.tsx
```diff
@@ -17,71 +18,81 @@ export default {
...
-    </Accordion.Item>
-    <Accordion.Item title={args['Accordion.Item Title #3']}>
-      <Text as="p">{args['Accordion.Item Content #3']}</Text>
-    </Accordion.Item>
-  </Accordion>
-);
+export const Basic: Story = {
+  render: ({ align }: AccordionProps) => (
+    <Accordion align={align}>
+      <Accordion.Item title={'Accordion.Item Title #1'}>
```
veu: This used to contain the text "Accordion I". Now it’s using "Accordion.Item Content #1" which used to be the internally used key. I think we should stick with the former.

packages/components/popover/src/usePopover.ts
veu: Right, I think I misinterpreted this last time. It’s meant to be set through the context. I was looking for a way to pass it somewhere. Dismiss.

packages/components/tooltip/src/Tooltip.tsx
missating: Screenshot 2025-10-28 at 14 11 10

Just checked the `InlineEntryCard` storybook locally and I can see everything gets displayed as expected, like here: https://f36-storybook.contentful.com/?path=/story/components-card-inlineentrycard--overview

is that what you are referring to?

cf-remylenoir: The issue surfaced in the stories where the `InlineEntryCard` is part of a paragraph, e.g., https://f36-storybook.contentful.com/?path=/story/components-card-inlineentrycard--default.

missating: yes, I checked there too and the tooltip displays as expected

Screenshot 2025-10-28 at 14 27 50

cf-remylenoir: Sorry, I mixed up two things. I thought we were talking about the inline being broken because of the previously removed `as="span"`.

cf-remylenoir: I meant the keyboard navigation does not trigger the tooltip anymore, since the code was removed. I am not sure why Kathrin made this decision.

packages/components/tooltip/src/TooltipContent.tsx
missating: The `inline` still works without the span.

packages/components/table/src/TableCell/TableCell.tsx
```diff
@@ -57,12 +57,14 @@ function _TableCell(
     sortButtonAriaLabel,
     ...otherProps
   }: TableCellProps,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   forwardedRef: React.Ref<any>,
```
veu: We could replace `any` with `HTMLElement`, right?

packages/components/typography/src/Text/Text.tsx
```diff
@@ -49,8 +49,10 @@ function wordBreakStyle() {
 export type TextProps<E extends React.ElementType = typeof TEXT_DEFAULT_TAG> =
   PolymorphicProps<TextInternalProps, E, 'color'>;
 
-function _Text<E extends React.ElementType = typeof TEXT_DEFAULT_TAG>(
+function TextBase<E extends React.ElementType = typeof TEXT_DEFAULT_TAG>(
   props: TextProps<E>,
+  // as this is a polymorphic base component we can not narrow down the type
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   ref: React.Ref<any>,
```
veu, line 56: We should be able to at least use `HtmlElement`.

packages/components/button/src/Button/Button.tsx
```diff
@@ -12,15 +12,17 @@ import { Spinner } from '@contentful/f36-spinner';
...
 
 export type ButtonProps<
   E extends React.ElementType = typeof BUTTON_DEFAULT_TAG,
 > = PolymorphicProps<ButtonInternalProps, E, 'disabled'>;
 
-function _Button<E extends React.ElementType = typeof BUTTON_DEFAULT_TAG>(
+function ButtonBase<E extends React.ElementType = typeof BUTTON_DEFAULT_TAG>(
   props: ButtonProps<E>,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   ref: React.Ref<any>,
```
veu: `HTMLElement`?

packages/components/button/stories/IconButton.stories.tsx
```diff
@@ -94,7 +93,7 @@ export const ColoredIconInTransparentIconButton = () => {
           icon={
             <Icon
               as={icons.StarIcon}
-              variant={isActive ? 'primary' : 'warning'}
+              color={isActive ? 'colorPrimary' : 'colorWarning'}
```
veu: Not sure why but this doesn’t work.

packages/components/button/src/IconButton/IconButton.tsx
```diff
@@ -63,8 +63,9 @@ export type IconButtonProps<
   E extends React.ElementType = typeof ICON_BUTTON_DEFAULT_TAG,
 > = PolymorphicProps<ExtendedIconButtonProps, E, 'disabled'>;
 
-function _IconButton<
+function IconButtonBase<
   E extends React.ElementType = typeof ICON_BUTTON_DEFAULT_TAG,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
 >(props: IconButtonProps<E>, ref: React.Ref<any>) {
```
veu: `HTMLElement`?

packages/components/forms/src/FormControl/FormControl.tsx
```diff
@@ -41,6 +41,7 @@ function _FormControl<
     testId = 'cf-ui-form-control',
     ...otherProps
   }: FormControlProps<E>,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
veu: `HTMLElement`?

packages/components/navlist/src/NavListItem/NavListItem.tsx
```diff
@@ -32,8 +32,9 @@ export type NavListItemProps<
   E extends React.ElementType = typeof NAV_LIST_ITEM_DEFAULT_TAG,
 > = PolymorphicProps<NavListItemInternalProps, E>;
 
-function _NavListItem<
+function NavListItemBase<
   E extends React.ElementType = typeof NAV_LIST_ITEM_DEFAULT_TAG,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
 >(props: NavListItemProps<E>, ref: React.Ref<any>) {
```
veu: `HTMLElement`?

packages/components/drag-handle/src/DragHandle.tsx
```diff
@@ -69,8 +69,9 @@ export type DragHandleProps<
   E extends ElementType = typeof DRAG_HANDLE_DEFAULT_TAG,
 > = PolymorphicProps<DragHandleInternalProps, E>;
 
-function _DragHandle<E extends ElementType = typeof DRAG_HANDLE_DEFAULT_TAG>(
+function DragHandleBase<E extends ElementType = typeof DRAG_HANDLE_DEFAULT_TAG>(
   props: DragHandleProps<E>,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   ref: Ref<any>,
```
veu: `HTMLElement`?

veu: `HTMLElement`?

packages/components/forms/stories/FormControl.stories.tsx
```diff
@@ -22,168 +22,221 @@ export default {
...
 
-export const WithCheckboxGroup = (args: ComponentProps<typeof FormControl>) => {
-  return (
-    <>
-      <FormControl as="fieldset" {...args}>
-        <FormControl.Label as="legend" marginBottom="none">
-          Select your ingredients
-        </FormControl.Label>
-        <Paragraph>No extra costs</Paragraph>
+          {args.isInvalid && (
```
veu: In the old story all controls were highlighted and all errors shown. We should remove the checks for `args.isInvalid` and get back to the original state.

packages/core/src/Box/Box.tsx
```diff
@@ -78,8 +77,10 @@ export function useBox<E extends React.ElementType = typeof BOX_DEFAULT_TAG>(
   };
 }
 
-function _Box<E extends React.ElementType = typeof BOX_DEFAULT_TAG>(
+function BoxBase<E extends React.ElementType = typeof BOX_DEFAULT_TAG>(
   props: BoxProps<E>,
+  // as this is a polymorphic base component we can not narrow down the type
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   ref: React.Ref<any>,
```
veu: `HTMLElement`?

eslint.config.mjs
```diff
@@ -1,147 +1,157 @@
...
-  },
-  env: {
-    jest: true,
-    browser: true,
-    node: true,
+export default [
+  {
+    ignores: [
+      'node_modules/**',
+      '.next/**',
```
veu: This and `next-env.d.ts` no longer exist.

scripts/.storybook/preview.tsx
```diff
@@ -0,0 +1,37 @@
...
+    (Story) => {
+      return (
+        <>
+          <GlobalStyles />
+          <Story />
+        </>
+      );
+    },
+  ],
+  parameters: {
```
veu, line 17: Is removing the [order in the sidebar](https://github.com/contentful/forma-36/blob/main/scripts/.storybook/preview.js#L19) intentional?

packages/components/accordion/stories/Accordion.stories.tsx
```diff
@@ -17,71 +18,81 @@ export default {
...
-            <Text as="p">{args['Accordion.Item Content #2']}</Text>
-          </Accordion.Item>
-          <Accordion.Item title={args['Accordion.Item Title #3']}>
-            <Text as="p">{args['Accordion.Item Content #3']}</Text>
-          </Accordion.Item>
-        </Accordion>
+        <Flex>
+          <Accordion align="start">
+            <Accordion.Item title={'Accordion Title'}>
+              <Text as="p">Accordion.Item Content #1</Text>
```
veu: We still have the key instead of the previous text in a couple places.

packages/components/button/src/Button/Button.tsx
```diff
@@ -75,28 +67,24 @@ function _Button<E extends React.ElementType = typeof BUTTON_DEFAULT_TAG>(
           // We need to pass the color to the icons to enable the usaged of the V5 icons
           // it may change in the future
           color:
-            (variant === 'transparent' &&
-              icon.props.variant === undefined &&
-              icon.props.color) ||
-            'currentColor',
+            (variant === 'transparent' && icon.props.color) || 'currentColor',
           // we want to allow variants for icons for transparent buttons
```
veu: This comment refers to removed code.

packages/components/forms/stories/FormControl.stories.tsx
missating: I'll check this in the story branch and see if its different there.

scripts/.storybook/main.ts
```diff
@@ -10,16 +10,25 @@ function getAbsolutePath(value: string): any {
   return dirname(require.resolve(join(value, 'package.json')));
 }
 const config: StorybookConfig = {
-  stories: ['./docs/**/*.stories.mdx', '../../packages/**/*.stories.@(ts|md)x'],
+  stories: [
+    './docs/**/*.stories.mdx',
+    '../../packages/**/*.stories.@(mdx|tsx|ts|jsx|js)',
```
veu: The old file pattern was enough. We don’t have any of the other files.

packages/forma-36-tokens/tools/build.js
```diff
@@ -71,16 +71,15 @@ const buildScssTokens = (srcPath, tokens) => {
...
+  return fse.outputFile(srcPath, cjs);
+};
 
-    module.exports = ${JSON.stringify(tokens, null, 2)}
-    `,
-  );
+const buildIndexMJS = (srcPath, tokens) => {
+  const objectLiteral = JSON.stringify(tokens, null, 2);
+  const esm = `// ESM build for @contentful/f36-tokens\nconst tokens = ${objectLiteral};\nexport default tokens;\nexport { tokens };\n`;
+  return fse.outputFile(srcPath.replace(/index\.js$/, 'index.mjs'), esm);
```
veu: Let’s replace this line with `return fse.outputFile(srcPath, mjs);`

packages/forma-36-tokens/tools/build.js
```diff
@@ -244,5 +243,6 @@ const generateIndex = (paths, extension) => {
   allTokens = camelCaseKeys(allTokens);
 
   await buildIndexJS('dist/index.js', allTokens);
+  await buildIndexMJS('dist/index.js', allTokens);
```
veu: https://github.com/contentful/forma-36/pull/3186/files#r2477247044
...and use `'dist/index.mjs'` here

packages/components/navbar/src/NavbarItem/NavbarItem.tsx
```diff
@@ -41,8 +41,9 @@ export type NavbarItemProps<
   E extends React.ElementType = typeof NAVBAR_ITEM_DEFAULT_TAG,
 > = PolymorphicProps<NavbarItemOwnProps, E>;
 
-function _NavbarItem(
+function NavbarItemBase(
   props: NavbarItemProps<React.ElementType>,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   ref: React.Ref<any>,
```
veu: One more `any`.

packages/components/multiselect/src/Multiselect.tsx
veu: Checked the logic and it’s all working as intended.

packages/components/multiselect/src/MultiselectOption.tsx
veu: This parameter is no longer used.

packages/components/modal/src/Modal.tsx
veu: I was wrong. There is another check before this code that checks for function and executes it so this check is just for TypeScript.

packages/components/tooltip/src/TooltipTrigger.tsx
veu: Floating UI doesn’t appear to add the role but this code does.

packages/forma-36-tokens/tools/build.js
missating: I don't understand this, are we getting rid of the esm build?

packages/components/multiselect/src/MultiselectOption.tsx
missating: `MultiselectOptionProps` this is passed to the `MultiselectOption` is that what you mean?

veu: `MultiselectOption` now reads `searchValue` from the `MultiselectContext` and ignores the `searchValue` prop making the `searchValue` prop obsolete.

packages/forma-36-tokens/tools/build.js
veu: No, the current version is a hack that gets `'dist/index.js'` as input and then replaces the extension with `.mjs`. Passing `'dist/index.mjs'` directly has the same outcome and we can remove the hack.

scripts/.storybook/preview.tsx
missating: I'm not sure but I'm adding it again.

packages/components/tooltip/src/Tooltip.tsx
missating: fixed

packages/components/button/stories/IconButton.stories.tsx
missating: fixed

PR #3193, opened 2025-10-09
feat: update modal component package to use react 19

missating:
# Purpose of PR

This PR updates the `modal component` package to be compatible with `react 19`

- Updated to `emotion 11` and replaced `css imports` with `@emotion/css`
- In the `ModalLauncher` updated `react-dom api`  to match `react-dom 19 api`: (https://react.dev/reference/react-dom#removed-apis)
- Update Stories to match v9

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2031,6 +2031,15 @@
         "node": ">=6"
       }
     },
+    "node_modules/@babel/register/node_modules/pify": {
+      "version": "4.0.1",
+      "resolved": "https://registry.npmjs.org/pify/-/pify-4.0.1.tgz",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -61,8 +61,6 @@
     "@types/jest": "^30.0.0",
     "@types/jest-axe": "^3.5.9",
     "@types/node": "^20",
-    "@types/react": "^19.1.12",
-    "@types/react-dom": "^19.1.9",
```
Lelith, line 65: is there a specific reason why you needed to remove the react version? is it a leftover from another branch? So far it seems not to have any side-effects when i tried it out locally

PR #3204, opened 2025-10-24
feat(f36-ai-components): add AIChatInput component

seanf-ctfl:
# Purpose of PR

https://contentful.atlassian.net/browse/PIC-707

This PR adds a new component `AIChatInput`, which will be used to capture the user's prompt to be sent to an LLM.

The implementation depends on [TipTap](https://tiptap.dev/docs), a library that allows users to create customised input editors.

This component can be customized in several ways:

1. The user can inject any additional prompt input tools that they may have for their use case (e.g., file upload, model selection, etc.)
2. The user can add any additional TipTap editor extensions. This will allow them to customise the editor behavior (e.g., Custom [mentions](https://tiptap.dev/docs/editor/extensions/nodes/mention) ) 

### Stories

image

image

image

image

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/f36-ai-components/package.json
+++ b/packages/f36-ai-components/package.json
@@ -19,11 +19,10 @@
   "dependencies": {
     "@contentful/f36-components": "^5.6.0",
     "@contentful/f36-button": "^5.6.0",
-    "@contentful/f36-card": "^5.6.0",
     "@contentful/f36-core": "^5.0.0",
-    "@contentful/f36-icons": "^5.6.0",
     "@contentful/f36-header": "^5.6.0",
     "@contentful/f36-layout": "^5.6.0",
+    "@contentful/f36-icons": "^5.6.0",
     "@contentful/f36-tokens": "^5.0.0",
     "@contentful/f36-utils": "^5.2.0",
     "@tiptap/extension-document": "^2.26.3",


packages/f36-ai-components/src/AIChatInput/AIChatInput.tsx
```diff
@@ -0,0 +1,222 @@
...
+          placeholder={placeholder}
+          initialContent={initialContent}
+          onUpdate={onUpdate}
+          onSubmit={onSubmit}
+          editorExtensions={editorExtensions}
+          editorRef={editorRef}
+        />
+        <Flex justifyContent="space-between" alignItems="end">
+          <InputGroup spacing="spacingS" testId={`${testId}-input-tool-group`}>
+            {promptInputTools}
```
elliotmassen, line 71: I wonder if we might want to be stricter with how we handle input tools here. Currently the consuming code can provide anything here, but this could lead to inconsistent styling and experiences for the end-user. Could we instead pass an array of objects, each defining an icon, a click handler, etc? That way the component could control how the button is styled to ensure they always match the designs. What do you think about this an approach?

packages/f36-ai-components/src/AIChatInput/AIChatInput.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+export interface AIChatInputProps extends CommonProps {
+  /** Initial content of the input field. See [here](https://tiptap.dev/docs/editor/api/editor#content) */
+  initialContent?: UseEditorOptions['content'];
+  /** Handler for input field updates */
+  onUpdate?: UseEditorOptions['onUpdate'];
+  /** Placeholder text for the input field */
+  placeholder?: string;
+  /** Indicates if the AI is currently streaming a response */
+  isSteaming?: boolean;
```
elliotmassen: Typo here:


```suggestion
  isStreaming?: boolean;
```

packages/f36-ai-components/stories/AIChatInput.stories.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+const DEFAULT_ARGS: Partial<AIChatInputProps> = {
+  style: { width: '400px' },
+  placeholder: 'Type your message...',
+};
+
+const Template: React.FC<
+  Omit<AIChatInputProps, 'setValue' | 'isSteaming' | 'onSubmit' | 'onStop'>
+> = (args) => {
+  const [isSteaming, setIsSteaming] = React.useState(false);
```
elliotmassen: I can't seem to toggle the `isStreaming` control in Storybook. I faced a similar issue with my PR and have [a solution](https://github.com/contentful/forma-36/blob/520b42e9d8c0e82d1b05dbf83725d5c69aaf540b/packages/f36-ai-components/stories/AIChatLayout.stories.tsx#L56-L62) using `useEffect`.

packages/f36-ai-components/src/AIChatInput/AIChatInput.tsx
seanf-ctfl: There are trade-offs here between flexibility and consistency.

Personally, I would like to side on flexibility and keep the interface open as we do not know what components the client will want to use. For example, in the '_[With Input Tools](https://5fd1dda724cc620021ace8c5-bxuolsfwty.chromatic.com/?path=/story/components-aichatinput--with-input-tools)_' story, I show that the client may want to use a combination of input types (buttons & selections, etc).

So I would like to keep it flexible to allow for this. It would be great to get your thoughts too @mshaaban0

packages/f36-ai-components/src/AIChatInput/AIChatInput.tsx
seanf-ctfl: Fixed in next commit

packages/f36-ai-components/stories/AIChatInput.stories.tsx
seanf-ctfl: Nice! fixed in next commit

seanf-ctfl:
> Nice one - this is really cool!
> 
> I had a minor comment about textarea when it overflows - I wonder if we might consider eeking out a little more height and/or a soft gradient at the bottom so it's not a hard cutoff? Happy to go either way with these!
> 
> Image

Thanks ! I updated this in the latest commit to reduce the margin to give the input more height. I didn't add a gradient at the end, as the user's active cursor will be at the bottom, so an end gradient may distract the user's editing

packages/f36-ai-components/src/AIChatInput/AIChatInput.tsx
```diff
@@ -0,0 +1,222 @@
...
+  return <EditorContent editor={editor} ref={editorContentRef} />;
+};
+
+type AIChatSubmitButtonProps = Pick<
+  AIChatInputProps,
+  'isStreaming' | 'onSubmit' | 'onStop'
+> &
+  Required<Pick<AIChatInputProps, 'testId' | 'onSubmit' | 'editorRef'>>;
+
+const AIChatSubmitButton: React.FC<AIChatSubmitButtonProps> = ({
```
mshaaban0: nit: separate file :)

packages/f36-ai-components/src/AIChatInput/AIChatInput.tsx
```diff
@@ -0,0 +1,222 @@
...
+  );
+}
+
+type AIChatInputTextAreaProps = Pick<
+  AIChatInputProps,
+  'placeholder' | 'initialContent' | 'onUpdate' | 'editorExtensions'
+> &
+  Required<Pick<AIChatInputProps, 'onSubmit' | 'editorRef'>>;
+
+const AIChatInputTextArea: React.FC<AIChatInputTextAreaProps> = ({
```
mshaaban0: nit: separate file :)

packages/f36-ai-components/src/AIChatInput/AIChatInput.tsx
mshaaban0: Yeah, unfortunately we can only provide guidance and examples.. they can still mess it up if they really want to :)

packages/f36-ai-components/src/AIChatInput/AIChatInput.tsx
```diff
@@ -0,0 +1,222 @@
...
+    // Block submits while streaming
+    if (isStreaming) return;
+    props.onSubmit?.(...args);
+  };
+  const internalEditorRef = useRef<Editor>(null);
+  const editorRef = props.editorRef || internalEditorRef;
+
+  return (
+    <Box testId={testId} className={className} style={style} ref={ref}>
+      <Card className={styles.aiChatInputCard}>
```
mshaaban0: Missing box shadow on the wrapper 💅🏼 

Designs
Image
Implementation
Image

packages/f36-ai-components/src/AIChatInput/AIChatInput.tsx
```diff
@@ -0,0 +1,222 @@
...
+  editorRef,
+}) => {
+  const styles = getStyles();
+  const stopIcon = <Box className={styles.stopIcon} />;
+  const submitButton = (
+    <IconButton
+      testId={`${testId}-submit-button`}
+      onClick={() => onSubmit(editorRef.current)}
+      variant="primary"
+      aria-label="send prompt"
```
mshaaban0: let's also add `title` attribute to these buttons please 🙏🏼

packages/f36-ai-components/src/AIChatInput/AIChatInput.tsx
seanf-ctfl: Done in next commit

packages/f36-ai-components/src/AIChatInput/AIChatInput.tsx
seanf-ctfl: Done in next commit

packages/f36-ai-components/src/AIChatInput/AIChatInput.tsx
seanf-ctfl: Good catch, fixed

packages/f36-ai-components/src/AIChatInput/AIChatInput.tsx
seanf-ctfl: done

PR #3212, opened 2025-11-04
feat: add mention support to ai chat input [PIC-710]

seanf-ctfl:
# Purpose of PR

https://contentful.atlassian.net/browse/PIC-708

This PR updates the `AIChatInput` with mention support.

If clients wish to enable mention support, they can define a `mentionConfig`. This will allow them to provide an implementation for the search feature.

## Screenshots

image

image

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/.github/workflows/publish-ai-package-alpha.yml
+++ b/.github/workflows/publish-ai-package-alpha.yml
@@ -1,0 +1,90 @@
+name: Publish AI Package (Alpha)
+
+on:
+  push:
+    branches:
+      - PIC-694-forma36-create-aichatlayout-component
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      id-token: write
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '22'
+          cache: 'npm'
+
+      - name: Retrieve NPM Token from Vault
+        id: vault
+        uses: hashicorp/vault-action@v2.4.3
+        with:
+          url: ${{ secrets.VAULT_URL }}
+          role: ${{ github.event.repository.name }}-github-action
+          method: jwt
+          path: github-actions
+          exportEnv: false
+          secrets: |
+            secret/data/github/github_packages_write GITHUB_PACKAGES_WRITE_TOKEN | GITHUB_PACKAGES_WRITE_TOKEN ;
+            github/token/${{ github.event.repository.name }}-semantic-release token | SEMANTIC_RELEASE_TOKEN ;
+
+      - name: Configure Git
+        run: |
+          git config user.name "${{ github.actor }}"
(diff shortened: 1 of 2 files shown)


packages/f36-ai-components/stories/AIChatInput.stories.tsx
```diff
@@ -122,70 +141,44 @@ export const WithInputTools = {
...
-      ) : (
-        <pre>No suggestions</pre>
-      )}
-    </>
+    <Template
+      {...args}
+      editorRef={editorRef}
+      promptInputTools={
+        <Tooltip
+          content="Use this button or the '@' key specify a content type"
```
toniwowtscherk: @seanf-ctfl  Can we fix the typo to "Use this button or the '@' key to specify a content type" ?

toniwowtscherk:
@seanf-ctfl  great job already! Two observations I made: 

1. If larger content in the input field, the suggestion dropdown appears way underneath the input field. It should appear right underneath or (even better) above the @ 

Screenshot 2025-11-04 at 16 36 07


2. If I type something after the @ that is not included in the dropdown, a strange small grey bar appears in the line under the input 
Screenshot 2025-11-04 at 16 38 48

packages/f36-ai-components/stories/AIChatInput.stories.tsx
seanf-ctfl: Fixed in latest commit

seanf-ctfl:
> @seanf-ctfl thanks for fixing these things.
> 
> Can we change the'@' button to a secondary button style to match the design?
> 
> I also realized following behavior and trying to give you an instruction to reproduce:
> 
> 1. Type @ or press the button
> 2. Type a key that does not reflect a first letter of one of the content types in the dropdown (i.e. 'a')
> 3. Press backspace
> 
> => dropdown appears at the top left of the screen
> 

> Also, I have seen that the dropdown now only appears if I type '@'. If I press the button, it is only appearing if I type a key that represents the first letter of a content type. The expected behavior here would be that it shows up in both cases immediately.

Thank you for the spot.
I have updated these now ([storybook](https://5fd1dda724cc620021ace8c5-fkovbskzme.chromatic.com/?path=/story/components-aichatinput--with-mention-support))

packages/f36-ai-components/src/AIChatInput/AIChatInput.styles.ts
```diff
@@ -19,6 +19,11 @@ export const getStyles = () => {
         float: 'left',
         height: '0',
       },
+      '& .mention': {
+        backgroundColor: tokens.gray200,
+        padding: '2px 4px',
+        borderRadius: '4px',
```
mshaaban0: We couldn't use tokens for these two?

mshaaban0 (approved):
Beautiful.. Thanks 🤩

packages/f36-ai-components/src/AIChatInput/AIChatInput.styles.ts
seanf-ctfl: Added in the latest commit

PR #3216, opened 2025-11-04
feat(f36-ai-components): add AIChatHistory component

elliotmassen:
# Purpose of PR

This PR adds a chat history. This component will be used to display the user's previous threads with the agent. Threads can be optionally sorted into two or three groups, available by a tab selector at the top. The number of groups has been limited in order to ensure there is enough space to display group labels.

This component is intended to be used _within_ the chat layout and therefore it has no border styling nor box shadows included within the component itself.

| Three groups | Two groups | No groups |
| --- | --- | --- |
| Screenshot 2025-11-06 at 14 56 03 | Screenshot 2025-11-06 at 14 55 57 | Screenshot 2025-11-06 at 14 56 09 |

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/f36-ai-components/stories/AIChatHistory.stories.tsx
+++ b/packages/f36-ai-components/stories/AIChatHistory.stories.tsx
@@ -1,3 +1,4 @@
+import { Tooltip } from '@contentful/f36-components';
 import {
   CheckCircleIcon,
   CheckIcon,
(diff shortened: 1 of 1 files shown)


packages/f36-ai-components/src/AIChatHistory/AIChatHistory.tsx
```diff
@@ -0,0 +1,178 @@
...
+        event.preventDefault();
+        newIndex = filteredThreads.length - 1;
+        break;
+      case 'Enter':
+      case ' ': {
+        event.preventDefault();
+        const focusedThread = filteredThreads[focusedThreadIndex];
+        if (focusedThread?.onThreadClick) {
+          focusedThread.onThreadClick();
+        }
```
seanf-ctfl: __nit__ this can be simplified to one line
(if undefined, the execution will be skipped)
```suggestion
focusedThread?.onThreadClick();       
```

packages/f36-ai-components/src/AIChatHistory/AIChatHistoryThread/AIChatHistoryThread.tsx
```diff
@@ -0,0 +1,76 @@
...
+    testId = `cf-ui-ai-chat-history-thread-${thread.id}`,
+  } = props;
+  const styles = getStyles();
+
+  const handleThreadClick = () => {
+    thread.onThreadClick?.();
+  };
+
+  return (
+    <Box
```
seanf-ctfl: Should we use a button for this ? [here](https://5fd1dda724cc620021ace8c5-ywhhfncyfu.chromatic.com/?path=/story/components-button-components--overview)

The benefit would be that we get the animations on click, which are missing currently.

We could also make this list a `Menu`, and then we would also get the key navigation by default

WDYT ?

elliotmassen: Good shout - I went for `NavList` as it felt like the semantically closest option - which in turn makes it a button 👍

toniwowtscherk (approved):
@elliotmassen  great work! Let's go! 🚀

PR #3219, opened 2025-11-06
feat: add square split horizontal icon [TOL-2711]

YvesRijckaert:
# Purpose of PR

Adding missing square split horizontal icon


cf-remylenoir (changes requested):
Thanks @YvesRijckaert! Please run the build command to update the icons package root index imports.

PR #3224, opened 2025-11-11
feat(f36-ai-components): Update AIChatLayout to support multiple headers

elliotmassen:
# Purpose of PR

This PR updates the `AIChatLayout` component to support multiple headers. When the header is changed the new header is transitioned to via a new slider component. This same slider component can also be re-used by consuming code to slider the layout contents as well. This facilitates a "history" view, in which both the header and contents transition when the user clicks a history button in the header.

https://github.com/user-attachments/assets/4324a86f-b79d-4dff-8e99-d2166fcf8fee

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/f36-ai-components/stories/AIChatLayout.stories.tsx
+++ b/packages/f36-ai-components/stories/AIChatLayout.stories.tsx
@@ -9,6 +9,7 @@ import { AIChatInput } from '../src/AIChatInput/AIChatInput';
 import { AIChatLayout } from '../src/AIChatLayout/AIChatLayout';
 import { getStyles } from '../src/AIChatLayout/AIChatLayout.styles';
 import { AIChatMessage } from '../src/AIChatMessage/AIChatMessage';
+import { AIChatSidePanel } from '../src/AIChatSidePanel/AIChatSidePanel';
 import { Slider } from '../src/Slider/Slider';
 import {
   mockChatMessages,
@@ -244,15 +245,7 @@ export const WithChangableHeader = ({ icon, title, variant }) => {
         duration={300}
       >
         {isHistoryMode ? (
-          <div
-            style={{
-              width: '100%',
-              height: '100%',
-              display: 'flex',
-              flexDirection: 'column',
-              padding: '0px 10px',
-            }}
-          >
+          <AIChatSidePanel>
             <AIChatHistory
               threads={threadsWithActions}
               groups={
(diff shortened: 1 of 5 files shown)


packages/f36-ai-components/src/AIChatLayout/AIChatLayout.tsx
```diff
@@ -78,51 +107,178 @@ function _AIChatLayout(props: AIChatLayoutProps, ref: Ref<HTMLDivElement>) {
...
+      !currentIcon &&
+      !currentTitle &&
+      currentButtonsLeft.length === 0 &&
+      currentButtonsRight.length === 0
+    ) {
+      return undefined;
+    }
+
+    const id = `${currentTitle || 'header'}-${JSON.stringify(
+      [...currentButtonsLeft, ...currentButtonsRight].map((b) => b.ariaLabel),
```
nikunj-tilva: I was wondering if we can avoid using JSON.stringify somehow? Since ariaLabel is a optional property this could break

packages/f36-ai-components/stories/AIChatLayout.stories.tsx
```diff
@@ -146,11 +167,155 @@ export const Basic = ({
...
+      variant={variant}
+      header={isHistoryMode ? historyHeader : defaultHeader}
+      testId="chat-layout"
+    >
+      <Slider
+        contentState={{
+          id: isHistoryMode ? 'history' : 'chat',
+          content: isHistoryMode ? (
+            <div
+              style={{
```
mshaaban0, line 270: Is this needed for it to be styled correctly? I think we should account for these inside the components themselves `<AIChatHistory />` and `<AIChatMessage />`

Do you think we need a wrapper for `<AIChatMessage />` and `<AIChatMessages />`? the goal should be the least amount of configuration possible for customers to compose these components

packages/f36-ai-components/src/AIChatLayout/AIChatLayout.tsx
elliotmassen: Good call - I've now simplified this to only rely on the `slideKey` as it was having unexpected consequences using the previous approach as well.

toniwowtscherk (approved):
@elliotmassen thanks for fixing! 🚀

packages/f36-ai-components/stories/AIChatLayout.stories.tsx
elliotmassen: I've added an `AIChatSidePanel` component to encapsulate the styling that supports the history view. I would agree about adding an `AIChatMessageList` or similar to wrap the styles that support the messages too - this would definitely reduce the amount of boilerplate required. Let's chat about it on Thursday 👍

PR #3232, opened 2025-11-13
feat(Pagination): add support for custom total items labels

cf-remylenoir:
# Purpose of PR

We need to support the total items text localization, and as such, enabling its customization through the properties is necessary.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/components/pagination/src/Pagination.tsx
+++ b/packages/components/pagination/src/Pagination.tsx
@@ -30,7 +30,7 @@ type NavigationButtonsProps = {
    */
   previousAriaLabel?: string;
 };
-export interface PaginationProps extends CommonProps {
+interface BasePaginationProps extends CommonProps {
   /**
    * Sets which page is active on the Pagination
    * @default 0
@@ -47,10 +47,6 @@ export interface PaginationProps extends CommonProps {
    * @default 20
    */
   pageLength?: number;
-  /**
-   * Total amount of items the pagination is applied to.
-   */
-  totalItems?: number;
   /**
    * Sets if the View per page selector is shown
    * @default false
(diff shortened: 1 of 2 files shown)


packages/components/pagination/src/Pagination.tsx
```diff
@@ -85,6 +85,14 @@ export interface PaginationProps extends CommonProps {
    * @default { labelNext: "Next", labelPrevious: "Previous", ariaLabelNext: "To next page", ariaLabelPrevious: "To previous page" }
    */
   navigationButtonsProps?: NavigationButtonsProps;
+  /**
+   * Labels for the text when total items are shown
```
Lelith: Maybe we can add an example in the description so it's easier to understand what part of the text is. Also mentioning that this is the postfix and not the full text with both variables is expected here:
```suggestion
   * Postfix Labels for the current number of items when total items are shown e.g., `of ${totaltems} items`
```

cf-remylenoir: I refactored the logic to pass only 1 string; otherwise, localization will not be possible if we need to switch the order of the sentence.

.changeset/chatty-tigers-visit.md
```diff
@@ -0,0 +1,5 @@
+---
+"@contentful/f36-pagination": patch
+---
+
+feat(Pagination): add support for custom total items labels
```
cf-remylenoir:
```suggestion
feat(Pagination): add support for custom total items label
```

packages/components/pagination/src/Pagination.test.tsx
```diff
@@ -91,6 +91,19 @@ describe('Pagination', () => {
     expect(screen.getByText('21 - 40 of 55 items')).toBeInTheDocument();
   });
 
+  it('renders custom range text with totalItems', () => {
```
cf-remylenoir:
```suggestion
  it('renders the custom total items label', () => {
```

PR #3235, opened 2025-11-17
feat(Pagination): add support for custom total items label

cf-remylenoir:
# Purpose of PR

We need to support the total items text localization, and as such, enabling its customization through the properties is necessary.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/.changeset/rare-bottles-clean.md
+++ b/.changeset/rare-bottles-clean.md
@@ -1,0 +1,5 @@
+---
+"@contentful/f36-pagination": patch
+---
+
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -134,5 +134,36 @@
   "volta": {
     "node": "18.20.1",
     "npm": "8.3.0"
+  },
+  "optionalDependencies": {
+    "@esbuild/aix-ppc64": "^0.25.5",
```
veu: We should add a rule for dependabot to keep these in sync.

PR #3250, opened 2025-11-25
feat: added chat history empty state component [PIC-783]

nikunj-tilva:
- Added ChatHistoryEmptyState component

### DIFF
--- a/packages/f36-ai-components/src/AIChatHistoryEmptyState/AIChatHistoryEmptyState.tsx
+++ b/packages/f36-ai-components/src/AIChatHistoryEmptyState/AIChatHistoryEmptyState.tsx
@@ -1,9 +1,10 @@
 import React from 'react';
 import { Flex, Text } from '@contentful/f36-components';
 import { ClockCounterClockwiseIconIcon } from '@contentful/f36-icons';
+import { type CommonProps } from '@contentful/f36-core';
 import { getStyles } from './AIChatHistoryEmptyState.styles';
 
-interface AIChatHistoryEmptyStateProps {
+export interface AIChatHistoryEmptyStateProps extends CommonProps {
   /**
    * The current filter state (e.g., 'paused', 'processing', 'done')
    * Used in default title and description if not provided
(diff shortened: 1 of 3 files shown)


packages/f36-ai-components/src/AIChatHistoryEmptyState/AIChatHistoryEmptyState.tsx
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
+import { Flex, Text } from '@contentful/f36-components';
+import { ClockCounterClockwiseIconIcon } from '@contentful/f36-icons';
+import { getStyles } from './AIChatHistoryEmptyState.styles';
+
+interface AIChatHistoryEmptyStateProps {
```
elliotmassen: Let's extend `CommonProps` and that way we get `className`, `testId` and `style` props like the other components.

packages/f36-ai-components/src/AIChatHistoryEmptyState/AIChatHistoryEmptyState.tsx
```diff
@@ -0,0 +1,57 @@
...
+  /**
+   * The current filter state (e.g., 'paused', 'processing', 'done')
+   * Used in default title and description if not provided
+   */
+  state?: string;
```
elliotmassen, line 12: [nit] What happens if no `state`, `title` or `description` is provided? Could we do something to strengthen the prop interface to ensure we always get enough props to render the texts?

packages/f36-ai-components/src/AIChatHistoryEmptyState/AIChatHistoryEmptyState.tsx
nikunj-tilva: Great Suggestion! Just updated

packages/f36-ai-components/src/AIChatHistoryEmptyState/AIChatHistoryEmptyState.tsx
nikunj-tilva: > [nit] What happens if no `state`, `title` or `description` is provided? Could we do something to strengthen the prop interface to ensure we always get enough props to render the texts?

Screenshot 2025-11-25 at 12 56 30

This is default text we render in case none of the properties are provided.

PR #3254, opened 2025-11-26
feat(f36-ai-components): add AIChatMessageList component

elliotmassen:
# Purpose of PR

This PR adds a new `AIChatMessageList` component that can be used to wrap a list of `AIChatMessage`'s. The message list handles auto-scrolling out-of-the-box. It provides a prop which can be used to either disable or customise the scrolling behaviour. The ref is forwarded as usual, which means that the consuming code could feasibly implement its own custom scrolling behaviour if required. 

This PR also introduces a `AIChatConversation` component as a container for both message list and user input with the required styling and layout. It also tweaks some message and user input styling to get things pixel perfect ✨ 

These new components mean that consuming code can get our designed AI Chat out-of-the-box without requiring additional code for styling and layout.

https://github.com/user-attachments/assets/d17d2725-8434-4bfb-8fdd-12378462584b

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/f36-ai-components/package.json
+++ b/packages/f36-ai-components/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@contentful/f36-ai-components",
   "private": false,
-  "version": "0.0.1-alpha.24",
+  "version": "0.0.1-alpha.25",
   "description": "A collection of components for use in AI applications, such as conversational interfaces.",
   "license": "MIT",
   "files": [
--- a/packages/f36-ai-components/src/AIChatHistory/AIChatHistory.tsx
+++ b/packages/f36-ai-components/src/AIChatHistory/AIChatHistory.tsx
@@ -4,6 +4,7 @@ import React, { forwardRef, Ref, useEffect, useRef, useState } from 'react';
 import { getStyles } from './AIChatHistory.styles';
 import { AIChatHistoryTabs } from './AIChatHistoryTabs';
 import { AIChatHistoryThread } from './AIChatHistoryThread';
+import { AIChatHistoryEmptyState } from '../AIChatHistoryEmptyState';
 
 export interface MessageThread {
   id: string;
@@ -119,6 +120,10 @@ function _AIChatHistory(props: AIChatHistoryProps, ref: Ref<HTMLDivElement>) {
 
   const renderGroupedThreads = () => {
     if (!hasGroups) {
+      if (threads.length === 0) {
+        return <AIChatHistoryEmptyState />;
+      }
+
       return (
         <NavList
           ref={scrollContainerRef}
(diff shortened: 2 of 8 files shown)


packages/f36-ai-components/src/AIChatMessageList/AIChatMessageList.tsx
```diff
@@ -0,0 +1,114 @@
...
+  /**
+   * Defines the scroll behavior of the message list.
+   * If set to 'none', no automatic scrolling will occur.
+   * If an object is provided, the `onScrollHeightChange` callback will be invoked
+   * whenever the scroll height changes. The callback receives a boolean indicating
+   * whether the user is near the bottom of the message list, and should return
+   * a boolean indicating whether to scroll to the bottom.
+   * @default { onScrollHeightChange: () => true }
+   */
+  scrollBehavior?:
```
nikunj-tilva, line 27: Nice!

nikunj-tilva (approved):

packages/f36-ai-components/src/AIChatMessageList/AIChatMessageList.tsx
```diff
@@ -0,0 +1,114 @@
...
+    const isNearTheBottom =
+      innerRef.current.scrollHeight -
+        innerRef.current.scrollTop -
+        innerRef.current.clientHeight <=
+      100;
+
+    if (scrollBehavior.onScrollHeightChange(isNearTheBottom)) {
+      scrollMessageListToBottom();
+    }
+  }, [scrollBehavior, innerRef.current?.scrollHeight]);
```
Rowadz, line 90: Just wondering, would this prevent the user from scrolling while a message is still being rendered?

For example, if `isNearTheBottom` is `true` while a message is being streamed and the user tries to scroll up, what happens?

How does the scroll behavior work when messages appear chunk by chunk?
I think it might be worth adding a Storybook example to demonstrate this. What do you think?

packages/f36-ai-components/stories/AIChatMessageList.stories.tsx
```diff
@@ -0,0 +1,101 @@
...
+    height: number;
+  }
+> = ({ numberOfChildren, width, height, ...componentProps }) => {
+  const containerRef = useRef<AIChatMessageListRef>(null);
+
+  return (
+    <>
+      <Button onClick={() => containerRef.current?.scrollToBottom()}>
+        Scroll to bottom
+      </Button>
```
Rowadz, line 80: I think the Storybook example would be better if it had a real input and simulated a chunked/streamed response when the user hits send, wdyt?

That way we can test all kinds of cases.

packages/f36-ai-components/src/AIChatMessageList/AIChatMessageList.tsx
Rowadz: I see, doing this fixes it, and gives us the control, coool

```ts
    <AIChatMessageList
        {...componentProps}
        scrollBehavior={{
          onScrollHeightChange: (isNearTheBottom) => isNearTheBottom,
        }}
      >
```

maybe it is worth having `isAtTheBottom`

PR #3261, opened 2025-12-05
chore: fix regression in Card component [UFO-2071]

veu:
# Purpose of PR

Fix regression introduced in v6 alphas that caused the _Card_ component to throw out its children and discard their state.

## PR Checklist

- [ ] Tests are passing


packages/components/autocomplete/src/Autocomplete.tsx
```diff
@@ -360,7 +360,8 @@ function AutocompleteBase<ItemType>(
     ...inputProps
   } = getInputProps({ ref: inputRef });
   const toggleProps = getToggleButtonProps();
-  const menuProps = getMenuProps();
+  const menuRef = useRef(null);
```
veu, line 363: This is unrelated to the other changes. Downshift complains when not passing a ref here despite everything working so let’s pass a ref to make downshift happy.

packages/components/autocomplete/src/Autocomplete.test.tsx
```diff
@@ -72,7 +72,7 @@ describe('Autocomplete', () => {
       expect(list.childElementCount).toBe(12);
 
       // Type one letter in the input to open the list
-      await user.type(screen.getByRole('textbox'), 'a');
```
veu, line 75: The role changed (correctly) due the downshift update in https://github.com/contentful/forma-36/pull/3259 but the failing tests weren’t visible in that PR because of unrelated errors that we hadn’t fixed yet.

packages/components/card/src/Card/Card.test.tsx
```diff
@@ -31,6 +32,33 @@ describe('Card', () => {
...
+            select fruit
+          </button>
+        </Card>
+      );
+    };
+
+    const user = userEvent.setup();
+    render(<StatefulCard />);
+
+    await user.click(screen.getByText('select fruit'));
```
Lelith, line 58: This is just a minor point, but it is not significant for this test. I usually prefer selecting buttons by the role 'button' rather than by the text, because text can change quickly and might cause the test to fail due to frequent copy changes instead of actual functionality changes.

veu: But in this case the text is directly baked into the test. Why would the test setup change with copy?

Lelith: @veu yes, that's what I meant with "not significant in your case." 
It's just my "living by" principle :)

PR #3262, opened 2025-12-08
fix: reenable Copybutton variant [UFO-2070]

veu:
# Purpose of PR

Reenable passing the `variant` to the `CopyButton` component. This functionality got lost in an earlier refactoring.

## PR Checklist

- [x] Tests are passing

### DIFF
--- a/packages/components/copybutton/src/CopyButton.tsx
+++ b/packages/components/copybutton/src/CopyButton.tsx
@@ -4,7 +4,6 @@ import React, {
   type MouseEventHandler,
   type FocusEventHandler,
 } from 'react';
-import tokens from '@contentful/f36-tokens';
 import { CopySimpleIcon } from '@contentful/f36-icons';
 import { type ExpandProps } from '@contentful/f36-core';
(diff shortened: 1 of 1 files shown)


packages/components/copybutton/src/CopyButton.tsx
cf-remylenoir, line 1: We need to fix the color mapping for the various variants.

https://github.com/user-attachments/assets/cd5996a5-67f2-4b97-87eb-4d2385e3de93

veu: It’s actually not the _CopyButton_ that’s broken but the underlying _IconButton_. I’ll have to investigate.

veu: Correction: the _CopyButton_ is actually overwriting the icon color which is why it looks like the variant is ignored in _IconButton_. Fix coming up.

veu: _CopyButton_ now uses the same variants as _IconButton_.

PR #3351, opened 2026-02-26
docs: add deprecation process description

Lelith:
## Purpose

Forma 36 lacked a documented deprecation process, leaving it unclear how and when to phase out components, props, and features. This PR establishes a consistent, transparent process that gives consumers adequate migration time while keeping the design system moving forward.

By defining explicit timelines, consumers gain the predictability they need to plan migrations on their own schedule, without being caught off guard by sudden breaking changes.

### New: `DEPRECATION.md`
A comprehensive guide covering:
- **When to deprecate** — deprecation is reserved for breaking changes only; non-breaking additions don't require it
- **Two-phase timeline** — deprecate in a minor release (minimum one major version cycle or 6 months), remove in a major release
- **Implementation patterns** — TypeScript `@deprecated` JSDoc tags, dev-only `console.warn` examples for both components and props
- **Changeset conventions** — deprecations are `minor` bumps, removals are `major`
- **Migration guide requirements** — before/after examples, prop mapping tables, codemod instructions, and support contacts

### New: `DEPRECATIONLOG.md`
A structured log to track the lifecycle of all deprecated features — from the deprecation date and version through to planned or completed removal. Gives consumers and contributors a single place to check what's currently deprecated and when it is scheduled for removal, making it easier to prioritize migrations.

### Updated: `README.md`
Adds a "Deprecation process" section with a short summary and a link to `DEPRECATION.md`.

### Updated: `RELEASES.md`
Adds a note linking to the deprecation guide for contributors creating changesets.

### DIFF
--- a/DEPRECATION.md
+++ b/DEPRECATION.md
@@ -141,7 +141,7 @@ if (process.env.NODE_ENV === 'development' && oldProp !== undefined) {
 
 - Announce deprecation in release notes
 - Consider posting in community channels
-- For major deprecations, consider a blog post
+- Consider a blog post for deprecations that affect more than one component
 
 ## Implementation Guidelines
 
--- a/DEPRECATION.md
+++ b/DEPRECATION.md
@@ -108,7 +108,7 @@ Include deprecation notices in the package CHANGELOG:
 - **ComponentName**: Deprecated in favor of NewComponent. Will be removed in next major version
 ```
 
-### Deprecationlog
+### Deprecation Log
 
 Document the deprecated feature or component in the Deprecation log
 
--- a/DEPRECATION.md
+++ b/DEPRECATION.md
@@ -54,6 +54,7 @@ We follow a **gradual deprecation approach** aligned with [semantic versioning](
 - Remove the deprecated feature in the next major version
 - Update documentation and examples
 - Update migration guides
+- Update deprecation log
 
 ### Example Timeline
 
@@ -93,7 +94,7 @@ Update the following locations:
 Add a prominent deprecation notice at the top of the component page:
 
 ```markdown
-> **⚠️ Deprecated**: This component is deprecated and will be removed in v5.0.0.
+> **⚠️ Deprecated**: This component is deprecated and will be removed in next major version
 > Please use [NewComponent](link) instead. See the [migration guide](#migration).
 ```
 
@@ -104,7 +105,7 @@ Include deprecation notices in the package CHANGELOG:
 ```markdown
 ### Deprecated
 
-- **ComponentName**: Deprecated in favor of NewComponent. Will be removed in v5.0.0.
+- **ComponentName**: Deprecated in favor of NewComponent. Will be removed in next major version
 ```
 
 ### Deprecationlog
@@ -118,7 +119,7 @@ Add console warnings in development mode:
 ```typescript
 if (process.env.NODE_ENV === 'development') {
   console.warn(
-    'Warning: `OldComponent` is deprecated and will be removed in v5.0.0. ' +
+    'Warning: `OldComponent` is deprecated and will be removed in next major version ' +
       'Please use `NewComponent` instead. ' +
       'See migration guide: https://f36.contentful.com/components/old-component#migration',
   );
@@ -130,7 +131,7 @@ For deprecated props on components still in use:
 ```typescript
 if (process.env.NODE_ENV === 'development' && oldProp !== undefined) {
   console.warn(
-    '`oldProp` on ComponentName is deprecated and will be removed in v5.0.0. ' +
+    '`oldProp` on ComponentName is deprecated and will be removed in next major version ' +
       'Please use `newProp` instead.',
   );
 }
@@ -198,7 +199,7 @@ Structure the changeset with clear messaging:
 
 ### Changeset Guidelines
 
-- **Version bump**: Deprecations are `minor` or `feature` changes (new warnings, not breaking)
+- **Version bump**: Deprecations are `minor` changes (new warnings, not breaking)
 - **Label clearly**: Start with `**Deprecated**` in the description
 - **Provide context**: Explain what's deprecated and what to use instead
 - **Link to docs**: Always include a link to migration guidance
@@ -217,7 +218,7 @@ Every deprecation must include a migration guide. The guide should contain:
 ### 2. Before/After Examples
 
 ```tsx
-// ❌ Deprecated - will be removed in v5.0.0
+// ❌ Deprecated - will be removed in next major version
 <OldButton type="primary">Click me</OldButton>
 
 // ✅ Recommended
@@ -290,7 +291,12 @@ export const Button: React.FC<ButtonProps> = ({
   if (process.env.NODE_ENV === 'development' && buttonType) {
     console.warn(
       'Button: `buttonType` prop is deprecated. Use `variant` instead. ' +
-      'This prop will be removed in v5.0.0.'
+      'This prop will be removed in the next major version'
+    );
+  } else if ( process.env.NODE_ENV === 'development' && buttonType && variant) {
+     console.warn(
+      'Button: `buttonType` prop is deprecated in favor of `variant` and will be ignored.' +
+      'This prop will be removed in the next major version'
     );
   }
 
@@ -313,7 +319,7 @@ import { Card } from './Card';
 
 /**
  * @deprecated Use `Card` from '@contentful/f36-card' instead.
- * Will be removed in v5.0.0.
+ * Will be removed in next Major Version
  *
  * @see {@link Card} for the replacement component
  * @example
@@ -328,7 +334,7 @@ import { Card } from './Card';
 export const LegacyCard: React.FC<LegacyCardProps> = (props) => {
   if (process.env.NODE_ENV === 'development') {
     console.warn(
-      'LegacyCard is deprecated and will be removed in v5.0.0. ' +
+      'LegacyCard is deprecated and will be removed in next major version ' +
       'Please migrate to Card. ' +
       'See: https://f36.contentful.com/components/card#migration'
     );
--- a/DEPRECATION.md
+++ b/DEPRECATION.md
@@ -1,0 +1,411 @@
+# Deprecation Process
+
+This document outlines the process for deprecating components, props, and features in Forma 36. Following a consistent deprecation process ensures that consumers of the design system have adequate time to migrate away from deprecated APIs while maintaining stability and clear communication.
+
+## Table of Contents
+
+- [When to Deprecate](#when-to-deprecate)
+- [Deprecation Timeline](#deprecation-timeline)
+- [Communication Process](#communication-process)
+- [Implementation Guidelines](#implementation-guidelines)
+- [Creating Changesets for Deprecations](#creating-changesets-for-deprecations)
+- [Migration Guide Requirements](#migration-guide-requirements)
+- [Examples](#examples)
+
+## When to Deprecate
+
+**Deprecation is specifically for breaking changes.** The deprecation process allows us to introduce breaking changes in a controlled manner by giving consumers advance notice and time to migrate. If a change can be made without breaking existing functionality (e.g., adding new props, extending accepted values, adding new components), it does not require deprecation.
+
+Consider deprecating a component, prop, or feature when:
+
+- **Replacement available**: A better alternative exists that provides the same or improved functionality
+- **Design system evolution**: The component no longer aligns with current design principles
+- **Performance issues**: A more performant solution is available
+- **Accessibility improvements**: A more accessible alternative exists
+- **API simplification**: Consolidating similar functionality to reduce confusion
+- **Technical debt**: Removing outdated patterns or dependencies
+
+**Do NOT deprecate when:**
+
+- No clear migration path exists
+- The feature is still widely used without a suitable alternative
+- The change would cause excessive disruption without significant benefit
+- The change can be performed non-breaking e.g. adding additional properties or extending accepted values
+
+## Deprecation Timeline
+
+We follow a **graduated deprecation approach** aligned with [semantic versioning](https://semver.org/):
+
+### Phase 1: Deprecation (Minor Release)
+
+- Mark the feature as deprecated
+- Add runtime console warnings (development only)
+- Update documentation with deprecation notices
+- Provide migration guide
+- Feature remains fully functional
+
+**Minimum duration**: At least **one major version cycle** or **6 months**, whichever is longer
+
+### Phase 2: Removal (Major Release)
+
+- Remove the deprecated feature in the next major version
+- Update documentation and examples
+- Update migration guides
+- Create codemod if applicable (for complex migrations)
+
+### Example Timeline
+
+```text
+v4.10.0 (Minor) - Component deprecated, warnings added
+v4.11.0 - v4.x  - Deprecated but functional
+v5.0.0  (Major) - Component removed
+```
+
+## Communication Process
+
+Effective communication is crucial for successful deprecations. Follow these steps:
+
+### 1. Internal Discussion
+
+- Discuss deprecation with the team
+- Document reasons and alternatives
+- Plan migration strategy
+- Consider impact on known consumers
+
+### 2. Documentation Updates
+
+Update the following locations:
+
+#### Component Documentation
+
+Add a prominent deprecation notice at the top of the component page:
+
+```markdown
+> **⚠️ Deprecated**: This component is deprecated and will be removed in v5.0.0.
+> Please use [NewComponent](link) instead. See the [migration guide](#migration).
+```
+
+#### API Documentation
+
+Mark deprecated props with JSDoc comments:
+
+```typescript
+/**
+ * @deprecated Use `newProp` instead. This prop will be removed in v5.0.0.
+ */
+oldProp?: string;
+```
+
+#### Changelog
+
+Include deprecation notices in the package CHANGELOG:
+
+```markdown
+### Deprecated
+
+- **ComponentName**: Deprecated in favor of NewComponent. Will be removed in v5.0.0.
+```
+
+### 3. Runtime Warnings
+
+Add console warnings in development mode:
+
+```typescript
+if (process.env.NODE_ENV === 'development') {
+  console.warn(
+    'Warning: `OldComponent` is deprecated and will be removed in v5.0.0. ' +
+      'Please use `NewComponent` instead. ' +
+      'See migration guide: https://f36.contentful.com/components/old-component#migration',
+  );
+}
+```
+
+For deprecated props on components still in use:
+
+```typescript
+if (process.env.NODE_ENV === 'development' && oldProp !== undefined) {
+  console.warn(
+    '`oldProp` on ComponentName is deprecated and will be removed in v5.0.0. ' +
+      'Please use `newProp` instead.',
+  );
+}
+```
+
+### 4. Community Communication
+
+- Announce deprecation in release notes
+- Consider posting in community channels
+- For major deprecations, consider a blog post
+
+## Implementation Guidelines
+
+### TypeScript Support
+
+Use TypeScript's `@deprecated` JSDoc tag for IDE support:
+
+```typescript
+/**
+ * @deprecated Use NewComponent instead. Will be removed in v5.0.0.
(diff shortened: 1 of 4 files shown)


DEPRECATION.md
```diff
@@ -0,0 +1,411 @@
...
+**Do NOT deprecate when:**
+
+- No clear migration path exists
+- The feature is still widely used without a suitable alternative
+- The change would cause excessive disruption without significant benefit
+- The change can be performed non-breaking e.g. adding additional properties or extending accepted values
+
+## Deprecation Timeline
+
+We follow a **graduated deprecation approach** aligned with [semantic versioning](https://semver.org/):
```
veu: What does "graduated" mean here? It’s not mentioned again below and I can’t quite make sense of it.

DEPRECATION.md
```diff
@@ -0,0 +1,411 @@
...
+
+### Phase 1: Deprecation (Minor Release)
+
+- Mark the feature as deprecated
+- Add runtime console warnings (development only)
+- Update documentation with deprecation notices
+- Provide migration guide
+- Feature remains fully functional
+
+**Minimum duration**: At least **one major version cycle** or **6 months**, whichever is longer
```
veu, line 50: Should we introduce a separate required waiting period for deprecation? My fear is that we won’t remember ourselves and break this inadvertently.

DEPRECATION.md
```diff
@@ -0,0 +1,411 @@
...
+
+### Phase 2: Removal (Major Release)
+
+- Remove the deprecated feature in the next major version
```
veu, line 54: That won’t work if we have a minimum duration of 6 months. We might have to wait a couple major versions if we start releasing faster. Or we should think about how to introduce breaking changes differently, ie. as a new component that can be used in parallel.

DEPRECATION.md
```diff
@@ -0,0 +1,411 @@
...
+- Update documentation and examples
+- Update migration guides
+- Create codemod if applicable (for complex migrations)
```
veu: I’d move code mods to phase 1. We have to do it anyway so we might as well do it when we start recommending to stop using the old feature.

DEPRECATION.md
veu: Just read that you’re planning to have the replacement ready in the same version as the deprecation. In that case ignore the second part of my previous comment.
Instead I’d suggest to keep a deprecation log somewhere in the project so we can check which deprecated features are old enough to be removed when we release a new major version, rather than having to search old changelogs.

DEPRECATION.md
```diff
@@ -0,0 +1,411 @@
...
+
+### Example Timeline
+
+```text
+v4.10.0 (Minor) - Component deprecated, warnings added
+v4.11.0 - v4.x  - Deprecated but functional
+v5.0.0  (Major) - Component removed
```
veu, line 64: Might be good to have an example where the next major release is less than 6 months after the deprecation so we have to wait for another major release.

DEPRECATION.md
```diff
@@ -0,0 +1,411 @@
...
+```
+
+## Communication Process
+
+Effective communication is crucial for successful deprecations. Follow these steps:
+
+### 1. Internal Discussion
+
+- Discuss deprecation with the team
+- Document reasons and alternatives
```
veu, line 84: If we add a deprecation log we could put the documenation there as well.

DEPRECATION.md
```diff
@@ -0,0 +1,411 @@
...
+### 2. Documentation Updates
+
+Update the following locations:
+
+#### Component Documentation
+
+Add a prominent deprecation notice at the top of the component page:
+
+```markdown
+> **⚠️ Deprecated**: This component is deprecated and will be removed in v5.0.0.
```
veu: We can’t predict when it will be removed if we have a minimum duration.

DEPRECATION.md
```diff
@@ -0,0 +1,411 @@
...
+> Please use [NewComponent](link) instead. See the [migration guide](#migration).
+```
+
+#### API Documentation
+
+Mark deprecated props with JSDoc comments:
+
+```typescript
+/**
+ * @deprecated Use `newProp` instead. This prop will be removed in v5.0.0.
```
veu: We mention this again below with more detail. Do we need both?

DEPRECATION.md
```diff
@@ -0,0 +1,411 @@
...
+  type?: 'primary' | 'secondary';
+
+  /** The variant of the component */
+  variant?: 'primary' | 'secondary';
+}
+```
+
+### Component Deprecation
+
+Create a wrapper that imports the new component:
```
veu: What is this section for? Are we just renaming the component here? We should be explicit about what approach to use when.

DEPRECATION.md
```diff
@@ -0,0 +1,411 @@
...
+  if (process.env.NODE_ENV === 'development') {
+    console.warn('OldComponent is deprecated...');
+  }
+  return <NewComponent {...convertProps(props)} />;
+};
+```
+
+### Export Strategy
+
+Keep deprecated exports available but marked:
```
veu: Why do we need this? Are there situations where marking the component as deprecated is not enough?

DEPRECATION.md
```diff
@@ -0,0 +1,411 @@
...
+
+### 6. Support Contact
+
+Provide a way to get help:
+
+```markdown
+If you encounter issues during migration, please:
+
+- Check our [GitHub Issues](https://github.com/contentful/forma-36/issues)
+- Send us direct feedback via [our form](https://forms.gle/j3oC9GuhxEJ1DF4k6)
```
veu: I wouldn’t do this. Putting contact information all over the place means it will be a pain to update.

DEPRECATION.md
```diff
@@ -0,0 +1,411 @@
...
+    );
+  }
+
+  // Map old props to new props
+  const newProps = {
+    padding: props.padding === 'large' ? '24px' : '12px',
+    ...props,
+  };
+
+  return <Card {...newProps} />;
```
veu, line 349: I don’t understand why the old component should wrap the new one. This sounds like an edge case that won’t be compatible with most component deprecations.

DEPRECATION.md
```diff
@@ -0,0 +1,411 @@
...
+};
+````
+
+### Example 3: Changeset for Deprecation
```
veu: This should be part of Example 2.

DEPRECATION.md
Lelith: Oh i think i meant gradually :)

DEPRECATION.md
Lelith: What do you mean with seperate? I thought as we have 6 month maintenance for older versions, 6 months deprecation cycle (minimum) would fit in.

Removal of deprecations will *always* be a major version update. But if it happens that the major version update is sooner than 6 months, it has to wait another major version. 

With our new plans on releasing major versions more regularly, I don't think that there will be too much of a confusion? But maybe the wording could be better.

DEPRECATION.md
Lelith: I think a deprecation log sounds very sensible. Will add that as well :)

DEPRECATION.md
Lelith: We need more long-term / mid-term planning for this for sure. But if we start to follow a more regular release schedule, it should be more predictable.

We could also, in the beginning, just give an estimate, and update that estimate according to our planning.

DEPRECATION.md
Lelith: That's what I did for the `Menu.SubmenuTrigger` because the component itself got removed, but has no change in the API, which is why i decided to keep the breaking changes minimal and still offer that component import, but with the new Element underneath. 

Might be quite the edge-case, so maybe we do not need to mention it.

DEPRECATION.md
Lelith: Yeah, it's the same as Prop deprecation, will remove

DEPRECATION.md
Lelith: Hm, I am not sure if there might be a difference depending on where you import it. I will have to try it out in an experiment :)

DEPRECATION.md
Lelith: It's the form we have on the f36 website. Maybe I can just write "Use the feedback form on the f36 website? instead of linking the form directly?

DEPRECATION.md
Lelith: you are right, it was an edge case, will remove it.

DEPRECATION.md
```diff
@@ -0,0 +1,411 @@
...
+- Remove the deprecated feature in the next major version
+- Update documentation and examples
+- Update migration guides
+- Create codemod if applicable (for complex migrations)
+
+### Example Timeline
+
+```text
+v4.10.0 (Minor) - Component deprecated, warnings added
+v4.11.0 - v4.x  - Deprecated but functional
```
BraunreutherA, line 63: Will the Component get maintained during that period or does it stop with the depracation?

Lelith: I believe we should maintain. So it will receive all security-related updates as well as bug fixes.

DEPRECATION.md
Lelith: I tried it and deprecating the component where its defined is enough. so will remove that

DEPRECATION.md
Lelith: @BraunreutherA in the last forma36 office hours we agreed that deprecation means "maintenance" mode and it will receive important security updates for the time beeing. 

I added that earlier in the document

DEPRECATION.md
```diff
@@ -0,0 +1,365 @@
...
+- The change can be performed non-breaking e.g. adding additional properties or extending accepted values
+
+## Deprecation Timeline
+
+We follow a **gradual deprecation approach** aligned with [semantic versioning](https://semver.org/):
+
+### Phase 1: Deprecation (Minor Release)
+
+- Mark the feature as deprecated
+- Add an entry for this feature or component into the deprecation log
```
veu:
```suggestion
- Add an entry for this feature or component in the deprecation log
```

DEPRECATION.md
```diff
@@ -0,0 +1,365 @@
...
+- Feature remains fully functional
+- Feature is in maintenance mode and will receive security updates
+
+**Minimum duration**: At least **one major version cycle** or **6 months**, whichever is longer
+
+### Phase 2: Removal (Major Release)
+
+- Remove the deprecated feature in the next major version
+- Update documentation and examples
+- Update migration guides
```
veu, line 56: And update deprecation log so we don’t try to remove it twice. ;)

DEPRECATION.md
```diff
@@ -0,0 +1,365 @@
...
+```
+
+#### Changelog
+
+Include deprecation notices in the package CHANGELOG:
+
+```markdown
+### Deprecated
+
+- **ComponentName**: Deprecated in favor of NewComponent. Will be removed in v5.0.0.
```
veu: I’d still go with "Will be removed in a future major version" since we don’t know which one it’ll be.

DEPRECATION.md
```diff
@@ -0,0 +1,365 @@
...
+
+**Deprecated**
+
+- `OldButton`: Deprecated in favor of `Button` with `variant` prop. Will be removed in v5.0.0
+- Migration guide available at: https://f36.contentful.com/components/button#migration
+```
+
+### Changeset Guidelines
+
+- **Version bump**: Deprecations are `minor` or `feature` changes (new warnings, not breaking)
```
veu: In which cases would we use `minor` or `feature`?

DEPRECATION.md
```diff
@@ -0,0 +1,365 @@
...
+   */
+  buttonType?: 'primary' | 'secondary' | 'positive' | 'negative';
+}
+
+export const Button: React.FC<ButtonProps> = ({
+  variant,
+  buttonType,
+  ...props
+}) => {
+  if (process.env.NODE_ENV === 'development' && buttonType) {
```
veu, line 291: Tiny thing but if `variant` was introduced as the replacement for `buttonType` then I’d throw an error if they’re used together because that silently ignores `buttonType`.

README.md
```diff
@@ -45,6 +46,12 @@ Use `npm run-script commit`. This uses the [Commitzen](https://github.com/commit
 
 Read more about changeset [here](RELEASES.md)
 
+## Deprecation process
+
+We follow a structured deprecation process to ensure smooth transitions when components or APIs are being phased out. Deprecations are announced at least one major version in advance, with clear migration paths and runtime warnings.
```
veu: "APIs" sounds weird in this context so I’d use the wording from the deprecation document.
```suggestion
We follow a structured deprecation process to ensure smooth transitions when components, props, and features are being phased out. Deprecations are announced at least one major version in advance, with clear migration paths and runtime warnings.
```

DEPRECATION.md
Lelith: ah just forgot to update it here.

DEPRECATION.md
Lelith: What I meant is: adding the new prop and deprecating the old one is not yet the breaking change, but the "minor" version update. 

or do you mean i should add more speaking examples?

veu: I was thinking that if you give the option to use one or the other, then it would be good to explain which one to use when.
But thinking about it more the deprecation itself should always be a `minor` change. It could be part of a bigger change that would be a `feature` but that’s how Semver works already. So I’d just say `minor` here.

Lelith: ah, i used that wording interchangeable. But you are right `minor` is the correct term in this case. will update

DEPRECATION.md
Lelith: @veu i added another if case with a different message if someone added buttonType and variant

veu (approved):

cf-remylenoir (approved):

DEPRECATION.md
```diff
@@ -0,0 +1,371 @@
...
+- Mark the feature as deprecated
+- Add an entry for this feature or component in the deprecation log
+- Add runtime console warnings (development only)
+- Update documentation with deprecation notices
+- Provide migration guide
+- Create codemod if applicable
+- Feature remains fully functional
+- Feature is in maintenance mode and will receive security updates
+
+**Minimum duration**: At least **one major version cycle** or **6 months**, whichever is longer
```
denkristoffer, line 50: What does this mean, exactly? Is this the duration of the deprecation? If so, why make it time based when we have semver?

DEPRECATION.md
```diff
@@ -0,0 +1,371 @@
...
+
+### Phase 2: Removal (Major Release)
+
+- Remove the deprecated feature in the next major version
+- Update documentation and examples
+- Update migration guides
+- Update deprecation log
```
denkristoffer, line 57: This was also mentioned in phase 1. Maybe we can be a bit more concrete and mention in just a few words what's supposed to go in the log at each stage.

DEPRECATION.md
```diff
@@ -0,0 +1,371 @@
...
+v6.0.0  (Major) - Component removed
+```
+
+## Communication Process
+
+Effective communication is crucial for successful deprecations. Follow these steps:
+
+### 1. Internal Discussion
+
+- Discuss deprecation with the team
```
denkristoffer: Which team?

DEPRECATION.md
```diff
@@ -0,0 +1,371 @@
...
+- Document reasons and alternatives
+- Plan migration strategy
+- Consider impact on known consumers
```
denkristoffer, line 86: I'd think of this as the first thing to do 😅

DEPRECATION.md
```diff
@@ -0,0 +1,371 @@
...
+
+Include deprecation notices in the package CHANGELOG:
+
+```markdown
+### Deprecated
+
+- **ComponentName**: Deprecated in favor of NewComponent. Will be removed in next major version
+```
+
+### Deprecationlog
```
denkristoffer:
```suggestion
### Deprecation Log
```

DEPRECATION.md
```diff
@@ -0,0 +1,371 @@
...
+
+Document the deprecated feature or component in the Deprecation log
+
+### 3. Runtime Warnings
+
+Add console warnings in development mode:
+
+```typescript
+if (process.env.NODE_ENV === 'development') {
```
denkristoffer, line 120: This won't run in the browser. Maybe let's check how other libraries handle it.

DEPRECATION.md
```diff
@@ -0,0 +1,371 @@
...
+      'Please use `newProp` instead.',
+  );
+}
+```
+
+### 4. Community Communication
+
+- Announce deprecation in release notes
+- Consider posting in community channels
+- For major deprecations, consider a blog post
```
denkristoffer: What's a "major deprecation"?

DEPRECATIONLOG.md
```diff
@@ -0,0 +1,31 @@
...
+- **Deprecated In**: v5.x.x
+- **Removal Planned**: next Major
+- **Reason**: [Brief explanation - e.g., "Replaced by improved component with better accessibility"]
+- **Migration**: Use `NewComponent` instead
+- **Guide**: [Migration documentation link]
+-->
+
+---
+
+## Removed Deprecations
```
denkristoffer, line 22: What's the value of tracking these after they're gone?

denkristoffer (commented):
Looks good overall, added some comments and questions. Big one is this migration guide that gets mentioned: where does it live? How do we structure it?

DEPRECATIONLOG.md
Lelith: For people to know exactly which version something was removed in, esp. if they did not update across multiple versions.

DEPRECATION.md
Lelith: What do you mean?

If we add this to the component body and you run Storybook or user_interface locally, it will appear in the console. So it will give others more opportunities to notify when something is deprecated.

image

DEPRECATION.md
Lelith: The time constraint is that others, such as partners or other teams, have a defined grace period for updates. This makes us and our development cycle more predictable.

DEPRECATION.md
Lelith: I put a template as example in the deprecation log itself, i did not want to clutter this document too much. I was also thinking that as a next step we could provide a script that would execute these steps for you.

DEPRECATION.md
Lelith: maybe we could rephrase this as:
```suggestion
- Discuss deprecation with the forma36 owners
```

it should be also possible for outside contributors to propose a deprecation / breaking change.

DEPRECATION.md
Lelith: how about:

```suggestion
- Consider a blog post for deprecations that affect more than one component
```

DEPRECATION.md
denkristoffer: I agree we should make our deprecation policy predictable, but introducing both a version-based and time-based boundary adds complexity to release planning. We would have to track both the version cycle and the 6-month window.

Since we already support previous majors for 6 months, users have a clear migration window. Do you see specific advantages of the dual boundary, or scenarios where this approach is especially useful?

Lelith: It's helpful when we want to do major releases more often, e.g., every quarter. So a deprecation might have been introduced a couple of weeks prior to a major release, which is too short for the 6-month cycle. This will enable more people to adopt a new major version quickly because there was enough time for migration.

To keep track of these cycles, we introduce the deprecation log. There, you would enter when a deprecation was introduced.

denkristoffer: But the previous version will still be there, supported for another 6 months as well. I think this creates extraneous process for no clear benefit.

Lelith: The key benefit is that everyone can quickly update to new major versions without needing extra resources to upgrade deprecated features at the moment. This happens frequently in our org. By guaranteeing support for at least 6 months, independent of other developments, others will benefit and can plan their work more effectively.

It's about proper planning and predictability, since we don't have a long-term development plan or fixed release dates.

denkristoffer (approved):

DEPRECATION.md
denkristoffer: My concern was that the library can be used in non-compiled code directly in the browser, but I guess it's an edge-case that we won't really see.

PR #3202, opened 2025-10-20
feat(f36-ai-components): add f36-ai-components package

elliotmassen:
## Summary

- Introduces the `@contentful/f36-ai-components` package — a collection of components for building AI conversational interfaces (chat layouts, message rendering, input with mentions, reasoning indicators, side panels, and more)
- Merges FormaV6 changes from `main` (React 19, Emotion v11 migration) into the ai-components branch
- Migrates all 28 files from `emotion` v10 to `@emotion/css` v11
- Bumps all `@contentful/f36-*` dependencies to `^6.0.0` and updates React peer deps to `^18.3.0 || >=19.1.0`
- Fixes React 19 type compatibility (`blockquote` HTMLQuoteElement typing, `Editor` type import from `@tiptap/core`, icon `variant` prop removal, controlled `Menu` requiring `onClose`)
- Upgrades `react-markdown` v7 → v9, `rehype-raw` v6 → v7, `remark-gfm` v3 → v4 for React 19 JSX namespace compatibility
- Updates Storybook 9 imports (`@storybook/addon-actions` → `storybook/actions`, `@storybook/react` → `@storybook/react-vite`)
- Renames internal `_Component` functions to `ComponentBase` suffix to satisfy `react-hooks/rules-of-hooks` linter
- Removes alpha publishing workflow and integrates package into the standard changeset release flow (added to `fixed` version group)

## New Components

- `AIChatLayout` — full-page chat layout with side panel support
- `AIChatMessage` — markdown-rendered user/assistant messages with action buttons
- `AIChatMessageList` — auto-scrolling message list with streaming support
- `AIChatInput` — TipTap-based rich text input with `@mention` support
- `AIChatHistory` — tabbed thread history with search
- `AIChatReasoning` — expandable reasoning/thinking indicator with animation
- `AIChatSidePanel` — collapsible side panel
- `AIChatSuggestionPill` — clickable suggestion chips
- `AIChatConversation` — conversation container
- `AIChatArtifactMessage` — artifact display in messages

## Test plan

> Chromatic deployment url: https://5fd1dda724cc620021ace8c5-shjeybujej.chromatic.com/?path=/docs/components-aichat-aichatartifactmessage--documentation
- [x] Build passes (`packages/core` and `packages/f36-ai-components`)
- [x] CI tests and lint pass
- [x] Storybook visual verification (animations, markdown, input, layout, history)
- [x] No changes outside `f36-ai-components` vs main (except `package-lock.json`)
[...]

PR #3400, opened 2026-05-04
chore: ignore ai-components-package publishing while releasing [PIC-694]

nikunj-tilva:
### Purpose

After merging [ai-components PR 
](https://github.com/contentful/forma-36/pull/3202), release step tries to publish it but since its a alpha package we do not want to publish it to npm

### Changes
- Added `ai-components` package to ignore in change set config
- Marked `ai-components` as private

### DIFF
--- a/packages/f36-ai-components/package.json
+++ b/packages/f36-ai-components/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@contentful/f36-ai-components",
-  "private": false,
+  "private": true,
   "version": "0.0.1",
   "description": "A collection of components for use in AI applications, such as conversational interfaces.",
   "license": "MIT",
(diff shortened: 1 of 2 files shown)


packages/f36-ai-components/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@contentful/f36-ai-components",
-  "private": false,
```
denkristoffer, line 3: With this change the package will refuse to publish, is that intended?

nikunj-tilva: Good point! Just reverted. Since its already part of ignore list we are safe.

cf-remylenoir (approved):

denkristoffer (approved):

PR #3407, opened 2026-05-19
feat(pill): add PillNext component

matthet:
## Summary

- Renames the existing `Pill` component internally to `LegacyPill` with `@deprecated` JSDoc notices, while preserving the `Pill` export for backwards compatibility
- Introduces a new `PillNext` component exported from `@contentful/f36-pill` with four variants (secondary, primary, warning, negative), leading icon with tooltip for warning/negative, removable end-icon with disabled state, and label truncation with tooltip
- Matches the visual spec from the Refreshed Pill design doc (see GROOT-2759): 16px border-radius, 32px height, correct spacing tokens, variant-specific backgrounds/borders/icons

## Screenshots
Screenshot 2026-05-20 at 16 21 12

Screenshot 2026-05-26 at 12 30 44

Screenshot 2026-05-20 at 16 27 04

Screenshot 2026-05-20 at 16 27 18

closing button hover (circle) - note also that no circle appears on hover of disabled button close icon
Screenshot 2026-05-20 at 16 27 38

Screenshot 2026-05-26 at 12 31 37

## Test plan

- [x] All 18 new PillNext tests pass (variants, remove button, disabled state, accessibility/axe)
- [x] All 9 existing LegacyPill tests pass unchanged
- [x] Package builds successfully (CJS + ESM + DTS)
- [x] `Pill` export unchanged — no breaking changes for consumers
- [x] Manual Storybook verification of all variants and states
- [x] Visual review against Figma spec


packages/components/pill/src/PillNext/PillNext.styles.ts
```diff
@@ -0,0 +1,98 @@
...
+      marginRight: tokens.spacing2Xs,
+    }),
+    labelWrapper: css({
+      display: 'inline-flex',
+      minWidth: 0,
+      maxWidth: '100%',
+    }),
+    label: css({
+      color: tokens.gray900,
+      fontSize: '14px',
```
Lelith: If this is not already the default font-size and needs to be overwritten font-size the token should be used.

```suggestion
      fontSize: tokens.fontSizeM
```

packages/components/pill-next/src/PillNext.styles.ts
```diff
@@ -0,0 +1,98 @@
...
+};
+
+export function getPillNextStyles(variant: PillNextVariant) {
+  const { background, border } = variantStyles[variant];
+
+  return {
+    pill: css({
+      display: 'inline-flex',
+      alignItems: 'center',
+      borderRadius: '16px',
```
Lelith, line 38: is this part of the new border-radius tokens? If yes, we should make a note to update it as soon as we introduce the new tokens in the next major version

packages/components/pill/src/PillNext/PillNext.styles.ts
```diff
@@ -0,0 +1,98 @@
...
+      fontWeight: 500,
+      lineHeight: '20px',
+      letterSpacing: '-0.154px',
```
Lelith: This looks a bit like a magic number, we do have tokens for changed letter spacing, to make it wider. But not to shrink them.

packages/components/pill/src/PillNext/PillNext.styles.ts
```diff
@@ -0,0 +1,98 @@
...
+    }),
+    labelWrapper: css({
+      display: 'inline-flex',
+      minWidth: 0,
+      maxWidth: '100%',
+    }),
+    label: css({
+      color: tokens.gray900,
+      fontSize: '14px',
+      fontWeight: 500,
```
Lelith: this is the token fontWeightMedium, which should already be the default font-weight

packages/components/pill/src/PillNext/PillNext.styles.ts
```diff
@@ -0,0 +1,98 @@
...
+      flexShrink: 0,
+      marginRight: tokens.spacing2Xs,
+    }),
+    labelWrapper: css({
+      display: 'inline-flex',
+      minWidth: 0,
+      maxWidth: '100%',
+    }),
+    label: css({
+      color: tokens.gray900,
```
Lelith: should already be default font-color?

packages/components/pill/src/PillNext/PillNext.styles.ts
```diff
@@ -0,0 +1,98 @@
...
+      fontSize: '14px',
+      fontWeight: 500,
+      lineHeight: '20px',
```
Lelith: this is the default lineHeight, i think that can be removed

packages/components/pill/src/PillNext/PillNext.styles.ts
```diff
@@ -0,0 +1,98 @@
...
+      overflow: 'hidden',
+    }),
+    removeButton: css({
+      display: 'inline-flex',
+      alignItems: 'center',
+      justifyContent: 'center',
+      width: '24px',
+      height: '24px',
+      padding: '4px',
+      borderRadius: '16px',
```
Lelith: In the design, the button is round, so we should use a percentage radius  of 50% to make sure its actually round also when people zoom. 

```suggestion
      borderRadius: '50%',
```

packages/components/pill-next/src/PillNext.styles.ts
```diff
@@ -0,0 +1,98 @@
...
+      color: tokens.gray900,
+      fontSize: '14px',
+      fontWeight: 500,
+      lineHeight: '20px',
+      letterSpacing: '-0.154px',
+      whiteSpace: 'nowrap',
+      textOverflow: 'ellipsis',
+      overflow: 'hidden',
+    }),
+    removeButton: css({
```
Lelith, line 65: Most of this style is an exact copy of the IconButton style. We can likely remove the entire styleset by using the IconButton component instead

packages/components/pill/src/PillNext/PillNext.styles.ts
```diff
@@ -0,0 +1,98 @@
...
+        backgroundColor: tokens.gray300,
+        mixBlendMode: 'luminosity' as const,
+        cursor: 'not-allowed',
+        color: tokens.gray500,
+      },
+    }),
+    removeIcon: css({
+      width: '16px',
+      height: '16px',
+    }),
```
Lelith: this can be removed by using the size property with value "small" on the Icon

packages/components/pill-next/src/PillNext.tsx
```diff
@@ -0,0 +1,116 @@
...
+  HTMLDivElement,
+  ExpandProps<PillNextProps>
+>((props, ref) => {
+  const {
+    label,
+    variant = 'secondary',
+    onRemove,
+    removeButtonLabel = 'Remove',
+    isRemoveDisabled = false,
+    tooltipContent,
```
Lelith, line 61: When we allow a tooltip, maybe it is also worth allowing tooltip overwrites for placements. In other instances. This could be exposed in the type like this:
```
  tooltipProps?: Omit<
    TooltipProps,
    'content' | 'children' | 'withTriggerWrapper'
  >;
```

packages/components/pill/src/PillNext/PillNext.styles.ts
```diff
@@ -0,0 +1,98 @@
...
+      whiteSpace: 'nowrap',
+      textOverflow: 'ellipsis',
+      overflow: 'hidden',
+    }),
+    removeButton: css({
+      display: 'inline-flex',
+      alignItems: 'center',
+      justifyContent: 'center',
+      width: '24px',
+      height: '24px',
```
Lelith: I can see in the design file, that the button is based on the small variant of the icon button, but then has a overwrite for the size. So I think we can keep that sizie overwrite, together with the overwrite for the border radius, to shape the `IconButton` into the variant we need.

packages/components/pill/src/PillNext/PillNext.tsx
```diff
@@ -0,0 +1,116 @@
...
+    setTextIsTruncated(node.scrollWidth > node.offsetWidth);
+  }, []);
+
+  const LeadingIcon = leadingIcons[variant];
+  const iconColor = leadingIconColors[variant];
+
+  const leadingIconElement = LeadingIcon ? (
+    <span className={styles.leadingIcon} style={{ color: iconColor }}>
+      <LeadingIcon />
+    </span>
```
Lelith: icon color should not be set as inline style on the outer span, but can be set directly as a prop on the Icon.

```suggestion
    <span className={styles.leadingIcon}>
      <LeadingIcon color={iconColor} />
    </span>
```

packages/components/pill/src/PillNext/PillNext.tsx
```diff
@@ -0,0 +1,116 @@
...
+      {onRemove && (
+        <button
+          type="button"
+          className={styles.removeButton}
+          onClick={onRemove}
+          disabled={isRemoveDisabled}
+          aria-label={removeButtonLabel}
+        >
+          <XIcon className={styles.removeIcon} />
+        </button>
```
Lelith: This should be an IconButton with specific overwrites for the size and border-radius styles.

packages/components/pill-next/src/PillNext.test.tsx
```diff
@@ -0,0 +1,140 @@
...
+      const mockOnRemove = jest.fn();
+      render(
+        <PillNext label="test" onRemove={mockOnRemove} isRemoveDisabled />,
+      );
+      fireEvent.click(screen.getByRole('button', { name: 'Remove' }));
+      expect(mockOnRemove).not.toHaveBeenCalled();
+    });
+  });
+
+  describe('tooltip', () => {
```
Lelith, line 95: We should also add a test for the truncated tooltip

packages/components/pill/src/PillNext/PillNext.tsx
```diff
@@ -0,0 +1,116 @@
...
+import type { PillNextVariant } from './PillNext.types';
+import { getPillNextStyles } from './PillNext.styles';
+
+export type PillNextInternalProps = CommonProps & {
+  label: string;
+  variant?: PillNextVariant;
+  onRemove?: () => void;
+  /** @default "Remove" */
+  removeButtonLabel?: string;
+  isRemoveDisabled?: boolean;
```
msieroslawska: I wonder if we should tie the disabled state explicitly to the button or the whole Pill - in case of the latter, it would be more fitting to name this prop in a more generic way (`isDisabled`).

packages/components/pill/src/PillNext/PillNext.tsx
```diff
@@ -0,0 +1,116 @@
...
+  /** Only rendered for variants with a leading icon (warning/negative). */
+  tooltipContent?: string;
+  value?: string;
```
msieroslawska: [Naming] Just looking at the props it is not entirely clear what `value` is for, since there is also `label`.

msieroslawska: Is it actually used?

packages/components/pill/src/PillNext/PillNext.styles.ts
matthet: Removed — it's the default; no override needed

packages/components/pill-next/src/PillNext.styles.ts
matthet: Added TODO comment to update when new tokens ship

packages/components/pill-next/src/PillNext.styles.ts
matthet: Replaced with `<IconButton variant="transparent" size="small" />` with size/radius overrides

packages/components/pill/src/PillNext/PillNext.styles.ts
matthet: Used size="small" on XIcon and leading icons, removed manual width/height styles

packages/components/pill-next/src/PillNext.tsx
matthet: Added `tooltipProps?: Omit<..., 'content' | 'children' | 'withTriggerWrapper'>`

packages/components/pill-next/src/PillNext.test.tsx
matthet: Added test that mocks scrollWidth > offsetWidth

packages/components/pill/src/PillNext/PillNext.tsx
matthet: Renamed (applies to whole pill)

packages/components/pill/src/PillNext/PillNext.tsx
matthet: 🙏🏽 Removed

packages/components/pill-next/src/PillNext.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+  return (
+    <div
+      className={cx(styles.pill, className)}
+      data-test-id={testId}
+      ref={ref}
+      {...otherProps}
+    >
+      {leadingIconElement &&
+        (tooltipContent ? (
```
msieroslawska, line 92: Tooltip should be accessible via tab and currently it's skipped. Is it possible to include it?

msieroslawska:
Visually when checking the components in storybook I see two issues:

1. Icons when wrapped in tooltip do not align with text anymore

Without tooltip (aligned):
Screenshot 2026-05-20 at 15 40 15

With tooltip (not aligned):
Screenshot 2026-05-20 at 15 40 19

2. Closing button hover is not a circle
Screenshot 2026-05-20 at 15 40 33

msieroslawska:
> ### ⚠️ No Changeset found
> Latest commit: [819faa9](https://github.com/contentful/forma-36/commit/819faa979bfa26d4737157c33841a1226b49f48b)
> 
> Merging this PR will not cause a version bump for any packages. If these changes should not result in a new version, you're good to go. **If these changes should result in a version bump, you need to add a changeset.**
> 
> This PR includes no changesets
> When changesets are added to this PR, you'll see the packages that this PR includes changesets for and the associated semver types
> 
> [Click here to learn what changesets are, and how to add one](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md).
> 
> [Click here if you're a maintainer who wants to add a changeset to this PR](https://github.com/contentful/forma-36/new/feat/pill-next-alpha?filename=.changeset/mighty-hounds-switch.md&value=---%0A%22%40contentful%2Ff36-pill%22%3A%20patch%0A---%0A%0Afeat(pill)%3A%20add%20PillNext%20component%20and%20rename%20existing%20Pill%20to%20LegacyPill%0A)

This message makes me wonder about the right way of releasing alpha components - is a changeset needed for that?

msieroslawska:
More visual aspects (sorry for not keeping the comments together)

1. Margin of the closing button does not seem to match Figma
2. Text weight / font does not look the same either
3. Hover color is different

Figma
Screenshot 2026-05-20 at 15 48 19

Storybook
Screenshot 2026-05-20 at 15 48 33

matthet:
> More visual aspects (sorry for not keeping the comments together)
> 
> 1. Margin of the closing button does not seem to match Figma
> 2. Text weight / font does not look the same either
> 3. Hover color is different
> 
> Figma Screenshot 2026-05-20 at 15 48 19
> 
> Storybook Screenshot 2026-05-20 at 15 48 33

Thank you for all the keen eye comments, and sorry this is a bit noisy! Also @Lelith. I'm gona do this more manually, clearly claude taking shortcuts and not as accurate on frontend.

msieroslawska:
> not as accurate on frontend

Tbh this is also a learning for design specs creation. Now seeing the discrepancies between storybook and figma I can tell why your claude interprets specs a certain way, as there are gaps in the specifics (eg. hover color on the close button).

packages/components/pill-next/src/PillNext.tsx
matthet: Done. When tooltipContent is provided, the leading icon span now gets tabIndex={0} (making it focusable via keyboard), plus role="img" and aria-label for accessibility.

packages/components/pill/src/PillNext/PillNext.styles.ts
matthet: Good catch — that was a Figma artifact Claude picked up. Likely known to you but Figma exposes SF Pro's (Apple's system font) native letter-spacing at 14px, but on the web the browser already handles spacing between characters for whatever system font is in use. It amounted to less than 1px total across a typical label with no visible effect. Removed it and checked the storybook renders again.

Lelith: very interesting. I thought that claude might have leveraged most of the css that figma is providing when you look at a design in dev mode.

Lelith:
> > ### ⚠️ No Changeset found
> > Latest commit: [819faa9](https://github.com/contentful/forma-36/commit/819faa979bfa26d4737157c33841a1226b49f48b)
> > Merging this PR will not cause a version bump for any packages. If these changes should not result in a new version, you're good to go. **If these changes should result in a version bump, you need to add a changeset.**
> > This PR includes no changesets
> > When changesets are added to this PR, you'll see the packages that this PR includes changesets for and the associated semver types
> > [Click here to learn what changesets are, and how to add one](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md).
> > [Click here if you're a maintainer who wants to add a changeset to this PR](https://github.com/contentful/forma-36/new/feat/pill-next-alpha?filename=.changeset/mighty-hounds-switch.md&value=---%0A%22%40contentful%2Ff36-pill%22%3A%20patch%0A---%0A%0Afeat(pill)%3A%20add%20PillNext%20component%20and%20rename%20existing%20Pill%20to%20LegacyPill%0A)
> 
> This message makes me wonder about the right way of releasing alpha components - is a changeset needed for that?

@msieroslawska 
Alpha releases work differently, they do not need a changeset. They instead have to be published completely by hand. We have a section on that here: https://github.com/contentful/forma-36/blob/main/RELEASES.md

Important is: 
* the package needs to be set to private
* before you can run the `npm prerelease` command, you have to run `npm run build` because the prerelease command does only publish whats locally build

Lelith:
@matthet the very last thing we need before we can publish this new component is updating the Readme files, explaining the deprecation and introducing the new PillNext component

Lelith:
@all-contributors please add @matthet as contributor.

Lelith:
@all-contributors add @matthet for code

Lelith (approved):
Looks good to me and can be merged and alpha-released after final Design approval :)

nastialia:
**Design QA**

❌ **Very long labels**
- We had this tag not round, but rather with a radius of 16px, so it looks round, but we allow it to be very large without breaking cc: @msieroslawska 
fefhjejhfehufheuhfuehfuewdwhudhwuhduwhduehfuheufheufheuf
Added it [here](https://www.figma.com/design/8un8Zv0egrfLQosvT2ZDKf/Tag-flow---UX-Improvements?node-id=1437-3505&t=uFk4XIlrAQGWgbv0-1) 

❌  **Tootips are not aligned centrally** 
Restricted access X
Deleted tag X

❌ **Icon button hover** 
Hover color should be gray300 and with the **Luminosity** effect, so it adapts to all colors

nastialia:
**Design QA**
❌ The hover effect is there, but it's disappearing after moving and defaulting to just gray300
https://github.com/user-attachments/assets/92c3d1a7-e0c3-4d97-a54f-2b6416b6dc82

nastialia:
Design QA passed! 🎉

PR #3417, opened 2026-06-01
fix(pill-next): support multiline label wrapping

matthet:
## Summary
- Allow long labels in PillNext to wrap to multiple lines instead of extending indefinitely
- Pill respects container width via `max-width: 100%` and grows vertically (`minHeight` instead of fixed `height`)
- Remove icon stays vertically centered when label wraps

## Open question
Should there be a fixed max-width cap (e.g. 400px) so the pill wraps even in very wide containers? Waiting on design input from Anastasia.

## How to review
Check the **LongLabels** story in Storybook — it shows a wide container (pill on one line) vs a narrower container (pill wraps to two lines).

Screenshot 2026-06-01 at 16 48 17

## Test plan
- [x] Verify short labels still render as single-line 32px pills
- [x] Verify long labels wrap correctly in narrow containers
- [x] Verify remove button stays vertically centered
- [x] Check warning/negative variants with leading icon still align correctly

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -42622,7 +42622,7 @@
     },
     "packages/components/pill-next": {
       "name": "@contentful/f36-pill-next",
-      "version": "6.0.1-alpha.4",
+      "version": "6.0.1-alpha.5",
       "license": "MIT",
       "dependencies": {
         "@contentful/f36-button": "^6.8.0",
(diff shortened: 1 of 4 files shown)


Lelith (approved):
Looks good to me!

nastialia:
Looks good!
And yes, let's make it adjustable depending on the size of the container!

nastialia:
After reviewing another PR, we have noticed that when the pill contains longer text, the top and bottom spacing is smaller than needed.
See below: 

image

Screenshot 2026-06-02 at 10 03 30

PR #3420, opened 2026-06-02
fix(pill-next): allow pill to shrink in flex containers

matthet:
## Summary
- Fixes long-label pills overflowing their flex container instead of wrapping text
- Flex items default to `min-width: auto` which prevents shrinking below content size
- Adding `min-width: 0` allows the pill to respect container width and trigger `word-break: break-word` on the label

## Context
Discovered in `user_interface` where `EntityTags` renders pills inside a `flex-wrap: wrap` container. Pills with long labels spanned the full container width without wrapping because the flex item couldn't shrink.
Screenshot 2026-06-02 at 14 58 24

## Fix
Screenshot 2026-06-02 at 15 12 24

## Test plan
- [x] Pills with short labels render unchanged
- [x] Pills with long labels in constrained flex containers wrap text to multiple lines
- [x] Verify in PillNext and Tag LongLabels stories


Lelith (approved):
Thanks for adding more examples to showcase the usage. This looks good to me!

PR #3423, opened 2026-06-03
chore: make pill-next package private as it is not stable

Lelith:
# Purpose of PR

when creating this new alpha component, we forgot to set the package to private, so it currently causes a publihsing error, because it can not yet be publicly published.


matthet (approved):
Thank you!!

PR #3425, opened 2026-06-03
feat(pill-next)!: replace onRemove with generic action icon slot

matthet:
[GROOT-2877](https://contentful.atlassian.net/browse/GROOT-2877)

## Summary

- Removes `onRemove`, `removeButtonLabel`, `removeButtonClassName`, and `removeIconColor` from PillNext and Tag
- Replaces with generic `actionIcon`, `onAction`, `actionButtonLabel`, and `actionButtonClassName` props
- Consumers pass whatever icon they need: `<XIcon />` for remove, `<PlusIcon />` for add, `<DotsThreeIcon />` for menu
- The action button uses `IconButton` with the same 24px circular styling (gray/300 hover, luminosity blend mode) — no manual overrides needed
- `IconButton` → `Button` handles icon sizing internally, so no `cloneElement` is needed
- Breaking change to the alpha API — no stable consumers exist

## Motivation

`onRemove` was just a special case of "render an action button with an icon". Having both `onRemove` and `actionIcon` doubled the prop surface for no benefit. The generic slot covers all use cases (remove, add, menu trigger) without prescribing semantics.

This also unifies PillNext/Tag with AiPill's existing `actionIcon`/`onAction` pattern (GROOT-2867).

## Screenshots

**Pill Next**
Screenshot 2026-06-03 at 16 37 14

**Tag**
Screenshot 2026-06-03 at 16 37 08

**AI Pill**
Screenshot 2026-06-03 at 16 40 50

## Test plan

- [x] PillNext and Tag tests pass (38 tests)
- [x] AiPill tests pass (unchanged — already uses actionIcon)
- [x] Documentation examples updated
- [x] TypeScript check passes
- [x] Visual check in Storybook — X, Plus, DotsThree all render correctly across variants

[GROOT-2877]: https://contentful.atlassian.net/browse/GROOT-2877
[GROOT-2867]: https://contentful.atlassian.net/browse/GROOT-2867


packages/components/pill-next/src/PillNext.tsx
```diff
@@ -117,16 +130,28 @@ export const PillNext = React.forwardRef<
...
+          <IconButton
+            variant="transparent"
+            size="small"
+            icon={<XIcon size="small" color={removeIconColor} />}
+            aria-label={removeButtonLabel}
+            onClick={onRemove}
+            isDisabled={isDisabled}
+            className={cx(styles.endButton, removeButtonClassName)}
+          />
+        )
```
Lelith: Hm, I'm thinking if it makes conceptually sense to have them both in the code in parallel, because the `Remove` functionality is basically the exact same as any other Action. The only "convenience" it offers is that it already sets the Icon, so you do not have to hand it over as a prop.

But all other props are basically the same, just with a more specific name. I don't think the convenience of the Icon is a valid enough reason to double the number of properties this component accepts. Because you can generate the remove functionality with the action-based properties as well. 

So I would rather replace the remove props with only the generic action props.

packages/components/pill-next/src/PillNext.tsx
```diff
@@ -117,16 +130,28 @@ export const PillNext = React.forwardRef<
...
       )}
```
cf-remylenoir: If the ' remove' and 'action' properties are mutually exclusive, we might want to add a `variant` property (or similar) and control the availability of the underlying properties with discriminated union types.

```TypeScript
type Variant = 
  | {
      variant: 'remove';
      onRemove?: () => void;
      removeButtonLabel?: string;
      ... // other remove props
      onAction?: never;
      actionButtonLabel?: never;
    }
  | {
      variant: 'action';
      onAction?: () => void;
      actionButtonLabel?: string;
      ... // other action props
      onRemove?:never;
      removeButtonLabel?: never;
    }

export type PillNextInternalProps = CommonProps & Variant & { ... // common props to all variants };
```

This way, we have full confidence that all properties of a given variant are passed; the consumer doesn't have to guess.

It also allows scalability beyond action and removal if necessary in the future.

There are examples in other components, e.g., the Header component with the variant.

packages/components/pill-next/src/PillNext.tsx
matthet: Absolutely. Thanks for calling that out. I did this a bit quick and bleary eyed yesterday, actually this is definitely the way to go - remove is just one possible action. It unintentionally got more sway given it was the example in the designs but now on usage, I see it's not a default use case.

packages/components/pill-next/src/PillNext.tsx
matthet: Thanks for the suggestion. Remove doesn't actually need to be something special here. I'll make the action generic and no specifics for remove as discussed above.

cf-remylenoir: Agree, we came to the same conclusion after discussing it this morning with the group.

Lelith (approved):
thanks for refactoring, this looks good to me.

PR #3430, opened 2026-06-08
chore: cleanup releases

Lelith:
## Purpose

Remove `@contentful/f36-codemod` from the Changesets ignore list so its pending patch changeset can be released through the normal Forma 36 release flow.

The codemod package was previously ignored together with manually released alpha packages, but it should be handled by Changesets. Because of that ignore entry, the existing codemod changeset was never consumed and the fix was not published.

This also cleans up stale changesets for alpha packages that were already released manually and should not be picked up by the standard release automation.

## Approach

- Remove `@contentful/f36-codemod` from `.changeset/config.json` `ignore`
- Keep the existing codemod changeset so the next release job can version and publish it
- Remove stale changesets for manually released alpha packages
- Leave manually released alpha packages ignored

## Expected Release Impact

On the next `main` release, Changesets should publish:

- `@contentful/f36-codemod`: `6.0.0` -> `6.0.1`

The existing `.changeset/shy-eggs-add.md` should be consumed by the release job.

The removed alpha-package changesets should have no release impact because those packages were released manually and remain ignored by Changesets.

## Testing

- Verified the remaining changeset targets `@contentful/f36-codemod` as a patch release
- Verified the removed changesets only targeted manually released alpha packages
- No runtime code changes

### DIFF
--- a/scripts/changesets/changelog-generate.js
+++ b/scripts/changesets/changelog-generate.js
@@ -17,6 +17,8 @@ const ignorePkgs = [
   '@contentful/f36-i18n-utils',
   '@contentful/f36-website',
   '@contentful/f36-docs-utils',
+  '@contentful/f36-pill-next',
+  '@contentful/ai-components',
 ];
 
 // Format package names as Start Case
--- a/scripts/changesets/changelog-generate.js
+++ b/scripts/changesets/changelog-generate.js
@@ -18,7 +18,7 @@ const ignorePkgs = [
   '@contentful/f36-website',
   '@contentful/f36-docs-utils',
   '@contentful/f36-pill-next',
-  '@contentful/ai-components',
+  '@contentful/f36-ai-components',
 ];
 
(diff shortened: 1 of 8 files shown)


scripts/changesets/changelog-generate.js
```diff
@@ -17,6 +17,8 @@ const ignorePkgs = [
   '@contentful/f36-i18n-utils',
   '@contentful/f36-website',
   '@contentful/f36-docs-utils',
+  '@contentful/f36-pill-next',
+  '@contentful/ai-components',
```
nikunj-tilva: This should be @contentful/f36-ai-components correct?

Lelith: right, thanks for pointing out the typo! fill update

nikunj-tilva (approved):

cf-remylenoir (approved):

PR #3431, opened 2026-06-08
chore: bump Storybook to latest

Lelith:
## Summary

Upgrades Storybook to the latest version and updates the Storybook Vite configuration to improve development performance.

### Performance issue
Opening the first story in storybook triggered Vite to transform the full `@contentful/f36-icons` package on demand. Since the icons package barrel re-exports every icon, this caused a long delay before stories became interactive.

This PR adds Vite dependency pre-bundling for the icon dependencies used by Storybook so that the cost is handled during dependency optimization instead of blocking the first story load.

## Changes

- Upgrade Storybook packages to the latest version.
- Update `scripts/.storybook/main.ts` for Storybook 10 compatibility.
- Pre-bundle `@contentful/f36-icons` and `@phosphor-icons/react` via Storybook’s `viteFinal` config.
- Reduce first-story load time in local Storybook development.

## Testing

- Started Storybook locally.
- Verified Storybook boots successfully.
- Verified opening stories is significantly faster after pre-bundling icons.


package.json
```diff
@@ -133,6 +133,10 @@
     "npm": ">=8.5.0",
     "yarn": "Please use npm"
   },
+  "optionalDependencies": {
+    "@emnapi/core": "1.11.0",
+    "@emnapi/runtime": "1.11.0"
```
cf-remylenoir: What part of the CI requires it?

Lelith: That was quite the rabbit hole, but the CI said that the emnapi packages where "out of sync" from the lockfile (see this run https://app.circleci.com/pipelines/github/contentful/forma-36/16266/workflows/6485de89-3b8b-4980-b4ea-d26cc916cca0/jobs/50038 )

emnapi seems to be a peer dependency introduced in Storybook 10 (somewhere traced back to wasm usage in storybook 10). When running npm install they did not get added in the way circle-ci expected and it complained that they are not there.  So claude suggested to either run the npm install on a linux machine or to try adding it as optional dependencies.

cf-remylenoir (approved):

cf-remylenoir (commented):
I took the liberty to rename it to `chore: bump Storybook to latest` as there isn't any feature introduction or component update, but solely the Storybook bump and loading optimization

veu (approved):

package.json
Lelith: Rebecca and I did dig even deeper. 
The dependency from the lock file was actually removed when updating the `vite` version. It was a peer dependency in the past from some `oxc-resolver` package. 
This is no longer used, and therefore the dependency was deleted from the lock file. Another `wasm` related package still wants to have `emnapi` but is marked as `externeous`, which is why it does not add it's own entry in the lock file. 

By marking it optional, we say "don't worry too much about that dependency," and we are done. We have now tried to have that based on a loose versioning scheme, so as not to always have to carry it around in the future, but that unfortunately did not work. 

I keep this comment open for future reference if it becomes relevant again.

PR #3454, opened 2026-07-06
feat(button): add tiny size variant

GigiIanni:
Adds a `tiny` (24px) size to Button following the Figma spec, using existing spacing/typography tokens. Includes Code Connect mapping, story, docs example, and test coverage.

# Purpose of PR

## PR Checklist

- [x] I have read the relevant `readme.md` file(s)
- [x] All commits follow our [Git commit message convention](https://github.com/contentful/forma-36/tree/main/packages/forma-36-react-components#commits)
- [x] Tests are added/updated/not required
- [x] Tests are passing
- [x] Storybook stories are added/updated/not required
- [x] Usage notes are added/updated/not required
- [x] Has been tested based on [Contentful's browser support](https://www.contentful.com/faq/about-contentful/#which-browsers-does-contentful-support)
- [x] Doesn't contain any sensitive information

PR #3488, opened 2026-07-13
feat(button): add tiny size variant to `IconButton`

GigiIanni:
# Purpose of PR

Adds a new `tiny` (24px) size variant to the **IconButton** component (with the companion `tiny` size on **Button** it builds on), following the Figma spec.

## Why

Contentful's denser UI surfaces — inline table cells, compact toolbars, tag/pill rows, and sidebars — increasingly need actionable icon buttons that fit within tight vertical rhythm. The smallest existing IconButton size (`small`, 32px) is still too tall for these contexts and forces teams into one-off CSS overrides.

A supported `tiny` (24px) size gives designers and engineers a token-based option for these space-constrained, high-density layouts instead of custom workarounds. It also complements Forma 36's existing high-density mode, where a 24px control aligns naturally with the tighter spacing scale.

### IconButton
- `tiny` size: `spacing2Xs` (4px) padding + `spacingL` (24px) min box → **24×24px** final box (incl. the 1px border)
- Icon auto-maps to the `tiny` (14px) icon size via Button's size mapping
- Figma Code Connect: added `Tiny` to the Default & Disabled variants
- Storybook: added `tiny` to the size control and to the Overview gallery
- Docs example + README ("3 sizes" → "4 sizes")
- Test coverage for the tiny box dimensions + a tiny-size axe scenario

### Button (companion change this builds on)
- `tiny` (24px) size using existing spacing/typography tokens
- Icon auto-maps to `tiny` (14px)
- Added an axe test across all sizes as a warning signal if the tiny sizing is ever changed to something too small

## PR Checklist

- [x] Tests are passing


packages/components/button/src/Button/Button.test.tsx
Lelith, line 32: :accessibility: could you maybe also add a separate axe test for the sizes of buttons (there is already one for variants, you can just copy paste that one and do it with all sizes instead of variants). So we have a warning signal if we ever change the font size or any other sizing meaning of `tiny` into something that is going to be too small.

packages/components/button/src/IconButton/IconButton.test.tsx
```diff
@@ -49,6 +49,22 @@ describe('IconButton', () => {
     expect(handleClick).not.toHaveBeenCalled();
   });
 
+  it('renders the tiny size', () => {
```
Lelith, line 52: :accessibility: Also for this variant, it would be great if you could add an `axe()` test scenario for the tiny IconButton

packages/components/button/stories/IconButton.stories.tsx
```diff
@@ -19,7 +19,7 @@ export default {
       control: 'select',
       options: ['', ...Object.keys(icons)],
     },
-    size: { control: 'select', options: ['small', 'medium', 'large'] },
+    size: { control: 'select', options: ['tiny', 'small', 'medium'] },
```
Lelith: we lost the `large` one on this update 

```suggestion
    size: { control: 'select', options: ['tiny', 'small', 'medium', 'large'] },
```

Lelith (commented):
Just some smaller things, overall it looks solid.

packages/components/button/src/Button/Button.test.tsx
GigiIanni: Good idea — added a `has no a11y issues across sizes` test that renders tiny/small/medium/large and runs axe over them, so we get a warning signal if the tiny sizing is ever reduced too far. Done in a4d9e9365.

packages/components/button/src/IconButton/IconButton.test.tsx
GigiIanni: Added a dedicated `is accessible in the tiny size (axe)` scenario for IconButton. Done in a4d9e9365.

packages/components/button/stories/IconButton.stories.tsx
GigiIanni: Good catch — restored `large` in the size control (`['tiny', 'small', 'medium', 'large']`). Done in a4d9e9365.

Lelith (approved):
thank you for your contribution! this looks good to me. 👍

cf-remylenoir (changes requested):
I believe your branch wasn't on the latest main branch; it appears you are reintroducing changes merged in https://github.com/contentful/forma-36/pull/3454? 

I understand you intended to introduce the tiny size only to the `IconButton` component?

cf-remylenoir (approved):

PR #3491, opened 2026-07-14
fix(popover): keep document tab order

denkristoffer:
# Purpose of PR

Fix a bug where `Popover` rendered inside React portals resets the focus order once the `Popover` closes.

## Cause
The bug is caused by the new `FloatingPortal` from `floating-ui`. It defaults to `preserveTabOrder={true}`, which creates tab key handlers that try to be smart about restoring focus order after the popover closes. With the portal that causes issues, but it seems to work to just leave the order as-is by disabling the option.

### DIFF
--- a/.changeset/lazy-popover-focus.md
+++ b/.changeset/lazy-popover-focus.md
@@ -1,0 +1,5 @@
+---
+'@contentful/f36-popover': patch
+---
+
+Fix portalled Popover tab order when using non-modal focus management.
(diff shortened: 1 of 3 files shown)


denkristoffer:
Closing this draft: the original user_interface repro path was removed/migrated, and we could not reproduce the focus-loss behavior in the current app even with the old portal behavior forced back on. The guard cleanup may still be worth revisiting later, but this PR should not stay open as a fix for UFO-1644.

PR #3502, opened 2026-07-21
chore(website): configure Osano script from environment

denkristoffer:
## Summary

- build the Osano script URL from standard Next.js public environment variables
- omit the consent script when either identifier is missing
- document local setup and include the identifiers in the Turbo build cache key

## Vercel rollout

- replace `OSANO_KEY` with `NEXT_PUBLIC_OSANO_CUSTOMER_ID`
- replace `OSANO_F36_WEBSITE_KEY` with `NEXT_PUBLIC_OSANO_CONFIGURATION_ID`
- configure separate Preview and Production values as needed, then redeploy each environment

## Testing

- `npm run tsc`
- `npm run lint` (passes with 12 existing warnings)


cf-remylenoir (approved):

packages/website/pages/_app.tsx
cf-remylenoir, line 1: Are the new keys already in Vercel? I wanted to confirm that the script is injected correctly; so far, I don't see it in the preview deployment, so I assume the keys aren't in Vercel.

denkristoffer: They are not, I am asking IT to allow separate non-sensitive environment variables for different environments first. So we can have the right script load for staging and production.

PR #3522, opened 2026-07-23
chore: remove unused avatar exports

denkristoffer:
## Summary
- remove four private Avatar exports reported by Knip
- delete the runtime-only `SIZES` constant and preserve its public literal-union type

Stacked on #3521. Export cleanup remains incremental; a single global exports/types CI check should be added only after repository-wide findings are resolved or classified.

## Verification
- `npm exec knip -- --workspace packages/components/avatar --include exports,types --no-config-hints`
- `npm run tsc`
- `npm run lint`
- `npm test -- --runInBand packages/components/avatar`
- `npm run build --workspace=@contentful/f36-avatar`


cf-remylenoir (approved):

Lelith (changes requested):
this has to be considered a breaking change.

Lelith (approved):
I just had a chat with Kristoffer, and we came to an agreement, that, while there is a possibility someone found these non-public exported types and functions, it should not being considered a breaking change. 

Meaning, we can go ahead and remove those exports

PR #3568, opened 2026-07-28
feat: new component TableNext with stackable layout

Lelith:
# Purpose of PR

Introduces `TableNext` as a new alpha table package for Forma 36.

`TableNext` is the responsive successor to the current `Table` component. It defaults to a horizontally scrollable layout and adds a `stackable` responsive mode for narrower containers, where each table row is displayed as a grouped set of cells. The default breakpoint for this containers behaviour is `700px` and can be overwritten with the `stackableBreakpoint` prop. The component uses native table markup and uses `columnTitles`prop to power the labels used in stacked layouts.

**Table Next stackable layout** 

image

## Usage examples

**Scrollable table with sortin**g 
```
      <TableNext> 
        <TableNext.Head>
          <TableNext.Row>
            <TableNext.Cell
              isSortable
              onClick={() => handleSort('name')}
              sortDirection={
                sorting?.column === 'name' ? sorting.direction : undefined
              }
            >
              Name
            </TableNext.Cell>
            <TableNext.Cell>Description</TableNext.Cell>
            <TableNext.Cell
              isSortable
              onClick={() => handleSort('updatedAt')}
              sortDirection={
                sorting?.column === 'updatedAt' ? sorting.direction : undefined
              }
            >
              Updated
            </TableNext.Cell>
            <TableNext.Cell>Status</TableNext.Cell>
          </TableNext.Row>
        </TableNext.Head>
        <TableNext.Body>
          {sortedRows.map((item) => (
            <TableNext.Row key={item.id}>
              <TableNext.Cell>{item.name}</TableNext.Cell>
              <TableNext.Cell>{item.description}</TableNext.Cell>
              <TableNext.Cell>{item.updatedAt}</TableNext.Cell>
              <TableNext.Cell>
                <EntityStatusBadge
                  entityStatus={item.status as 'published' | 'draft'}
                />
              </TableNext.Cell>
            </TableNext.Row>
          ))}
        </TableNext.Body>
      </TableNext>
```

**Stackable Table with custom breakpoint**

```
 <TableNext
            layout="stackable"
            columnTitles={[
              'Name',
              'Status',
              'Email',
              'Content type',
              'Updated by',
              'Last activity',
            ]}
            stackableBreakpoint={400}
            {...args}
          >
            <TableNext.Head>
              <TableNext.Row>
                <TableNext.Cell>Name</TableNext.Cell>
                <TableNext.Cell>Status</TableNext.Cell>
                <TableNext.Cell>Email</TableNext.Cell>
                <TableNext.Cell>Content type</TableNext.Cell>
                <TableNext.Cell>Updated by</TableNext.Cell>
                <TableNext.Cell>Last activity</TableNext.Cell>
              </TableNext.Row>
            </TableNext.Head>
            <TableNext.Body>
              {withContentTableData.map((item) => (
                <TableNext.Row key={item.name}>
                  <TableNext.Cell>{item.name}</TableNext.Cell>
                  <TableNext.Cell>
                    <EntityStatusBadge entityStatus={item.status} />
                  </TableNext.Cell>
                  <TableNext.Cell>{item.email}</TableNext.Cell>
                  <TableNext.Cell>{item.contentType}</TableNext.Cell>
                  <TableNext.Cell>{item.updatedBy}</TableNext.Cell>
                  <TableNext.Cell>{item.updated}</TableNext.Cell>
                </TableNext.Row>
              ))}
            </TableNext.Body>
          </TableNext>
```

This PR adds:

- new `@contentful/f36-table-next` package with compound components: `TableNext`, `TableNext.Head`, `TableNext.Body`, `TableNext.Row`, and `TableNext.Cell`
- responsive first `scrollable` behaviour 
- Layout `stackable` with container based responsive behaviour, default `700px` can be overwritten with `stackableBreakpoint`
- `inline` and `embedded` visual variants
- sticky header and sticky first-column support for scrollable tables
- accessibility coverage with `jest-axe`
- Storybook stories for default usage, loading states, variants, long content, stackable layouts, sticky columns, and sticky headers
- website documentation, examples, sidebar entry, live editor scope
- migration documentation from `Table` to `TableNext`, including current alpha limitations

Current limitations are documented in the migration guide: Layout `stackable` does not yet support sortable headers, check-all header checkboxes, or custom React content inside table headers.


packages/components/tablenext/examples/TableScrollableStickyFirstColumnExample.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+    return {
+      ...entry,
+      id: `entry-${index + 1}`,
+      updated: `${index + 1} days ago`,
+    };
+  });
+
+  return (
+    <TableNext layout="scrollable" isFirstColumnSticky>
```
Lelith:
```suggestion
    <TableNext isFirstColumnSticky>
```
scrollable is default

packages/components/tablenext/examples/TableScrollableStickyHeaderAndFirstColumnExample.tsx
```diff
@@ -0,0 +1,101 @@
...
+  return (
+    <div
+      style={{
+        display: 'flex',
+        flexDirection: 'column',
+        height: '320px',
+        overflowY: 'auto',
+      }}
+    >
+      <TableNext layout="scrollable" isFirstColumnSticky>
```
Lelith:
```suggestion
      <TableNext isFirstColumnSticky>
```

packages/components/tablenext/src/TableCell/TableCell.styles.ts
```diff
@@ -0,0 +1,59 @@
+import { css } from '@emotion/css';
```
Lelith, line 1: basically exact copy of original table

packages/components/tablenext/src/TableCell/TableCell.tsx
```diff
@@ -0,0 +1,150 @@
+import { cx } from '@emotion/css';
```
Lelith, line 1: exact copy of original table

packages/components/tablenext/src/TableCell/TableCellContext.ts
Lelith, line 1: exact copy of original table

packages/components/tablenext/src/TableHead/TableHead.styles.ts
```diff
@@ -0,0 +1,30 @@
...
+  sticky: css({
+    th: {
+      position: 'sticky',
+      top: 0,
+      zIndex: tokens.zIndexDefault,
+    },
+  }),
+  stackableHidden: (breakpoint: string) =>
+    css({
+      [`@container (width <= ${breakpoint})`]: {
```
Lelith, line 17: style only applies until breakpoint, no other overwrites required.

packages/components/tablenext/src/TableRow/TableRow.styles.ts
```diff
@@ -0,0 +1,75 @@
...
+                gridColumn: '1 / 3',
+                gridTemplateColumns: 'subgrid',
+              }
+            : {
+                display: 'flex',
+                flexDirection: 'column',
+              }),
+          borderBottom: `1px solid ${tokens.gray200}`,
+        },
+        td: {
```
Lelith, line 35: overwrite of cell styles in TableRow, to not have to spread the variable into the individual table cells.

packages/components/tablenext/src/TableRow/TableRow.tsx
```diff
@@ -0,0 +1,100 @@
...
+      stackableBreakpoint = '700px',
+    } = useTableContext();
+
+    const originalCells = Children.toArray(children);
+    const updatedCells: React.ReactNode[] = [];
+    const columnTitleCount = columnTitles?.length ?? 0;
+    const cellCount = Math.max(columnTitleCount, originalCells.length);
+
+    if (
+      process.env.NODE_ENV !== 'production' &&
```
Lelith: ensuring this text is only visible during development.

packages/components/tablenext/src/TableNext.tsx
```diff
@@ -0,0 +1,161 @@
...
+        >
+          {children}
+        </TableContextProvider>
+      </Box>
+    );
+
+    return (
+      <section
+        // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex -- scrollable region requires tabIndex for keyboard access (WCAG 2.1 SC 2.1.1)
+        tabIndex={0}
```
denkristoffer: I asked before on the other PR but for good measure, shouldn't this only apply to the scrollable table variant?

packages/components/tablenext/src/TableRow/TableRow.tsx
```diff
@@ -0,0 +1,100 @@
...
+      console.warn(
+        `[TableNext.Row] Some row cells do not have matching column titles, which can make the cell content harder to understand. ` +
+          `Received ${columnTitleCount} columnTitles for ${originalCells.length} row cells.`,
+      );
+    }
+
+    for (let i = 0; i < cellCount; i++) {
+      const title = columnTitles?.[i];
+
+      if (isStackable && i > 0 && title) {
```
denkristoffer: We should filter out header rows here, I think?

packages/components/tablenext/src/TableNext.tsx
Lelith: Yes, the tabindex is required as there could be a overflow-y scrollbar appearing. If we would add the tabindex only for isScrollable, axe would identify it as violation. And the section itself is required as a stable reference for the container-size based breakpoint.

packages/components/tablenext/src/TableRow/TableRow.tsx
Lelith: Good Idea! I now leveraged the tableCell context we use for identifying if a cell is a `th` or a `td` to only push the ghost-cells when its a body cell.

packages/components/tablenext/README.mdx
```diff
@@ -0,0 +1,116 @@
...
+typescript: ./src/TableNext.tsx,./src/TableCell/TableCell.tsx,./src/TableRow/TableRow.tsx,./src/TableBody/TableBody.tsx,
+---
+
+`TableNext` is the next iteration of the basic `Table` component. It is per default responsive by receiving a horizontal scrollbar. It additionally offers a second responsive layout mode called `stackable`. In the `stackable` layout, the Table data will be displayed in a grid, each table row represented as a single unit with all cells of this row stacked on top of each other. If a table header with titles is required, the column title is displayed next to the table cell.
+
+## Current limitations
+
+This component is still in an unstable alpha state. The `layout="stackable"` variant has the following restrictions:
+
+- **No sorting.** Sorting props (`isSortable`, `sortDirection`, `sortButtonAriaLabel`) have no effect in stackable mode. Use `layout="scrollable"` for sortable tables.
```
cf-remylenoir, line 17: Ideally, we should exclude these props when `layout="stackable"` with a discriminated union type to avoid confusion, but given the alpha state, it is acceptable.

packages/components/tablenext/package.json
cf-remylenoir, line 1: Should we add the private property right before merging, or are we fine exposing it publicly?

Lelith: good catch, i just forgot to add the private:true

packages/components/tablenext/src/TableNext.tsx
denkristoffer: The way I understand it, the stackable wrapper is only the container-query wrapper and doesn't set overflow/height, so it shouldn't become its own scroll region. Otherwise, we're adding an extra keyboard stop to stackable tables.

Lelith: I double checked again, the missing tabindex for the stackable variant falls back to missing from the LayoutBody. So technically speaking we should actually fix the tabindex for layout body, as this is the main scrolling container for the web-app.
However when the table is having the stickyHeader the scrollcontainer has to switch to the outer section, and for this case we need the tabindex on the table. 
I updated it with conditions and cleaned up the story code, so no false violations appearing for these stories.

packages/components/tablenext/README.mdx
Lelith: Yes, I thought about that one as well. But the problem is that the TableCell which receives all the information around sorting, has no knowledge of the parents property of layout. It would be available trough the context, but that is way to late for it to be useful in the Typescript type. 

And even if we clone the elements to be receiving the layout property later, it would still be too late for the user to recognize their mistake. So the only thing we can do is the error message.

packages/components/tablenext/examples/TableNextWithSortingExample.tsx
cf-remylenoir, line 1: I don't see the `.` on line 1, but apparently there is something going on here.

https://forma-36-git-feat-ufo2434stackabletablewithheadercellscombi.colorfuldemo.com/components/tablenext#sorting

Image

Lelith: The import path was wrong, it needs to be imported from the package, not a relative import.

cf-remylenoir (approved):
This looks pretty good, and as an alpha package we can always adjust on the fly in the upcoming weeks 🙂

PR #3582, opened 2026-08-04
Feat: enable navbar semantic tag overwrites

Lelith:
# Purpose of PR

Adds a new property `as` to the Navbar component, which allows overwriting of the default tag of the Navbar wrapper. Default is `header`.

This makes the Navbar more to also become child of a `banner` landmark.


packages/components/navbar/src/Navbar.tsx
```diff
@@ -64,14 +64,23 @@ type NavbarOwnProps = CommonProps & {
     labelPromotions?: string;
     labelAccount?: string;
   };
+  /**
+   * HTML element used as main wrapper of the navbar
+   * @default 'header'
+   */
+  as?: React.ElementType;
```
cf-remylenoir: Should we control what can be used in place of `header` instead of allowing anything?

Lelith: Good question, I think semantically & from an accessibility standpoint it only makes sense to allow `header` and `div`, maybe `section` as well. 
What do you think?

cf-remylenoir: I think `'header' | 'div'` would be enough. As per [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/section), the `section` tag requires a heading.

cf-remylenoir (approved):

PR #3589, opened 2026-08-05
chore: vitest

denkristoffer:
## Summary
- Add Vitest/jsdom as opt-in test infrastructure while keeping Jest as the default runner.
- Keep main's Vite 8 dependency and use `vitest.config.mts` so Vite 8 native config loading has no ESM/CommonJS warnings.
- Add a package-scoped `npm run test:vitest:image` command.
- Fully migrate `@contentful/f36-image` tests to Vitest: explicit imports, `vi` mocks, shared `axe-core` helper, and root `axe-core` dependency.
- Exclude the migrated image package from Jest so both runners can coexist during package-by-package migration.

## Validation
- `npm run test:vitest:image`: 1 file passed, 4 tests passed.
- `npm run test`: 111 suites passed, 690 tests passed, 1 todo. The converted image package is covered by Vitest instead.
- `npm run lint`: 0 errors; existing warnings remain.
- `npm exec knip`: no unused dependency failures; existing configuration hints remain.
- `npx prettier --check package.json jest.config.js vitest.config.mts scripts/test/expectNoA11yViolations.ts scripts/test/setupVitest.ts packages/components/image/package.json packages/components/image/src/Image.test.tsx`: passed.
- `npm run tsc`: fails only on known unresolved `@contentful/f36-table-next` imports in TableNext examples and LiveEditor.


Lelith:
When i run vitest it works as expected, but I also receive the following i warning 
`Your app (or one of its dependencies) is using an outdated JSX transform. Update to the modern JSX transform for faster performance: https://react.dev/link/new-jsx-transform`

is this something we have to worry about? this message is also super noisy on the other branches when more tests are running because it basically prints it once per test.

denkristoffer:
> When i run vitest it works as expected, but I also receive the following i warning `Your app (or one of its dependencies) is using an outdated JSX transform. Update to the modern JSX transform for faster performance: https://react.dev/link/new-jsx-transform`
> 
> is this something we have to worry about? this message is also super noisy on the other branches when more tests are running because it basically prints it once per test.

Vitest inherits the root TypeScript setting: `"jsx": "react"`, which selects the legacy JSX transform and emits `React.createElement(...)`. The warning does not indicate incorrect tests, but it noisy.

I think we should handle this in a follow-up, either a scoped Vitest change or a repo-wide fix updating the TypeScript and Babel JSX settings.

cf-remylenoir (approved):

PR #3616, opened 2026-08-10
chore: use automatic jsx runtime

denkristoffer:
# Purpose of PR

Switch component-library JSX emission to React's automatic runtime. Published tsup artifacts now import `react/jsx-runtime` rather than calling `React.createElement`; React peer ranges and public APIs are unchanged.

This should not affect consumers, as long as they are on one of our supported versions of React (18, 19)

Verified with:

- `pnpm --filter @contentful/f36-button build`
- `pnpm --filter @contentful/f36-components build`
- `pnpm test -- packages/components/button/src --runInBand`
- `pnpm tsc`

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -6,6 +6,7 @@ module.exports = (api) => {
         [
           '@babel/preset-react',
           {
+            runtime: 'automatic',
             targets: {
               node: 'current',
             },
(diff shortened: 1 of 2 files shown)


veu (approved):

Lelith (approved):
looks good to me :)

PR #3618, opened 2026-08-11
chore(website): ignore Contentful linkchecker 429s

denkristoffer:
# Purpose of PR

Prevent the Forma 36 website linkchecker job from failing when `contentful.com` or one of its subdomains returns HTTP 429 due to rate limiting. Other HTTP errors and 429 responses from unrelated domains remain blocking.

### DIFF
--- a/packages/website/scripts/check-links.mjs
+++ b/packages/website/scripts/check-links.mjs
@@ -5,12 +5,25 @@ import brokenLinkChecker from 'broken-link-checker';
 const { SiteChecker } = brokenLinkChecker;
 
 const DEFAULT_URL = 'http://localhost:3000';
-const EXCLUDED_KEYWORDS = [
+const DEFAULT_EXCLUDED_KEYWORDS = [
   'https://medium.com/contentful-design',
   'https://github.com/contentful/forma-36',
   'https://www.figma.com/@contentful',
   'https://react-hook-form.com',
 ];
+
+function getExcludedKeywords(
+  value = process.env.LINKCHECKER_EXCLUDED_KEYWORDS,
+) {
+  if (value === undefined) {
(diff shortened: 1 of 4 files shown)


packages/website/scripts/check-links.mjs
```diff
@@ -0,0 +1,90 @@
...
+
+const { SiteChecker } = brokenLinkChecker;
+
+const DEFAULT_URL = 'http://localhost:3000';
+const EXCLUDED_KEYWORDS = [
+  'https://medium.com/contentful-design',
+  'https://github.com/contentful/forma-36',
+  'https://www.figma.com/@contentful',
+  'https://react-hook-form.com',
+  'https://emotion.sh',
```
cf-remylenoir: The emotion domain is back online; we can remove it from the exclude list 😄 

```suggestion
```

cf-remylenoir (approved):

packages/website/scripts/check-links.mjs
```diff
@@ -0,0 +1,89 @@
...
+import brokenLinkChecker from 'broken-link-checker';
+
+const { SiteChecker } = brokenLinkChecker;
+
+const DEFAULT_URL = 'http://localhost:3000';
+const EXCLUDED_KEYWORDS = [
+  'https://medium.com/contentful-design',
+  'https://github.com/contentful/forma-36',
+  'https://www.figma.com/@contentful',
+  'https://react-hook-form.com',
```
cf-remylenoir, line 12: Ideally, we could control those from an environment variable (in Vercel) 🤔

denkristoffer: Good idea, I will add that.

denkristoffer: Added.

PR #3623, opened 2026-08-11
chore(website): remove private playground dependencies

denkristoffer:
# Purpose of PR

The public playground was requesting metadata for private alpha packages through the public npm registry. This caused Sandpack previews to fail before rendering, including the default Button example.

This removes the private `pill-next` and `table-next` packages from the public playground dependency list. Alpha component pages also hide the “Open in Playground” action because those packages are not available to Sandpack. Their inline examples and local documentation rendering remain available.

### DIFF
--- a/packages/website/components/LiveEditor/ComponentSource.tsx
+++ b/packages/website/components/LiveEditor/ComponentSource.tsx
@@ -34,6 +34,7 @@ import { Card, Button, CopyButton, Flex } from '@contentful/f36-components';
 import { theme } from './theme';
 import { formatSourceCode } from './utils';
 import * as coder from '../../utils/coder';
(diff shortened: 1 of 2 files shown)


cf-remylenoir (approved):
QA with success 

Image

Lelith:
The other playgrounds work again. But the ones for TableNext and PillNext are not working, because they of course also do not have access to these packages.
Is there a way we could disable the playground for alpha packages to avoid the bad user experience?

Lelith (approved):
let's ship the fix for the other playgrounds, so at least they are no longer broken because of the unavailable dependency

denkristoffer:
@Lelith added.

PR #3628, opened 2026-08-12
chore: migrate date and time components to Vitest

denkristoffer:
# Purpose of PR

Migrates the datepicker and datetime component packages from Jest to Vitest. This includes datetime utility tests, replaces Jest fake timers with Vitest equivalents, and makes fixed timezone expectations match the Vitest test environment.

This PR is based on `chore/vitest-avatar`.

### DIFF
--- a/packages/components/datetime/src/DateTime/DateTime.test.tsx
+++ b/packages/components/datetime/src/DateTime/DateTime.test.tsx
@@ -1,3 +1,4 @@
+import { describe, expect, it } from 'vitest';
 import React from 'react';
 import { render } from '@testing-library/react';
 import { DateTime } from './DateTime';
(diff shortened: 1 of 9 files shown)


packages/components/datetime/src/DateTime/DateTime.test.tsx
```diff
@@ -47,6 +48,6 @@ describe('DateTime', function () {
     const date = 1629240300000;
     const { container } = render(<DateTime date={date} />);
 
-    expect(container.textContent).toBe('Tue, 17 Aug 2021 at 3:45 PM');
+    expect(container.textContent).toBe('Wed, 18 Aug 2021 at 7:45 AM');
```
veu, line 51: Thinking out loud: is that a change in behavior or does the test setup have a different timezone?

denkristoffer: No it's just the test setup, I set the `TZ` for vitest to `Asia/Tokyo` in the initial PR. I didn't think about it until now, but I can restore it to `America/Los_Angeles` like it was in jest, if you prefer.

veu (approved):

packages/components/datetime/src/DateTime/DateTime.test.tsx
veu: All good, just checking that we didn’t overlook anything.

PR #3646, opened 2026-08-13
chore: remove Jest configuration and dependencies

denkristoffer:
# Purpose of PR

Complete the Jest to Vitest migration by migrating the codemod tests and removing the remaining Jest configuration, setup files, scripts, direct dependencies, and stale references. The Vitest suite has been validated with 112 passing files, 694 passing tests, and 1 todo.


cf-remylenoir (approved):

packages/forma-36-codemod/package.json
```diff
@@ -14,14 +14,13 @@
     "f36-codemod": "./bin/f36-codemod.mjs"
   },
   "scripts": {
-    "test": "jest",
```
veu, line 17: I can’t run this locally (`No test files found, exiting with code 1`). Is that expected?

packages/forma-36-codemod/transforms/__tests__/v5.test.js
```diff
@@ -1,14 +1,20 @@
 'use strict';
 
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
 const defineTest = require('jscodeshift/dist/testUtils').defineTest;
 
+globalThis.describe = describe;
```
veu, line 7: Out of interest, why do we need to add these explicitly here when it’s not needed for the other tests?

denkristoffer: The `defineTest` helper from `jscodeshift` uses the globals.

packages/forma-36-codemod/package.json
denkristoffer: As mentioned on Slack this is due to the separate script not finding the vitest config in the root. Since `forma-36-codemod` is the only package with its own test script, the simplest solution for now is to remove it and only support running it together with the other tests from the root.

veu (approved):

PR #3651, opened 2026-08-14
chore: convert Octokit release tooling to ESM

denkristoffer:
# Purpose of PR

Convert the Octokit release tooling to native ESM and upgrade Octokit from v3 to v5.

This renames both release scripts to `.mjs`, preserves their CLI behavior, exports their helpers for testing, and adds mocked Vitest coverage for read-only, write, cleanup, and error paths.

## Verification

- Focused release-tooling tests: 11 passed
- Full Vitest suite: 700 passed, 1 todo
- Prettier check passed
- Missing-token smoke command produced the expected error


scripts/changesets/smoke-test-release.test.mjs
```diff
@@ -0,0 +1,141 @@
...
+    expect(octokit.rest.repos.deleteRelease).not.toHaveBeenCalled();
+    expect(octokit.rest.git.deleteRef).not.toHaveBeenCalled();
+  });
+
+  it('runs the write path and cleans up the temporary release and tag', async () => {
+    const octokit = createOctokitMock();
+    const consoleLog = vi.spyOn(console, 'log').mockImplementation(() => {});
+
+    try {
+      await main({ octokit, shouldWrite: true });
```
veu, line 56: The test isn’t particularly helpful if it only calls `main` for the happy path and functions that `main` calls internally for the other cases. How do we know whether the internal functions will still be used in a rewrite? Especially an AI could accidentally break a feature and we wouldn’t see it if we don’t test a consistent abstraction layer.

denkristoffer: Thanks, that's a good callout. I've changed the test to go through `main` 😄

PR #3652, opened 2026-08-14
chore: align Renovate minimum release age

denkristoffer:
# Purpose of PR

Align Renovate's minimum release age with pnpm's project policy. The repository's `pnpm-workspace.yaml` requires packages to be 15 days old before installation, so Renovate should use the same minimum age.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,6 +1,7 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": ["local>contentful/renovate-config"],
+  "minimumReleaseAge": "15 days",
   "rangeStrategy": "update-lockfile",
   "labels": ["dependencies"]
 }


cf-remylenoir (commented):
Curious to know how does that work if there is a critical vulnerability patch to install?

renovate.json
```diff
@@ -1,12 +1,7 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": ["local>contentful/renovate-config"],
+  "minimumReleaseAge": "15 days",
```
cf-remylenoir, line 4: Curious to know how does that work if there is a critical vulnerability patch to install?

denkristoffer: So far Renovate has been adding exceptions like this:

https://github.com/contentful/forma-36/blob/a6eaa74232534d82eabb69f5a5ab70dd63f1c69c/pnpm-workspace.yaml#L39-L40

cf-remylenoir (approved):

PR #3672, opened 2026-09-04
chore: replace tsup with tsdown

veu:
# Purpose of PR

Replace `tsup` with its spiritual successor `tsdown` while keeping the output similar to avoid breaking changes.

There are multiple reasons to replace `tsup` with `tsdown`.
* `tsup` is no longer maintained
* `tsdown` is the spiritual successor for `tsup` which means it’s closest in setup
* `tsdown` supports `vite` plugins which gives us a way forward to integrate `lingui` with macros

## PR Checklist

- [x] Tests are passing


tsconfig.json
```diff
@@ -41,16 +33,11 @@
...
-    "types",
-    "**/*.ts",
-    "**/*.tsx"
-  ],
+  "include": ["packages", "types", "**/*.ts", "**/*.tsx"],
   "exclude": [
     "**/node_modules",
     "**/dist",
-    "tsup.config.ts",
+    "tsdown.config.ts",
```
denkristoffer:
```suggestion
    "tsdown.config.mts",
```

eslint.config.mjs
```diff
@@ -23,7 +23,7 @@ export default [
       '*.css.d.ts',
       'public/**',
       '.cache/**',
-      '**/tsup.config.ts',
+      '**/tsdown.config.ts',
```
denkristoffer:
```suggestion
      '**/tsdown.config.mts',
```

knip.json
```diff
@@ -19,7 +19,7 @@
   "entry": [
     "scripts/**/*.{js,ts,tsx,mjs}",
     "*.config.{js,mjs,ts}",
-    "tsup.config.ts"
+    "tsdown.config.ts"
```
denkristoffer:
```suggestion
    "tsdown.config.mts"
```

denkristoffer:
The increased bundle size seems to be because esbuild provides some compact helpers we don't have with Rolldown. Object spread and rest parameters, stuff like that. When I change the target from `es6` to `es2018` for native implementations of these I see a size reduction below what we had with `tsup`. So other than the changes above I think we can merge this, and then let's make a note of the target for next major release 👍

Lelith (approved):
I locally installed the build and it looks and behaves integrated the same as before.

